PR #44, opened 2022-03-07
refactor: job queue based around multi-producer single-consumer channel

WiredSound:
https://app.clickup.com/t/21pccn4

### DIFF
--- a/api/src/deployment.rs
+++ b/api/src/deployment.rs
@@ -1,5 +1,4 @@
-use crate::build::Build;
-use crate::{BuildSystem, ShuttleFactory};
+use crate::{build::Build, BuildSystem, ShuttleFactory};
 use anyhow::{anyhow, Context as AnyhowContext};
 use core::default::Default;
 use shuttle_common::project::ProjectConfig;
@@ -13,10 +12,10 @@ use std::fs::DirEntry;
 use std::io::Write;
 use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4, TcpListener};
 use std::path::{Path, PathBuf};
-use std::sync::{Arc, Mutex};
-use std::time::Duration;
+use std::sync::Arc;
 use tokio::sync::RwLock;
 use tokio::task::JoinHandle;
+use tokio::sync::{mpsc};
 
 use crate::database;
 use crate::router::Router;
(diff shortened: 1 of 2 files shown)


api/src/deployment.rs
```diff
@@ -365,8 +364,7 @@ impl DeploymentSystem {
     async fn initialise_from_fs(fs_root: &Path) -> Deployments {
         let mut deployments = HashMap::default();
         for project_dir in std::fs::read_dir(fs_root)
-            .unwrap() // safety: api can read the fs root dir
-            .into_iter() {
+            .unwrap() {
```
rimutaka, line 367: could still be not safe - fs root probably exists, but it doesn't mean the app can read it

lib/src/project.rs
```diff
@@ -29,9 +29,9 @@ fn deserialize_project_name<'de, D>(deserializer: D) -> Result<ProjectName, D::E
 impl ProjectName {
     pub fn is_valid(hostname: &str) -> bool {
         fn is_valid_char(byte: u8) -> bool {
-            (byte >= b'a' && byte <= b'z')
-                || (byte >= b'A' && byte <= b'Z')
-                || (byte >= b'0' && byte <= b'9')
+            (b'a'..=b'z').contains(&byte)
```
rimutaka, line 32: I wonder why. This is definitely more elegant, but is probably slower and takes up more memory.

api/src/auth.rs
```diff
@@ -99,7 +99,7 @@ impl UserDirectory {
     fn from_user_file() -> Self {
         let file_path = Self::users_toml_file_path();
         let file_contents: String = std::fs::read_to_string(&file_path)
-            .expect(&format!("this should blow up if the users.toml file is not present at {:?}", &file_path));
+            .unwrap_or_else(|_| panic!("this should blow up if the users.toml file is not present at {:?}", &file_path));
```
rimutaka, line 102: I'm curious to why replace expect with this construct.

lib/src/project.rs
WiredSound: I believe this was just a correction of [this Clippy lint.](https://rust-lang.github.io/rust-clippy/master/#manual_range_contains) You raise an interesting point about efficiency - I wonder if the compiler is clever enough to optimise it to be equivalent to the former under the hood or not.

api/src/auth.rs
WiredSound: Ahh this is again [a Clippy lint](https://rust-lang.github.io/rust-clippy/master/#expect_fun_call) :) Using `unwrap_or_else` avoids the unnecessary call to `format!` should `read_to_string` return `Ok`.

PR #69, opened 2022-03-16
fix: persistence regression postgres not ready

christoshadjiaslanis:
This regression occurs because the job processor loop starts before postgres is ready, resulting in us not being able to acquire a connection in `db_state.ensure().await` in `deployment.rs`.

Ideally `supervisord` would have dependencies where you can start program a after program b, but alas this is an open PR: https://github.com/Supervisor/supervisor/pull/1449/files so we'll have to employ the hacky solution of sleeping the job processor thread for 20 seconds on startup.

This is not great as we will have 20 seconds of application downtime, however it is better than not having persistence at all. 

Future improvements can employ a 'retry' system in `ensure()`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,8 @@ RUN cargo build --bin api
 
 FROM runtime
 COPY --from=builder /app/target/debug/api /usr/local/bin/shuttle-backend
+
 COPY docker/entrypoint.sh /bin/entrypoint.sh
+COPY docker/wait-for-pg-then /usr/bin/wait-for-pg-then
 COPY docker/supervisord.conf /usr/share/supervisord/supervisord.conf
 ENTRYPOINT ["/bin/entrypoint.sh"]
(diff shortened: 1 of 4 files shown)


api/src/deployment.rs
```diff
@@ -314,6 +314,12 @@ impl JobQueue {
             queue.push(deployment.clone());
             debug!("loading deployment: {:?}", deployment.meta);
         }
+        debug!("sleeping for postgres to start...");
+
+        // Postgres takes a little bit of time to start up so we'll postpone
+        // the job processor loop until then.
+
+        tokio::time::sleep(std::time::Duration::from_secs(20)).await;
```
brokad, line 322: Manually delaying the start of the loop like this may not be enough as that time would vary among environments and state of the underlying datadir.

Can we maybe use something like a connect with timeout and a test query to establish the cluster is ready instead? And after a global timeout if it didn't succeed, panic out so we can bubble up the error into the container's state.

PR #70, opened 2022-03-16
feat: set server response header so shuttle

chesedo:


brokad (approved):
LGTM, let's keep an eye out for when this gets deployed correctly to AWS in case the LB plays around with that.

PR #89, opened 2022-03-20
update users api

thecotne:
- return api keys for existing users
- add auth using `SHUTTLE_ADMIN_SECRET` env var

### DIFF
--- a/api/src/auth_admin.rs
+++ b/api/src/auth_admin.rs
@@ -1,0 +1,57 @@
+use rocket::http::Status;
+use rocket::request::{FromRequest, Outcome};
+use rocket::Request;
+use serde::{Deserialize, Serialize};
+
+use crate::auth::AuthorizationError;
+
+#[derive(Debug, PartialEq, Hash, Eq, Deserialize, Serialize, Responder)]
+pub struct AdminSecret(String);
+
+/// Parses an authorization header string into an AdminSecret
+impl TryFrom<Option<&str>> for AdminSecret {
+    type Error = AuthorizationError;
+
+    fn try_from(s: Option<&str>) -> Result<Self, Self::Error> {
+        match s {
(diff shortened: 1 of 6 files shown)


api/src/auth_admin.rs
```diff
@@ -0,0 +1,57 @@
...
+    type Error = AuthorizationError;
+
+    async fn from_request(req: &'r Request<'_>) -> Outcome<Self, Self::Error> {
+        let admin_secret = match AdminSecret::try_from(req.headers().get_one("Authorization")) {
+            Ok(admin_secret) => admin_secret,
+            Err(e) => return Outcome::Failure((Status::BadRequest, e)),
+        };
+        let shuttle_admin_secret = std::env::var("SHUTTLE_ADMIN_SECRET").expect(
+            "SHUTTLE_ADMIN_SECRET env var not found!"
+        );
```
chesedo, line 44: Can we put this in a `lazy_static!` to only read the environment variable once?

api/src/main.rs
```diff
@@ -31,12 +33,11 @@ use crate::deployment::DeploymentSystem;
...
-
-/// Creates a user if the username is available and returns the corresponding
-/// API key.
-/// Returns an error if the user already exists.
+/// Find user by username and return it's API Key.
+/// if user does not exist create it and update `users` state to `users.toml`.
+/// Finally return user's API Key.
 #[post("/users/<username>")]
-async fn create_user(username: String) -> Result<ApiKey, AuthorizationError> {
+async fn create_user(username: String, _admin: Admin) -> Result<ApiKey, AuthorizationError> {
```
chesedo, line 40: Same rename suggestion based on function comment: `get_or_create_user`.

chesedo (commented):
I'm happy with this. It looks really good @thecotne! Just some suggestions you are free to ignore.

PR #99, opened 2022-03-23
Improved examples DX

rimutaka:
The instructions for running examples were missing a few steps. I filled the gaps while I was at it. 

* print a msg about saving api-key
* updated README with missed steps
* updated Cargo.toml to avoid an error msgs

image

Given that we ask the devs to edit Shuttle.toml they will always have to either commit or use --allow-dirty. Might just give them that shortcut straight away.

And I added the missing meta in Cargo.toml to kill the warning.

### Unrelated

`Created At:         2022-03-23 02:12:35.044378437 UTC` in post-deploy would look better without nanoseconds 

`Created At:         2022-03-23 02:12:35 UTC`

Not sure that level of precision is needed.

### DIFF
--- a/cargo-shuttle/src/main.rs
+++ b/cargo-shuttle/src/main.rs
@@ -44,10 +44,16 @@ async fn login(login_args: LoginArgs) -> Result<()> {
         
         io::stdin().read_line(&mut input).unwrap();    
 
+        // read_line preserves CR/LF that need to be trimmed
+        let input = input.trim().to_string();
+        println!(
+            "Store the key in a safe location for future use with `cargo shuttle deploy --api-key {input}`"
+        );
+
         input
     });
 
-    config::create_with_api_key(api_key.trim().to_string())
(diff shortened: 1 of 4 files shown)


rimutaka:
@brokad, https://rimutaka-hello-world-rocket-app.shuttleapp.rs/ returns 404 after giving me this:

```
        Project:            rimutaka-hello-world-rocket-app
        Deployment Id:      406f024c-afe1-44a9-9be5-c13a7ab0764a
        Deployment Status:  DEPLOYED
        Host:               rimutaka-hello-world-rocket-app.shuttleapp.rs
        Created At:         2022-03-23 02:12:35.044378437 UTC
        Database URI:       postgres://user-rimutaka-hello-world-rocket-app:WW4BjBN5gHP2@pg.shuttle.rs/db-rimutaka-hello-world-rocket-app
``` 
Not sure it's supposed to be so or it's bug. FYI.

PR #101, opened 2022-03-23
refactor: loader

chesedo:
Put our loader code in services


service/tests/helpers/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+            .expect("failed to spawn a postgres container");
+        let mut stdout = child
+            .stdout
+            .take()
+            .expect("failed to get stdout of container");
+
+        let mut buf = [0; 2 << 12];
+        let mut current_pos = 0;
+
+        'wait: while let Ok(len) = stdout.read(&mut buf[current_pos..]) {
```
brokad, line 62: Is this needed? It seems to only bubble the spawned process's stdout so this could be done with `Stdio::inherit()` (which I believe is the default with `spawn`?) with potentially breaking when we see a particular line of log.

Maybe better to `while not timed out { docker exec container pg_isready; check exit code; sleep(100ms); }`?

service/tests/resources/not-shuttle/src/lib.rs
```diff
@@ -0,0 +1,36 @@
+/// This service cannot be hosted on shuttle since it is missing the entrypoint the shutlle macros would have added!!!
```
brokad, line 1: Any reason why we have all the wait code in there? Is this not equivalent as a test to if this was empty?

brokad (approved):
LGTM, great work @chesedo, just a couple nits from me :smile:

PR #106, opened 2022-03-25
refactor(api,common,cargo-shuttle): remove ProjectConfig and others

brokad:
This also
- Uses rocket guards for authentication
- Uses `ApiKey` struct instead of ad hoc string
- Remove `ProjectConfig`
- Gracefully handle failure of extraction of user archive
- Use `State<UserDirectory>` instead of `lazy_static`
- Tweak routes ahead of sandboxing
- Refactor and document `config.rs` in CLI
- Refactor `main.rs` in CLI to reduce redundancy


api/src/auth.rs
```diff
@@ -74,43 +79,95 @@ impl Display for AuthorizationError {
...
-            Ok(api_key) => api_key,
-            Err(e) => return Outcome::Failure((Status::BadRequest, e)),
-        };
-        match USER_DIRECTORY.user_for_api_key(&api_key) {
-            None => {
-                log::warn!("authorization failure for api key {:?}", &api_key);
-                Outcome::Failure((Status::Unauthorized, AuthorizationError::Unauthorized(())))
+#[async_trait]
+impl<'r> FromRequest<'r> for User {
+    type Error = Option<AuthorizationError>;
```
chesedo, line 96: Is it common to have an optional error?

api/src/auth.rs
```diff
@@ -119,71 +176,74 @@ impl UserDirectory {
     /// will return an error since the project does not belong to the current user
```
chesedo, line 176: Just need to update the comment two lines above this since we will now also fail if the user owns the project

api/src/auth.rs
```diff
@@ -119,71 +176,74 @@ impl UserDirectory {
...
+        };
 
-        self.save(&*users);
+        self.save();
 
-        Ok(ApiKey(api_key))
+        Ok(api_key)
     }
 
     /// Overwrites users.toml with a new `HashMap<String, User>`
```
chesedo, line 245:
```suggestion
    /// Overwrites users.toml with the latest users' field data
```

cargo-shuttle/src/main.rs
```diff
@@ -3,143 +3,132 @@ mod client;
...
-}
+            let _ = webbrowser::open(url);
 
-async fn delete() -> Result<()> {
-    let (api_key, project) = get_api_key_and_project()?;
-    client::delete(api_key, project)
-        .await
-        .context("failed to delete deployment")
-}
+            println!("If your browser did not automatically open, go to {url}");
```
chesedo, line 58: For some reason `{url}` keeps erroring on my system. Is the feature in rust stable yet?

chesedo (commented):
Just some minor questions and suggestions :smile:

api/src/auth.rs
brokad: I don't think so! In this case it's not a `std::error::Error` though, it's just the associated type for [`FromRequest`](https://docs.rs/rocket/0.5.0-rc.1/rocket/request/trait.FromRequest.html). I think the idea is that's what gets printed out in server-side logs and is why this only needs to be `Debug`.

That being said, let me see if I can get rid of that `Option` 🤔

cargo-shuttle/src/config.rs
```diff
@@ -1,79 +1,295 @@
...
+        C: Serialize + Default
+    {
+        if self.exists() { return Ok(()) }
+        let config = C::default();
+        self.save(&config)
+    }
+
+    fn open<C>(&self) -> Result<C>
+    where
+        C: for<'de> Deserialize<'de>
```
christoshadjiaslanis, line 36: D

cargo-shuttle/src/main.rs
chesedo: This is now working for me after using the Mozilla overlays for Rust

PR #111, opened 2022-03-28
feat: api gateway

brokad:


gateway/src/project.rs
```diff
@@ -0,0 +1,412 @@
...
+                    ctx.docker()
+                        .create_container(Some(opts), config)
+                        .and_then(|_| ctx.docker().inspect_container(&container_name, None))
+                        .await
+                } else {
+                    Err(err)
+                }
+            })
+            .await
+            .unwrap();
```
chesedo, line 243: Shouldn't these return the error - ie `?`

gateway/src/project.rs
```diff
@@ -0,0 +1,412 @@
...
+            .await
+            .or_else(|err| {
+                if matches!(err, DockerError::DockerResponseNotModifiedError { .. }) {
+                    // Already started
+                    Ok(())
+                } else {
+                    Err(err)
+                }
+            })
+            .unwrap();
```
chesedo, line 271:
```suggestion
            })?;
```

gateway/src/service.rs
```diff
@@ -0,0 +1,351 @@
...
+        }
+
+        let db = SqlitePool::connect(DB_PATH).await.unwrap();
+
+        query("CREATE TABLE IF NOT EXISTS projects (project_name TEXT PRIMARY KEY, account_name TEXT NOT NULL, project_state JSON NOT NULL)")
+            .execute(&db)
+            .await
+            .unwrap();
+
+        query("CREATE TABLE IF NOT EXISTS accounts (account_name TEXT PRIMARY KEY, key TEXT UNIQUE, super_user BOOLEAN DEFAULT FALSE)")
```
chesedo, line 78: Can we put all these in a file and execute them as a single query?

gateway/src/service.rs
```diff
@@ -0,0 +1,351 @@
...
+                    project_name,
+                    next.state(),
+                    account_name
+                );
+                if current_state != next.state() {
+                    debug!(
+                        "queuing more work for project_name={} (not done)",
+                        project_name
+                    );
+                    // TODO may deadlock?
```
chesedo, line 120: This shouldn't deadlock since projects don't interact with one another right?

gateway/src/service.rs
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    pub async fn account_name_from_key(&self, key: &Key) -> Result<AccountName, Error> {
+        let name = query("SELECT account_name FROM accounts WHERE key = ?1")
+            .bind(key)
+            .fetch_optional(&self.db)
+            .await
+            .unwrap()
+            .map(|row| row.try_get("account_name").unwrap())
+            .unwrap();  // TODO: user not found
```
chesedo, line 245: We might wanna handle this todo since it's used in the request guard?

gateway/src/service.rs
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    pub async fn create_user(&self, name: AccountName) -> Result<User, Error> {
+        let key = Key::new_random();
+        query("INSERT INTO accounts (account_name, key) VALUES (?1, ?2)")
+            .bind(&name)
+            .bind(&key)
+            .execute(&self.db)
+            .await
+            .unwrap(); // TODO: user already exists
```
chesedo, line 281: Reminder

gateway/src/service.rs
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    pub async fn is_super_user(&self, account_name: &AccountName) -> Result<bool, Error> {
+        let super_user = query("SELECT super_user FROM accounts WHERE account_name = ?1")
+            .bind(account_name)
+            .fetch_optional(&self.db)
+            .await
+            .unwrap()
+            .map(|row| row.try_get("super_user").unwrap())
+            .unwrap();  // TODO: user does not exist
```
chesedo, line 296: It seems safe to default so `false`?

gateway/src/service.rs
```diff
@@ -0,0 +1,351 @@
...
+            .map(|row| {
+                row.try_get::<ProjectName, _>("project_name").unwrap()
+            })
+    }
+
+    pub async fn create_project(&self, project_name: ProjectName, account_name: AccountName) -> Result<Project, Error> {
+        let project = SqlxJson(Project::Creating(project::ProjectCreating::new(
+            project_name.clone()
+        )));
+        // TODO
```
chesedo, line 316:
```suggestion

```

gateway/src/service.rs
```diff
@@ -0,0 +1,351 @@
...
+                }
+            }
+        });
+
+        // Queue up all the projects for reconciliation
+        for Work {
+            project_name,
+            project,
+            account_name
+        } in service.iter_projects().await
```
chesedo, line 140: It seems we will also start up projects users have deleted since we never remove them from the database and since `iter_projects()` will return projects in _all_ states.

port/src/main.rs
```diff
@@ -27,93 +23,108 @@ use structopt::StructOpt;
...
+            (s, "a ControlGuard was used in a route without a state")
+        }));
+        if let Some(auth) = request.headers().get_one("Authorization") {
+            let parts: Vec<&str> = auth.split(' ').collect();
+            if parts.len() != 2 {
+                return Outcome::Failure((Status::Unauthorized, "a request to the control plane was blocked: bad request"));
+            }
+            // unwrap ok because of explicit check above
+            let secret = *parts.get(1).unwrap();
+            if state.deployment_manager.is_authorized(secret) {
```
chesedo, line 52: Does this mean we will still need a firewall between projects? Because if each port has its own unique plane secret, then even if users found the internal IPs of other projects then they still won't have the correct plane secret to authenticate their requests?

chesedo (changes requested):
Sorry for all the comment :see_no_evil:

brokad:
Legend, thank you @chesedo for the review :smile:

Will address in a bit!

gateway/src/service.rs
```diff
@@ -0,0 +1,522 @@
...
+            .await?;
+        Ok(())
+    }
+
+    pub async fn key_from_account_name(&self, account_name: &AccountName) -> Result<Key, Error> {
+        let key = query("SELECT key FROM accounts WHERE account_name = ?1")
+            .bind(account_name)
+            .fetch_optional(&self.db)
+            .await?
+            .map(|row| row.try_get("key").unwrap())
```
bmoxb, line 327: Can use `get` instead of `try_get` to avoid repeating column name and unwrapping. Perhaps overkill but may be worth implementing `FromRow` on `Key` and then using `query_as`.
```suggestion
            .map(|row| row.get(0))
```

gateway/src/service.rs
```diff
@@ -0,0 +1,522 @@
...
+            .ok_or_else(|| Error::from(ErrorKind::UserNotFound))?;
+        Ok(key)
+    }
+
+    pub async fn account_name_from_key(&self, key: &Key) -> Result<AccountName, Error> {
+        let name = query("SELECT account_name FROM accounts WHERE key = ?1")
+            .bind(key)
+            .fetch_optional(&self.db)
+            .await?
+            .map(|row| row.try_get("account_name").unwrap())
```
bmoxb, line 337:
```suggestion
            .map(|row| row.get(0))
```

gateway/src/service.rs
```diff
@@ -0,0 +1,522 @@
...
+
+    pub async fn control_key_from_project_name(
+        &self,
+        project_name: &ProjectName,
+    ) -> Result<String, Error> {
+        let control_key = query("SELECT initial_key FROM projects WHERE project_name = ?1")
+            .bind(project_name)
+            .fetch_optional(&self.db)
+            .await?
+            .map(|row| row.try_get("initial_key").unwrap())
```
bmoxb, line 350:
```suggestion
        let (control_key,) = sqlx::query_as("SELECT initial_key FROM projects WHERE project_name = ?1")
            .bind(project_name)
            .fetch_optional(&self.db)
            .await?
```

gateway/src/service.rs
```diff
@@ -0,0 +1,522 @@
...
+        })
+    }
+
+    pub async fn is_super_user(&self, account_name: &AccountName) -> Result<bool, Error> {
+        let is_super_user = query("SELECT super_user FROM accounts WHERE account_name = ?1")
+            .bind(account_name)
+            .fetch_optional(&self.db)
+            .await?
+            .map(|row| row.try_get("super_user").unwrap())
+            .unwrap_or(false); // defaults to `false` (i.e. not super user)
```
bmoxb, line 395:
```suggestion
        let (is_super_user,) = sqlx::query_as("SELECT super_user FROM accounts WHERE account_name = ?1")
            .bind(account_name)
            .fetch_optional(&self.db)
            .await?
            .unwrap_or((false,)); // defaults to `false` (i.e. not super user)
```

gateway/src/service.rs
```diff
@@ -0,0 +1,522 @@
...
+    async fn iter_user_projects(
+        &self,
+        AccountName(account_name): &AccountName,
+    ) -> Result<impl Iterator<Item = ProjectName>, Error> {
+        let iter = query("SELECT project_name FROM projects WHERE account_name = ?1")
+            .bind(account_name)
+            .fetch_all(&self.db)
+            .await?
+            .into_iter()
+            .map(|row| row.try_get::<ProjectName, _>("project_name").unwrap());
```
bmoxb, line 408:
```suggestion
            .map(|row| row.get(0));
```

gateway/src/main.rs
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    pub fn custom<S: AsRef<str>>(kind: ErrorKind, message: S) -> Self {
+        Self {
+            kind,
+            source: Some(Box::new(io::Error::new(io::ErrorKind::Other, message.as_ref().to_string())))
+        }
+    }
+
+    pub fn kind(kind: ErrorKind) -> Self {
```
bmoxb, line 114: I've noticed inconsistent use  of `kind` and `From<ErrorKind::from` throughout the codebase - might be worth removing `kind` and just using `from` for consistency and out-of-the-box compatibility.

gateway/src/project.rs
```diff
@@ -0,0 +1,613 @@
...
+            _ => None,
+        }
+    }
+
+    pub async fn destroy<'c, C: Context<'c>>(self, ctx: &C) -> Result<(), Error> {
+        if let Some(ref container_id) = self.container_id() {
+            ctx.docker()
+                .stop_container(container_id, Some(StopContainerOptions { t: 30 }))
+                .await
+                .unwrap_or(());
```
bmoxb, line 167: A matter of preference but `let _ = ` appears to be the more common way to discard a `Result` at least in my experience.
```suggestion
            let _ = ctx.docker()
                .stop_container(container_id, Some(StopContainerOptions { t: 30 }))
                .await;
```

gateway/src/project.rs
```diff
@@ -0,0 +1,613 @@
...
+            ctx.docker()
+                .remove_container(
+                    container_id,
+                    Some(RemoveContainerOptions {
+                        force: true,
+                        ..Default::default()
+                    }),
+                )
+                .await
+                .unwrap_or(());
```
bmoxb, line 177:
```suggestion
            let _ = ctx.docker()
                .remove_container(
                    container_id,
                    Some(RemoveContainerOptions {
                        force: true,
                        ..Default::default()
                    }),
                )
                .await;
```

gateway/src/service.rs
```diff
@@ -0,0 +1,522 @@
...
+            }
+        }
+
+        service
+    }
+
+    pub async fn set_sender(&self, sender: Option<Sender<Work>>) -> Result<(), Error> {
+        *self.sender.lock().await = sender;
+        Ok(())
+    }
```
bmoxb, line 221: Does this need to return a `Result`?
```suggestion
    pub async fn set_sender(&self, sender: Option<Sender<Work>>) {
        *self.sender.lock().await = sender;
    }
```

gateway/src/service.rs
```diff
@@ -0,0 +1,522 @@
...
+    pub async fn find_project(&self, project_name: &ProjectName) -> Result<Project, Error> {
+        query("SELECT project_state FROM projects WHERE project_name=?1")
+            .bind(project_name)
+            .fetch_optional(&self.db)
+            .await?
+            .map(|r| {
+                r.try_get::<SqlxJson<Project>, _>("project_state")
+                    .unwrap()
+                    .0
+            })
```
bmoxb, line 305:
```suggestion
            .map(|r| r.get::<SqlxJson<Project>, _>(0).0)
```

gateway/src/auth.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+impl Key {
+    pub fn new_random() -> Self {
+        Self(
+            rand::thread_rng()
+                .sample_iter(&rand::distributions::Alphanumeric)
+                .take(16)
+                .map(char::from)
+                .collect::<String>(),
```
bmoxb, line 57: Will need to `use rand::distributions::DistString` but looks a bit prettier :)
```suggestion
            rand::distributions::Alphanumeric.sample_string(&mut rand::thread_rng(), 16)
```

bmoxb (commented):
Looks very nice :) Please excuse/feel free to ignore my incredibly nitpicky suggestions

gateway/src/service.rs
```diff
@@ -0,0 +1,522 @@
...
+
+            match self.service.update(&next).await {
+                Ok(_) => {}
+                Err(err) => info!("failed to update a state: {}\nstate: {:?}", err, next),
+            };
+
+            if !next.is_done() {
+                // Safety: Should only panic if `self.recv` has
+                // hanged, which at this point can't happen
+                self.send.send(next).await.unwrap();
```
chesedo, line 130: Will this block (deadlock) when the channel is full?

gateway/src/service.rs
```diff
@@ -0,0 +1,522 @@
...
+use crate::{auth::Key, AccountName};
+
+impl GatewayService {
+    /// Initialize `GatewayService` and its required dependencies.
+    ///
+    /// * `args` - The [`Args`] with which the service was
+    /// started. Will be passed as [`Context`] to workers and state.
+    pub async fn init(args: Args) -> Arc<Self> {
+        info!("docker host: tcp://127.0.0.1:2735");
+        let docker = Docker::connect_with_http_defaults().unwrap();
```
chesedo, line 157:
```suggestion
        let docker = Docker::connect_with_local_defaults().unwrap();
```
Just to make it work on Unix and Windows

chesedo (commented):
Oh boy, is that a lot of comments. Hopefully, some can be ignored :sweat_smile:

api/Containerfile.dev
```diff
@@ -3,22 +3,13 @@ RUN apt-get update &&\
     apt-get install -y supervisor
 RUN rustup component add rust-src
 
-FROM rust:buster AS chef
+FROM rust:buster AS builder
 RUN apt-get update &&\
     apt-get install -y protobuf-compiler
 WORKDIR app
-RUN cargo install cargo-chef
```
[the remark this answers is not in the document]
brokad: Oops! I made this change as the cache was getting me in trouble for a bit, then I forgot to revert it. Thanks!

gateway/src/main.rs
```diff
@@ -0,0 +1,314 @@
...
+//       starting up~~
+
+#![allow(warnings)]
+
+#[macro_use]
+extern crate async_trait;
+
+#[macro_use]
+extern crate log;
+extern crate core;
```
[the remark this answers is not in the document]
brokad: That's weird! I guess my IDE added that!

gateway/src/main.rs
brokad: I'll refactor the `From` to dispatch through `Error::kind`.

gateway/src/main.rs
```diff
@@ -0,0 +1,314 @@
...
+pub struct ProjectName(pub String);
+
+impl<'de> Deserialize<'de> for ProjectName {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>
+    {
+        String::deserialize(deserializer)?
+            .parse()
+            .map_err(|_err| todo!())
```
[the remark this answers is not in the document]
brokad: Probably is 😄

gateway/src/project.rs
```diff
@@ -0,0 +1,613 @@
...
+                }
+            }
+        });
+
+        config.host_config = deserialize_json!({
+            "Mounts": [{
+                "Target": "/opt/shuttle",
+                "Source": format!("{prefix}{project_name}_vol"),
+                "Type": "volume"
+            }]
```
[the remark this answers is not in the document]
brokad: Yes, indeed. Instead these are stored as retained as docker volumes.

PR #122, opened 2022-04-06
add support for --working-directory and --name parameters to `cargo shuttle status` and friends

alsuren:
Hi all. I added a --name argument to `cargo shuttle status`, so that I could check whether my project was up without `cd`ing into my project dir. I also added a --working-directory arg, while I was at it.

## Open questions:

- [x] Are these good names for command-line flags?

for `--working-directory`, I did a quick:
```bash
 for f in  /Users/alsuren/.cargo/bin/cargo* ; do 
    echo $f
    $f --help | grep dir
done
```
and the only thing that supports any kind of dir arg is:
```
/Users/alsuren/.cargo/bin/cargo-watch
    -C, --workdir <workdir>        Change working directory before running command [default: crate root]
```
cargo subcommands tend to accept a --manifest-path flag, instead
```
        --manifest-path <manifest-path>      Specify path to Cargo.toml
```
(I only implemented `--working-directory` flag because it felt like a pure refactor, and I like to front-load my PRs with refactors if I can. I can kill it off if you want)

- [x] Should I also update the documentation? 

`--name` would allow the "Change the name of your service" section of the docs (https://docs.rs/shuttle-service/latest/shuttle_service/#change-the-name-of-your-service) to be run without making any changes to the git repo (I feel like there was some feedback about needing the `--allow-dirty` flag to deploy the examples, but I can't find the issue now). 

- [x] Should I add some tests?

I feel like it should be pretty easy to write some assert_cmd tests for the failing structopt cases (https://rust-cli.github.io/book/tutorial/testing.html#testing-cli-applications-by-running-them), to make sure the error messages for `--working-directory=something/that/doesnt/exist` makes sense, and same for names that are not valid project names. I can also add some tests for the `name` fallback code.

### DIFF
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -1,3 +1,9 @@
+use std::{
+    ffi::{OsStr, OsString},
+    fs::canonicalize,
+    path::PathBuf,
+};
+
 use structopt::StructOpt;
 
 #[derive(StructOpt)]
@@ -17,6 +23,8 @@ pub struct Args {
     )]
     /// Run this command against the api at the supplied url
     pub api_url: Option<String>,
+    #[structopt(flatten)]
+    pub project_args: ProjectArgs,
     #[structopt(subcommand)]
     pub cmd: Command,
 }
@@ -51,4 +59,22 @@ pub struct AuthArgs {
 pub struct DeployArgs {
     #[structopt(long, about = "allow dirty working directories to be packaged")]
     pub allow_dirty: bool,
+}
+
+fn parse_working_directory(working_directory: &OsStr) -> Result<PathBuf, OsString> {
+    canonicalize(working_directory)
+        .map_err(|e| format!("could not turn {working_directory:?} into a real path: {e}").into())
+}
+
+// Common args for subcommands that deal with projects.
+#[derive(StructOpt)]
+pub struct ProjectArgs {
+    #[structopt(
+        global = true,
+        long,
+        parse(try_from_os_str = parse_working_directory),
+        default_value = ".",
+        about = "specify the working directory"
+    )]
+    pub working_directory: PathBuf,
 }
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -4,6 +4,7 @@ use std::{
     path::PathBuf,
 };
 
+use shuttle_common::project::ProjectName;
 use structopt::StructOpt;
 
 #[derive(StructOpt)]
@@ -77,4 +78,6 @@ pub struct ProjectArgs {
         about = "specify the working directory"
     )]
     pub working_directory: PathBuf,
+    #[structopt(long, about = "specify the name of the project (overrides crate name)")]
+    pub name: Option<ProjectName>,
 }
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -41,7 +41,11 @@ pub struct ProjectArgs {
         about = "specify the working directory"
     )]
     pub working_directory: PathBuf,
-    #[structopt(long, about = "specify the name of the project (overrides crate name)")]
+    #[structopt(
+        global = true,
+        long,
+        about = "specify the name of the project (overrides crate name)"
+    )]
     pub name: Option<ProjectName>,
 }
 
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -30,6 +30,26 @@ pub struct Args {
     pub cmd: Command,
 }
 
+// Common args for subcommands that deal with projects.
+#[derive(StructOpt)]
+pub struct ProjectArgs {
+    #[structopt(
+        global = true,
+        long,
+        parse(try_from_os_str = parse_working_directory),
+        default_value = ".",
+        about = "specify the working directory"
+    )]
+    pub working_directory: PathBuf,
+    #[structopt(long, about = "specify the name of the project (overrides crate name)")]
+    pub name: Option<ProjectName>,
+}
+
+fn parse_working_directory(working_directory: &OsStr) -> Result<PathBuf, OsString> {
+    canonicalize(working_directory)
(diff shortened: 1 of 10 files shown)


christoshadjiaslanis:
Hey @alsuren this looks really good.

I think the `--working-directory` flag is a good idea. It might be cool as well to have a something like the functionality in git - where in any given directory it will go to it's parent to check if that is a git directory recursively until it reaches `/` such that you can do `git status` from any sub-directory of the root git directory which contains the `.git` folder.

In our case we don't really have a `.shuttle` marker or something like that - except for the optional `Shuttle.toml`.

Speaking of which - for the case where you want to change the name of your service without needing to change any underlying config files `--name` would work great. I'm curious as to how this came up?

alsuren:
> It might be cool as well to have a something like the functionality in git - where in any given directory it will go to it's parent to check if that is a git directory recursively until it reaches `/` such that you can do `git status` from any sub-directory of the root git directory which contains the `.git` folder.
>
> In our case we don't really have a `.shuttle` marker or something like that - except for the optional `Shuttle.toml`.

I have a feeling that `cargo metadata` will already do that for you. I can't remember whether it finds the workspace root or stops at the first Cargo.toml it finds. I will try writing a test over the weekend or something.

> Speaking of which - for the case where you want to change the name of your service without needing to change any underlying config files `--name` would work great. I'm curious as to how this came up?

I just wanted to deploy one of the example projects, to get a feel for how the platform worked. I got a name collision because it already existed. I then switched to the laptop that I brought on holiday with me and got the same genre of problem again. There might be other better ways to solve this though (like giving everyone their own namespace based on their GitHub username, so you can deploy all of the examples without name collisions?)

cargo-shuttle/src/config.rs
```diff
@@ -320,3 +334,81 @@ impl RequestContext {
...
+            .as_ref()
+            .unwrap()
+            .name
+            .as_ref()
+            .unwrap()
+            .to_string();
+
+        // FIXME: this is not the intended behaviour. We should fix this.
+        // This should really be "hello-world-axum-app", as above.
+        assert_eq!(name, "hello-world");
```
alsuren, line 392: I guess I should do this as part of this PR? I think it's an existing bug, so I would usually fix it as a separate PR, to avoid things dragging out too long unmerged. If we want it fixed as part of this PR then I can look at it later in the week (requires re-jigging when we call out to `cargo metadata`, I think).

cargo-shuttle/src/args.rs
```diff
@@ -38,4 +45,25 @@ pub struct AuthArgs {
...
+}
+
+fn parse_working_directory(working_directory: &OsStr) -> Result<PathBuf, OsString> {
+    canonicalize(working_directory)
+        .map_err(|e| format!("could not turn {working_directory:?} into a real path: {e}").into())
+}
+
+// Common args for subcommands that deal with projects.
+#[derive(StructOpt)]
+pub struct ProjectArgs {
```
chesedo, line 59: This can now be combined with the recently created [`Args`](https://github.com/shuttle-hq/shuttle/blob/6a13ff0525517a36043ba6f0bbfa0cd0f3047559/cargo-shuttle/src/args.rs#L12)

cargo-shuttle/src/config.rs
chesedo: We can keep this as a separate PR. Will just need to open an issue for it then.

chesedo (changes requested):
Thanks for this @alsuren. Just one change request for some code that changed just after this PR.

chesedo:
Personally, I prefer the `manifest-path` from `cargo` more, then `workdir`. But anything will do really

chesedo:
Oh, and yes, a doc update would be great :smile:

cargo-shuttle/src/args.rs
alsuren: rebased and force-pushed.

I added this to Args as 
```rust
    #[structopt(flatten)]
    pub project_args: ProjectArgs,
```
because it was nice to be able to pass it as a single arg into `RequestContext::load_local()`. This is the simplest approach, but has slightly awkward consequences in the UI (see below).

I also marked `name` and `working_directory` as `global = true`. This means that they will apply to all subcommands (unfortunately, this means you can do `cargo shuttle login --name=myproject` even though `login` doesn't know anything about projects).

(I did not touch api_url, because api_url feels like a kind-of hidden feature, that you only use for local dev, and probably shouldn't be in everyone's face)

I'm really tempted to put ProjectArgs back into the payloads of each of the project-related sub-command enums (this is where I originally put them). This does not allow them to be specified at the top level (`cargo shuttle --name=myproject status`). It would also be possible to put ProjectArgs in Args and also each of the subcommands, and merge them, but this feels like a maintenance headache.

cargo-shuttle/src/config.rs
alsuren: Done: https://github.com/shuttle-hq/shuttle/issues/163

cargo-shuttle/src/args.rs
chesedo: Ah, I see what you mean about `login` not using them. For now, I think it's fine to keep the code simpler.

I never knew about `global = true` so learned something new :smile: . The `api_url` will be used for self-deploys soon. But it can remain hidden for now.

Let's avoid an args merging maintenance headache :sweat_smile:

cargo-shuttle/src/config.rs
```diff
@@ -341,3 +355,81 @@ impl RequestContext {
...
+    #[test]
+    fn get_local_config_finds_name_in_shuttle_toml() {
+        let project_args = ProjectArgs {
+            working_directory: path_from_workspace_root("examples/axum/hello-world/"),
+            name: None,
+        };
+
+        let local_config = RequestContext::get_local_config(&project_args).unwrap();
+
+        // FIXME: make a helper for this?
```
chesedo, line 382: For three use cases, we should make this a helper. You can go as simple as a function call if you want to.

chesedo (changes requested):
Just the helper that's left. Thanks again for this!

chesedo (approved):

PR #123, opened 2022-04-08
Gracefully shutdown service when running `cargo shuttle delete`

smallstepman:
### The idea goes as follows:
1. grab server's shutdown handle, have it sitting on the receiver end of the `tokio::sync::oneshot` channel, and wrap everything in a tokio thread ([like here](https://github.com/getsynth/shuttle/commit/55a64e4cfdb88202e395302f1c9bc0068841221a#diff-a981df14dcb291acc010b3484e3911046bb9f568ebd28ed7c6c2bc4de23aa1faR420-R426))
2. assign transmiter end of the channel to [`SimpleService`'s field](https://github.com/getsynth/shuttle/commit/c3610f60403506f760f0a2142a9d034c515cbebd#diff-a981df14dcb291acc010b3484e3911046bb9f568ebd28ed7c6c2bc4de23aa1faR377) ([+ assignment itself](https://github.com/getsynth/shuttle/commit/c3610f60403506f760f0a2142a9d034c515cbebd#diff-a981df14dcb291acc010b3484e3911046bb9f568ebd28ed7c6c2bc4de23aa1faR413)), then have it passed around `SimpleService ==> dyn Service ==> Loader::load ==> DeploymentState ==> fn kill_deployment`
3. send value trought the channel to shutdown the service

### Current status:
Still validating PoC, expect to find some nonsense. 

[First approach](https://github.com/getsynth/shuttle/commit/c3610f60403506f760f0a2142a9d034c515cbebd) was kinda nice looking, but couldn't figure out how to walk around the issue of grabbing function pointer to trait object ([here](https://github.com/getsynth/shuttle/commit/c3610f60403506f760f0a2142a9d034c515cbebd#diff-a97dd9bb823b0963666fdfa86931d31afe901935b6a5d20686e7122199f2a315R62-R65)). Second approach (HEAD aka 55a64e4) looks like it's complete solution and compiles, but throws an error when trying to `deploy` hello-world example 
```
shuttle-app-1  | 2022-04-08 01:58:28,876 DEBG fd 9 closed, stopped monitoring <POutputDispatcher at 140577858553040 for <Subprocess at 140577858488152 with name shuttle-api in state STARTING> (stdout)>
```
not sure how to debug it yet...

### References (shutdown handles):
- [axum-server graceful_shutdown](https://docs.rs/axum-server/latest/src/graceful_shutdown/graceful_shutdown.rs.html#42)
- [Rocket (stage: `Ignite`)](https://docs.rs/rocket/0.5.0-rc.1/rocket/struct.Rocket.html#method.shutdown)
- [Rocket (stage: `Orbit`)](https://docs.rs/rocket/0.5.0-rc.1/rocket/struct.Rocket.html#method.shutdown-1)

### DIFF
--- a/api/src/deployment.rs
+++ b/api/src/deployment.rs
@@ -495,9 +495,15 @@ impl DeploymentSystem {
                     ..
                 }) = lock.take()
                 {
+                    if let Some(f) = *shutdown_handle {
+                        // TODO: ideally it would receive feedback regarding shutdown
+                        // finalization, and only then we'd move to aborting the thread
+                        // TODO: error handling
+                        f.send(true).ok();
+                    }
+                    // QUESTION: why abort thread instead of stopping the runtime?
                     handle.abort();
                     tokio::spawn(async move {
-                        shutdown_handle();
                         so.close().unwrap();
                     });
                 }
@@ -619,7 +625,7 @@ impl DeploymentState {
         port: Port,
         handle: ServeHandle,
         db_state: database::State,
-        shutdown_handle: fn(),
+        shutdown_handle: Box<Option<tokio::sync::oneshot::Sender<bool>>>,
     ) -> Self {
         Self::Deployed(DeployedState {
             so,
@@ -656,5 +662,5 @@ struct DeployedState {
     port: Port,
     handle: ServeHandle,
     db_state: database::State,
-    shutdown_handle: fn(),
+    shutdown_handle: Box<Option<tokio::sync::oneshot::Sender<bool>>>,
 }
--- a/api/src/deployment.rs
+++ b/api/src/deployment.rs
@@ -177,7 +177,7 @@ impl Deployment {
                                     debug!("{}: factory phase FAILED: {:?}", meta.project, e);
                                     DeploymentState::Error(e.into())
                                 }
-                                Ok((handle, so)) => {
+                                Ok((handle, so, shutdown_handle)) => {
                                     debug!("{}: factory phase DONE", meta.project);
                                     // Remove stale active deployments
                                     if let Some(stale_id) =
@@ -187,7 +187,13 @@ impl Deployment {
                                         context.deployments.write().await.remove(&stale_id);
                                     }
 
-                                    DeploymentState::deployed(so, port, handle, db_state)
+                                    DeploymentState::deployed(
+                                        so,
+                                        port,
+                                        handle,
+                                        db_state,
+                                        shutdown_handle,
+                                    )
                                 }
                             }
                         }
@@ -482,9 +488,16 @@ impl DeploymentSystem {
                 // library when the runtime gets around to it.
 
                 let mut lock = deployment.state.write().await;
-                if let DeploymentState::Deployed(DeployedState { so, handle, .. }) = lock.take() {
+                if let DeploymentState::Deployed(DeployedState {
+                    so,
+                    handle,
+                    shutdown_handle,
+                    ..
+                }) = lock.take()
+                {
                     handle.abort();
                     tokio::spawn(async move {
+                        shutdown_handle();
                         so.close().unwrap();
                     });
                 }
@@ -601,12 +614,19 @@ impl DeploymentState {
         Self::Loaded(loader)
     }
 
-    fn deployed(so: Library, port: Port, handle: ServeHandle, db_state: database::State) -> Self {
+    fn deployed(
+        so: Library,
+        port: Port,
+        handle: ServeHandle,
+        db_state: database::State,
+        shutdown_handle: fn(),
+    ) -> Self {
         Self::Deployed(DeployedState {
             so,
             port,
             handle,
             db_state,
+            shutdown_handle,
         })
     }
 
@@ -636,4 +656,5 @@ struct DeployedState {
     port: Port,
     handle: ServeHandle,
     db_state: database::State,
+    shutdown_handle: fn(),
 }
--- a/service/src/lib.rs
+++ b/service/src/lib.rs
@@ -271,6 +271,9 @@ pub trait Service: Send + Sync {
     ///
     /// The deployer expects this instance of [Service][Service] to bind to the passed [SocketAddr][SocketAddr].
     fn bind(&mut self, addr: SocketAddr) -> Result<(), error::Error>;
+    fn shutdown(&mut self) -> Result<(), error::Error> {
+        Ok(())
+    }
 }
 
 /// A convenience trait for handling out of the box conversions into [Service][Service] instances.
@@ -371,6 +374,7 @@ pub struct SimpleService<T> {
     service: Option<T>,
     builder: Option<StateBuilder<T>>,
     runtime: Runtime,
+    shutdown_handler: Option<tokio::sync::oneshot::Sender<bool>>,
 }
 
 impl<T> IntoService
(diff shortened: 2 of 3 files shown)


smallstepman:
one thing to add: before I attempted to implement this PR, I've built bunch of toy/sandbox things to test things out.
When I ran this one https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=35fb60cce6a66ac1788e5a38cf7f1dde , and send SIGHUP signal to the process, it happily tried to take down rocket server as instructed, however, it produces some warnings, which look kinda what I'd expect, but not 100%
```
🚀 Rocket has launched from http://127.0.0.1:8000
got signal HUP
()
killed it
its killing me
died
Received shutdown request. Waiting for pending I/O...
Warning: Server failed to shutdown cooperatively.                   <---- huh ??? ;/
   >> Server is executing inside of a custom runtime.
   >> Rocket's runtime is `#[rocket::main]` or `#[launch]`.
   >> Refusing to terminate runaway custom runtime.
```
last three lines (`>> `) would be gone if I were to use `#[rocket::main]` instead of 
```rust
    let r = tokio::runtime::Runtime::new().unwrap();
    r.spawn(
        rocket.launch(),
    );
```
Additionally, if after sending 1st signal I were to send another signal SIGTERM (terminate), this message will get logged
```
Warning: Received SIGTERM. Shutdown already in progress.
```

service/src/loader.rs
```diff
@@ -50,16 +52,21 @@ impl Loader {
         self,
         factory: &mut dyn Factory,
         addr: SocketAddr,
-    ) -> Result<(ServeHandle, Library), Error> {
+    ) -> Result<(ServeHandle, Library, ShutdownHandle), Error> {
         let mut service = self.service;
+        let shutdown_handle = service.shutdown_handle();
```
chesedo, line 57: The handle is created in `bind()`, so doesn't this get it too early?

service/src/lib.rs
```diff
@@ -417,6 +439,10 @@ impl Service for SimpleService<rocket::Rocket<rocket::Build>> {
             .map_err(error::CustomError::new)?;
         Ok(())
     }
+
+    fn shutdown_handle(&mut self) -> Box<Option<tokio::sync::oneshot::Sender<bool>>> {
```
brokad, line 443: Should we just make this `shutdown() -> Result<(), _>` (instead of returning a handle)? This may help each implementation to decide how to shut itself down.

api/src/deployment.rs
```diff
@@ -482,7 +488,20 @@ impl DeploymentSystem {
...
+                    ..
+                }) = lock.take()
+                {
+                    if let Some(f) = *shutdown_handle {
+                        // TODO: ideally it would receive feedback regarding shutdown
+                        // finalization, and only then we'd move to aborting the thread
+                        // TODO: error handling
+                        f.send(true).ok();
+                    }
+                    // QUESTION: why abort thread instead of stopping the runtime?
```
brokad, line 504: That is a good question, and I believe this to be an error.

Maybe we should not even have the `Handle` in there in the first place but just keep the `Service` inside `DeploymentState` and call its `::shutdown` on signalling for termination.

service/src/lib.rs
```diff
@@ -271,6 +271,10 @@ pub trait Service: Send + Sync {
     ///
     /// The deployer expects this instance of [Service][Service] to bind to the passed [SocketAddr][SocketAddr].
     fn bind(&mut self, addr: SocketAddr) -> Result<(), error::Error>;
+    fn shutdown_handle(&mut self) -> Box<Option<tokio::sync::oneshot::Sender<bool>>> {
+        // TODO: get rid of the default
```
brokad, line 275: TODO reminder :smile:

This gets hit, for example, when running the `shuttle-service` tests.

api/src/deployment.rs
smallstepman: I think having `Service` inside `DeploymentState` would be the most convenient, but also requires bigger refactor, therefore I myself held off from trying it before I showed you guys this one. I'm gonna attempt it next and see how it goes.

Worst case scenario we can pass around the handle to runtime and just shutdown the runtime. I'm sure it will work, but I'm also least excited about that option cause I'm not really certain what happens under the hood when we do that - it would be nice to know this, but that would require a lot more effort (at least for me, unless someone would be interested to help and research this together). Also I'm not sure if shutting down underlying runtime in Rocket vs Axum vs [...] would yield the same effect.

service/src/lib.rs
smallstepman: I also think that's better approach, and that's the first thing I tried, but couldn't figure this out: https://github.com/getsynth/shuttle/blob/c3610f60403506f760f0a2142a9d034c515cbebd/service/src/loader.rs#L60-L65 .
The issues is we're moving `service: Box<dyn Service>` into tokio thread, therefore we're no longer able to access it or its methods.

smallstepman:
Test are passing, however, calling `cargo shuttle delete` will currently freeze the API server.

service/src/lib.rs
smallstepman: perhaps I should leave that out :smile: 

resolving since we no longer use `fn shutdown_handle`

api/src/deployment.rs
smallstepman: [3a78937](https://github.com/getsynth/shuttle/pull/123/commits/3a789379f6fa3adcabd476ae56c54ef8d5fd30c7) resolves it

service/src/loader.rs
smallstepman: no longer an issue

api/src/deployment.rs
```diff
@@ -488,21 +484,15 @@ impl DeploymentSystem {
...
-                        // TODO: ideally it would receive feedback regarding shutdown
-                        // finalization, and only then we'd move to aborting the thread
-                        // TODO: error handling
-                        f.send(true).ok();
-                    }
-                    // QUESTION: why abort thread instead of stopping the runtime?
-                    handle.abort();
+                if let DeploymentState::Deployed(DeployedState { so, service, .. }) = lock.take() {
+                    // TODO: error handling
+                    let mut service = service.lock().unwrap();
```
smallstepman, line 489: The api is getting frozen because we're not able to acquire lock here. That happens because `bind()` is blocking.

service/src/lib.rs
```diff
@@ -365,8 +366,12 @@ where
         self.runtime
             .block_on(launched)
             .map_err(error::CustomError::new)?;
```
smallstepman, line 368: this will ofc block, therefore `fn bind` will never return, therefore `Mutex lock` acquired in `Loader.load` in `tokio::spawn_blocking` call will never get released.

service/src/loader.rs
```diff
@@ -42,31 +42,24 @@ impl Loader {
...
+        let service = Arc::clone(&self.service);
 
         // We cannot use spawn here since that blocks the api completely. We suspect this is because `bind` makes a blocking call,
         // however that does not completely makes sense as the blocking call is made on another runtime.
-        let handle = tokio::task::spawn_blocking(move || service.bind(addr));
-
-        // i want this
-        // this is just a mock
-        // let shutdown_handle = || {};
+        tokio::task::spawn_blocking(move || service.lock().unwrap().bind(addr));
```
smallstepman, line 60: lock forever

chesedo: Lol, fun :sweat: Let's try not blocking...

PR #126, opened 2022-04-09
Tests for cargo-shuttle

alsuren:
This is based on top of https://github.com/getsynth/shuttle/pull/122. It is also pretty opinionated. Probably a bit too opinionated for a first contribution, so it might be worth getting that PR merged first, and then deciding which bits of this PR you want to keep, and which to re-work.

I originally developed it as https://github.com/alsuren/shuttle/pull/1, but I don't have any self-hosted runners available over there, so I have opened up a draft pr over here.

> > It might be cool as well to have a something like the functionality in git - where in any given directory it will go to it's parent to check if that is a git directory recursively until it reaches `/` such that you can do `git status` from any sub-directory of the root git directory which contains the `.git` folder.
> > In our case we don't really have a `.shuttle` marker or something like that - except for the optional `Shuttle.toml`.
> 
> I have a feeling that `cargo metadata` will already do that for you. I can't remember whether it finds the workspace root or stops at the first Cargo.toml it finds. I will try writing a test over the weekend or something.

I wrote the above test. It kind-of passes and kind-of fails: cargo-metadata correctly finds the crate root, but then we don't use that information to find the Shuttle.toml. I left it in there as `fixme_running_in_src_subdir_finds_crate_but_fails_to_find_config()` so you can see what I mean.

things that are ugly:

- [ ] there are probably a bunch of unix-isms floating about, that will break on windows. I will wait for a set windows CI build errors before thinking about how to solve these things.
- [ ] something odd is going with the dead_code lint in `clippy --tests` for cargo-shuttle.
- [ ] I have gone all-in on the integration-tests-only approach for cargo-shuttle, following matklad's blog post. A few of them are things that I would normally write as unit tests. I can turn them into unit tests if you prefer.
- [ ] the bin-only -> bin+lib split is potentially controversial, and may also cause merge conflicts with other work that's going on at the same time. I can keep cargo-shuttle as a bin-only crate, and convert the `config` tests into "unit" tests, if you prefer.
- [ ] `fixme_running_in_src_subdir_finds_crate_but_fails_to_find_config()` is kind-of a failing test. I suspect that the fix won't be too bad, but this PR series is bigger than I would like already (given that this is still my first code contribution to the project), so I would prefer to fix it later.
- [ ] we should probably work out a subset of the `cargo-test` workflow that can be run on the github OSS free-tier windows/mac/linux runners.


alsuren:
I have rolled the changes into https://github.com/getsynth/shuttle/pull/122 (using unit tests rather than integration tests with a main+lib split). I will work on the cache + github runner stuff in my own time, on https://github.com/alsuren/shuttle/pull/2

PR #144, opened 2022-04-20
feat: implement secrets

WiredSound:
I spent a considerable amount of time trying out different ways of doing this but ultimately settle on a pretty simple system wherein `PgPool` implements a `SecretStore` which fetches secrets stored in a table appropriately named `secrets` in the database. Storing secrets directly in the user-accessible database allows for them to simply set secrets by just connecting to the database directly. Secrets are also read from a `Secrets.toml` file at time of deployment.


service/src/secrets.rs
```diff
@@ -0,0 +1,90 @@
...
+    for<'c> usize: ColumnIndex<<DB as Database>::Row>,
+    for<'c> Query<'c, Postgres, PgArguments>: Execute<'c, DB>
+{
+    // TODO: Don't restrict to Postgres types above.
+
+    const GET_QUERY: &'static str;
+    const SET_QUERY: &'static str;
+    const CREATE_TABLE_QUERY: &'static str;
+
+    /// Read the secret with the given key from the database. Will an error if a secret with the
```
chesedo, line 50:
```suggestion
    /// Read the secret with the given key from the database. Will error if a secret with the
```

service/src/lib.rs
```diff
@@ -165,6 +165,11 @@ use tokio::runtime::Runtime;
 pub mod error;
 pub use error::Error;
 
+#[cfg(feature = "sqlx-postgres")]
+pub mod secrets;
+#[cfg(feature = "sqlx-postgres")]
```
chesedo, line 170: Should this maybe be behind a secrets feature gate?

api/Cargo.toml
```diff
@@ -29,4 +29,4 @@ base64 = "0.13.0"
 rand = "0.8.5"
 
 shuttle-common = { version = "0.2.6", path = "../common" }
-shuttle-service = { version = "0.2.6", path = "../service", features = [ "loader" ] }
+shuttle-service = { version = "0.2.6", path = "../service", features = [ "loader", "sqlx-postgres" ] }
```
chesedo, line 32: Is it possible to not turn the postgres feature on? Maybe the secret store should be in common instead?

api/src/args.rs
```diff
@@ -30,7 +30,7 @@ pub struct Args {
     )]
     pub(crate) bind_addr: IpAddr,
     #[structopt(long, about = "Fully qualified domain name deployed services are reachable at", parse(try_from_str = parse_fqdn))]
-    pub(crate) proxy_fqdn: FQDN,
+    pub(crate) proxy_fqdn: FQDN
```
chesedo, line 33: For some reason the rustfmt file seems to not be respected by your formater :joy:

cargo-shuttle/src/client.rs
```diff
@@ -138,6 +157,33 @@ pub(crate) async fn deploy(
...
 
+pub(crate) async fn secrets(
+    api_url: ApiUrl,
+    api_key: &ApiKey,
+    project: &ProjectName,
+    secrets: HashMap<String, String>
+) -> Result<()> {
+    if secrets.is_empty() {
+        Ok(())
+    }
```
chesedo, line 168:
```suggestion
    if secrets.is_empty() {
        return Ok(());
    }
```
This is called a guard clause and will allow you to remove the nesting created by the `else` block (to be removed manually :sweat_smile: )

chesedo (changes requested):
Looking good!! Left some minor suggestions

api/Cargo.toml
WiredSound: But then that would surely just move the Postgres dependency into common instead - would that still be advantageous?

chesedo (approved):

PR #150, opened 2022-04-25
add CODE_OF_CONDUCT.md and CONTRIBUTING.md

thecotne:
should be ok other then "Running Locally" in `CONTRIBUTING.md`

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,81 @@
+# Code of conduct
+
+* We are committed to providing a friendly, safe and welcoming environment for
+all, regardless of level of experience, gender identity and expression, sexual
+orientation, disability, personal appearance, body size, race, ethnicity, age,
+religion, nationality, or other similar characteristics.
+* Please avoid using overtly sexual aliases or other nicknames that might
+detract from a friendly, safe and welcoming environment for all.
+* Please be kind and courteous. There’s no need to be mean or rude.
+* Respect that people have differences of opinion and that every design or
+implementation choice carries a trade-off and numerous costs. There is seldom
+a right answer.
+* Please keep unstructured critique to a minimum. If you have solid ideas you
+want to experiment with, make a fork and see how it works.
+* We will exclude you from interaction if you insult, demean or harass anyone.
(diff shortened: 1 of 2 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,81 @@
...
+That is not welcome behavior. We interpret the term “harassment” as including
+the definition in the
+[Citizen Code of Conduct](https://github.com/stumpsyn/policies/blob/master/citizen_code_of_conduct.md);
+if you have any lack of clarity about what might be included in that concept,
+please read their definition. In particular, we don’t tolerate behavior that
+excludes people in socially marginalised groups.
+* Private harassment is also unacceptable. No matter who you are, if you feel
+you have been or are being harassed or made uncomfortable by a community member,
+please contact one of the channel ops or any of the
+[shuttle moderators](mailto:mods@shuttle.rs) immediately. Whether you’re a
```
christoshadjiaslanis, line 25: I don't think we have mods@shuttle.rs - consider changing to opensource@shuttle.rs

jmwill86 (approved):

jmwill86:
I've just run through these on a fresh install and everything work correctly, content all reads well. A couple of suggestions, and things I had minor issues with (which I'm happy to contribute one this has been merged, or sooner if there's a good way to do this):

CONTRIBUTING.md:

- Would it be worth linking to the [Docker install docs](https://docs.docker.com/get-started/overview/) upon mentioning docker-compose just so people know where to look if they're new to Docker.
- How to set Shuttle up with a non-example project might be helpful, detailing what `--manifest-path` should point to and maybe mentioning the optional `--working-directory` flag?
- Maybe mention the `axum-shuttle-test` can be found/changed in Cargo.toml or Shuttle.toml for the `curl --header "Host: axum-shuttle-test.shuttleapp.rs" localhost:8000/`command.
- It's also be good to know how I can view my project in a browser. I had the curl command working, but is there an easy to find URL too? (I haven't looked into this much, so it may be obvious in the code or other docs somewhere)
- Addition of a link to this page from the homepage somewhere.

Pretty painless setup though, worked well and without too much issue. Would be a great addition when this is added. 😄

PR #153, opened 2022-04-27
blog: dev log 1

christoshadjiaslanis:
TODO:

- [x] make sure state machine image renders


www/_blog/2022-04-27-dev-log-1.mdx
```diff
@@ -0,0 +1,164 @@
...
+    // Application Code
+}
+
+```
+
+In the [previous DevLog](http://localhost:3000/blog/2022/04/22/dev-log-0) we started the journey of building the shuttle MVP. We went over the design and implementation of the `cargo` subcommand which deploys cargo projects to shuttle. This has been a race against the clock, so corners were cut and tradeoffs were made. A similar theme emerges in this DevLog which covers the deployment state machine. We're going to think about compiling and deploying user code, while also covering one of my favourite design patterns in Rust.
+
+## Deployment State
+
+shuttle exposes an HTTP endpoint under `POST /deploy`. This endpoint receives a series of bytes from [`cargo shuttle`](https://crates.io/crates/cargo-shuttle) which correspond to a packaged cargo project (basically a compressed tarball with a bunch of `.rs` files).
```
chesedo, line 36:
```suggestion
shuttle exposes an HTTP endpoint under `POST /deploy`. This endpoint receives a series of bytes, from [`cargo shuttle`](https://crates.io/crates/cargo-shuttle), which correspond to a packaged cargo project (basically a compressed tarball with a bunch of `.rs` files).
```
I almost thought the "which" related to cargo-shuttle

www/_blog/2022-04-27-dev-log-1.mdx
```diff
@@ -0,0 +1,164 @@
...
+}
+```
+
+you'll also notice that there is no mutation happening here. We found it cleaner to simply drop the old state and construct a new one (although we did try).
+
+## Conclusion
+
+In the case of shuttle, using enum variants and structs to represent states in a state machine worked well. The states were distinct and clear, and for the most part the transitions are clean and self-contained.
+
+So what do you think about enum variants as states in a state machine? What would have done differently?
```
chesedo, line 162:
```suggestion
So what do you think about enum variants as states in a state machine? What would you have done differently?
```

chesedo (changes requested):
Wow, really nice!! :star_struck: Just two comments

PR #155, opened 2022-04-28
fix: routing table start

christoshadjiaslanis:
This should fix the issue around the routing table not working when the API restarts.


christoshadjiaslanis:
TODO
- [ ] needs some local testing
- [ ] consider adding an integration test to examine cross-restart behaviour

Issue #133, opened 2022-04-14
feature: support `tower::Service`


christoshadjiaslanis:
Add shuttle support for `tower::Service`.

PR #159, opened 2022-05-03
feat: Tower/Hyper integration

bmoxb:
## Problemos:

* CI doesn't work as compiling with all features results in conflicting trait implementations.
* Can't kill deployments for some reasons.

closes https://github.com/shuttle-hq/shuttle/issues/133


examples/tower/hello-world/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[dependencies]
+shuttle-service = { version = "0.2", features = ["web-tower"] }
+tokio = "0.1"
+serde = { version = "1.0", features = ["derive"] }
+hyper = { version = "0.14", features = ["server"] }
+tower = { version = "0.4", features = ["make", "util"] }
+http = "0.2"
+http-body = "0.4"
+tower-http = { version = "0.3", features = ["trace", "compression-gzip"] }
```
chesedo, line 19: Are all of these used? Like `serde` and `http-body`?

service/src/lib.rs
```diff
@@ -444,6 +447,54 @@ where
...
+
+        Ok(())
+    }
+
+    fn bind(&mut self, addr: SocketAddr) -> Result<ServeHandle, error::Error> {
+        let service = self.service.take().expect("service has already been bound");
+
+        let future = async move {
+            let shared = tower::make::Shared::new(service);
+            hyper::Server::bind(&addr).serve(shared).await
```
chesedo, line 485: Can't this code move to the `spawn`?

.github/workflows/fmt.yml
```diff
@@ -19,6 +19,10 @@ jobs:
           components: rustfmt
       - run: cargo fmt --all -- --check
   clippy_test:
+    strategy:
+      fail-fast: true
+      matrix:
+        features: ["web-axum", "web-rocket", "web-tide", "web-tower"]
```
chesedo, line 25: Are the other features like `sqlx-postgres` still running through clippy somehow?

chesedo (commented):
Looks great Mario! Just have a few questions

service/src/lib.rs
marioidival: fixed

examples/tower/hello-world/Cargo.toml
marioidival: fixed

.github/workflows/fmt.yml
chesedo: For future maintainability, I'm a bit worried because we are losing the clippy checks on the other features like `sqlx-postgres` and `loader`. Maybe we should add an extra clippy check which does all features except the frameworks?

.github/workflows/fmt.yml
```diff
@@ -27,4 +31,5 @@ jobs:
           toolchain: stable
           override: true
           components: clippy
-      - run: cargo clippy --tests --all-targets --all-features -- --D warnings
+      - name: Run test with ${{ matrix.features}}
+        run: cargo clippy --tests --all-targets --features=${{ matrix.features }} -- --D warnings --no-deps
```
chesedo, line 35:
```suggestion
        run: cargo clippy --tests --all-targets --features=${{ matrix.features }} --no-deps -- --D warnings
```
I might be wrong, but think `--no-deps` goes on the clippy side

chesedo (changes requested):

.github/workflows/fmt.yml
marioidival: @chesedo we can create two steps to clippy check. One of web with this matrix and another one for the others.

chesedo (approved):

PR #161, opened 2022-05-05
refactor: shutdown

chesedo:
Allow services to be shutdown by blocking less

Makes #123 obsolete


NodarD:
Task linked: [CU-2344pwa Deployments don't get killed](https://app.clickup.com/t/2344pwa)

api/src/deployment.rs
```diff
@@ -485,6 +485,10 @@ impl DeploymentSystem {
                 let mut lock = deployment.state.write().await;
                 if let DeploymentState::Deployed(DeployedState { so, handle, .. }) = lock.take() {
                     handle.abort();
+                    handle
+                        .await
+                        .map_err(|e| DeploymentApiError::Internal(e.to_string()))?
```
christoshadjiaslanis, line 490: If you abort and then `map_err(...)?` the `JoinHandle` will result in an `Err(JoinError)` since the task is cancelled no?

[Awaiting a [cancelled](https://docs.rs/tokio/latest/tokio/task/struct.JoinError.html#method.is_cancelled) task might complete as usual if the task was already completed at the time it was cancelled, but most likely it will fail with a cancelled JoinError.](https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html#impl).

This is fine but the issue is the `?` which will bubble the error up and return an API error. Am I missing something here?

christoshadjiaslanis: You also need to apply this diff in the imports (line 18):

```
-use shuttle_service::loader::{Loader, ServeHandle};
+use shuttle_service::loader::Loader;
+use shuttle_service::ServeHandle;
```

otherwise the code doesn't compile

service/src/lib.rs
```diff
@@ -289,7 +293,7 @@ pub trait Service: Send + Sync {
     /// This function is run exactly once on each instance of a deployment.
     ///
     /// The deployer expects this instance of [Service][Service] to bind to the passed [SocketAddr][SocketAddr].
-    fn bind(&mut self, addr: SocketAddr) -> Result<(), error::Error>;
+    fn bind(&mut self, addr: SocketAddr) -> Result<ServeHandle, error::Error>;
```
christoshadjiaslanis, line 296: Probably need to bump the version since the API is breaking.

christoshadjiaslanis (changes requested):

api/src/deployment.rs
chesedo: Good catch, I missed the `JoinError::Cancelled`

chesedo: Good catch, I missed the `JoinError::Cancelled`

service/src/lib.rs
chesedo: I can, but I'm breaking it again in the PR that follows this one. Should we rather only bump once?

christoshadjiaslanis: Fair enough - I'm not sure to be honest but we need a consistent strategy. Tbf we have been consistent in bumping with distinct commits for that purpose, so let's stick to that for now I guess :)

christoshadjiaslanis (approved):

PR #162, opened 2022-05-07
ci: fix formatting checks

### LINKED ISSUE #98: Add rustfmt.toml
My VSCode kept re-formatting the code on Ctrl+Shift+I creating huge diffs. All contributors should share the same formatting rules.

See https://rust-lang.github.io/rustfmt/

The alternative is to accept that the diffs will contain formatting changes potentially reverting back and forth depending on who's got what version of `rustfmt` installed.

WiredSound:
Some of the options specified in the current `rustfmt.toml` do not work on the current stable toolchain which the CI format is set to use. Changing to nightly for the format check means that the check can now actually pass with proper formatting.

This does raise the question though as to whether it may be more sensible to just remove the nightly options in `rustfmt.toml` to ensure no weird comparability issues that may arise through mixing and matching use of stable vs nightly compiler.

Closes #98


chesedo:
My local dev for shuttle uses stable. So personally, I would prefer the formater to also use stable to have less of an inconvenience. :man_shrugging:

PR #166, opened 2022-05-10
fix: failing cargo shuttle deploy returns successful exit code

### LINKED ISSUE #138: failing cargo shuttle deploy returns successful exit code
if I forget to do the lib.crate-type = ["cdylib"] dance, `cargo shuttle deploy` gives me:
```
   Compiling reqwest v0.11.9
   Compiling dancelist v0.1.0 (/opt/unveil/crates/dancelist)
    Finished dev [unoptimized + debuginfo] target(s) in 1m 37s

        Project:            dancelist
        Deployment Id:      d3a0e264-f785-42d5-bf85-0d39aba5fb36
        Deployment Status:  ERROR: "a cdylib was not created"
        Host:               dancelist.shuttleapp.rs
        Created At:         2022-04-15 06:38:17.692974962 UTC
```
but `cargo-shuttle`'s exit code is still "success" ($? == 0) in this case :-( . This means that bash's && and || chains do the wrong thing. This also means you can't build useful CI pipelines, etc.

_Originally posted by @alsuren in https://github.com/qwandor/dancelist/issues/2#issuecomment-1099906351_

bmoxb:
`cargo shuttle deploy` and `cargo shuttle status` will now return shell error/failure code (i.e. non-zero value) should the deployment in question be in the 'error' state.

Closes #138


cargo-shuttle/src/main.rs
```diff
@@ -51,8 +61,8 @@ impl Shuttle {
             Command::Deploy(deploy_args) => self.deploy(deploy_args).await,
             Command::Status => self.status().await,
             Command::Delete => self.delete().await,
-            Command::Auth(auth_args) => self.auth(auth_args).await,
-            Command::Login(login_args) => self.login(login_args).await,
+            Command::Auth(auth_args) => self.auth(auth_args).await.map(|_| DeploymentStateMeta::Deployed),
+            Command::Login(login_args) => self.login(login_args).await.map(|_| DeploymentStateMeta::Deployed),
```
chesedo, line 65: Maybe we should go for something like `CommandStatus` to prevent this unexpected tagging on auth and login?

chesedo: Also, why can't we just use `Result::Err`?

chesedo (commented):
Nice, thanks for this @bmoxb! I'm wondering how we can make this cleaner :thinking:

PR #168, opened 2022-05-12
fix: API crash on panic occurring in a `Service` implementation

bmoxb:
Internal ticket: https://app.clickup.com/t/290zt1h

* Prevent any panics in a deployment's `Service` implementation's `build` or `bind` functions from crashing the entire API.
* Add checks that ensure a deployment's `Cargo.toml` does not specify aborting panics (which would again result in API crash regardless of above).
* Note that malicious users can still crash the API through non-`panic!` methods - this is considered outside the scope of this PR.


service/tests/resources/bind-panic/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+crate-type = ["cdylib"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[workspace]
+
+[dependencies]
+shuttle-service = { path = "../../../" }
+async-trait = "0.1.56"
+log = "0.4.17"
```
chesedo, line 16:
```suggestion
```
Code does not seem to use `log`

service/tests/resources/build-panic/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+crate-type = ["cdylib"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[workspace]
+
+[dependencies]
+shuttle-service = { path = "../../../" }
+async-trait = "0.1.56"
+log = "0.4.17"
```
chesedo, line 16:
```suggestion
```

service/src/loader.rs
```diff
@@ -66,12 +74,35 @@ impl Loader {
         let mut service = self.service;
         let logger = Box::new(Logger::new(tx, deployment_id));
 
-        service.build(factory, logger).await?;
+        AssertUnwindSafe(service.build(factory, logger))
+            .catch_unwind()
+            .await
+            .unwrap_or(Err(Error::BuildPanic))?;
```
chesedo, line 80: Is it possible to also get the panic message with `map_err`?

service/src/loader.rs
```diff
@@ -66,12 +74,35 @@ impl Loader {
...
-        Ok((handle, self.so))
+        // Start service on this side of the FFI
+        let handle = tokio::spawn(async move {
+            let bound = AssertUnwindSafe(async { service.bind(addr) })
+                .catch_unwind()
+                .await;
+
+            send.send(bound.is_ok()).unwrap();
+
+            if let Ok(b) = bound {
```
chesedo, line 94: Maybe we can get the panic message here too by using `match`?

service/src/loader.rs
```diff
@@ -66,12 +74,35 @@ impl Loader {
...
+                b?.await?
+            } else {
+                Err(anyhow!("panic in `Service::bound`"))
+            }
+        });
+
+        if matches!(recv.await, Ok(true)) {
```
chesedo, line 101: Doesn't this block the `load()` function? Ie, the code in deployment.rs will `await` until this oneshot resolves?

chesedo: The implication being, does this block a second deploy?

chesedo (commented):
Nicely done @bmoxb! I mostly have questions :smile:

service/src/loader.rs
chesedo: Ok, if `load()` returns immediately then I think this will be fine

service/src/loader.rs
```diff
@@ -36,6 +40,11 @@ pub struct Loader {
 }
 
 impl Loader {
+    #[allow(dead_code)]
+    pub fn new(service: Box<dyn Service>, so: Library) -> Self {
+        Loader { service, so }
+    }
+
```
chesedo, line 47:
```suggestion
```
This does not appear to be used :man_shrugging:

service/src/loader.rs
```diff
@@ -66,12 +75,42 @@ impl Loader {
         let mut service = self.service;
         let logger = Box::new(Logger::new(tx, deployment_id));
 
-        service.build(factory, logger).await?;
+        fn map_any_to_panic_string(a: &dyn Any) -> String {
+            a.downcast_ref::<&str>()
+                .map(|x| x.to_string())
+                .unwrap_or_else(|| "<no panic message>".to_string())
+        }
```
chesedo, line 82: Maybe we should not go with nested functions, but rather put this at the bottom of this file?

chesedo (commented):
Just a small refactor :sweat_smile:

service/src/loader.rs
bmoxb: Ah, my bad, a holdover from earlier testing :)

PR #167, opened 2022-05-10
refactor: conditional db

chesedo:
Only create a DB when needed

Waits on #161

PR #180, opened 2022-05-18
feat: AWS RDS

chesedo:
Adds support for Postgres, Mysql and MariaDB AWS RDS

~~Needs #167~~

## Breaking change
All old deploys that use the `PgPool` needs to add attribute configuration to explicitly ask for a shared Postgres resource as follow:
```rust
#[shuttle_service::main]
async fn rocket(#[shared::Postgres] pool: PgPool) -> Result<Rocket<Build>, shuttle_service::Error> {
    //          ^^^^^^^^^^^^^^^^^^^
    //                   |
    //                   This is the part to add

    // Business as usual here
}
```


bmoxb (approved):
All looks cool to me, the `#[shared::Postgres]`, etc. proc-macro annotations are very funky :)

PR #181, opened 2022-05-19
Feature/add native deps libtorch

### LINKED ISSUE #135: feature: specify native dependencies to deployment runtime
Some deployments may require native dependencies such as (e.g. `libtorch`). 

There needs to be a mechanism to specify these native dependencies in the project source and them being provided during deployment.

marioidival:
Task linked: [CU-26gu8gh As a user I would like to be able to specify and add native dependencies (say libtorch)](https://app.clickup.com/t/26gu8gh)

closes #135


NodarD:
Task linked: [CU-26gu8gh As a user I would like to be able to specify and add native dependencies (say libtorch)](https://app.clickup.com/t/26gu8gh)

Dockerfile.dev
```diff
@@ -4,6 +4,14 @@ RUN apt-get update &&\
 RUN pg_dropcluster $(pg_lsclusters -h | cut -d' ' -f-2 | head -n1)
 RUN rustup component add rust-src
 
+WORKDIR /opt
+ENV LIBTORCH_URL=https://download.pytorch.org/libtorch/cpu/libtorch-cxx11-abi-shared-with-deps-1.11.0%2Bcpu.zip
```
chesedo, line 8: I can see this is working locally. Since our prod server runs `arm64` I am wondering if this download is for `x86_64`? Maybe we should try install torch in the `apt install` section?

chesedo: The first bullet point under [Getting Started](https://github.com/LaurentMazare/tch-rs#getting-started) also says it will use the system-wide default :man_shrugging: 
Alternatively, we just have to figure a workaround for a `arm64` download (if that is the problem)

PR #186, opened 2022-05-26
feat: simple local run

chesedo:
This is the first steps to getting local runs working using the following command:
```
cargo shuttle run
```

Currently services relaying on a DB is not supported


NodarD:
Task linked: [CU-2cta5rn Simple `cargo shuttle run` (ignore factory for now)](https://app.clickup.com/t/2cta5rn)

cargo-shuttle/src/factory.rs
```diff
@@ -0,0 +1,11 @@
+use async_trait::async_trait;
+use shuttle_service::Factory;
+
+pub struct LocalFactory {}
+
+#[async_trait]
+impl Factory for LocalFactory {
+    async fn get_sql_connection_string(&mut self) -> Result<String, shuttle_service::Error> {
+        todo!()
```
brokad, line 9: If we want to break this down into two PRs for velocity, we could just `Err` out here and say services with database dependencies are not (yet) supported by `cargo shuttle run` - try deploying instead!

brokad (approved):
LGTM! And thanks for cleaning up `cargo-shuttle` here and there :smile:

Issue #175, opened 2022-05-16
feature: cargo shuttle init


christoshadjiaslanis:
Introduce an `init` function to `cargo-shuttle` that will initialise the barebones of a project (similar to `cargo init`). `cargo shuttle init <name>`

This should probably:
1. Add an `Init(ProjectName)` variant to `Command` in `args.rs` 
2. Call `cargo init --lib` via the `cargo` crate
3. Automatically add the `crate-type = ["cdylib"] ` to `cargo.toml`
4. Add the `shuttle-service` deps top `cargo.toml`.

Bonus points - have a separate `cargo init` command for each framework supported. For example with Rocket:
1. `cargo shuttle init-rocket <name>`
2.  As above, but also add the feature flags: `shuttle-service = { version = "0.2", features = ["web-rocket"] }`
3.  Also create `hello-world` scaffolding
```rust
#[macro_use]
extern crate rocket;

use rocket::{Build, Rocket};

#[get("/")]
fn index() -> &'static str {
    "Hello, world!"
}

#[shuttle_service::main]
async fn rocket() -> Result<Rocket<Build>, shuttle_service::Error> {
    Ok(rocket::build().mount("/hello", routes![index]))
}
```

nahuakang:
I'd be happy to claim this if no one else is working on it yet and it's not high priority time-wise on shuttle's roadmap :)

brokad:
> I'd be happy to claim this if no one else is working on it yet and it's not high priority time-wise on shuttle's roadmap :)

Thanks @nahuakang!

> Edit: After asking around and doing some investigation, it seems that cargo init [writes naively to files](https://github.com/rust-lang/cargo/blob/39ad1039d9e3e1746177bf5d134af4c164f95528/src/cargo/ops/cargo_new.rs#L766-L793) and we could use [toml-edit](https://crates.io/crates/toml-edit) for modifying an existing Cargo.toml file.

Sounds like a plan!

PR #192, opened 2022-05-27
Issue 175: `cargo shuttle init` (without bonus)

nahuakang:
## Motivation

This PR adds `cargo shuttle init` feature to #175. This is a **partial solution** to the feature as it does not include separate `cargo shuttle init` commands for each supported framework, e.g. `axum` and `rocket`.

I hope to continue on these separate commands on a separate PR and then start refactoring `Shuttle.init` method inside `cargo-shuttle/src/main.rs` as I think we could extract out some logic and add unit tests to the operations on `Cargo.toml`.

## Changes

1. Add an `Init(InitArgs)` variant to `Command` in `args.rs`
2. Call `cargo init --lib` via the `cargo` crate
3. Automatically add the `crate-type = ["cdylib"]` to `Cargo.toml`
4. Add the `shuttle-service` dependency to `Cargo.toml`

## How it works

- Run `cargo shuttle init` initializes `Cargo.toml` and `src/` in the current directory (similar to `cargo init`)
- Run `cargo shuttle init newproject` initializes in `./newproject` directory; if the directory `newproject` does not exist, it will be created
- Run `cargo shuttle init /full/path/to/newproject` initializes `Cargo.toml` and `src/` in the given path; if any directory does not exist, it will be recursively created

## To-Dos

- Update documentation?


cargo-shuttle/src/args.rs
```diff
@@ -46,25 +46,22 @@ pub struct ProjectArgs {
...
     Deploy(DeployArgs),
-    #[structopt(about = "view the status of a shuttle project")]
-    Status,
-    #[structopt(about = "view the logs of a shuttle project")]
-    Logs,
     #[structopt(about = "delete the latest deployment for a shuttle project")]
     Delete,
-    #[structopt(about = "create user credentials for the shuttle platform")]
-    Auth(AuthArgs),
+    #[structopt(about = "create a new shuttle project in an existing directory")]
```
chesedo, line 57:
```suggestion
    #[structopt(about = "create a new shuttle project")]
```
From the PR description I understand it will create the directory when it does not exist yet :sweat_smile:

cargo-shuttle/src/args.rs
```diff
@@ -84,3 +81,26 @@ pub struct DeployArgs {
     #[structopt(long, about = "allow dirty working directories to be packaged")]
     pub allow_dirty: bool,
 }
+
+#[derive(StructOpt)]
+pub struct InitArgs {
+    #[structopt(
+        about = "the path to initialize",
```
chesedo, line 88:
```suggestion
        about = "the path to initialize a new shuttle project",
```

cargo-shuttle/src/main.rs
```diff
@@ -50,12 +52,13 @@ impl Shuttle {
...
-            Command::Status => self.status().await,
-            Command::Logs => self.logs().await,
-            Command::Delete => self.delete().await,
             Command::Auth(auth_args) => self.auth(auth_args).await,
+            Command::Delete => self.delete().await,
+            Command::Deploy(deploy_args) => self.deploy(deploy_args).await,
+            Command::Init(init_args) => self.init(init_args).await,
             Command::Login(login_args) => self.login(login_args).await,
+            Command::Logs => self.logs().await,
+            Command::Status => self.status().await,
```
chesedo, line 61: The ordering is nice. Is there a reason you decided to order them other than for readability?
I'm thinking, since Rust will match them in order, maybe we would rather want the most popular commands to be first?

cargo-shuttle/src/main.rs
```diff
@@ -106,6 +109,38 @@ impl Shuttle {
         .context("failed to delete deployment")
     }
 
+    async fn init(&self, args: InitArgs) -> Result<()> {
+        // Call `cargo init --lib` to generate files
+        let path = args.path.to_str().unwrap();
+        process::Command::new("cargo")
+            .args(["init", "--lib", path])
+            .spawn()?
+            .wait()?;
```
chesedo, line 118: I know the cargo API documentation is a bit lacking, but maybe we can interface with it directly?
https://docs.rs/cargo/latest/cargo/ops/fn.init.html

cargo-shuttle/src/main.rs
```diff
@@ -106,6 +109,38 @@ impl Shuttle {
...
+
+        // Insert `crate-type = ["cdylib"]` array into `[lib]` table
+        let crate_type_array = Array::from_iter(["cdylib"].into_iter());
+        let mut lib_table = Table::new();
+        lib_table["crate-type"] = Item::Value(Value::Array(crate_type_array));
+        cargo_doc["lib"] = Item::Table(lib_table);
+
+        // Insert shuttle-service to `[dependencies]` table
+        let mut dep_table = Table::new();
+        dep_table["shuttle-service"]["version"] = value(env!("CARGO_PKG_VERSION"));
```
chesedo, line 134: I might be wrong, but won't this get the package version of `cargo-shuttle` while we want the package version of `shuttle-service`?

Maybe we can add a constant to `shuttle-service` which gets its value from `CARGO_PKG_VERSION` and then use that constant here.

chesedo (commented):
This is looking real good. Thanks @nahuakang !

cargo-shuttle/src/main.rs
chesedo: Almost forgot. Here is an example of us interacting with `cargo` directly
https://github.com/shuttle-hq/shuttle/blob/main/api/src/build.rs#L213

cargo-shuttle/src/main.rs
nahuakang: That makes sense, too. I was more thinking about alphabetical order in case one wants to find a command when one doesn't know a lot about the commands. However, I think you're probably right that, if the coder is familiar with everything, order by popularity is better :)

nahuakang:
@chesedo I opted to use `cargo-edit` on version `"0.8.0"` for fetching the latest `shuttle-service` version (could use `cargo` to do the same but it'd add quite a bit more code since we'd fetch all `Summary` instances of `shuttle-service` and then manually rank them to extract the latest one).

However, I cannot use `cargo-edit = "0.9.0"` due to a dependency conflict from `cargo` (not sure if we care?):

```sh
Caused by:
  failed to select a version for `libgit2-sys`.
      ... required by package `cargo v0.59.0`
      ... which satisfies dependency `cargo = "^0.59.0"` of package `cargo-shuttle v0.3.0 (/Users/nahua/projects/shuttle-hq/shuttle/cargo-shuttle)`
  versions that meet the requirements `^0.12.24` are: 0.12.26+1.3.0, 0.12.25+1.3.0, 0.12.24+1.3.0

  the package `libgit2-sys` links to the native library `git2`, but it conflicts with a previous package which links to `git2` as well:
  package `libgit2-sys v0.13.0+1.4.1`
      ... which satisfies dependency `libgit2-sys = "^0.13.0"` of package `git2 v0.14.0`
      ... which satisfies dependency `git2 = "^0.14"` of package `cargo-edit v0.9.0`
      ... which satisfies dependency `cargo-edit = "^0.9.0"` of package `cargo-shuttle v0.3.0 (/Users/nahua/projects/shuttle-hq/shuttle/cargo-shuttle)`
  Only one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. Try to adjust your dependencies so that only one package uses the links ='libgit2-sys' value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.

  failed to select a version for `libgit2-sys` which could resolve this conflict
```

cargo-shuttle/src/main.rs
```diff
@@ -106,6 +109,53 @@ impl Shuttle {
...
+
+        // Mimick `cargo init` behavior and log status or error to shell
+        match init_result {
+            Ok(project_kind) => {
+                cargo_config
+                    .shell()
+                    .status("Created", format!("{} (shuttle) package", project_kind))?;
+            }
+            Err(e) => return Err(e),
+        }
```
chesedo, line 126:
```suggestion
        let init_result = cargo::ops::init(&opts, &cargo_config)?;

        // Mimick `cargo init` behavior and log status or error to shell
        cargo_config
            .shell()
            .status("Created", format!("{} (shuttle) package", init_result))?;
```

chesedo (approved):
I'm all happy with this @nahuakang.

If updating to cargo v0.62 is easy and fixes the update for `cargo-edit`, then feel free to commit it on this PR. Else I'm happy to merge with the older version of `cargo-edit`

chesedo (approved):
Awesome, thanks for this @nahuakang!

Issue #120, opened 2022-04-06
feature: cargo shuttle run


christos-h:
This is a proposal to be able to run a `shuttle` app locally without needing to deploy.

0xkelvin:
hi, i am going to take this issue.

brokad:
Closing as done, except where in favour of #497 and #496.

PR #196, opened 2022-05-31
feat: local run with DB

chesedo:
Adds DB support to local run

For #120


NodarD:
Task linked: [CU-2cta64h Local runs creates all DB as docker containers](https://app.clickup.com/t/2cta64h)

cargo-shuttle/src/factory.rs
```diff
@@ -1,14 +1,231 @@
...
+            ..Default::default()
+        });
+        let mut output = self.docker.create_image(create_image_options, None, None);
+
+        while let Some(line) = output.next().await {
+            let info = line.expect("failed to create image");
+
+            if let Some(id) = info.id.as_ref() {
+                match layers.iter_mut().find(|item: &&mut CreateImageInfo| {
+                    item.id.is_some() && item.id.as_deref().unwrap() == id
```
bmoxb, line 179: I think you can do a direct comparison instead of unwrapping here.

```suggestion
                    item.id.as_deref() == Some(id)
```

bmoxb (changes requested):
Super cool :) One minor problem I've encountered however is when running an example which uses a Postgres database, I get an 'address already in use' error if I have an instance of Postgres already running locally on my computer. Perhaps it would be a good idea to run Postgres on a non-standard port to avoid this?

bmoxb (approved):

PR #199, opened 2022-06-03
feat: provisioner

chesedo:
A microservice for handling the provisioning of all resources


bmoxb (approved):
LGTM :) The use of `format!` in the CREATE ROLE queries unnerves me slightly still but not really a big concern for now I reckon.

chesedo:
> The use of `format!` in the CREATE ROLE queries unnerves me slightly still but not really a big concern for now I reckon.

I hear you big time! :worried: Hopefully the injection test will help us see when it becomes injectable :sweat_smile:

Issue #197, opened 2022-06-01
bug: calling `get_secrets` in setup code fails


chesedo:
Currently calling `get_secrets` works from routes. But calling it in the `#[shuttle_service::main]` causes the secret to not be found.

It seems we are not setting the secrets up early enough.

https://discord.com/channels/803236282088161321/965974106652688394/981436093683613776

PR #201, opened 2022-06-03
fix: calling `get_secrets` in setup code fails

bmoxb:
https://github.com/shuttle-hq/shuttle/issues/197

**TODO: Handle in `Loader::load` instead of codegen** - need to work out Tokio runtime problems.

* The set of secrets read from `Secrets.toml` are now sent as a HTTP header instead of in a separate request.
* That set of initial secrets is then set immediately before the user's setup function is called.

### Considerations

* Is passing the received set of secrets through the full deployment pipeline all the way to the service build function the best approach? Is there a simpler way to achieve the say outcome?
* Secrets from `Secrets.toml` are now sent as a HTTP header with JSON encoding - this should not result in an invalid HTTP request but may need to test some edge cases (e.g., weird unicode characters). If it is a problem, then a solution could be base64 encoding.

### DIFF
--- a/api/src/deployment.rs
+++ b/api/src/deployment.rs
@@ -12,7 +12,7 @@ use futures::prelude::*;
 use libloading::Library;
 use rocket::data::ByteUnit;
 use rocket::{tokio, Data};
-use shuttle_common::project::ProjectName;
+use shuttle_common::project::{ProjectName, InitialSecrets};
 use shuttle_common::{
     DeploymentApiError, DeploymentId, DeploymentMeta, DeploymentStateMeta, Host, LogItem, Port,
 };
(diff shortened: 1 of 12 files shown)


api/src/deployment.rs
```diff
@@ -49,10 +49,15 @@ impl Deployment {
         }
     }
 
-    fn from_bytes(fqdn: &str, project: ProjectName, crate_bytes: Vec<u8>) -> Self {
+    fn from_bytes(
+        fqdn: &str,
+        project: ProjectName,
+        crate_bytes: Vec<u8>,
+        initial_secrets: InitialSecrets,
```
chesedo, line 56: Won't the `Secrets.toml` file also be in `crate_bytes`? Could we possibly read them from there on the `api` end?

PR #202, opened 2022-06-06
feature: check shuttle service version before deploy

marioidival:
Approach:

The idea is given to all users an endpoint to get the current version of shuttle-service used by the API, then the clients just can make a call on this endpoint and see the current version.


cargo-shuttle/src/client.rs
```diff
@@ -69,6 +69,32 @@ pub(crate) async fn status(api_url: ApiUrl, api_key: &ApiKey, project: &ProjectN
     Ok(())
 }
 
+pub(crate) async fn version(api_url: ApiUrl) -> Result<String> {
+    let client = get_retry_client();
+    let mut api_url = api_url;
```
chesedo, line 74:
```suggestion
pub(crate) async fn version(mut api_url: ApiUrl) -> Result<String> {
    let client = get_retry_client();
```
I sometimes get mixed up with `mut` too. Also, only think this works, but I can be wrong too :see_no_evil:

chesedo: Also, maybe we should call this function something like `service_version` to prevent it from being confused with the client version?

cargo-shuttle/src/client.rs
```diff
@@ -69,6 +69,32 @@ pub(crate) async fn status(api_url: ApiUrl, api_key: &ApiKey, project: &ProjectN
...
+
+    if response_status == StatusCode::OK {
+        return Ok(response_text);
+    }
+
+    Err(anyhow!(
+        "status: {}, body: {}",
+        response_status,
+        response_text
+    ))
```
chesedo, line 95: Can we put the `Err` in an `else` branch to be able to remove the `return` in the `if`?

chesedo: Maybe we can move the `await` on line 85 to inside the if too?

cargo-shuttle/src/lib.rs
```diff
@@ -210,6 +213,53 @@ impl Shuttle {
         .context("failed to deploy cargo project")
     }
 
+    async fn check_lib_version(&self) -> Result<()> {
+        let cargo_tree = PCmd::new("cargo")
```
chesedo, line 217: PR #192 uses `toml-edit` to [set the shuttle service version](https://github.com/shuttle-hq/shuttle/pull/192/files#diff-7fea87037f7c8fa590c68cd9e295ce99aa94dca3eab2be9a935dfda0a90cd5d0R111). I wonder if we can use the same library to get the version and make this easier?

chesedo (commented):
@marioidival this is looking great!! I mostly have questions on small things

cargo-shuttle/src/lib.rs
```diff
@@ -210,6 +215,29 @@ impl Shuttle {
...
+    async fn check_lib_version(&self, project_args: ProjectArgs) -> Result<()> {
+        let cargo_path = project_args.working_directory.join("Cargo.toml");
+        let cargo_doc = read_to_string(cargo_path.clone())?.parse::<Document>()?;
+        let current_shuttle_version = &cargo_doc["dependencies"]["shuttle-service"]["version"];
+        let service_semver = Version::parse(current_shuttle_version.as_str().unwrap())?;
+        let mut server_version = String::new();
+
+        client::shuttle_version(self.ctx.api_url())
+            .await
+            .map(|response| server_version = response)?;
```
chesedo, line 227:
```suggestion
        let server_version =  client::shuttle_version(self.ctx.api_url()).await?;
```
Hopefully I understand this correctly :see_no_evil:

cargo-shuttle/src/lib.rs
```diff
@@ -210,6 +215,29 @@ impl Shuttle {
...
+
+        let server_semver = VersionReq::parse(&server_version)?;
+
+        if server_semver.matches(&service_semver) {
```
chesedo, line 231: Does this also match on something like `v0.3.2` and `v0.3.43`?

cargo-shuttle/src/lib.rs
```diff
@@ -210,6 +215,29 @@ impl Shuttle {
...
+            Ok(())
+        } else {
+            Err(anyhow!(
+                "Update your shuttle-version to {}",
```
chesedo, line 235:
```suggestion
                "Your shuttle_service version is outdated. Update your shuttle_service version to {} and try to deploy again",
```

chesedo (commented):
Just one question on how the matching works

cargo-shuttle/src/lib.rs
marioidival: thanks!

cargo-shuttle/src/lib.rs
marioidival: thanks!

chesedo (approved):
This is looking awesome, thanks @marioidival

PR #203, opened 2022-06-07
feat: locate root dir

### LINKED ISSUE #152: bug: running `cargo shuttle status` outside a cargo project panics
The error message looks like this:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: CargoMetadata { stderr: "error: could not find `Cargo.toml` in `/path/to/folder` or any parent directory\n" }', /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-shuttle-0.2.5/src/config.rs:224:10
```

### LINKED ISSUE #163: cargo shuttle ignores Shuttle.toml if run in subdir of crate
The logic to find Cargo.toml works fine, but we don't then use the location of Cargo.toml to search for a Shuttle.toml (and will continue as if it's not there). This means that it will have the wrong project name.

I wrote a test for this in https://github.com/shuttle-hq/shuttle/pull/122 to help when implementing this. See https://github.com/shuttle-hq/shuttle/pull/122#discussion_r866546185_

It's very much an edge case, so might not be worth fixing.

### LINKED ISSUE #195: feature: `cargo-shuttle` commands should work from subdirectory of project root
Certain `cargo-shuttle` commands (like `deploy`) don't work if your `PWD` is not the project root. 

We would want commands like `cargo shuttle deploy` to work from arbitrary subdirectories of the project root.

ivancernja:
**Feature description**
The command `cargo-shuttle status`, and other `cargo-shuttle` commands that contain the service name, now return the right service name if there's a `Shuttle.toml` file with a custom `name` set inside.

Fixes #195 
Fixes #163 
Closes #152 

**Additional changes**
The test on line 457 has been fixed and the comment has been removed.

**Suggestion**
I'd suggest that once the Shuttle.toml has been found, it prints out to the user something along the lines of;
"A Shuttle.toml has been found. Setting your project name to NEW_SERVICE_NAME"


cargo-shuttle/src/config.rs
```diff
@@ -294,8 +294,26 @@ impl RequestContext {
...
+
+        // Iterates through the directories until it finds the Cargo.toml file (project root) and returns the path
+        pub fn find_root_directory(dir: &Path) -> Option<PathBuf> {
+            for ancestor in dir.ancestors() {
+                if ancestor.join("Cargo.toml").exists() {
+                    return Some(ancestor.to_path_buf());
+                }
+            }
+            None
+        }
```
chesedo, line 309: Maybe this function does not need to be nested inside `get_local_config`? Maybe it can go all the way at the bottom of `lib.rs` (see other comment for why `lib.rs`)

chesedo: It also does not need to be `pub`lic

cargo-shuttle/src/config.rs
```diff
@@ -294,8 +294,26 @@ impl RequestContext {
...
+
+        if let Some(working_directory) = root_directory_path {
+            project_args.working_directory = working_directory;
+        } else {
+            panic!("Shuttle.toml not found.");
```
chesedo, line 314:
```suggestion
            return Err(anyhow!("could not locate the root of a cargo project. Are you inside a cargo project? You can also use `--working-directory` to locate your cargo project."));
```
This should make the error message a bit friendlier.

cargo-shuttle/src/config.rs
```diff
@@ -294,8 +294,26 @@ impl RequestContext {
     }
 
     pub fn get_local_config(
-        project_args: &ProjectArgs,
+        project_args: &mut ProjectArgs
     ) -> Result<Config<LocalConfigManager, ProjectConfig>> {
+        let root_directory_path = find_root_directory(&project_args.working_directory);
```
chesedo, line 299: I see the function (`load_local()`) that calls this function (`get_local_config()`) also uses `working_directory` to read the secrets. So maybe we should move this fix even earlier in the call stack so that the secrets are read correctly too?

I'm thinking of moving it all the way up to `load_project()` in `lib.rs`. This would mean the test will have to move to `lib.rs` and will need to change to test against `load_project()`.

chesedo (changes requested):
Thanks for this @ivancernja! I just made some refactoring suggestions

cargo-shuttle/src/config.rs
ivancernja: Moved to lib.rs, removed public.

cargo-shuttle/src/config.rs
ivancernja: Error message updated.

cargo-shuttle/src/config.rs
ivancernja: I've separated the `find_root_directory()` function and wrote a test for it.

`load_project()` now calls it instead.

ivancernja:
This PR should fix: https://github.com/shuttle-hq/shuttle/issues/152 as well.

cargo-shuttle/src/lib.rs
```diff
@@ -242,6 +260,7 @@ impl Shuttle {
             .run_cargo_package(args.allow_dirty)
             .context("failed to package cargo project")?;
 
+
```
chesedo, line 263: Double new line? `cargo fmt --all` might automatically fix this for you

cargo-shuttle/src/lib.rs
```diff
@@ -339,3 +358,41 @@ impl Shuttle {
...
+    fn find_root_directory_returns_proper_directory() {
+        let working_directory = path_from_workspace_root("examples/axum/hello-world/src");
+        let deep_working_directory = path_from_workspace_root("examples/axum/hello-world/target/debug");
+
+        let root_dir = Shuttle::find_root_directory(&working_directory).unwrap();
+        let deep_root_dir = Shuttle::find_root_directory(&deep_working_directory).unwrap();
+
+        assert_eq!(root_dir, path_from_workspace_root("examples/axum/hello-world/"));
+        assert_eq!(deep_root_dir, path_from_workspace_root("examples/axum/hello-world/"));
+    }
```
chesedo, line 384: Rather make this two tests.

chesedo (commented):
Just a few polish points left. But this is perfect, thanks @ivancernja

cargo-shuttle/src/lib.rs
ivancernja: Sure, will separate them!

re: failing if the `target` does not exist. I assumed that no one would remove the `examples` folder. If someone does that, then the other tests will fail as well since we run the tests against `examples/axum`. And iirc, the examples always come with the `target` folder already inside, no?

cargo-shuttle/src/config.rs
```diff
@@ -273,7 +273,7 @@ impl RequestContext {
     /// Ensures that if `--name` is not specified on the command-line, and either the project
     /// file does not exist, or it has not set the `name` key then the `ProjectConfig` instance
     /// has `ProjectConfig.name = Some("crate-name")`.
-    pub fn load_local(&mut self, project_args: &ProjectArgs) -> Result<()> {
+    pub fn load_local(&mut self, project_args: &mut ProjectArgs) -> Result<()> {
```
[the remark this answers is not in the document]
ivancernja: Hmm. I think they are still needed. Could you elaborate on why they wouldn't be needed?

cargo-shuttle/src/lib.rs
```diff
@@ -122,8 +123,25 @@ impl Shuttle {
...
+            if ancestor.join("Cargo.toml").exists() {
+                return Some(ancestor.to_path_buf());
+            }
+        }
+        None
+    }
+
+    pub fn load_project(&mut self, project_args: &mut ProjectArgs) -> Result<()> {
         trace!("loading project arguments: {project_args:?}");
+        let root_directory_path = Shuttle::find_root_directory(&project_args.working_directory);
```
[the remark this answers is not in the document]
ivancernja: Good point, resolved!

cargo-shuttle/src/lib.rs
ivancernja: VSCode has let me down. Fixed, thanks! :)

cargo-shuttle/src/lib.rs
ivancernja: By the way, the test/variable naming is horrible in that test. Please advise something better, my brain malfunctioned.

```rust

#[test]
    fn find_root_directory_returns_proper_directory() {
        let working_directory = path_from_workspace_root("examples/axum/hello-world/src");

        let root_dir = Shuttle::find_root_directory(&working_directory).unwrap();

        assert_eq!(
            root_dir,
            path_from_workspace_root("examples/axum/hello-world/")
        );
    }

    #[test]
    fn find_root_directory_returns_proper_directory_from_a_deeper_directory() {
        let deep_working_directory =
            path_from_workspace_root("examples/axum/hello-world/target/debug");

        let deep_root_dir = Shuttle::find_root_directory(&deep_working_directory).unwrap();

        assert_eq!(
            deep_root_dir,
            path_from_workspace_root("examples/axum/hello-world/")
        );
    }

```

cargo-shuttle/src/config.rs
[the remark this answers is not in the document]
chesedo: Because nothing below this level changes `project_args` :smile:

cargo-shuttle/src/lib.rs
chesedo: > By the way, the test/variable naming is horrible in that test. Please advise something better, my brain malfunctioned.

I don't mind the variable names. They can even be the same in both tests...

ivancernja: Excerpt from the Discord conversation:
```
@ivancernja : re: target folder not being there for the tests. I can run it against rocket/authentication/src which is deeper than hello_world/src -- do you think that makes sense or should we scratch that test?

@chesedo : Hmm, we can scratch it if there isn't a deeply nested folder in the examples
```

Removing the `deeper` test for now.

cargo-shuttle/src/config.rs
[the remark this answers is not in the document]
ivancernja: @chesedo You are right. I've removed the unnecessary `mut`.

chesedo (approved):
Awesome, thanks again for this @ivancernja

PR #211, opened 2022-06-10
chore: bump automation

chesedo:
Having messed up the release today, I decided to just create scripts to make the future easier....


release.sh
```diff
@@ -0,0 +1,44 @@
...
+
+    cargo set-version --workspace $version
+    git commit -am "chore: v$version"
+}
+
+function update-examples-versions()
+{
+    local version=$1
+
+    rg "shuttle-service = \{ version" --files-with-matches -g '!www/*' | xargs sed -i "s/shuttle-service = { version = \"[[:digit:]]*.[[:digit:]]*.[[:digit:]]*\"/shuttle-service = { version = \"$version\"/g"
```
brokad, line 22: This strongly depends on the formatting of the `shuttle-service` entry in the `Cargo.toml` - would it be potentially less brittle to iterate over `examples/**/Cargo.toml` and run `cargo add shuttle-service@$version`?

publish.sh
```diff
@@ -0,0 +1,37 @@
...
+
+    echo $version | rg "\d+\.\d+\.\d+" || { echo "first argument must be in the form x.y.z"; exit 1; }
+
+    publish-folder "common"
+    publish-folder "codegen"
+    publish-folder "service"
+    publish-folder "cargo-shuttle"
+
+    git tag "v$version"
+    git push --tags
```
brokad, line 32: May want to just `git push v$version` instead of all tags.

brokad:
Can we also maybe move this to a `scripts/`?

scripts/publish.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+#
+# Little script to publish to crates.io
+# Usage: publish.sh x.y.z
+#
+# Dependencies: git, ripgrep
```
chesedo, line 6: I'm a bit lost where `cargo set-version` is used in this script :sweat_smile:

brokad: Haha oops, sorry it's only used in the other one! :smile:

brokad (approved):

Issue #213, opened 2022-06-13
bug: `cargo-shuttle`fails to parse a dependency without a patch version


chesedo:
Our version checker in `cargo-shuttle` fails to parse a `shuttle-service` version when it is missing a patch like the following definition:

```toml
shuttle-service = { version = "0.2", features = ["sqlx-postgres", "web-rocket"] }
```

chesedo:
The quick fix for now is to specify a full version like

```toml
shuttle-service = { version = "0.2.2", features = ["sqlx-postgres", "web-rocket"] }
```

jmwill86:
Is there anything I need to do with this to close it @chesedo? I dont think there's anything more to be done.

chesedo:
@jmwill86 I also don't see anything more to do. Just missed that this can be closed :smile:

PR #214, opened 2022-06-15
feat: check shuttle service version on server side

marioidival:
checking the shuttle service version on server side


api/src/build.rs
```diff
@@ -188,3 +194,23 @@ fn extract_tarball(crate_path: &Path, project_path: &Path) -> Result<()> {
...
+
+fn check_shuttle_version(working_directory: &Path) -> anyhow::Result<()> {
+    let cargo_path = working_directory.join("Cargo.toml");
+    let cargo_doc = read_to_string(cargo_path)?.parse::<Document>()?;
+    let current_shuttle_version = &cargo_doc["dependencies"]["shuttle-service"]["version"];
+    let service_semver = Version::parse(current_shuttle_version.as_str().unwrap())?;
+    let server_version = Version::parse(shuttle_service::VERSION)?;
+
+    let version_required = format!("{}.{}", server_version.major, server_version.minor);
+    let server_semver = VersionReq::parse(&version_required)?;
```
brokad, line 206: Can you maybe parameterise this? That way we can set, from the command-line arguments, the required version for deployment. If the argument is not given, then use a default.

api/src/build.rs
```diff
@@ -188,3 +194,23 @@ fn extract_tarball(crate_path: &Path, project_path: &Path) -> Result<()> {
         Ok(())
     }
 }
+
+fn check_shuttle_version(working_directory: &Path) -> anyhow::Result<()> {
+    let cargo_path = working_directory.join("Cargo.toml");
+    let cargo_doc = read_to_string(cargo_path)?.parse::<Document>()?;
+    let current_shuttle_version = &cargo_doc["dependencies"]["shuttle-service"]["version"];
+    let service_semver = Version::parse(current_shuttle_version.as_str().unwrap())?;
+    let server_version = Version::parse(shuttle_service::VERSION)?;
```
brokad, line 203: Nit:  few of these can be unwrapped instead of handled when they come from from `shuttle-service` and not user input.

api/src/build.rs
```diff
@@ -188,3 +194,23 @@ fn extract_tarball(crate_path: &Path, project_path: &Path) -> Result<()> {
         Ok(())
     }
 }
+
+fn check_shuttle_version(working_directory: &Path) -> anyhow::Result<()> {
+    let cargo_path = working_directory.join("Cargo.toml");
+    let cargo_doc = read_to_string(cargo_path)?.parse::<Document>()?;
+    let current_shuttle_version = &cargo_doc["dependencies"]["shuttle-service"]["version"];
+    let service_semver = Version::parse(current_shuttle_version.as_str().unwrap())?;
```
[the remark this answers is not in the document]
chesedo: This can also link to #213

api/src/args.rs
```diff
@@ -39,6 +39,12 @@ pub struct Args {
         default_value = "5001"
     )]
     pub(crate) provisioner_port: Port,
+    #[structopt(
+        long,
+        about = "MSSV - Minimum supported Shuttle Version",
+        default_value = shuttle_service::VERSION
+    )]
+    pub(crate) shuttle_version: String,
```
brokad, line 47: It seems to me the first time this is validated as a valid `semver::Version` is when it's parsed in the `get_shuttle_service_from_user_crate` function, which would only happen by issuing a user error whereas the error is actually on our end.

Would it not be better to have this `: String` be replaced by `: semver::Version` so that it gets validated there? I believe this should work since `semver::Version` is `FromStr`.

brokad (commented):
LGTM, thank you @marioidival! Just one comment :smile:

api/src/args.rs
marioidival: @brokad it's working! Thanks for that

PR #220, opened 2022-06-21
feat: build incoming services

bmoxb:
[Building of incoming services](https://app.clickup.com/t/2h84xfe)

* [x] Un-archive and build Cargo projects coming in
* [x] Create marker files in preparation of loading
* [x] Delete old build if any
* [x] Unit testing

## Marker Files

The work around that C came up with to avoid the fact that linked libraries were not being properly reloaded was to give each new `.so` file a random name and then use a well-known 'marker file' to indicate the name of the linked library `.so` file.

Looking at the code of `libloading` (dependency that handles `.so` loading for us), the `dlopen` is likely the cause of the issue. The [man pages](https://man7.org/linux/man-pages/man3/dlopen.3.html) state:

> The cache file /etc/ld.so.cache (maintained by [ldconfig(8)](https://man7.org/linux/man-pages/man8/ldconfig.8.html)) is checked to see whether it contains an entry for filename.

I'm trying to figure out an a way of either not add a library to the cache or otherwise ignore it.

## Notes

* Possibility of (minor) merge collisions based on changes to `error` module.
* Introduces new dependencies for un-archiving `.tar.gz` data (`flate2` and `tar`) and for creating random `.so` file names (`rand`).


NodarD:
Task linked: [CU-2h84xfe Building of incoming services](https://app.clickup.com/t/2h84xfe)

deployer/src/deployment/queue.rs
```diff
@@ -79,3 +122,169 @@ impl fmt::Debug for Queued {
...
+    for entry in archive.entries()? {
+        let mut entry = entry?;
+        let path: PathBuf = entry.path()?.components().skip(1).collect();
+        entry.unpack(dest.as_ref().join(path))?;
+    }
+
+    Ok(())
+}
+
+/// Check for an existing marker file is the specified project directory and
```
chesedo, line 141:
```suggestion
/// Check for an existing marker file in the specified project directory and
```

deployer/src/deployment/queue.rs
```diff
@@ -79,3 +122,169 @@ impl fmt::Debug for Queued {
...
+#[cfg(test)]
+mod tests {
+    use std::path::Path;
+    use tokio::fs;
+
+    use super::MARKER_FILE_NAME;
+
+    #[tokio::test]
+    async fn extract_tar_gz_data() {
+        let p = Path::new("/tmp/shuttle-extraction-test");
```
chesedo, line 187: We might want to use this crate to create temp directories instead :man_shrugging: 
https://docs.rs/tempdir

chesedo (approved):
LGTM, getting one step closer :smile:

bmoxb:
Many thanks, @chesedo !

PR #224, opened 2022-06-24
feat: send build logs (deployer)

bmoxb:
Provides an endpoint for both streaming via WebSockets live build logs as Cargo compiles a service, as well as an HTTP GET endpoint for fetching all build logs so far at that point in time.

Potential merge conflicts due to changes to `DeploymentManager::queue_push` and `DeploymentManager::run_push`.

## Locks

* Current implementation uses locks - not desirable! However, the locking is independent of the build/load pipeline so that should not be slowed down.
* Not sure if it would be possible to achieve this without locking as it is necessary to map deployment names after the deployment has already disappeared down the build pipeline.


deployer/src/deployment/build_logs.rs
```diff
@@ -0,0 +1,200 @@
...
+
+struct Deployment {
+    /// A weak pointer to the [`BuildLogSender`] stored in a [`BuildLogWriter`].
+    /// It is a weak pointer so that it is possible to identify when a build
+    /// has failed/ended (and thus close a WebSocket connection sending build
+    /// logs). The dropping of `BuildLogWriter` will result in the closing of
+    /// the build log broadcast channel as this weak pointer cannot keep the
+    /// channel open.
+    sender: Weak<BuildLogSender>,
+    logs_so_far: Arc<RwLock<Vec<String>>>,
```
chesedo, line 85: Shouldn't we be pushing the logs to the log table rather than keeping them in memory?

chesedo (commented):
This looks good to me. Although I might be missing something, so have a question :see_no_evil:

NodarD:
Task linked: [CU-2h84y8b Send build logs](https://app.clickup.com/t/2h84y8b)

deployer/src/deployment/build_logs.rs
bmoxb: Sorry @chesedo , do you have any suggestions of how to approach this with regards to your logging implementation? I'm not failure with `tracing` and am struggling to get it work how I want :sweat_smile:

chesedo: When I was working on the tests for #235, it appeared the build logs are automatically caught by tracing (and therefore persistence). We may just need to edit the parsing of these logs in `deploy_layer.rs` a bit :sweat_smile:

brokad:
Closing in favor of #265.

PR #225, opened 2022-06-24
refactor: service

chesedo:
Make `shuttle_service::Service` easier to implement by users

## Breaking changes
- Axum service no longer use the `SyncWrapper` shim
- `declare_service!` no longer exists. `#[shuttle_service::main]` should fill all needs
- Any custom services need to change as follow:
  -  Remove the `build()` method
  - `bind()` changed to the following
    ```rust
    async fn bind(mut self: Box<Self>, addr: SocketAddr) -> Result<(), shuttle_service::Error>;
    ```
    These changes are:
    - add **`async`**
    - remove the ref (**`&`**) on `self`
    - add **`: Box<Self>`** to `self`
    - change the Ok variant of the return result to **`()`**
  - Remove the `IntoService` impl
  - (Optional) drop the async_trait dependency and use `shuttle_service::async_trait` instead


NodarD:
Task linked: [CU-2ghbe5d Can't return a custom impl of `shuttle_service::Service` from `shuttle::main`](https://app.clickup.com/t/2ghbe5d)

service/src/lib.rs
```diff
@@ -326,40 +326,20 @@ pub trait ResourceBuilder<T> {
...
 /// A tokio handle the service was started on
-pub type ServeHandle = JoinHandle<Result<(), anyhow::Error>>;
+pub type ServeHandle = JoinHandle<Result<(), error::Error>>;
 
 /// The core trait of the shuttle platform. Every crate deployed to shuttle needs to implement this trait.
 ///
-/// Use the [declare_service!][crate::declare_service] macro to expose your implementation to the deployment backend.
+/// Use the [main][main] macro to expose your implementation to the deployment backend.
+///
+/// TODO: our current state machine in the api crate stores this service and can move it across
```
brokad, line 335: TODO will end up in the docs

brokad (approved):
LGTM.

Nit which may help newcomers to this: `StateBuilder<T>` is basically `ServiceBuilder<T: Service>`. I wonder if we should document these internal types to make the crate more readable?

PR #226, opened 2022-06-27
feat: docker-compose deploys

brokad:
This
- Uses `docker-compose` for all deploys
- Uses a common `Containerfile` for building all runtime images
- Adds a `docker-compose` override for (local) development
- Refactors the E2E tests to run against the `docker-compose`d deployment
- Includes instructions for how to set up an local dev environment for working on `shuttle` as well as running the required tests.

## TODOs
- [ ] Required changes to the TF files
- [ ] Required changes to the CI


e2e/tests/integration/helpers/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+            .unwrap()
+            .to_path_buf()
+    };
+    static ref DOCKER: PathBuf = which::which("docker").unwrap();
+    static ref DOCKER_COMPOSE: PathBuf = which::which("docker-compose").unwrap();
+    static ref CARGO: PathBuf = which::which("cargo").unwrap();
+    static ref LOCAL_UP: () = {
+        let docker_bake = WORKSPACE_ROOT.join("docker-bake.hcl");
+        let docker_bake_override = WORKSPACE_ROOT.join("e2e/docker-bake-override.hcl");
+        let docker_compose = WORKSPACE_ROOT.join("docker-compose.dev.yml");
```
chesedo, line 27: How do the containers shutdown after the test?

chesedo (commented):
Can't wait for this :tada: Just one concern for cleaning up after the tests

e2e/tests/integration/helpers/mod.rs
brokad: They don't, they stay up to make the next one (potentially) faster. But I can run a quick clean up `docker compose down` if this sounds like a better idea?

chesedo: I see... hmm, I guess we will have to rebuild and restart them somehow when the API or provisioner code changes

chesedo: Maybe e2e should also not run with every PR but only version branches :man_shrugging:

brokad: > 

That should not be necessary, as images will then be rebuilt and `docker-compose` will redeploy the services with the new images.

brokad: > Maybe e2e should also not run with every PR but only version branches

That sounds a bit scary to me for now as our test suite for each service or API element is not super exhaustive, we could end up with frequent merges to main that don't actually work and we would only notice it when we do a release.

Maybe we could break down the test CI action so we have two jobs (and two ticks), that way we can get quicker feedback from unit tests and still rely on the E2E tick before we merge?

PR #230, opened 2022-06-30
feat: store '.so' file names in database instead of using marker files

bmoxb:
* Compiled '.so' files are still given random names but a service deployment can be mapped to a '.so' file name via a new `libs` database table.
* All '.so' files are moved to a common directory instead of being left in the project directory.

### DIFF
--- a/deployer/src/deployment/queue.rs
+++ b/deployer/src/deployment/queue.rs
@@ -84,11 +84,11 @@ impl Queued {
 
         info!("Removing old build (if present)");
 
-        remove_old_build(&persistence, &self.name).await?;
+        remove_old_lib(&persistence, &self.name, LIBS_PATH).await?;
 
         info!("Moving built library and storing its location in the database");
 
-        store_build(&persistence, &self.name, so_path).await?;
+        store_lib(&persistence, &self.name, so_path, LIBS_PATH).await?;
 
         let built = Built { name: self.name };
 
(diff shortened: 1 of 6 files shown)


deployer/src/deployment/queue.rs
```diff
@@ -117,53 +117,52 @@ fn extract_tar_gz_data(data: impl Read, dest: impl AsRef<Path>) -> Result<()> {
...
+async fn remove_old_lib(
+    persistence: &Persistence,
+    name: &str,
+    libs_path: impl AsRef<Path>,
+) -> Result<()> {
+    if let Some(lib_name) = persistence.get_last_built_lib(name).await? {
+        let lib_path = libs_path.as_ref().join(lib_name);
+
+        if lib_path.exists() {
+            fs::remove_file(lib_path).await?;
```
chesedo, line 131: I'm wondering, what will happen when a new deployment version fails to start during run. We will want to return to a previous instance, right?

Maybe we shouldn't delete old 'so' files?

deployer/src/persistence.rs
```diff
@@ -45,13 +45,18 @@ impl Persistence {
                 state INTEGER          -- Enum indicating the current state of the deployment.
             );
 
+            CREATE TABLE IF NOT EXISTS libs (
+                name TEXT PRIMARY KEY, -- Name of the service.
+                lib TEXT               -- The name of linked library produced the last time the service was built.
+            );
```
chesedo, line 51: Maybe we can remove this table if we give each incoming deployment an id and store the id in the deployments table? Then we can use the same id when renaming the 'so' file?

chesedo (commented):
Oh nice!! This will be easier

deployer/src/deployment/queue.rs
bmoxb: @chesedo Very point - perhaps the deletion logic should be moved into the loading/execution stage then?

deployer/src/persistence.rs
bmoxb: @chesedo Sorry if I'm misunderstanding, but are you saying that each deployment would then just have a single '.so' filename even after re-deployment? I don't think that would work due to our library caching/reloading issue (i.e., each new compiled '.so' file needs to have a unique name to be correctly loaded) :/

chesedo: Yes, each deployment would have a single 'so' for that deployment. This of course assumes we are sticking with the terminology of a deployment being a version of a service

chesedo: So technically I guess we should have 3 tables:
- Service (S) with the service name
- Deployments (D) with the deployment id. Here exists a (D) many-to-one (S) link
- Logs (L) with the logs for a deployment. Here exists a (L) many-to-one (D) link

The last two tables can always store the service name again if we want the DB view to remain easy (I guess the S table is then no longer needed really)

chesedo:
Will be replaced by #242

PR #231, opened 2022-06-30
Add error handling blog post

kaleidawave:
Hopefully the errors are in the examples and not in the post


www/_blog/2022-06-30-error-handling.mdx
```diff
@@ -0,0 +1,407 @@
...
+}
+```
+
+Running the above causes:
+
+```
+thread 'main' panicked at 'error!', examples\panics.rs:2:5
+```
+
+They are similar to `throw` in JavaScript and other languages, in that the don't need a annotation on the function to run and they can pass through function boundaries. However in Rust, panics cannot be recovered from, there is no way to incept a panic in the current thread.
```
jhawkesworth, line 32: typos: 'in that the don't need a annotation' -> 'in that they don't need an annotation'?

www/_blog/2022-06-30-error-handling.mdx
```diff
@@ -0,0 +1,407 @@
...
+```
+|     send_message();
+|     ^^^^^^^^^^^^^^^
+= note: `#[warn(unused_must_use)]` on by default
+= note: this `Result` may be an `Err` variant, which should be handled
+```
+
+### Examples of `Result` in the standard library
+
+Result can be found in most libraries. On of my favorite examples is the return type of the [FromStr::from_str](https://doc.rust-lang.org/std/str/trait.FromStr.html#tymethod.from_str) trait method. With [str::parse](https://doc.rust-lang.org/std/primitive.str.html#method.parse) (which uses the `FromStr` trait) we can do the following:
```
jhawkesworth, line 92: Typo: 'On of my favourite' -> One of my favorite' ?

www/_blog/2022-06-30-error-handling.mdx
```diff
@@ -0,0 +1,407 @@
...
+
+Here we can see when we type in a number we get a `Ok` variant with the number else we get a [ParseFloatError](https://doc.rust-lang.org/std/num/struct.ParseFloatError.html)
+
+### Files, networks and databases
+
+**All errors occur when you interact with the outside world or things outside of the Rust runtime**. One of the places where a lot of errors can occur is interacting with the file system. The `File::open` function attempts to open a file. This can fail for many reasons, either the filename being attempted to open is invalid, the file does not exist or a permission error. Otherwise it returns a Ok variant with a [File](https://doc.rust-lang.org/std/fs/struct.File.html) which can be read or written to.
+
+##### Aliasing Results and errors
+
+For many results that have the same error you can use type aliases to simplify things. All functions which operate on the FS have the same errors (file not exists, invalid permissions). [io::Result](https://doc.rust-lang.org/std/io/type.Result.html) is a alias over a result but means that every function does not have to specify the error type:
```
jhawkesworth, line 136: Consider not abbreviating file system to FS here, for clarity

www/_blog/2022-06-30-error-handling.mdx
```diff
@@ -0,0 +1,407 @@
...
+
+```rust
+pub type Result<T> = Result<T, io::Error>;
+```
+
+If you have a API which has a common error type it may be a good idea to consider this pattern.
```
jhawkesworth, line 142: 'a API'  should perhaps be 'an API'?

www/_blog/2022-06-30-error-handling.mdx
```diff
@@ -0,0 +1,407 @@
...
+        .map_err(|_| "read file error");
+    send_to_db()?;
+    Ok(())
+}
+```
+
+The nice thing about the question mark operator on Result is that it implicitly runs [Into::into](https://doc.rust-lang.org/std/convert/trait.Into.html#tymethod.into) (which is automatically implemented for [From](https://doc.rust-lang.org/std/convert/trait.From.html) implementors) on the error value. So we can don't have to worry about converting the error before we use the operator:
+
+```rust
+// This derive means we get a a into implementation for `std::io::Error -> MyError`
```
jhawkesworth, line 179: 'get a a into' -> 'get an into'?

www/_blog/2022-06-30-error-handling.mdx
```diff
@@ -0,0 +1,407 @@
...
+
+### [The error trait](https://doc.rust-lang.org/std/error/trait.Error.html)
+
+[The Error trait is implemented for many errors](https://doc.rust-lang.org/std/error/trait.Error.html#implementors) and provides a unified API for information on errors. The Error trait is a bit needy and requires that the error implements both [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html) and [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html). While it can be cumbersome to implement we will see some helper libraries for doing so later on.
+
+In the standard library [VarError](https://doc.rust-lang.org/std/env/enum.VarError.html) (for reading environment variables) and [ParseIntError](https://doc.rust-lang.org/std/num/struct.ParseIntError.html) (for parsing a string slice as a integer) are different errors. When we interact them we need to differentiate between the types because they have different properties and different stack sizes. To build a combination of them we could build a sum type using a enum. Alternatively we can use dynamically dispatched traits which handled varying stack sized items and other type information.
```
jhawkesworth, line 198: 'traits which handled varying' -> 'traits which handle varying'

www/_blog/2022-06-30-error-handling.mdx
```diff
@@ -0,0 +1,407 @@
...
+
+Using the above mentioned try syntax (`?`) we can convert the the above errors to be dynamically dispatched. This makes it easy to handle different errors without building enums to combine errors.
```
jhawkesworth, line 200: 'convert the the above' -> 'convert the above'

www/_blog/2022-06-30-error-handling.mdx
```diff
@@ -0,0 +1,407 @@
...
+}
+```
+
+[(main function can return Result)](https://github.com/rust-lang/rust/pull/49162)
+
+While this is a easy way to handle errors, it isn't easy to differentiate between the types and can make handling errors in libraries hard. More information on this later.
+
+### Methods on Result
+
+Result and Option contains many useful functions. Here are some of my commonly used:
```
jhawkesworth, line 217: consider either 'some commonly used functions' or 'some functions I use commonly'?

www/_blog/2022-06-30-error-handling.mdx
```diff
@@ -0,0 +1,407 @@
...
+
+```rust
+fn get_first(vec: &Vec<i32>) -> Result<&i32, NotInVec> {
+    vec.first().ok_or_else(|| NotInVec)
+}
+```
+
+### Error handling for iteration 
+
+Using results in iterator chains can be a little confusing. Luckily Result implements collect. We can use this to short circuit a iterator if a error occurs. In the following, if all the `parse`s succeed then we get collected vec of numbers result. If one fails then it instead returns a Result with the failing Err.
```
jhawkesworth, line 250: 'short circuit a iterator if a error occurs' -> 'short circuit an iterator if an error occurs'.  I think I am starting to sound like clippy in pedantic mode :-)

www/_blog/2022-06-30-error-handling.mdx
```diff
@@ -0,0 +1,407 @@
...
+
+`todo!()`, `unimplemented!()`, `unreachable!()` are all wrappers for `panic!()` which but are specialized to their situation. Panics have a special `!` never type, which means they can be passed of as any type.
+
+```rust
+fn func_i_havent_written_yet() -> u32 {
+    todo!()
+}
+```
+
+Sometimes there is Rust code which the compiler cannot properly infer is valid. For those using a `unreachable!` panic can be used:
```
jhawkesworth: 'a unreachable' -> 'an unreachable'?

www/_blog/2022-06-30-error-handling.mdx
```diff
@@ -0,0 +1,407 @@
...
+// result.unwrap()
+
+let value = if let Ok(value) = result {
+    value
+} else {
+    panic!("Unwrapped!")
+};
+```
+
+The uses cases for this is developer error and in situations the compiler cannot infer. If you are just trying something and don't want to set up a full error handling system then they can be used to ignore compiler warnings.
```
jhawkesworth: -> 'this are developer error and'  (to agree with plural 'use cases')

PR #232, opened 2022-06-30
Refactor/locate root dir

ivancernja:
Cleaner implementation of the `locate_root_directory()` function.


cargo-shuttle/src/config.rs
```diff
@@ -431,24 +431,24 @@ mod tests {
 
     #[test]
     fn get_local_config_finds_name_in_shuttle_toml() {
-        let project_args = ProjectArgs {
+        let mut project_args = ProjectArgs {
```
chesedo, line 434: These `mut`s should not be needed :thinking: ?

chesedo (changes requested):
Thanks for this @ivancernja! It seems some of the original code changes did not pull through

PR #238, opened 2022-07-05
Implement `cargo shuttle init --axum|rocket|tide|tower`

### LINKED ISSUE #175: feature: cargo shuttle init

nahuakang:
## Motivation
This PR closes #175. It adds 4 flags to `cargo shuttle init` command and generates the corresponding boilerplate code for each framework.

## Changes
Added 3 crates: `url`, `indoc` and `tempfile`.

- `url` is needed for handling `cargo-edit`-related operations.
- `indoc` is required to format raw string literals nicely with the code so that we can indent raw string literals in sync with the code indentation.
- `tempfile` is a dev dependency for generating and deleting temporary files for integration tests that require such file generations.


cargo-shuttle/src/init.rs
```diff
@@ -0,0 +1,342 @@
...
+                "Hello, world!"
+            }
+
+            #[shuttle_service::main]
+            async fn axum() -> shuttle_service::ShuttleAxum {
+                let router = Router::new().route("/hello", get(hello_world));
+                let sync_wrapper = SyncWrapper::new(router);
+
+                Ok(sync_wrapper){}
+            }"#},
```
chesedo, line 228: Is it possible to put these in static/template files, rather than in the code?

cargo-shuttle/src/init.rs
```diff
@@ -0,0 +1,342 @@
...
+use url::Url;
+
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum Framework {
+    Axum,
+    Rocket,
+    Tide,
+    Tower,
+    Default,
```
chesedo, line 21: What will the idea with `Default` be? Will it still generate a template, or just error?

chesedo (commented):
This is taking shape nicely, thanks @nahuakang. I like the direction you took here :smile:

cargo-shuttle/src/init.rs
nahuakang: Oh good point! Yes that would be a great idea. I can just read the files and then write the content into `lib.rs`. And it makes testing easier. Thanks!

cargo-shuttle/src/init.rs
nahuakang: Thanks for asking. I think it just won't generate anything. I used it as a condition. If this enum option is not necessary, do you have any suggestion to handle when user doesn't specify a flag? 🙏

cargo-shuttle/src/init.rs
nahuakang: Btw, should the `static/templates` files live inside `shuttle/cargo-shuttle/` directory?

nahuakang:
@chesedo Hey Pieter! I added some unit tests (skipping the functions that involve IO-related traits, which are a bit harder to mock for the time-being). Tested the installed cli app locally and it's behaving properly.

Let me know what you think :)

cargo-shuttle/src/init.rs
```diff
@@ -0,0 +1,650 @@
...
+
+        let expected = indoc! {r#"
+            [dependencies]
+            shuttle-service = { version = "1.0", features = ["web-axum"] }
+            axum = "1.0"
+            sync_wrapper = "1.0"
+        "#};
+
+        assert_eq!(cargo_toml.to_string(), expected);
+    }
```
chesedo, line 579: Nice, I like these :smile:

cargo-shuttle/src/lib.rs
```diff
@@ -82,42 +82,17 @@ impl Shuttle {
...
+        let path = args.path.clone();
+        init::cargo_init(path.clone())?;
+        init::process_cargo_init(path.clone())?;
+
+        let framework = init::get_framework(&args);
+        match framework {
+            Some(framework) => {
+                init::framework_init(&path, framework)?;
+            },
+            None => {},
```
chesedo, line 94: Is the idea here to do a no-op when the user did not choose a framework?

cargo-shuttle/src/init.rs
```diff
@@ -0,0 +1,650 @@
...
+
+    // Truncate Cargo.toml and write the updated `Document` to it
+    let mut cargo_toml = File::create(cargo_toml_path.clone())?;
+    cargo_toml.write_all(cargo_doc.to_string().as_bytes())?;
+
+    Ok(())
+}
+
+/// Generates framework-specific dependencies to `Cargo.toml` and boilerplate code for `src/lib.rs`.
+pub fn framework_init(project_path: &PathBuf, framework: Box<dyn ShuttleInit>) -> Result<()> {
```
chesedo, line 355: Since this function and the previous one (`process_cargo_init`) both setup the `manifest_path` and `url` and both then write out a toml file, I'm wondering if it is possible to combine them to only have these steps happen once? Maybe the framework code can move into `process_cargo_init`?

This all depends on the no-op comment...

cargo-shuttle/tests/integration/init.rs
```diff
@@ -19,7 +19,13 @@ async fn cargo_shuttle_init(path: &str) -> anyhow::Result<CommandOutcome> {
...
                 name: None,
             },
-            cmd: Command::Init(InitArgs { path }),
+            cmd: Command::Init(InitArgs {
+                axum: false,
+                rocket: false,
+                tide: false,
+                tower: false,
+                path,
+            }),
```
chesedo, line 28: I am happy with all the unit tests. An extra one that tests the whole sequence and that the `Cargo.toml` and `lib.rs` is correct would be nice. This can be for any framework as there is no need to test all of them in this way I think.

Maybe this test can be expanded to check the contents of `Cargo.toml` and `lib.rs`? [tempdir](docs.rs/tempdir/) can maybe be used to create the `working_directory` for this test :man_shrugging:

chesedo (commented):
This seems about done @nahuakang. I just have a few possible polish points

cargo-shuttle/src/lib.rs
nahuakang: Yes, no-op if the framework isn't there since `cargo init --lib` was called and that suffices. Any suggestion to improve?

chesedo: Nope, this fine. Just wanted to make sure it is intentional :smile:

cargo-shuttle/src/init.rs
nahuakang: So if the no-op part stays, I can ignore this comment? :D

cargo-shuttle/src/init.rs
```diff
@@ -0,0 +1,650 @@
...
+    let mut dependencies = Table::new();
+
+    // Set "shuttle-service" version to `[dependencies]` table
+    set_inline_table_dependency_version(
+        "shuttle-service",
+        &mut dependencies,
+        &manifest_path,
+        &url,
+        get_latest_dependency_version,
+    );
```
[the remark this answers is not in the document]
nahuakang: You're right. Thanks for pointing this out!

cargo-shuttle/src/init.rs
chesedo: Lol, I think I was wrong about an earlier request to remove the `Default` enum variant :sweat_smile: 

If we create a `ShuttleInitNoOp` and return it instead of `None` in `get_framework`, then we should be able to make `framework: Box<dyn ShuttleInit>` an argument of `process_cargo_init` right?

cargo-shuttle/tests/integration/init.rs
```diff
@@ -19,17 +39,68 @@ async fn cargo_shuttle_init(path: &str) -> anyhow::Result<CommandOutcome> {
...
         })
         .await
 }
 
-#[tokio::test]
-async fn basic_init() {
-    cargo_shuttle_init("tests/tmp/basic-init").await.unwrap();
+#[cfg(test)]
+mod init_integration_tests {
```
chesedo, line 54: The `cfg` check and namespacing should not be needed since these are in the `tests` directory

chesedo (commented):
Just a small tests fixup :smile:

chesedo (approved):
This is turned out real nice, thanks @nahuakang!

PR #242, opened 2022-07-06
refactor: deployer db

chesedo:
Adds an id field to deployments. This field is then used to also name 'so' files

Close #230 

Fixes ENG-55


brokad (approved):
👌

PR #243, opened 2022-07-06
Fix webpage link for the url-shortener example

lilianmoraru:
The webpage link to the url-shortener example is outdated - this updates it, similar to the other 2 examples, using the `rocket` example.


chesedo (approved):
Thanks for this @lilianmoraru :smile:

PR #244, opened 2022-07-07
chore: docker compose

chesedo:
Use docker-compose file overrides rather than extends


brokad (approved):
LGTM, thank you :smile:

PR #247, opened 2022-07-09
infra: placement constraints

brokad:


docker-compose.yml
```diff
@@ -1,7 +1,6 @@
 version: "3"
 volumes:
   shuttle-backend-vol:
-    external: true
```
chesedo, line 4: Shouid we also remove the `docker volume create shuttle-backend-vol` from `CONTRIBUTING.md`?

PR #249, opened 2022-07-09
fix: compiler error undeclared

Xavientois:
The postgres example was failing to compile due to the use of the `shared` module/crate:

```
error[E0433]: failed to resolve: use of undeclared crate or module `shared`
  --> src/lib.rs:52:19
   |
52 | async fn rocket(#[shared::Postgres] pool: PgPool) -> shuttle_service::ShuttleRocket {
   |                   ^^^^^^ use of undeclared crate or module `shared`
```

This removes that annotation to allow for successful compilation.


Xavientois:
This looks like it causes the CI to fail. I will investigate why the annotation works in CircleCI, but not locally

chesedo:
Replaced by #255

PR #265, opened 2022-07-18
feat: deployer build logs

chesedo:
Replaces #224 

Fixes ENG-32


deployer/src/deployment/mod.rs
```diff
@@ -79,13 +81,13 @@ impl Pipeline {
...
     /// deployments between the aforementioned tasks.
-    fn new(kill_send: KillSender) -> Pipeline {
+    fn new(kill_send: KillSender, log_recorder: impl LogRecorder) -> Pipeline {
         let (queue_send, queue_recv) = mpsc::channel(QUEUE_BUFFER_SIZE);
         let (run_send, run_recv) = mpsc::channel(RUN_BUFFER_SIZE);
 
         let run_send_clone = run_send.clone();
 
-        tokio::spawn(async move { queue::task(queue_recv, run_send_clone).await });
+        tokio::spawn(async move { queue::task(queue_recv, run_send_clone, log_recorder).await });
```
brokad, line 90: Nit: I know this is not from this PR, but I don't think the `async move` and the inner `.await` are required.

deployer/src/deployment/deploy_layer.rs
```diff
@@ -30,7 +31,7 @@ use super::{
 };
 
 /// Records logs for the deployment progress
-pub trait LogRecorder {
+pub trait LogRecorder: Clone + Send + 'static {
```
brokad, line 34: Not that it matters too much, but `LogRecorder` seems very close to [`futures::sink::Sink::<Log>`](https://docs.rs/futures/latest/futures/sink/trait.Sink.html).

deployer/src/persistence.rs
```diff
@@ -135,6 +155,23 @@ impl Persistence {
...
+    pub fn get_build_log_subscriber(&self) -> Receiver<BuildLog> {
+        self.build_log_send.subscribe()
+    }
+
+    pub async fn get_build_logs(&self, name: &str) -> Result<Vec<BuildLog>> {
+        let logs: Vec<Log> =
+            sqlx::query_as("SELECT * FROM logs WHERE name = ? OR state = ? ORDER BY timestamp")
+                .bind(name)
+                .bind(State::Building)
+                .fetch_all(&self.pool)
```
brokad, line 168: Do we want an upper bound on this to prevent some crazy edge case?

deployer/src/handlers.rs
```diff
@@ -65,3 +73,47 @@ async fn delete_service(
...
+    Extension(persistence): Extension<Persistence>,
+    Path(name): Path<String>,
+    ws_upgrade: ws::WebSocketUpgrade,
+) -> axum::response::Response {
+    ws_upgrade.on_upgrade(move |s| websocket_handler(s, persistence, name))
+}
+
+async fn websocket_handler(mut s: WebSocket, persistence: Persistence, name: String) {
+    let mut log_recv = persistence.get_build_log_subscriber();
+    let backlog = persistence.get_build_logs(&name).await.unwrap();
```
brokad, line 94: Safe?

deployer/src/persistence.rs
```diff
@@ -59,14 +62,27 @@ impl Persistence {
...
         let handle = tokio::spawn(async move {
             while let Some(log) = log_recv.recv().await {
                 match log.r#type {
-                    LogType::Event => insert_log(&pool_cloned, log).await.unwrap(),
+                    LogType::Event => {
+                        if log.state == State::Building {
+                            if let Some(build_log) = Into::into(&log) {
+                                build_log_send_clone
+                                    .send(build_log)
+                                    .expect("failed to broadcast build log");
```
brokad, line 80: In this, a thread is spawned which single handedly manages the flow of of `Log`s to places where it needs to go. If one of these `.expect` or `.unwrap` actually panics, I guess the service is bricked but it will keep running instead of restarting, or fixing the condition.

We should perhaps avoid things that could panic in here (and maybe log them out as errors instead), and later on grab this returned `handle` and monitor it?

brokad (approved):
LGTM :smile:

deployer/src/deployment/deploy_layer.rs
```diff
@@ -84,6 +85,42 @@ impl From<Log> for DeploymentInfo {
     }
 }
 
+impl From<&Log> for Option<BuildLog> {
```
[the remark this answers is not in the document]
chesedo: I made the `Log::to_build_log` change but also need to use the `to_build_log` function (not method) [here](https://github.com/shuttle-hq/shuttle/pull/265/files#diff-0d281763868328b5543da731e24aaf26c398eef8bb3eb695e068aae73c1973b0R28)

deployer/src/deployment/deploy_layer.rs
chesedo: Do you see a benefit for us converting this to a `Sink` (like buffering)? Maybe I should add it as a ticket under the deployer project?

deployer/src/persistence.rs
chesedo: I'm not sure yet :thinking: For now I'm adding a todo comment and linear ticket to test this when all the pieces are together

chesedo: ... that `OR` should be an `AND` though :man_facepalming:

PR #266, opened 2022-07-18
refactor: erroring on raw response (deployer)

chesedo:
Fixes: ENG-57

### DIFF
--- a/cargo-shuttle/Cargo.toml
+++ b/cargo-shuttle/Cargo.toml
@@ -17,7 +17,6 @@ chrono = "0.4.19"
 colored = "2.0.0"
 crossterm = "0.23.2"
 dirs = "4.0.0"
-env_logger = "0.9.0"
 futures = "0.3.21"
 log = "0.4.17"
 portpicker = "0.1.1"
@@ -31,6 +30,8 @@ structopt = "0.3.26"
 tokio = "1.19.2"
 toml = "0.5.9"
 toml_edit = "0.14.4"
+tracing = "0.1.35"
+tracing-subscriber = { version = "0.3.11", features = ["env-filter"] }
 uuid = { version = "1.1.1", features = ["v4"] }
 webbrowser = "0.7.1"
 
--- a/cargo-shuttle/src/client.rs
+++ b/cargo-shuttle/src/client.rs
@@ -37,7 +37,7 @@ pub(crate) async fn auth(mut api_url: ApiUrl, username: String) -> Result<ApiKey
     error!(
         text = response_text,
         status = %response_status,
-        "failed to authenicate with server"
+        "failed to authenticate with server"
     );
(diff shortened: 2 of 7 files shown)


cargo-shuttle/src/client.rs
```diff
@@ -33,11 +34,12 @@ pub(crate) async fn auth(mut api_url: ApiUrl, username: String) -> Result<ApiKey
...
 
-    Err(anyhow!(
-        "status: {}, body: {}",
-        response_status,
-        response_text
-    ))
+    error!(
+        text = response_text,
+        status = %response_status,
+        "failed to authenicate with server"
```
brokad, line 40:
```suggestion
        "failed to authenticate with server"
```

cargo-shuttle/src/client.rs
```diff
@@ -220,6 +223,12 @@ async fn to_api_result(res: Response) -> Result<DeploymentMeta> {
     let text = res.text().await?;
     match serde_json::from_str::<DeploymentMeta>(&text) {
         Ok(meta) => Ok(meta),
-        Err(_) => Err(anyhow!("{}", text)),
+        Err(error) => {
+            error!(
+                error = &error as &dyn std::error::Error,
+                text, "failed to parse deployment meta"
+            );
+            Err(anyhow!("could not parse server response"))
```
brokad, line 231: Does using `error.context(anyhow!(...))` provide a nicer output?

cargo-shuttle/Cargo.toml
```diff
@@ -31,6 +30,8 @@ structopt = "0.3.26"
 tokio = "1.19.2"
 toml = "0.5.9"
 toml_edit = "0.14.4"
+tracing = "0.1.35"
```
brokad, line 33: `tracing` is taking over 😉

cargo-shuttle/src/client.rs
```diff
@@ -87,11 +89,12 @@ pub(crate) async fn shuttle_version(mut api_url: ApiUrl) -> Result<String> {
...
-            "status: {}, body: {}",
-            response_status,
-            res.text().await?
-        ))
+        error!(
+            text = res.text().await?,
+            status = %response_status,
+            "failed to get shuttle version from server"
+        );
+        Err(anyhow!("failed to get shuttle version from server",))
```
brokad, line 97: Same question as a comment below?

brokad (approved):

cargo-shuttle/src/client.rs
chesedo: Yip, it gives a bit more of a "backtrace". Thanks!

cargo-shuttle/src/client.rs
chesedo: Not quite, there is no `Result` here to call `.context()` on - ie nothing to "backtracify" so nothing to make nicer

PR #267, opened 2022-07-18
Log to tracing migration

SonicZentropy:
- Migrates all usages of 'log' to tracing
- Adds custom serde derive for tracing::Level

Someone PLEASE try this out before merging!  It works fine as far as I can tell, but I don't know Shuttle well enough yet to be certain there's not a huge chunk of logs somewhere missing.


SonicZentropy:
Def no merging, looks like it broke some rando tests comparing log output

service/src/loader.rs
```diff
@@ -11,7 +11,7 @@ use cargo::ops::{compile, CompileOptions};
 use cargo::util::homedir;
 use cargo::Config;
 use libloading::{Library, Symbol};
-use log::trace;
+use tracing::log::trace;
```
chesedo, line 14: I can't find a `tracing::log` module in the docs. I can find a `tracing::trace` though. Maybe I'm just looking in the wrong place?

service/src/logger.rs
```diff
@@ -21,7 +21,7 @@ impl Logger {
     }
 }
 
-impl log::Log for Logger {
+impl tracing::log::Log for Logger {
```
chesedo, line 24: We can go with this, it would also be great if this can be redesigned to be a tracing subscriber or layer instead. I'm not quite sure how difficult that will be though :man_shrugging:

cargo-shuttle/src/main.rs
```diff
@@ -4,8 +4,6 @@ use clap::Parser;
 
 #[tokio::main]
 async fn main() -> Result<()> {
-    env_logger::init();
```
chesedo, line 7: Shouldn't the fmt subscriber be registered here?

chesedo (commented):
Thanks for starting this @SonicZentropy. This seems to be great progress. I just have some high-level notes.

service/src/loader.rs
SonicZentropy: It's hidden behind the "log" feature on tracing.  The difference is somewhat confusing enough I ended up in the Tokio discord while doing this same thing last time.  The `tracing::trace!` version accumulates the logs internally unless you attach some sort of subscriber (I'm using tracing_subscriber as a dev dependency in some of the tests).  Enabling the "log" feature unveils the `tracing::log::` module where the macros behave as a drop-in replacement for the normal type of logging

service/src/logger.rs
SonicZentropy: It's not, I'd prefer that way myself since my old code I cribbed most of this from does it that way already :)  I decided to start off as simple as possible PoC, but I'd also much prefer the layered approach.  I'll trade it out and people can decide which to stick with

cargo-shuttle/src/main.rs
SonicZentropy: I am still not certain.  I have a //TODO marker under it to find out, because the docs are not overly helpful either way.  I'm still working through e2e test fixes and expect to see something exploding eventually

SonicZentropy:
Self-reminder to tag the rest of the serde alternative Layer defs based on https://www.reddit.com/r/rust/comments/w3q1oq/things_i_wish_i_had_known_about_serde_json/ before this gets merged

oddgrd:
Thanks for starting on this, but this has been completed in other PRs. :slightly_smiling_face:

PR #268, opened 2022-07-19
Fixes e2e Readme link

SonicZentropy:
Tiny housekeeping fix I ran into

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -1,7 +1,7 @@
 # Overview
-This crate runs all the end-to-end tests for shuttle. These tests must run against a local dev environment, so you first have to set that up by following [these instructions](../README.md#working-on-shuttle).
+This crate runs all the end-to-end tests for shuttle. These tests must run against a local dev environment, so you first have to set that up by following [these instructions](https://github.com/shuttle-hq/shuttle/blob/main/CONTRIBUTING.md).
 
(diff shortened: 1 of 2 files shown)


e2e/README.md
```diff
@@ -1,7 +1,7 @@
 # Overview
-This crate runs all the end-to-end tests for shuttle. These tests must run against a local dev environment, so you first have to set that up by following [these instructions](../README.md#working-on-shuttle).
+This crate runs all the end-to-end tests for shuttle. These tests must run against a local dev environment, so you first have to set that up by following [these instructions](https://github.com/shuttle-hq/shuttle/blob/main/CONTRIBUTING.md).
```
chesedo, line 2: Any reason for linking to `main`? I'm a bit worried about having `rc`s and stuff in the future that might have different instructions for a temporary period.

chesedo (approved):
Thanks for the fixup @SonicZentropy. I'm happy merging this as is, just want to check something.

e2e/README.md
SonicZentropy: I copied it straight from the base README.  I see what you mean, so I think perhaps we should change both of them to local?

chesedo: Yes, if the change makes sense, then let's change both :+1:

SonicZentropy: Fixed!

chesedo (approved):
Thanks for the cleanup @SonicZentropy

PR #273, opened 2022-07-21
refactor: plugins

chesedo:
This turns our service feature flags into separate crates. The long game here will be to allow vendors to write their own plugins...

### TODO
- [x] Update CI

### Breaking changes
This changes the main function argument attributes yet again :smiley: . These changes are as follow:

| Old attribute | New attribute |
| ---------------- | ------------------ |
|shared::Postgres | shuttle_shared_db::Postgres |
|shared::MongoDb | shuttle_shared_db::MongoDb |
|aws::rds::Postgres | shuttle_aws_rds::Postgres |
|aws::rds::MySql | shuttle_aws_rds::MySql |
|aws::rds::MariaDB | shuttle_aws_rds::MariaDB |
|persist | shuttle_persist::Persist |

The old feature flag on `shuttle_service` also needs to be replaced with the following crates definitions:

| Old feature flag | Extra crate definition |
| -------------------- | --------------------------- |
| sqlx-postgres | shuttle-shared-db = { version = "0.5.2", features = ["postgres"] } |
| mongo-integration | shuttle-shared-db = { version = "0.5.2", features = ["mongo"] }|
| sqlx-aws-postgres | shuttle-aws-rds = { version = "0.5.2", features = ["postgres"] } |
| sqlx-aws-mysql | shuttle-aws-rds = { version = "0.5.2", features = ["mysql"] } |
| sqlx-aws-mariadb | shuttle-aws-rds = { version = "0.5.2", features = ["mariadb"] } |
| secrets | shuttle-secrets = "0.5.2" |
| persist | shuttle-persist = "0.5.2" |


.circleci/config.yml
```diff
@@ -85,6 +85,9 @@ commands:
             cat\<< EOF > ~/.cargo/config.toml
             [patch.crates-io]
             shuttle-service = { path = "$PWD/service" }
+            shuttle-aws-rds = { path = "$PWD/services/aws-rds" }
```
brokad, line 88: Nit: we now have two top-level directories that are very closely named: `service` and `services`.

Perhaps we should rename `services` to `plugins` or `builders`?

service/src/lib.rs
```diff
@@ -220,26 +220,9 @@ pub mod logger;
 
 pub use shuttle_common::database;
 
-#[cfg(any(feature = "sqlx-postgres", feature = "mongodb-integration",))]
+#[cfg(any(feature = "mongodb-integration"))]
```
brokad, line 223: `mongodb` is still a feature flag of `shuttle-service` since it was added later

service/src/lib.rs
```diff
@@ -282,7 +265,7 @@ extern crate shuttle_codegen;
 /// struct MyState(PgPool);
 ///
 /// #[shuttle_service::main]
-/// async fn rocket(#[shared::Postgres] pool: PgPool) -> ShuttleRocket {
+/// async fn rocket(#[shuttle_shared_db::Postgres] pool: PgPool) -> ShuttleRocket {
```
brokad, line 268: We should document how these attributes work if we haven't already. AFAICT, the proc macro just expects these to be type addresses to something implementing `ResourceBuilder<T>` where `T` is the type of the argument you want to instantiate.

If we document this, people could write their own hook to grab the `Factory` and manage resources however they want.

brokad (commented):
Holy conflict mackerel! I hope I resolved this without creating more issues (apart from the ones outlined below, due to the time difference between when this was introduced and support for additional resources we have added since then).

Couple more things not specifically pointed at in the comments below:
- Web frameworks do not seem to have been modularised at all?
- The CLI is missing a small tweak for secrets I think
- Serenity is not modularised either (due to time difference)
- The modules don't seem to call on their dependencies through fully qualified versions, so I guess this will need to be added before publishing
- On that topic, will the publish scripts handle the version strings in the new crates correctly?

service/src/lib.rs
chesedo: Document is now up at https://github.com/shuttle-hq/shuttle/pull/273/files#diff-a981df14dcb291acc010b3484e3911046bb9f568ebd28ed7c6c2bc4de23aa1faR304

PR #279, opened 2022-07-23
refactor: main macro to own module

Xavientois:
Opening this PR to POC various elements of the feature and check in about implementation details before moving forward with them.


codegen/src/lib.rs
```diff
@@ -1,392 +1,17 @@
...
-
-#[derive(Debug, PartialEq)]
-struct Input {
-    /// The identifier for a resource input
-    ident: Ident,
-
-    /// The shuttle_service path to the builder for this resource
-    builder: Path,
+pub fn main(attr: TokenStream, item: TokenStream) -> TokenStream {
+    main::r#impl(attr, item)
```
Xavientois, line 10: @NodarD  In order to keep the implementation for the `main` and `lambda` macros separate, I moved the implementation for `main` into its own file (the actual proc_macro function must remain in the lib root). The implementation for `lambda` is a `todo`, but if we want to merge this refactor, I can remove it from this branch and just keep the moved implementation of `main`

chesedo: I'm good with merging this refactor first :smile:

codegen/src/main/mod.rs
```diff
@@ -0,0 +1,390 @@
...
+            }
+        };
+
+        assert_eq!(actual.to_string(), expected.to_string());
+    }
+
+    #[test]
+    fn ui() {
+        let t = trybuild::TestCases::new();
+        t.compile_fail("tests/ui/*.rs");
```
chesedo, line 388: Should these maybe be moved too? Maybe something like `tests/ui/main/*.rs` :man_shrugging:

chesedo (commented):
LGTM with an initial refactor

codegen/src/main/mod.rs
Xavientois: Sure! So you're thinking that they would be run as integration tests rather than unit tests?

chesedo: For these, they can be integration or unit tests really - can't quite remember why I made them unit tests :smile: 

I was thinking more of moving the `ui` files since another set will eventually be added to test the codegen for the lambda macro

Xavientois: Oh, I see what you mean. Sure! That makes sense

chesedo (approved):

PR #285, opened 2022-07-26
added missing symbol

ivancernja:


chesedo (approved):
Thanks for this @ivancernja :smile:

Issue #283, opened 2022-07-26
check_lib_version can panic on deploy without informative message


kaleidawave:
The current method uses indexing to pull out the version of `Cargo.toml`. This can panic if the key is not present:

https://github.com/shuttle-hq/shuttle/blob/d4af3674de5f77477eda7febbbd7935add7f1672/cargo-shuttle/src/lib.rs#L257

If the `Cargo.toml` does not have a dependencies table or shuttle-service is not present then this expression panics. It would better to catch these and produce a readable error. Can use [`Document::get`](https://docs.rs/toml_edit/0.14.4/toml_edit/struct.Document.html#method.get) to do that.

The second question is whether `shuttle-service = "0.4.0"` is an error because there are no features specified. While adding this change should shuttle suggest that a feature needs to present?

chesedo:
I think `shuttle-service = "0.4.0"` should be correct, because that's how I think a user implemented service would specify the shuttle_service dependency.

@kaleidawave sure, you can make the change :smile:

oddgrd:
We removed this check in 0.12, so closing this.

PR #287, opened 2022-07-27
Add better error handling for shuttle-service version checking #283

kaleidawave:
Adds better error handling for when checking shuttle-service version checking #283. Removes potential for indexes to panic and adds detailed information for incorrect results.

**I haven't tested this update yet. I also didn't quite know how to add testing for this feature sooo 😆**

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -25,7 +25,7 @@ use factory::LocalFactory;
 use semver::{Version, VersionReq};
 use shuttle_service::loader::{build_crate, Loader};
 use tokio::sync::mpsc;
-use toml_edit::Document;
+use toml_edit::{Document, Item};
 use uuid::Uuid;
 
 #[macro_use]
@@ -254,10 +254,31 @@ impl Shuttle {
     async fn check_lib_version(&self, project_args: ProjectArgs) -> Result<()> {
         let cargo_path = project_args.working_directory.join("Cargo.toml");
         let cargo_doc = read_to_string(cargo_path.clone())?.parse::<Document>()?;
-        let current_shuttle_version = &cargo_doc["dependencies"]["shuttle-service"]["version"];
-        let service_semver = match Version::parse(current_shuttle_version.as_str().unwrap()) {
+
+        let shuttle_service_entry = cargo_doc
+            .get("dependencies")
+            .ok_or_else(|| anyhow!("Missing dependencies section in Cargo.toml"))?
+            .get("shuttle-service")
+            .ok_or_else(|| anyhow!("Missing shuttle_service dependency in Cargo.toml"))?;
+
+        let current_shuttle_version = match shuttle_service_entry {
+            Item::ArrayOfTables(_)
+            Item::None => {
+                return Err(anyhow!("Invalid entry for shuttle_service"));
+            }
+            Item::Value(_) => shuttle_service_entry,
+            Item::Table(table) => table
+                .get("version")
+                .ok_or_else(|| anyhow!("Missing version key"))?
+        };
+
+        let version_string = current_shuttle_version
+            .as_str()
+            .ok_or_else(|| anyhow!("Invalid version as string"))?;
+        
+        let service_semver = match Version::parse(version_string) {
             Ok(version) => version,
-            Err(error) => return Err(anyhow!("Your shuttle-service version ({}) is invalid and should follow the MAJOR.MINOR.PATCH semantic versioning format. Error given: {:?}", current_shuttle_version.as_str().unwrap(), error.to_string())),
+            Err(error) => return Err(anyhow!("Your shuttle-service version ({}) is invalid and should follow the MAJOR.MINOR.PATCH semantic versioning format. Error given: {:?}", version_string, error.to_string())),
         };
 
         let server_version = client::shuttle_version(self.ctx.api_url()).await?;


kaleidawave:
Hmm I guess I am not handling a `Value::InlineTable`... 

I think I instead need to be looking at parsing a [manifest and using the dependency method on that](https://docs.rs/cargo/latest/cargo/core/manifest/struct.Manifest.html#method.dependencies) instead of ad hoc dependency parsing as there is [a lot of places the version could be in](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#choosing-features)

kaleidawave:
Awesome looks like it is passing all the tests and can be merged! :)

I decided to use the existing toml_edit crate in the end as the cargo crate is quite integrated and getting dependency information requires linking with the filesystem and stuff that seemed unnecessary complicated.

chesedo (changes requested):
These are great. Thanks @kaleidawave!

I see you also figured the tests out and love them :heart:

kaleidawave:
Awesome, yes those changes are good. Have un-nested some of the functions and expanded out the tests a lot more. Now to see if they still pass 😬

cargo-shuttle/src/lib.rs
```diff
@@ -342,7 +390,10 @@ pub enum CommandOutcome {
 mod tests {
     use crate::args::ProjectArgs;
     use crate::Shuttle;
+    use semver::Version;
     use std::path::PathBuf;
+    use std::str::FromStr;
+    use toml_edit::Document;
```
chesedo, line 396: Maybe some of these should be in the `check_lib_version` module?

cargo-shuttle/src/lib.rs
```diff
@@ -377,4 +428,131 @@ mod tests {
...
+            )
+            .unwrap();
+
+            Shuttle::check_lib_version(
+                Document::from_str("[dependencies]\nshuttle-service = \"0.4.5\"").unwrap(),
+                &version_four,
+            )
+            .unwrap();
+
+            let version_four_point_one = Version::parse("0.4.3").unwrap();
```
chesedo, line 525: `version_four_piont_three` :see_no_evil:

cargo-shuttle/src/lib.rs
```diff
@@ -377,4 +428,131 @@ mod tests {
...
+                Document::from_str("[dependencies]\nshuttle-service = \"0.5\"").unwrap(),
+                &version_four_point_one,
+            )
+            .unwrap();
+
+            Shuttle::check_lib_version(
+                Document::from_str("[dependencies]\nshuttle-service = \"2\"").unwrap(),
+                &version_four_point_one,
+            )
+            .unwrap();
```
chesedo, line 555: Hmm, should these be vaild? These are behind the server and we have no backward compatibility at this time

cargo-shuttle/src/lib.rs
```diff
@@ -377,4 +428,131 @@ mod tests {
...
         );
     }
+
+    mod check_lib_version {
+        use super::*;
+
+        #[test]
+        fn invalid_manifests() {
+            Shuttle::check_lib_version(Document::new(), &Version::parse("0.4.1").unwrap())
+                .unwrap_err();
```
chesedo, line 438: For all the `unwrap_err()` tests, maybe we should check the error is what we expect?

cargo-shuttle/src/lib.rs
```diff
@@ -251,20 +256,63 @@ impl Shuttle {
...
+                .get("version")
+                .ok_or_else(|| anyhow!("Missing 'version' key"))
+        }
+
+        fn get_item_as_str(item: &Item) -> Result<&str> {
+            item.as_str()
+                .ok_or_else(|| anyhow!("Expected string, found {}", item))
+        }
+
+        fn get_dependency_version_string<'a>(
```
chesedo, line 271: Ok, this is better. Any reason why the 2 levels of nesting is still needed?

chesedo (commented):
Just a bit worried about one of the test cases :smile:

cargo-shuttle/src/lib.rs
kaleidawave: Purely stylistic 😁 . Can be better with code folding. I don't think the functions are that useful outside of the `check_lib_version` function so that is why they are in that scope. Can change if it looks complicated

cargo-shuttle/src/lib.rs
kaleidawave: Yeah the tests are a little complicated to grapple. I think it says that if the clients shuttle_service is version 0.5 or 2 then it works on a server *running* 0.4.3 (because at the moment it is just a >= check ignoring patch versions). That might not be valid but at the same if they have a version 2 of shuttle service, will the server be running 0.4.3?

kaleidawave:
Awesome have responded. Yes this is quite critical, don't want to start rejecting all builds because of a error with the version logic 😆. So should rigorous with the tests here

cargo-shuttle/src/lib.rs
chesedo: Hmm... good point. Let's keep this for now then

cargo-shuttle/src/lib.rs
chesedo: I see, this is fine :smile:

cargo-shuttle/src/lib.rs
kaleidawave: This is a good conversation. Have written this in a comment in the tests

chesedo (approved):
:partying_face: Seems all good to go, thanks @kaleidawave

chesedo:
@kaleidawave we have some tests failing to compile :smile:

kaleidawave:
Ooof that was a silly error on my part. Is fixed now :)

chesedo:
@kaleidawave it seems there is a few extra failing tests in `cargo-shuttle/src/lib.rs` :smile:

kaleidawave:
Ooof another silly error. Also after some local testing I realise `Version` is a fully qualified version and that cargo dependencies specify a requirement instead (`VersionReq`). So I think I will also add a fix that allows version to be `*`, `0.5`, `0.5.*` and so on to be compatible with the ones Cargo supports. 

But [VersionReq](https://docs.rs/semver/latest/semver/struct.VersionReq.html) doesn't have order defined... When I have a bit more time next week I will do some reading on semver and write some custom matching logic

chesedo:
Most of this has been made out of date by #377. Closing for now, but we can open it again for the better error messages

PR #293, opened 2022-08-03
feat: profanity filter & added "shuttle.rs" to the reserved list of project names

Butch78:
This ensures users aren't using profanity for the names of their shuttle projects. It also creates an easy way to add to the reserved project name list.


common/src/project.rs
```diff
@@ -124,6 +144,8 @@ pub mod tests {
             ".invalid",
             "invalid.name",
             "invalid.name.",
+            "test-fuck-fuck",
```
christoshadjiaslanis, line 147: consider changing to `crap` as it is still profanity and little less abrasive :smiley:

common/src/project.rs
```diff
@@ -41,14 +47,26 @@ impl std::fmt::Display for ProjectName {
...
 
-        !(hostname.bytes().any(|byte| !is_valid_char(byte))
+        fn is_profanity_free_and_not_reserved(hostname: &str) -> bool {
+
+            let mut reserved_words = HashSet::new();
+            // Reserved words can be easily added here. 
+            reserved_words.insert("shuttle.rs".to_string());
+
+
+            let censor = Censor::Standard + Censor::Sex + Censor::Custom(reserved_words);
```
chesedo, line 65: Maybe this can be replaced with an `once_cell` to not initialize it each time?

common/src/project.rs
```diff
@@ -41,14 +47,26 @@ impl std::fmt::Display for ProjectName {
...
                 || (b'0'..=b'9').contains(&byte)
                 || byte == b'-'
         }
 
-        !(hostname.bytes().any(|byte| !is_valid_char(byte))
+        fn is_profanity_free_and_not_reserved(hostname: &str) -> bool {
+
+            let mut reserved_words = HashSet::new();
+            // Reserved words can be easily added here. 
+            reserved_words.insert("shuttle.rs".to_string());
```
chesedo, line 62: Creating the `HashSet` from an array ([the last doc example](https://doc.rust-lang.org/std/collections/struct.HashSet.html)) may make it even easier to add reserve words in the future :man_shrugging:

chesedo (commented):
Wow, nice and simple! Thanks @Butch78

common/src/project.rs
Butch78: Fixed! :)

chesedo (approved):
Nice one @Butch78!

chesedo:
@Butch78 it seems we only need to run a `cargo fmt --all` here

common/src/project.rs
```diff
@@ -48,7 +56,19 @@ impl ProjectName {
...
+
+            let censor = Censor::Standard
+                + Censor::Sex
+                + Censor::Custom(INSTANCE.get().expect("Reserved words not set").clone());
+            return !censor.check(hostname);
+        }
+
+        !(hostname
+            .bytes()
+            .any(|byte| !is_valid_char(byte) | !is_profanity_free_and_not_reserved(hostname))
```
coszio, line 71: I think you meant to add the check outside of the `any` closure to only check it once?

common/src/project.rs
```diff
@@ -48,7 +56,19 @@ impl ProjectName {
...
             || hostname.ends_with('-')
```
chesedo, line 72:
```suggestion
            .any(|byte| !is_valid_char(byte))
            || !is_profanity_free_and_not_reserved(hostname)
            || hostname.ends_with('-')
```
I think this is what @coszio means. That way the same `is_profanity_free_and_not_reserved(hostname)` is not made with every byte check

common/src/project.rs
Butch78: Agreed good save @coszio 👍

common/src/project.rs
Butch78: Easy fix, I'll do it now! :)

chesedo:
@Butch78 we just need to merge in the underscore check that was merged to `main` recently

PR #295, opened 2022-08-05
fix(blog): typos

oddgrd:
Nice post! :smile: Here's a few corrections for some typos I found as I was reading it.


chesedo (approved):
Great catch, thanks @oddgrd

PR #296, opened 2022-08-05
docs(examples): update url-shortener example README

coszio:
Found out that the README of the url-shortener example had an outdated command, which I updated in this quick PR.


chesedo (commented):
Thanks for the fixup @coszio

Issue #262, opened 2022-07-14
Support underscore in project name


jacobsvante:
To harmonize with crate naming I think Shuttle supporting underscore in the project name as well would be reasonable. After all, underscore is a legal character in a domain name.

MRE:

```
cargo install cargo-shuttle@0.4.0
git clone git@github.com:shuttle-hq/shuttle.git
cd shuttle/examples/axum/hello-world
rm Shuttle.toml
sed -i.bkp 's/hello-world/hello_world/g' Cargo.toml
cargo shuttle run
```

Last line gives the following output:
```
Error:
`hello_world` is an invalid project name. project name must
1. not start or end with `-`.
2. not contain any characters outside of the alphanumeric range, except for `-`.
3. not be empty.
```

coszio:
I can take this one

chesedo:
Closes by #299

PR #299, opened 2022-08-09
feat: support underscore in project name

coszio:
## Motivation 

As suggested in #262, naming a project subdomain should allow underscores within the name in the same way it allows hyphens.

In the issue it is stated that the intention is to harmonize with crate naming, but such naming is a still ongoing discussion whether `kebab-case` or `snake_case` should be recommended ([which can be seen here](https://github.com/rust-lang/api-guidelines/discussions/29)). Since one or the other is not standardized, I think shuttle should allow the underscore as well.

## Implementation

Underscores are treated in the same way as hyphens: they can only be within the name of the subdomain, not at the end nor at the start. 
Also, when looking at the code I swapped the implementation of a valid character with a `matches!` macro to use more idiomatic Rust.

### DIFF
--- a/common/src/project.rs
+++ b/common/src/project.rs
@@ -9,8 +9,8 @@ use serde::{Deserialize, Deserializer, Serialize};
 /// Project names should conform to valid Host segments (or labels)
 /// as per [IETF RFC 1123](https://datatracker.ietf.org/doc/html/rfc1123).
 /// Initially we'll implement a strict subset of the IETF RFC 1123, concretely:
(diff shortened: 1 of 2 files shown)


chesedo (approved):
Nice, and thanks for the match cleanup @coszio

chesedo:
@coszio this just needs a quick `cargo fmt --all`

coszio:
@chesedo Totally missed it

coszio:
This also closes #262, forgot to add it in the description :)

PR #301, opened 2022-08-12
blog: Authentication tutorial blog post

kaleidawave:
Tutorial on some of the parts of building: https://github.com/kaleidawave/axum-shuttle-postgres-authentication-demo


www/_blog/2022-08-11-authentication-tutorial.mdx
```diff
@@ -0,0 +1,516 @@
...
+            *store = Some(User { username });
+        }
+    }
+    store.as_ref()
+}
+```
+
+Here because we store the user internally it means that if another middleware gets the user and then a handler gets the user it results in one database request not two!
+
+[The full demo includes registering the middleware](https://github.com/kaleidawave/axum-shuttle-postgres-authentication-demo/blob/ba71a914055f312636581f5e82172b1078e7b9eb/src/lib.rs#L53-L68).
```
chesedo, line 443: Maybe we want to include the registration code at the minimum to complete this post?

chesedo (commented):
I just have comment and questions... some might be completely wrong though :sweat_smile:

www/_blog/2022-08-11-authentication-tutorial.mdx
```diff
@@ -0,0 +1,597 @@
...
+async fn axum(
+    #[shared::Postgres] pool: Database
+) -> ShuttleAxum {
+    let router = Router::new()
+        .route("/", get(index))
+        .route("/styles.css", get(styles))
+        .layer(Extension(Arc::new(tera)))
+        .layer(database)
+    Ok(sync_wrapper::SyncWrapper::new(get_router(pool)))
+}
```
chesedo, line 201:
```suggestion
) -> ShuttleAxum {
    // Build tera as before
    
    let router = Router::new()
        .route("/", get(index))
        .route("/styles.css", get(styles))
        .layer(Extension(Arc::new(tera)))
        .layer(pool)
    
    // Wrap and return router as before
}

```

www/_blog/2022-08-11-authentication-tutorial.mdx
```diff
@@ -0,0 +1,597 @@
...
+        data.get("password"),
+        data.get("confirm_password"),
+    ) {
+        if password != confirm_password {
+            return Err(error_page(&SignupError::PasswordsDoNotMatch));
+        }
+
+        // We will create users here
+
+        Ok(())
```
chesedo, line 228: It's a little hard to follow how the code pieces below need to be added here. Maybe this should be a function call (`try_to_add_user()`) and the code pieces below be in the block of this called function?

chesedo: Or maybe there should be three comment lines here?
1. Hash password
1. Insert user and get user id
1. Create session and set cookie

chesedo: Or 3+ function calls with the code pieces below defining each function :man_shrugging:

www/_blog/2022-08-11-authentication-tutorial.mdx
```diff
@@ -0,0 +1,597 @@
...
+
+This is unlike [Pseudo-random number generators where you can predict the next *random* number given a start point and the algorithm](https://www.youtube.com/watch?v=-h_rj2-HP2E). This could be a problem if we have our token we can get the session token of the person who logged in after us really easy and thus impersonate them.
+
+To start the random we have to set up a seed. We will be doing something similar to the creation of the Tera instance. We must wrap it in an arc and a mutex because generating new identifiers requires mutable access. We then start this in our main function with `let random = ChaCha8Rng::seed_from_u64(OsRng.next_u64());` and add it as an extension to our router chain with `.layer(Extension(Arc::new(Mutex::new(random))))`.
```
chesedo, line 340: Maybe we should rather have the code context from line ~188 again over the textual description?

www/_blog/2022-08-11-authentication-tutorial.mdx
```diff
@@ -0,0 +1,597 @@
...
+    session_token
+}
+```
+
+In the full demo, we use the [new type pattern](https://www.shuttle.rs/blog/2022/07/28/patterns-with-rust-types#the-new-type-pattern) over a u128 to make this easier, but we will stick with a u128 type here.
+
+Now we have our token we need to package it into a cookie value. We will do it in the simplest way possible, using `.to_string()`. We will send a response that does two things, sets this new value and returns/redirects us back to the index page. Still in the signup code we now do:
+
+```rust
+let session_token = new_session(database, random, user_id);
```
chesedo, line 372: From the reader's perspective, where does `random` come from?

www/_blog/2022-08-11-authentication-tutorial.mdx
chesedo: A second code snippet might be needed here to show the update for the `post_signup` function signature. (This relates to the `random` comment)

www/_blog/2022-08-11-authentication-tutorial.mdx
```diff
@@ -0,0 +1,597 @@
...
+```
+
+Now we have our value we now need to make a query that gets user information. We can do that using SQL joins
+
+```sql
+SELECT id, username 
+FROM users 
+JOIN sessions ON user_id = id 
+WHERE session_token = $1;
+```
```
chesedo, line 412: I think this can safely be removed as it's repeated in `get_user()` below

www/_blog/2022-08-11-authentication-tutorial.mdx
```diff
@@ -0,0 +1,597 @@
...
+        .insert(AuthState(session_token.map(|v| (v, None, database))));
+
+    next.run(req).await
+}
+```
+
+Then we create a method on `AuthState` which makes the database request.
+
+```rust
+pub async fn get_user(&mut self) -> Option<&User> {
```
chesedo, line 441: Maybe just include the `impl AuthState {` line for better context here

www/_blog/2022-08-11-authentication-tutorial.mdx
```diff
@@ -0,0 +1,597 @@
...
+.route("/login", get(get_login).post(post_login))
+```
+
+## Deployment
+
+This is great, we now have a full authenticated site. But we have no users, our friends can't log in on our localhost. We want to put it on the interwebs and with our set up shuttle service main function set up it is easy.
+
+`cargo shuttle deploy`
+
+Now you can go ahead with these concepts and add login, logout, delete, and listing user functionality. [The full demo implements these if you are looking for clues](https://github.com/kaleidawave/axum-shuttle-postgres-authentication-demo).
```
chesedo, line 579:
```suggestion
Now you can go ahead with these concepts and add delete, and listing user functionality. [The full demo implements these if you are looking for clues](https://github.com/kaleidawave/axum-shuttle-postgres-authentication-demo).

```

www/_blog/2022-08-11-authentication-tutorial.mdx
```diff
@@ -0,0 +1,597 @@
...
+        }
+
+        let session_token = new_session(database, random, user_id);
+
+        let resp = http::Response::builder()
+            .status(http::StatusCode::SEE_OTHER)
+            .header("Location", "/")
+            .header("Set-Cookie", format!("session_token={}; Max-Age=999999", session_token))
+            .body(http_body::Empty::new())
+            .unwrap()
```
chesedo, line 557: Maybe there should be a function (`set_cookie`) with this code for both handlers to use?

www/_blog/2022-08-11-authentication-tutorial.mdx
chesedo: The function calls suggestion also ties in with the `set_cookie` refactor comment

chesedo (commented):
This seems like a lot of comments, but there are actually only 2 main ideas with me babbling and trying to figure out how best to solve them :see_no_evil:

www/_blog/2022-08-11-authentication-tutorial.mdx
```diff
@@ -366,24 +410,57 @@ pub(crate) async fn new_session(database: &Database, random: Arc<Mutex<ChaCha8Rn
 
 In the full demo, we use the [new type pattern](https://www.shuttle.rs/blog/2022/07/28/patterns-with-rust-types#the-new-type-pattern) over a u128 to make this easier, but we will stick with a u128 type here.
 
-Now we have our token we need to package it into a cookie value. We will do it in the simplest way possible, using `.to_string()`. We will send a response that does two things, sets this new value and returns/redirects us back to the index page. Still in the signup code we now do:
+Now we have our token we need to package it into a cookie value. We will do it in the simplest way possible, using `.to_string()`. We will send a response that does two things, sets this new value and returns/redirects us back to the index page. We will create a utility function for doing this:
```
ivancernja, line 413: * Now that we have our token, we need to package it

PR #302, opened 2022-08-14
fix: make profanity filter zealous, add tests to CI

oddgrd:
Fixes the new profanity filter test that failed because it wasn't zealous enough. Alternatively words can be added to or removed from the `censor` crate's word lists easily with the addition/subtraction operators, e.g. `Zealous - "bad word"`. So if you think the `Zealous` list is too strict I can remove words.

I've also included the `shuttle-common` crate in the Github `test` workflow so these new tests run in CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,7 +35,7 @@ jobs:
   platform_tests:
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


chesedo (approved):
Nice and quick, thanks @oddgrd

oddgrd:
Oops, it seems hello-world is too crude now. I'll just remove "hell" from the list.

PR #303, opened 2022-08-14
feat: serenity integration

### LINKED ISSUE #281: Add serenity as a service for building Discord bots
[Serenity](https://github.com/serenity-rs/serenity) is a framework for building [Discord bots](https://discord.com/developers/docs/intro#bots-and-apps). Implementing `Service` for it under a flag would make it possible to host discord bots on Shuttle.

```rust
use serenity::Client;

#[shuttle_service::main]
async fn init() -> impl shuttle_service::SerenityService {
    let token = env::var("DISCORD_TOKEN").expect("token");
    let intents = GatewayIntents::non_privileged() | GatewayIntents::MESSAGE_CONTENT;

    let mut client = Client::builder(token, intents)
        .event_handler(...)
        .framework(...)
        .await
        .expect("Error creating client");

    Ok(client)
}
```

This should be a simple addition Serenity's client has a simple start function which selects and connects to a free port via web sockets:

```rust
#[async_trait]
impl Service for serenity::client::Client {
    async fn bind(mut self: Box<Self>, _addr: SocketAddr) -> Result<(), error::Error> {
        self
            .start()
            .await
            .map_err(error::CustomError::new)
    }
}
```

oddgrd:
This PR implements `Service` for `serenity::Client`. This enable Discord bot-developers using Serenity to easily deploy bots with with or without databases. I've added two usage examples, a basic `hello-world` bot and a `todo-list` bot with `shared::Postgres` for persistence.

~~The integration works in it's current state with a valid token, but is a WIP as I am trying to figure out how to write tests for it. It's not straight-forward as the bot listens for commands via web-socket from Discord's servers. A valid discord token is also needed for the bot client to start.~~

Closes #281


kaleidawave:
Looks great so far! Would also be good if we could get `cargo shuttle init --serenity` as well. Would make writing a bot super simple!

examples/serenity/hello_world/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+#[shuttle_service::main]
+async fn serenity() -> shuttle_service::ShuttleSerenity {
+    // Configure the client with your Discord bot token in the environment.
+    let token = std::env::var("DISCORD_TOKEN").expect("Expected a token in the environment");
+
+    // Set gateway intents, which decides what events the bot will be notified about
+    let intents = GatewayIntents::GUILD_MESSAGES
+        | GatewayIntents::DIRECT_MESSAGES
```
chesedo, line 31: I'm not sure if the `DIRECT_MESSAGES` intent does anything :man_shrugging:

examples/serenity/postgres/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+#[async_trait]
+impl EventHandler for Bot {
+    async fn message(&self, ctx: Context, msg: Message) {
+        // The user_id of the user sending a command
+        let user_id = msg.author.id.0 as i64;
+
+        // Add a new todo using `~todo add <note>` and persist it in postgres
+        if let Some(note) = msg.content.strip_prefix("~todo add") {
```
chesedo, line 25: I wonder if it is worth modelling this example after the [slash commands](https://github.com/serenity-rs/serenity/blob/current/examples/e14_slash_commands/src/main.rs) to allow invoking these with `/todo add` instead of `~todo add`?

examples/serenity/hello_world/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+use std::env;
+
+struct Bot;
+
+#[async_trait]
+impl EventHandler for Bot {
+    async fn message(&self, ctx: Context, msg: Message) {
+        if msg.content == "!hello" {
+            if let Err(why) = msg.channel_id.say(&ctx.http, "world!").await {
+                println!("Error sending message: {:?}", why);
```
chesedo, line 14: Maybe all the `println!` in the examples should be normal log `info!` or `error!` instead? This way they will appear in the runtime logs

chesedo (commented):
Oh, this is exciting @oddgrd!! Just some minor questions :smiley:

examples/serenity/hello_world/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+
+[dependencies]
+serenity = { version = "0.11.5", default-features = false, features = ["client", "gateway", "rustls_backend", "model"] }
+shuttle-service = { version = "0.5.0", features = ["bot-serenity", "secrets", "sqlx-postgres"] }
+sqlx = { version = "0.6", features = ["runtime-tokio-native-tls", "postgres"] }
+tracing = "0.1.36"
```
chesedo, line 12: This should still be `log` until we completed the switch to tracing (else they will be lost if I'm remember correctly)

service/Cargo.toml
```diff
@@ -24,6 +24,7 @@ paste = "1.0.8"
 poem = { version = "1.3.40", optional = true }
 regex = "1.6.0"
 rocket = { version = "0.5.0-rc.2", optional = true }
+serenity = { version = "0.11.5", default-features = false, features = ["client", "gateway", "rustls_backend", "model"], optional = true }
```
chesedo, line 27: Only the `client` feature is used by `shuttle_service`, right?

examples/serenity/hello_world/Cargo.toml
oddgrd: It seems to work with the tracing Serenity uses under the hood, but I'll change it for now if there is some issue with this currently.

service/Cargo.toml
oddgrd: One would think so, but I tried removing all but `client` and adding them back one by one. It actually won't compile without all of them. :thinking:

chesedo (approved):

PR #311, opened 2022-08-19
chore: dependencies upgrades

chesedo:
Upgrades a bunch of dependencies. Also closes #297


examples/rocket/postgres/Cargo.toml
```diff
@@ -9,4 +9,4 @@ edition = "2021"
 rocket = { version = "0.5.0-rc.1", features = ["json"] }
```
oddgrd, line 9: Maybe these too while you're at it? :slightly_smiling_face:

PR #314, opened 2022-08-23
fix: init rocket dependency version

### LINKED ISSUE #313: cargo shuttle init --rocket --hello-world
When using the init rocket example you'll get the following error:

image

This is due to the wrong dependencies for Rocket being added to the cargo.toml as shown belown:
image

It should instead be ```0.5.0-rc.2"```

coszio:
fixes #313, which needed the prerelease version of `rocket`.

Thanks for the tip @chesedo, and for spotting the issue @Butch78 :smile:

### DIFF
--- a/cargo-shuttle/src/init.rs
+++ b/cargo-shuttle/src/init.rs
@@ -415,7 +415,8 @@ fn set_inline_table_dependency_version(
     flag_allow_prerelease: bool,
     get_dependency_version_fn: GetDependencyVersionFn,
 ) {
-    let dependency_version = get_dependency_version_fn(crate_name, flag_allow_prerelease, manifest_path, url);
+    let dependency_version =
+        get_dependency_version_fn(crate_name, flag_allow_prerelease, manifest_path, url);
     dependencies[crate_name]["version"] = value(dependency_version);
 }
 
--- a/cargo-shuttle/src/init.rs
+++ b/cargo-shuttle/src/init.rs
@@ -425,7 +425,10 @@ type GetDependencyVersionFn = fn(&str, &Path, &Url) -> String;
 /// Gets the latest version for a dependency of `crate_name`.
 /// This is a wrapper function for `cargo_edit::get_latest_dependency` function.
 fn get_latest_dependency_version(crate_name: &str, manifest_path: &Path, url: &Url) -> String {
-    let latest_version = get_latest_dependency(crate_name, false, manifest_path, Some(url))
+    // some crates need to be on the prerelease version
+    let flag_allow_prerelease = matches!(crate_name, "rocket");
+
+    let latest_version = get_latest_dependency(crate_name, flag_allow_prerelease, manifest_path, Some(url))
         .unwrap_or_else(|_| panic!("Could not query the latest version of {}", crate_name));
     let latest_version = latest_version
         .version()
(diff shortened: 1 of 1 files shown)


cargo-shuttle/src/init.rs
```diff
@@ -425,7 +425,10 @@ type GetDependencyVersionFn = fn(&str, &Path, &Url) -> String;
 /// Gets the latest version for a dependency of `crate_name`.
 /// This is a wrapper function for `cargo_edit::get_latest_dependency` function.
 fn get_latest_dependency_version(crate_name: &str, manifest_path: &Path, url: &Url) -> String {
-    let latest_version = get_latest_dependency(crate_name, false, manifest_path, Some(url))
+    // some crates need to be on the prerelease version
+    let flag_allow_prerelease = matches!(crate_name, "rocket");
```
chesedo, line 429: This does fix it for rocket, but I'm not sure if this is the best approach for long-term maintenance though. This function will be "injected" into the template generation process. It, therefore, doing this "side-effect" seem to be a bit unexpected.

Can't a `bool` rather be added to the argument list of `GetDependencyVersionFn`? This will mean also updating `set_key_value_dependency_version`

chesedo (changes requested):
Thanks for the speedy PR @coszio!

cargo-shuttle/src/init.rs
coszio: You are right, doing it within the template makes more sense. I did consider that, but then figured that we already had the `crate_name` in the arguments. 

Maintainance-wise is simpler to just add the boolean in the template, yes.

Will update soon

cargo-shuttle/src/init.rs
```diff
@@ -390,8 +395,10 @@ fn set_key_value_dependency_version(
     manifest_path: &Path,
     url: &Url,
     get_dependency_version_fn: GetDependencyVersionFn,
+    flag_allow_prerelease: bool,
 ) {
-    let dependency_version = get_dependency_version_fn(crate_name, manifest_path, url);
+    let dependency_version =
+        get_dependency_version_fn(crate_name, flag_allow_prerelease, manifest_path, url);
```
coszio, line 401: Made the `flag_allow_prerelease` bool as a function argument for `set_value_dependency_version` --which sets the version as `crate_name = "version"`-- to be able to pass it within the implementation of each framework. However, now the `set_inline_table_dependency_version` --which sets the version as `crate_name = { version = "version" }`-- does not receive the same arguments, and becomes asymmetrical. 

Should I also update the latter?

coszio: Nvm, I just answered my own question :facepalm:

cargo-shuttle/src/init.rs
```diff
@@ -725,4 +751,30 @@ mod shuttle_init_tests {
...
+            &url,
+            get_latest_dependency_version,
+        );
+
+        let version = dependencies["rocket"].as_str().unwrap();
+
+        let expected = get_latest_dependency("rocket", true, &manifest_path, Some(&url))
+            .expect("Could not query the latest version of rocket")
+            .version()
+            .expect("no shuttle-service version found")
```
chesedo, line 775:
```suggestion
            .expect("no rocket version found")
```

chesedo (approved):
This is great, thanks @coszio!

PR #329, opened 2022-09-07
feat: Added UpdateShuttle command to update ShuttleCommand

Butch78:
I added Subcommand for Updating ShuttleCargo from within Cargo Shuttle, I wasn't sure how to test this and if this is the best strategy but I thought this would be a great conversation starter :)

### DIFF
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -52,7 +52,7 @@ pub enum Command {
     Init(InitArgs),
     /// view the status of a shuttle project
     Status,
-    /// Update Cargo shuttle 
+    /// Update Cargo shuttle
     UpdateShuttle,
     /// view the logs of a shuttle project
     Logs,
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -52,6 +52,8 @@ pub enum Command {
     Init(InitArgs),
     /// view the status of a shuttle project
     Status,
+    /// Update Cargo shuttle 
+    UpdateShuttle,
     /// view the logs of a shuttle project
     Logs,
     /// delete the latest deployment for a shuttle project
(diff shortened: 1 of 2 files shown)


cargo-shuttle/src/args.rs
```diff
@@ -52,6 +52,8 @@ pub enum Command {
     Init(InitArgs),
     /// view the status of a shuttle project
     Status,
+    /// Update Cargo shuttle
```
brokad, line 55:
```suggestion
    /// update cargo-shuttle
```

cargo-shuttle/src/args.rs
```diff
@@ -52,6 +52,8 @@ pub enum Command {
...
+    UpdateShuttle,
```
brokad, line 56:
```suggestion
    Update,
```

brokad: I like the semantics of `cargo shuttle update` to get the new version of shuttle installed, at which point this is basically `cargo install cargo-shuttle` (or whatever binary distro we end up doing down the line); and `cargo shuttle upgrade` to bump the `shuttle-service` version of people's projects.

cargo-shuttle/src/client.rs
```diff
@@ -62,6 +62,14 @@ pub(crate) async fn delete(
     Ok(())
 }
 
+pub(crate) async fn update_shuttle() -> Result<()> {
+    // run this command: "cargo install cargo-shuttle"
+
+    clap::command!("cargo", "install", "cargo-shuttle").run()?;
```
brokad, line 68: This does not build for me! Looking at the [docs](https://docs.rs/clap/latest/clap/macro.command.html) I think this is supposed to be used to create a stub `Command` instance with things like author and version pre-filled from `Cargo.toml`.

brokad:
Happy to reopen this if work restarts on here.

PR #337, opened 2022-09-09
Add colour to `cargo shuttle run` when building project

Milo123459:
This runs `cargo build --release` with `--color always` to provide colour rather then plain white text.


Milo123459:
not sure why ci is failing

chesedo:
@Milo123459 nice! This currently only adds coloured output for the tests, right?

If I remember correctly, we somehow have to figure out how to turn on colour when we call [`cargo::compile` here](https://github.com/shuttle-hq/shuttle/blob/60011c41b59e9e937fcabb8335b16f1268226135/service/src/loader.rs#L156)

chesedo:
Oh, don't worry too much about the CI failing for now. That's a known issue we'll have to fix another day :sweat_smile:

Milo123459:
updated, any other comments?

common/src/version.rs
```diff
@@ -23,6 +23,8 @@ pub fn get_shuttle_service_from_user_crate<P: AsRef<Path>>(manifest_path: P) ->
             "--format",
             "{p}",
         ])
+        .arg("--color")
```
chesedo, line 26: Hmm... This turns on colouring when we check the versions. It still should not affect `cargo shuttle run` when you try it locally right?

Milo123459: You told me to make it work in the compile function ig, so it does

chesedo: I'm confused, how is this the compile function I linked to? This is the `get_shuttle_service_from_user_crate` function

Milo123459: oh god, I got mixed up, sorry

chesedo:
@Milo123459 are you going to take another swing at this?

PR #338, opened 2022-09-12
refactor: change references of `project` to `service`

oddgrd:
Changes references of `project` to `service` to bring the naming up to date with the new architecture in the `feat/deployer` branch.

### DIFF
--- a/deployer/Cargo.toml
+++ b/deployer/Cargo.toml
@@ -2,7 +2,7 @@
 name = "shuttle-deployer"
 version = "0.1.0"
 edition = "2021"
-description = "Service with instances created per project for handling the compliation, loading, and execution of Shuttle services"
+description = "Service with instances created per project for handling the compilation, loading, and execution of Shuttle services"
 
 [dependencies]
 anyhow = "1.0.58"
--- a/deployer/src/deployment/provisioner_factory.rs
+++ b/deployer/src/deployment/provisioner_factory.rs
@@ -1,7 +1,7 @@
 use std::collections::BTreeMap;
 
 use async_trait::async_trait;
-use shuttle_common::{database, project::ProjectName, DatabaseReadyInfo};
+use shuttle_common::{database, project::ProjectName as ServiceName, DatabaseReadyInfo};
 use shuttle_proto::provisioner::{
     database_request::DbType, provisioner_client::ProvisionerClient, DatabaseRequest,
 };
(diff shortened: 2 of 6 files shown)


deployer/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "shuttle-deployer"
 version = "0.1.0"
 edition = "2021"
-description = "Service with instances created per project for handling the compliation, loading, and execution of Shuttle services"
+description = "Service with instances created per service for handling the compilation, loading, and execution of Shuttle services"
```
brokad, line 5:
```suggestion
description = "Service with instances created per project for handling the compilation, loading, and execution of Shuttle services"
```

deployer/README.md
```diff
@@ -1,6 +1,6 @@
 # Deployment System
 
-Service that manages the building, loading, and deployment of the Shuttle service(s) that make up a user's Shuttle project.
+Service that manages the building, loading, and deployment of the Shuttle service(s) that make up a user's Shuttle service.
```
brokad, line 3:
```suggestion
Service that manages the building, loading, and deployment of the Shuttle service(s) that make up a user's Shuttle project.
```

deployer/src/args.rs
```diff
@@ -2,7 +2,7 @@ use clap::Parser;
 use fqdn::FQDN;
 use shuttle_common::Port;
 
-/// Program to handle the deploys for a single project
+/// Program to handle the deploys for a single service
```
brokad, line 5:
```suggestion
/// Program to handle the deploys for a single project
```

brokad (commented):
The idea we had is that, within a *project*, a user can have one or more *services*. Exactly one of those *services* is exposed over HTTP at `${project_name}.shuttleapp.rs` - let's call it the *http service* of the user's *project*.

This means we still need `ProjectName` because projects still exist and the url in fact corresponds to the project name.

Then the idea is that, by default, the first deployment that a user makes under a new project is the *http service* of that project, and carries the same name as the project itself. So if we want to be pedantic we can have `type ServiceName = ProjectName` and replace where appropriate in the deployer.

TL;DR: We need to keep `ProjectName` and related `shuttle_common` items. We just need to change names and types, inside the deployer, of things that relate to `shuttle_service::Service` implementations, from *project* to *service*.

I've only commented the diff on a couple of those, because I figured it was easier to explain in the comment than inline in the review! I realise this was not super clear a priori, sorry :smile:

oddgrd:
Ah, I see! I'll update it, thanks for the review. :smiley:

deployer/src/deployment/provisioner_factory.rs
```diff
@@ -101,7 +101,7 @@ impl<R: ResourceRecorder, S: SecretGetter> Factory for ProvisionerFactory<R, S>
         let db_type: DbType = db_type.into();
 
         let request = Request::new(DatabaseRequest {
-            project_name: self.project_name.to_string(),
+            project_name: self.service_name.to_string(),
```
oddgrd, line 104: Renaming this field means we will also need to rename it in the protobuffer and `provisioner`, not sure if that is what we want.

brokad: I agree with you, we shouldn't rename this.

brokad (approved):
LGTM 😄

PR #339, opened 2022-09-12
feat(deployer): split up deployer error enum

oddgrd:
Split up `deployer/src/error`, since the handlers use almost none of its variants.

### DIFF
--- a/deployer/src/handlers/error.rs
+++ b/deployer/src/handlers/error.rs
@@ -1,0 +1,42 @@
+use std::error::Error as StdError;
+
+use axum::http::{header, HeaderValue, StatusCode};
+use axum::response::{IntoResponse, Response};
+use axum::Json;
+
+use serde::{ser::SerializeMap, Serialize};
+use serde_json::json;
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Persistence failure: {0}")]
+    Persistence(#[from] crate::persistence::PersistenceError),
+}
+
+impl Serialize for Error {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let mut map = serializer.serialize_map(Some(2))?;
+        map.serialize_entry("type", &format!("{:?}", self))?;
+        map.serialize_entry("msg", &self.source().unwrap().to_string())?;
+        map.end()
+    }
+}
+
+impl IntoResponse for Error {
+    fn into_response(self) -> Response {
+        (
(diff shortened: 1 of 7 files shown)


deployer/src/handlers/error.rs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+impl Serialize for Error {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let mut map = serializer.serialize_map(Some(2))?;
+        map.serialize_entry("type", &format!("{:?}", self))?;
+        map.serialize_entry("msg", &self.source().unwrap().to_string())?;
```
brokad, line 23: As per `thiserror` doc `.source().unwrap()` is safe only for variants that have `#[from]` or `#[source]` so we should be a bit careful - perhaps we can just `.unwrap_or_default` or something?

deployer/src/handlers/error.rs
```diff
@@ -0,0 +1,42 @@
...
+    Persistence(#[from] crate::persistence::PersistenceError),
+}
+
+impl Serialize for Error {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let mut map = serializer.serialize_map(Some(2))?;
+        map.serialize_entry("type", &format!("{:?}", self))?;
```
brokad, line 22: I know you moved this over from somewhere else, so not really part of this PR, but I couldn't help but notice:

Why `{:?}` and why `self`? Won't this inline the entire debug disposition of the error as its "type"? Would it be better to have a `match` and display the variant name?

brokad (approved):
LGTM 😄 

The two comments I have actually came from code you inherited in the diff!

deployer/src/handlers/error.rs
oddgrd: Yeah, this will have the entire debug representation as the "type". With an example error the serialized output will be: `{"message":{"type":"Persistence(Database(PoolTimedOut))","msg":"Database error: pool timed out while waiting for an open connection"}}`. It's a bit verbose and I can change this to just match and return variant name, but I think the explicitness of this is kinda nice for internal errors.

deployer/src/handlers/error.rs
oddgrd: I agree, this is a good point. Since we implement `Display` through `thiserror` for all the deployer errors, we can do `&self.source().unwrap_or(self).to_string()`. This way we will get either the source or the `Display` representation. If we add a new error to `handlers::Error`, this will not compile if the new error doesn't implement `Display`.

PR #343, opened 2022-09-15
feat: shell completions

XyLyXyRR:


cargo-shuttle/src/args.rs
```diff
@@ -50,6 +51,15 @@ pub enum Command {
     Deploy(DeployArgs),
     /// create a new shuttle project
     Init(InitArgs),
+    /// generate shell completions
+    Generate {
+        /// which shell
+        #[clap(short, long, env)]
+        shell: Option<Shell>,
```
chesedo, line 58: Rather than `Option` can the default be set to bash?

cargo-shuttle/src/args.rs
```diff
@@ -50,6 +51,15 @@ pub enum Command {
...
+        /// output file
+        #[clap(short, long, env)]
+        output: Option<PathBuf>,
```
chesedo, line 61: Same here, can the default be set to stdout and the `Option` be dropped?

cargo-shuttle/src/lib.rs
```diff
@@ -154,6 +158,23 @@ impl Shuttle {
         .context("failed to delete deployment")
     }
 
+    async fn complete(&self, shell: Option<Shell>, output: Option<PathBuf>) -> Result<()> {
+        let name = env!("CARGO_PKG_NAME");
+        tokio::task::spawn_blocking(move || -> Result<()> {
```
chesedo, line 163: I'm missing why the spawn is needed. What happens when it is removed?

chesedo (changes requested):
This is quite interesting. Thanks @XyLyXyRR

cargo-shuttle/src/lib.rs
```diff
@@ -154,6 +158,18 @@ impl Shuttle {
         .context("failed to delete deployment")
     }
 
+    async fn complete(&self, shell: Shell, output: Option<PathBuf>) -> Result<()> {
+        let name = self.ctx.project_name().as_str();
+        let mut app = Command::command();
+        let mut buf = match output {
+            Some(v) => Either::Left(File::create(v)?),
+            None => Either::Right(stdout()),
```
chesedo, line 166: Can you find a way to drop the use of `Either`?

cargo-shuttle/src/main.rs
```diff
@@ -1,11 +1,18 @@
...
 
 #[tokio::main]
 async fn main() -> Result<()> {
     env_logger::init();
 
+    env::var("SHELL").ok().map(|x| {
+        PathBuf::from(x)
+            .file_name()
+            .map(|x| env::set_var("SHELL", x))
+    });
```
chesedo, line 14: Why is this needed?

chesedo (commented):
Sorry for the extra comments :sweat_smile:

cargo-shuttle/src/main.rs
chesedo: Since the `FromStr` commit code has been removed, I think it is safe to remove this too right?

chesedo: For the panic, I think that line was correct as `env!("CARGO_PKG_NAME")` since it should get the name of cargo-shuttle and not the project if I understand correctly :man_shrugging:

chesedo: The only thing still blocking this PR for is modifying the `SHELL` env var since it will be an unexpected side effect (seeing that shuttle is not a shell utility). I think this should rather be fixed as a PR to `clap_complete` to not introduce these lines to shuttle

cargo-shuttle/src/main.rs
```diff
@@ -1,6 +1,7 @@
 use anyhow::Result;
 use cargo_shuttle::{Args, CommandOutcome, Shuttle};
 use clap::Parser;
+use std::{env, path::PathBuf};
```
chesedo, line 4:
```suggestion
```
This likely causes the CI to fail since they are unused

chesedo (approved):
Very nice, thanks for this @XyLyXyRR!

chesedo:
@XyLyXyRR Thanks for this. Sorry, I got side tracked a bit by the 0.6.0 release. I'm merging this today

chesedo:
@XyLyXyRR would you mind merging main into this?

PR #355, opened 2022-09-27
feat(common): clean up deps passed to user crates

oddgrd:
Add feature flags to `shuttle-common` to reduce the dependencies passed through to user crates who depend on `shuttle-service`. I also removed the unused `regex` dep from `shuttle_service`, and put `async-std` behind the `web-tide` feature as it is only used there.

Without trimming these dependencies, building the `poem-hello-world` example took ~32s. After trimming it took ~18s, a ~43% improvement (on Ubuntu 22 with Intel Co i7-9700K). There was an issue with removing the rocket dependency, but it was solved by #324.


common/src/lib.rs
```diff
@@ -1,21 +1,24 @@
...
 pub mod resource;
+#[cfg(feature = "secret")]
 pub mod secret;
+#[cfg(feature = "service")]
 pub mod service;
+#[cfg(feature = "version")]
 pub mod version;
-
+#[cfg(feature = "log")]
 pub use log::Item as LogItem;
```
chesedo, line 15: Maybe we should just have a single "models" feature for all the response models send between gateway/deployer and the client? We could even put them in a "models" or whatever namespace...

.gitignore
```diff
@@ -18,7 +18,7 @@ Cargo.lock
 .DS_Store
 
 # Terraform
-**/secret*
+# **/secret*
```
oddgrd, line 21: Do we still need this? It ignores `secret.rs` in `common::models`.

PR #357, opened 2022-09-28
feat: support release profile

### LINKED ISSUE #346: Deploy in release mode
When deploying, the app is build in debug mode.

Building in release mode reduce the size of the binaries and the runtime speed.

christoshadjiaslanis:
We now support building in release mode:

Defaults to building with the `dev` profile (debug mode):
```bash
cargo shuttle deploy
```

optionally deploy with a release profile:
```bash
cargo shuttle deploy --release
```

`cargo shuttle run` has no option for `--release` currently and only builds with the `dev` profile.

Also introduced a new HTTP header to the API called `BuildConfig` with the idea that it is an extensible DTO for sending build configurations from the client to the server.

Closes #346.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4907,8 +4907,6 @@ dependencies = [
  "anyhow",
  "async-trait",
  "base64 0.13.0",
- "cargo",
- "cargo-util",
  "chrono",
  "clap",
  "env_logger",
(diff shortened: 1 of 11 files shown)


christoshadjiaslanis:
@chesedo I think this is ready to merge - do I just need to `cargo fmt`?

PR #358, opened 2022-09-28
feat: added field in Cargo.toml to avoid accidental publishing

### LINKED ISSUE #352: `shutte init` : add `publish: false` in cargo.toml
I suggest adding `publish: false` in the `cargo shuttle init` generated cargo.toml to avoid accidental `cargo deploy`.

christoshadjiaslanis:
Closes #352.


chesedo (approved):
Great, thanks @christoshadjiaslanis

chesedo:
@christoshadjiaslanis I was expecting some test to fail with this update :confused: Since none is failing it means we are missing a test case. Would you be able to add a unit test for `cargo_shuttle_init()`?

PR #368, opened 2022-10-03
fix: postgres container renaming

oddgrd:
The Postgres container was renamed from `pg` to `postgres` in #362, but the name wasn't changed in all the places it needed to, causing name resolution failures in the `provisioner`.

Note: there are more places the `pg` abbreviation is used, like file names and CLI args. Happy to change these too if we want to be completely consistent.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,13 +37,13 @@ ifeq ($(PROD),true)
 DOCKER_COMPOSE_FILES=-f docker-compose.yml
 STACK=shuttle-prod
 APPS_FQDN=shuttleapp.rs
-DB_FQDN=pg.shuttle.rs
+DB_FQDN=postgres.shuttle.rs
 CONTAINER_REGISTRY=public.ecr.aws/shuttle
 else
 DOCKER_COMPOSE_FILES=-f docker-compose.yml -f docker-compose.dev.yml
 STACK=shuttle-dev
 APPS_FQDN=unstable.shuttleapp.rs
-DB_FQDN=pg.unstable.shuttle.rs
+DB_FQDN=postgres.unstable.shuttle.rs
 CONTAINER_REGISTRY=public.ecr.aws/shuttle-dev
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -37,13 +37,13 @@ ifeq ($(PROD),true)
 DOCKER_COMPOSE_FILES=-f docker-compose.yml
 STACK=shuttle-prod
 APPS_FQDN=shuttleapp.rs
-DB_FQDN=postgres.shuttle.rs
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -37,13 +37,13 @@ ifeq ($(PROD),true)
...
 APPS_FQDN=shuttleapp.rs
-DB_FQDN=pg.shuttle.rs
+DB_FQDN=postgres.shuttle.rs
 CONTAINER_REGISTRY=public.ecr.aws/shuttle
 else
 DOCKER_COMPOSE_FILES=-f docker-compose.yml -f docker-compose.dev.yml
 STACK=shuttle-dev
 APPS_FQDN=unstable.shuttleapp.rs
-DB_FQDN=pg.unstable.shuttle.rs
+DB_FQDN=postgres.unstable.shuttle.rs
```
brokad, line 46: I believe that change also requires changing the tf files for route53 in `terraform-aws-shuttle`.

terraform/modules/shuttle/systemd/system/shuttle-pg.service.tftpl
```diff
@@ -11,7 +11,7 @@ Restart=always
 ExecStartPre=/usr/bin/docker pull postgres:11
 ExecStart=/usr/bin/docker run --rm \
 			      --network shuttle-net \
-			      --name pg \
+			      --name postgres \
```
brokad, line 14: I'm not sure why this is necessary for `v0.6.0rc1`? Or was this done to stay consistent across?

These tf files are now obsolete, the new ones being in `terraform-aws-shuttle`.

docker-compose.dev.yml
```diff
@@ -6,9 +6,9 @@ services:
       - /bin/bash
       - -c
       - |
-        until </dev/tcp/pg/5432
+        until </dev/tcp/postgres/5432
```
brokad, line 9: Thank you, sorry I forgot about this guy in #362!

brokad (commented):
Thank you @oddgrd 😄

terraform/modules/shuttle/systemd/system/shuttle-pg.service.tftpl
oddgrd: Ah! I'll leave this out then.

docker-compose.dev.yml
oddgrd: No problem!

Makefile
oddgrd: I'll leave this out as well then, for now at least.

oddgrd:
Closing this since Pieter fixed this in 0.6.0 in #369

PR #413, opened 2022-10-19
bug: remove user volume when project is deleted

chesedo:
Closes ENG-215


chesedo:
We can't do this as users will lose running deployments when using the `cargo shuttle project rm; cargo shuttle project new` sequence

PR #418, opened 2022-10-21
feat: project name must be lowercase

oddgrd:
Restrict project names to lowercase, and to be within the permitted length for hostname segments: 63 chars. Additionally, restrict using underscores in a project names, as it's not valid in hostnames.

Note: when we merge this we can remove the ad-hoc validation in gateway: https://github.com/shuttle-hq/shuttle/blob/main/gateway/src/service.rs#L464-L475


iulianbarbu:
@oddgrd , how should we approach this? It's still relevant?

jonaro00:
Closing in favor of #1223

PR #426, opened 2022-10-24
Feat/update contributing

oddgrd:
Update the contribution guide. 

When doing this, I noticed we don't do any clean up after e2e tests. This is fine for CI, but locally we will have conflicts for the admin user (name PK, key UNIQUE) as well as conflicts for the projects (a project with that name already exists). There will also currently be ~10 running deployers running after the tests complete, which won't be cleaned up with `make down`. 

To remedy this I did:
1. ~~If the admin username already exists, just update the key. This is a quick fix with a caveat: if a super-user is created with a different username but a non-unique key, the tests will fail. But if the tests are run without modifying the setup code, this shouldn't be a problem.~~ Remove `ensure_success` from the superuser creation in test setup.
2. Implement `Drop` for `helpers::Services`: when the test completes `project rm` will be called, and this should occur even if the test fails.


gateway/src/main.rs
```diff
@@ -138,12 +138,20 @@ async fn init(db: SqlitePool, args: InitArgs) -> io::Result<()> {
...
-        .execute(&db)
-        .await
-        .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+    query(
+        r#"
+    INSERT INTO accounts (account_name, key, super_user) 
+    VALUES (?1, ?2, 1)
+    ON CONFLICT (account_name)
+    DO
+        UPDATE SET key = ?2
```
chesedo, line 147: This might be a bit dangerous. It means that anyone who gains unauthorized access to the VM will be able to cycle any key to their choosing. Maybe we should rather drop the `ensure_success` when creating the user in e2e (and also change the username to `test`)?

chesedo (commented):
This is looking real great, thanks @oddgrd!!

I just have one concern

gateway/src/main.rs
oddgrd: That's a very good point, thanks Pieter! I'll implement your suggestion.

PR #446, opened 2022-10-31
Add docker-compose extra flags param in Makefile

brokad:


Makefile
```diff
@@ -71,7 +71,7 @@ postgres:
 	       $(POSTGRES_EXTRA_PATH)
 
 docker-compose.rendered.yml: docker-compose.yml docker-compose.dev.yml
-	$(DOCKER_COMPOSE_ENV) $(DOCKER_COMPOSE) $(DOCKER_COMPOSE_FILES) -p $(STACK) config > $@
+	$(DOCKER_COMPOSE_ENV) $(DOCKER_COMPOSE) $(DOCKER_COMPOSE_FILES) $(DOCKER_COMPOSE_FLAGS) -p $(STACK) config > $@
```
chesedo, line 74: I might be overlooking where we are setting `DOCKER_COMPOSE_FLAGS`? :see_no_evil: Ie how do we use this change?

brokad: Oh no no we're not using it otherwise, but it lets you add flags to this via setting the env variable from outside.

This is useful to me so I can tweak deployments further when I roll them out locally (with a `docker-compose.local.yml` of sorts) 😄

PR #447, opened 2022-10-31
fix: gateway state drifts, health checks and project recreation

brokad:


brokad:
This is a pretty big diff. And even though all of it is tested somehow, it'd definitely be nice to run this in staging for a while before committing it to a release.

gateway/src/lib.rs
```diff
@@ -237,29 +235,42 @@ pub trait EndStateExt<'c>: EndState<'c> {
...
+}
 
-pub trait IntoEndState<'c, E>
+pub trait IntoTryState<E>
 where
-    E: EndState<'c>,
+    E: TryState,
 {
-    fn into_end_state(self) -> Result<E, Infallible>;
+    fn into_try_state(self) -> Result<E, Infallible>;
```
chesedo, line 257: Maybe we should no longer call this generic `E` as it can be confused with e for error.

gateway/src/service.rs
```diff
@@ -402,7 +398,41 @@ impl GatewayService {
...
+            .bind(&project_name)
+            .bind(&account_name)
+            .fetch_optional(&self.db)
+            .await?
+        {
+            // If the project already exists and belongs to this account
+            let project = row.get::<SqlxJson<Project>, _>("project_state").0;
+            if project.is_destroyed() {
+                // But is in `::Destroyed` state, recreate it
+                let initial_key = row.get::<String, _>("initial_key");
```
chesedo, line 412: Should we consider cycling the initial key here?

gateway/src/service.rs
```diff
@@ -591,35 +568,110 @@ pub mod tests {
...
+
+        let mut task = svc
+            .new_task()
+            .account(neo.clone())
+            .project(matrix.clone())
+            .build();
+
+        while let TaskResult::Pending(_) = task.poll(()).await {
+            // keep polling
+        }
```
chesedo, line 637: Is there some build-in timeout to prevent this from polling indefinitely?

chesedo (commented):
This is looking great, thanks @brokad!

Just left some comments

gateway/src/service.rs
brokad: Not in the tests no, that's a good point. There is a timeout for `Project` tasks but not generically for these. I'll quickly add that.

gateway/src/service.rs
brokad: No prob, will do.

oddgrd (commented):
Looks very good to me, thanks! I guess in the end it was an M :smile:

chesedo (approved):

oddgrd (approved):

PR #450, opened 2022-11-01
WIP feat: start runtime from deployer

oddgrd:
First draft of starting the runtime and deploying the service in it via deployer.


runtime/src/legacy/mod.rs
```diff
@@ -58,8 +58,8 @@ impl Runtime for Legacy {
         &self,
         request: Request<StartRequest>,
     ) -> Result<Response<StartResponse>, Status> {
-        let port = 8001;
-        let address = SocketAddr::new(Ipv4Addr::LOCALHOST.into(), port);
+        let service_port = 8001;
```
chesedo, line 61: I wasn't very smart with these port numbers. Sorry :see_no_evil: 

So I think we have 4 to dance around at this point:
1. 8001 which is the control plane in deployer
2. 8000 which is the proxy in deployer
3. The random port deployer (used to) start the service on
4. 8001 which I had here again. Maybe just move this to 6001 or 7001

(For local I would start provisioner on 5001)

deployer/src/lib.rs
```diff
@@ -54,12 +55,15 @@ pub async fn start(
...
 
-    info!("Binding to and listening at address: {}", args.api_address);
-
-    axum::Server::bind(&args.api_address)
-        .serve(make_service)
-        .await
-        .unwrap_or_else(|_| panic!("Failed to bind to address: {}", args.api_address));
+    select! {
+        _ = tokio::spawn(shuttle_runtime::start_legacy()) => {
+            info!("Legacy runtime stopped.")
```
chesedo, line 60: So this would be some [`Command::new()`](https://doc.rust-lang.org/std/process/struct.Command.html) to start the binary in a new process.

chesedo (commented):
Added some comment that will hopefully unstuck you :smile:

deployer/Cargo.toml
```diff
@@ -54,6 +54,10 @@ version = "0.7.0"
 path = "../service"
 features = ["loader"]
 
+[dependencies.shuttle-runtime]
+version = "0.1.0"
+path = "../runtime"
+
```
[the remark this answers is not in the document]
oddgrd: Ah! That makes sense.

proto/runtime.proto
```diff
@@ -31,7 +31,8 @@ message StartResponse {
   // Was the start successful
   bool success = 1;
 
+  // todo: find a way to add optional flag here
   // Optional port the service was started on
   // This is likely to be None for bots
-  optional uint32 port = 2;
```
oddgrd, line 36: This was causing an error about optional not being allowed in proto3 syntax, even though it should be in the versions we use. Ref: https://github.com/bufbuild/buf/issues/275

This meant I also ended up adding a temp hardcoded port to the `next` runtime bot.

deployer/src/lib.rs
```diff
@@ -54,12 +55,23 @@ pub async fn start(
     );
     let make_service = router.into_make_service();
 
-    info!("Binding to and listening at address: {}", args.api_address);
+    // todo: don't hardcode this path
+    let runtime_dir = PathBuf::from_str("/home/oddgrd/Documents/shuttle/target/debug").unwrap();
```
chesedo, line 59: Can we make this `MANIFEST_PATH` + "target/debug" possibly?

chesedo (commented):
This seems good to me, thanks @oddgrd.

We might just want to ensure the path to runtime will work for me and D too :smile:

deployer/src/lib.rs
oddgrd: Woops, sure thing!

PR #451, opened 2022-11-01
fix: wrap around common::ProjectName for parsing

brokad:


gateway/src/lib.rs
```diff
@@ -12,8 +12,6 @@ use axum::response::{IntoResponse, Response};
 use axum::Json;
 use bollard::Docker;
 use futures::prelude::*;
-use once_cell::sync::Lazy;
-use regex::Regex;
```
chesedo, line 16: Are we able to drop regex in `Cargo.toml` too?

chesedo (commented):
I think this is good to go, just checking if we can trim `Cargo.toml` too :smile:

PR #456, opened 2022-11-03
misc: add more helpful flags to Makefile

brokad:
And do not run "images" target when targeting "up". 

On my end, this still triggers a build of the images initially, unless when `DOCKER_COMPOSE_FLAGS=--no-build` is set, which matches the [documented behavior](https://docs.docker.com/engine/reference/commandline/compose_up/#description). Contributing guidelines also make it clear you want to run `make images` first in order to sync up images with their most recent versions, and the CircleCI workflows that do `make up` and `make test` also run `make images` prior. So this should not require changing anything else.


chesedo (approved):
Seems all good to me. LGTM

PR #457, opened 2022-11-03
UNC Paths and compilation on Windows

rutexd:
Fixes:

1. Gateway service compile error: Since Shuttle is Docker and Linux first oriented project, Shuttle gateway service uses Docker linux-first functions which defined through conditional compiling by cfg. This is the issue under windows, since you have to cut out project member from compilation by hand.
`Replacing Docker::connect_with_unix` with `Docker::connect_with_local` does the same thing as before for unix but allows to be compiled under windows, for cases where if you want to complie Shuttle-CLI or work with other non-server stuff.

2. Fixes (partially) Windows paths for cargo-shuttle deployer. Since Rust tries to support new Windows UNC paths, which are new and compatibility breaking, this support isnt perfect and rust has some bugs with it. As the example - cargo package may break the package on git initialized project in some unclear conditions and generate instead of working package - package filled with zeros only because of UNC path.

### DIFF
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -1,6 +1,6 @@
 use std::{
     ffi::OsStr,
-    fs::{canonicalize, create_dir_all},
+    fs::{create_dir_all},
     io::{self, ErrorKind},
     path::PathBuf,
 };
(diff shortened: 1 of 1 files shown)


cargo-shuttle/src/args.rs
```diff
@@ -183,7 +183,26 @@ pub struct InitArgs {
 
 // Helper function to parse and return the absolute path
 fn parse_path(path: &OsStr) -> Result<PathBuf, io::Error> {
-    canonicalize(path).map_err(|e| {
+    #[cfg(target_os = "windows")]
```
chesedo, line 186: @rutexd I really want to merge this. But will prefer a solution where we don't use `cfg`. Ie the same non-canonicalize code should work on both Windows and Unix. If you can push such an update, then we can merge this

chesedo (changes requested):

PR #458, opened 2022-11-03
feat(gateway): add account_tier column

brokad:


chesedo (approved):
Seems all good to me. `cargo fmt` is just complaining

PR #460, opened 2022-11-03
Feat: revive via gateway endpoint

oddgrd:
Add a "admin/revive" endpoint to the gateway API, and remove it from the gateway cli.

### DIFF
--- a/gateway/src/api/latest.rs
+++ b/gateway/src/api/latest.rs
@@ -9,6 +9,7 @@ use axum::routing::{any, get};
 use axum::{Json as AxumJson, Router};
 use http::StatusCode;
 use serde::{Deserialize, Serialize};
+use shuttle_common::models::error::ErrorKind;
 use shuttle_common::models::{project, user};
 use tokio::sync::mpsc::Sender;
 use tower_http::trace::TraceLayer;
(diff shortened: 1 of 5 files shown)


chesedo (approved):
I'm happy with this, thanks @oddgrd! Will just wait on your tests

gateway/src/api/latest.rs
```diff
@@ -167,6 +177,7 @@ pub fn make_api(service: Arc<GatewayService>, sender: Sender<BoxedTask>) -> Rout
         )
         .route("/users/:account_name", get(get_user).post(post_user))
         .route("/projects/:project/*any", any(route_project))
+        .route("/admin/revive", get(revive_projects))
```
brokad, line 180:
```suggestion
        .route("/admin/revive", post(revive_projects))
```
seems more idiomatic? Under the idea that you'd have `/admin/revive/0`, `/admin/revive/1` etc to describe operations. Obviously we don't have that yet, but I think this would be more intuitive than using `get` for starting side-effects.

gateway/src/api/latest.rs
```diff
@@ -153,6 +154,15 @@ async fn get_status(Extension(sender): Extension<Sender<BoxedTask>>) -> Response
         .unwrap()
 }
 
+async fn revive_projects(
```
brokad, line 157: From here you should grab `sender` and feed that down `exec::revive`. The only reason why it wasn't there before if because it was initially written to run stand-alone as a subcommand. Now it's part of the API, it should use the same worker queue.

gateway/src/project.rs
```diff
@@ -849,9 +849,8 @@ pub mod exec {
 
     use super::*;
 
-    pub async fn revive(gateway: GatewayService) -> Result<(), ProjectError> {
+    pub async fn revive(gateway: Arc<GatewayService>) -> Result<(), ProjectError> {
```
brokad, line 852: Probably better to send this down the global work queue that is used by other API endpoints as well, instead of creating our own separately.

For that, grab `sender` here and use `TaskBuilder::send`.

gateway/src/api/latest.rs
oddgrd: Agreed. I wasn't sure about this one since there's no body, but I looked around now and there seems to be some consensus on empty POST being okay for cases like this.

brokad (approved):

PR #466, opened 2022-11-07
feat: make deployer only answer its own project

chesedo:


oddgrd (approved):
Looks good! Just one question from a tracing noob: does this add the project name to the tracing layer?

chesedo:
@oddgrd, not yet. I'll spin up another PR for a bunch of tracing stuff

gateway/src/project.rs
```diff
@@ -615,9 +615,7 @@ pub struct Service {
 
 impl Service {
     pub fn from_container(container: ContainerInspectResponse) -> Result<Self, ProjectError> {
-        let container_name = safe_unwrap!(container.name.strip_prefix("/")).to_string();
-
-        let resource_name = safe_unwrap!(container_name.strip_suffix("_run")).to_string();
+        let name = safe_unwrap!(container.config.labels.get("project.name")).to_string();
```
chesedo, line 618: @brokad I tried to trace where `Service::name` will be used but it seems to be only for the health check. However, can you still confirm I'm not breaking anything here?

brokad: You may be breaking existing containers that were not deployed with the label.

chesedo: Thanks @brokad!! I've changed it to a backwards compatible implementation

PR #468, opened 2022-11-08
fix: add timeout to health checks

brokad:


gateway/src/worker.rs
```diff
@@ -57,7 +61,16 @@ impl Worker<BoxedTask> {
 
         while let Some(mut work) = self.recv.recv().await {
             loop {
-                match work.poll(()).await {
+                let timeout = time::sleep(Duration::from_secs(TASK_MAX_IDLE_TIMEOUT));
+                let mut poll = work.poll(());
+                let res = tokio::select! {
+                    res = &mut poll => res,
+                    _ = timeout => {
+                        warn!("a task has been running for a long time");
```
chesedo, line 69: Is it possible to log the project that timed out?

chesedo (approved):
Looks good, LGTM

gateway/src/worker.rs
brokad: Ah, yes, absolutely. Coming up.

PR #469, opened 2022-11-08
WIP feat: count recent start events before restart

oddgrd:
Count recent start events before restarting a stopped container, to prevent it from looping restarts indefinitely.

I tested that it does indeed not attempt to start if it has started 3 times in the last 15 minutes.

### DIFF
--- a/gateway/src/project.rs
+++ b/gateway/src/project.rs
@@ -690,37 +690,40 @@ where
     type Error = ProjectError;
 
     async fn next(self, ctx: &Ctx) -> Result<Self::Next, Self::Error> {
-        let container_id = self
-            .container
-            .id
-            .as_deref()
-            .expect("container should have ID");
+        let container = self.container;
 
-        // Filter and count the `start` events for this project in the last 15 minutes
-        let start_event_count = ctx
+        let since = (chrono::Utc::now() - chrono::Duration::minutes(15))
+            .timestamp()
+            .to_string();
+        let until = chrono::Utc::now().timestamp().to_string();
+
+        // Filter and collect `start` events for this project in the last 15 minutes
+        let start_events = ctx
             .docker()
-            .events(Some(EventsOptions::<String> {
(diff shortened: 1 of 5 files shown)


gateway/src/project.rs
```diff
@@ -687,9 +689,33 @@ where
     type Next = ProjectStarting;
     type Error = ProjectError;
 
-    async fn next(self, _ctx: &Ctx) -> Result<Self::Next, Self::Error> {
+    async fn next(self, ctx: &Ctx) -> Result<Self::Next, Self::Error> {
+        let container_id = self
```
brokad, line 693: Use the custom `safe_unwrap!` macro for this, as it will always propagate a standard internal error for these mismatches between docker API responses and what is published in their docs.

gateway/src/project.rs
```diff
@@ -687,9 +689,33 @@ where
...
+                since: Some(
+                    (chrono::offset::Utc::now() - chrono::Duration::minutes(15)).to_string(),
+                ),
+                until: Some(chrono::offset::Utc::now().to_string()),
+                filters: HashMap::from([
+                    ("container".to_string(), vec![container_id.to_string()]),
+                    ("event".to_string(), vec!["start".to_string()]),
+                ]),
+            }))
+            .count()
```
brokad, line 712: A possible pointer for why this is not working is that this is a stream of `Result`s. So you may be just counting a single error before the stream terminates?

gateway/src/project.rs
oddgrd: Nice, thanks!

gateway/src/project.rs
oddgrd: This was the case, and replacing it with `try_collect` showed me that I was using the wrong time format for the filters. :man_facepalming: Thanks!

chesedo (approved):
LGTM

PR #471, opened 2022-11-09
feat: TLS acceptor with SNI resolver

brokad:


gateway/src/tls.rs
```diff
@@ -0,0 +1,107 @@
...
+            Ok(PrivateKey(key))
+        }
+        _ => Err(Error::from_kind(ErrorKind::Internal)),
+    }?;
+
+    Ok((certs, private_key))
+}
+
+pub struct GatewayCertResolver {
+    keys: RwLock<HashMap<String, Arc<CertifiedKey>>>,
```
christos-h, line 42: Have you considered `dashmap` here? https://crates.io/crates/dashmap

I have to admit I'm not entirely sure what the performance trade-offs are in the Cert resolver - but it is a production ready drop-in replacement for `RwLock<HashMap<..>>` focusing on performance.

brokad: Ah! I have seen this before. I feel I need to read a bit more into it before using it blindly though. Is the `RwLock` similar to `tokio`'s? I.e. write-preferring?

christos-h: Dashmap doesn't have an asynchronous API as far as I can tell, which may disqualify it immediately but I don't really have an intuition on this.

Here are some benchmarks of dashmap against other popular hashmap implementations (including `RwLock<HashMap>` but this `RwLock` in `std`).

I'm really not an expert on this crate - but from my understanding the philosophy of dashmap is to minimise contention which is achieved by sharding the hashmap internally:

```rust
pub struct DashMap<K, V, S = RandomState> {
    shift: usize,
    shards: Box<[RwLock<HashMap<K, V, S>>]>,
    hasher: S,
}
```

christos-h: Also found this: https://tokio.rs/tokio/tutorial/shared-state which I thought was interesting:

> Note, std::sync::Mutex and not tokio::sync::Mutex is used to guard the HashMap. A common error is to unconditionally use tokio::sync::Mutex from within async code. An async mutex is a mutex that is locked across calls to .await.

> A synchronous mutex will block the current thread when waiting to acquire the lock. This, in turn, will block other tasks from processing. However, switching to tokio::sync::Mutex usually does not help as the asynchronous mutex uses a synchronous mutex internally.

> As a rule of thumb, using a synchronous mutex from within asynchronous code is fine as long as contention remains low and the lock is not held across calls to .await. Additionally, consider using [parking_lot::Mutex](https://docs.rs/parking_lot/0.10.2/parking_lot/type.Mutex.html) as a faster alternative to std::sync::Mutex.

brokad: Thank you! It's a lot of very useful info, I'll look into it 😄

gateway/src/args.rs
```diff
@@ -29,6 +35,9 @@ pub struct StartArgs {
     /// Address to bind the user plane to
     #[arg(long, default_value = "127.0.0.1:8000")]
     pub user: SocketAddr,
+    /// Allows to disable the use of TLS in proxying deployed services (DANGEROUS)
+    #[arg(long, default_value = "enable")]
+    pub use_tls: UseTls,
```
chesedo, line 40: Wouldn't a `bool` have done the same?

gateway/src/main.rs
```diff
@@ -126,9 +127,17 @@ async fn start(db: SqlitePool, args: StartArgs) -> io::Result<()> {
...
+    let (_resolver, tls_acceptor) = make_tls_acceptor();
+    let proxy_handle = if let UseTls::Disable = args.use_tls {
+        warn!("TLS is disabled in the proxy service. This is only acceptable in testing, and should *never* be used in deployments.");
+        axum_server::Server::bind(args.user).serve(proxy).boxed()
+    } else {
+        axum_server::Server::bind(args.user)
+            .acceptor(tls_acceptor)
+            .serve(proxy)
+            .boxed()
+    };
```
chesedo, line 140: Can we only initialise `tls_acceptor` in the `else` block since the `if` block does not use it? Ie moving line 131  to 135

gateway/src/tls.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    let mut server_config = ServerConfig::builder()
+        .with_safe_defaults()
+        .with_no_client_auth()
+        .with_cert_resolver(Arc::clone(&resolver) as Arc<dyn ResolvesServerCert>);
+    server_config.alpn_protocols = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
+
+    let rustls_config = RustlsConfig::from_config(Arc::new(server_config));
+
+    (resolver, RustlsAcceptor::new(rustls_config))
```
chesedo, line 111: Why is the `resolver` being returned?

chesedo (approved):
Seems all good to me. LGTM

gateway/src/tls.rs
brokad: It'll be needed to insert certificates to be resolved.

gateway/src/args.rs
brokad: It is more difficult to parameterise an optional flag with docker-compose command arguments than it is to parameterise an option.

gateway/src/proxy.rs
```diff
@@ -113,39 +203,134 @@ impl<'r> Service<&'r AddrStream> for MakeProxyService {
...
-                .layer(challenge_response_layer)
-                .service(proxy_service);
+pub struct UserServiceBuilder {
+    service: Option<Arc<GatewayService>>,
+    acme: Option<AcmeClient>,
+    tls_acceptor: Option<RustlsAcceptor<DefaultAcceptor>>,
+    bouncer_binds_to: Option<SocketAddr>,
+    user_binds_to: Option<SocketAddr>,
+    public: Option<FQDN>,
+}
```
chesedo, line 218: Thanks for all the builder cleanups :pray:

chesedo (commented):
I'm happy to give this a few spins around the block

gateway/src/tls.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        for item in rustls_pemfile::read_all(&mut BufReader::new(rd))
+            .map_err(|_| Error::from_kind(ErrorKind::Internal))?
+        {
+            match item {
+                Item::X509Certificate(cert) => chain.push(Certificate(cert)),
+                Item::ECKey(key) | Item::PKCS8Key(key) | Item::RSAKey(key) => {
+                    private_key = Some(PrivateKey(key))
+                }
+                _ => unreachable!(),
```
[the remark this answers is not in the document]
brokad: Mmmh, I wish I knew myself. Stand by 😄

PR #474, opened 2022-11-10
feat: verify project exists before sending destroy task

oddgrd:
This is a quick fix to the issue of `cargo project rm` always returning `destroying`. Draft while I look into improving cli feedback.

### DIFF
--- a/gateway/src/api/latest.rs
+++ b/gateway/src/api/latest.rs
@@ -124,6 +124,18 @@ async fn delete_project(
 ) -> Result<AxumJson<project::Response>, Error> {
     let project_name = project.clone();
 
+    let state = service.find_project(&project_name).await?;
+
+    let mut response = project::Response {
+        name: project_name.to_string(),
+        state: state.into(),
+    };
+
+    if response.state == shuttle_common::models::project::State::Destroyed {
+        return Ok(AxumJson(response));
+    }
+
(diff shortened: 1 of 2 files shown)


chesedo (commented):
This seems good to me. Is there a test we can update for this too?

oddgrd:
> This seems good to me. Is there a test we can update for this too?

Probably! I'll do that now. I'm looking into updating the cli feedback/ws status stream, but we can merge this first anyway.

gateway/src/api/latest.rs
```diff
@@ -356,6 +366,15 @@ pub mod tests {
             .await
             .unwrap();
 
+        // delete returns 404 for project that doesn't exist
+        router
+            .call(delete_project("resurrections").with_header(&authorization))
```
chesedo, line 371: Can we also add a test to delete "reloaded" twice? (it would more closely cover your code change)

chesedo (commented):
Nice, you found the tests :partying_face: Maybe we should also add a test to cover your code change though

gateway/src/api/latest.rs
oddgrd: Good point, I forgot to test that case. I added it now to the `end_to_end` test in `lib.rs`, since the test in `latest` doesn't progress the tasks.

PR #475, opened 2022-11-11
refactor: more metrics

chesedo:


oddgrd (approved):
LGTM! I hadn't heard of Valuable before, it looks like it will be a useful addition.

gateway/src/service.rs
```diff
@@ -202,21 +203,22 @@ impl GatewayService {
         Self { provider, db }
     }
 
-    pub async fn route_fqdn(&self, req: Request<Body>) -> Result<Response<Body>, Error> {
-        let fqdn = req
-            .headers()
-            .typed_get::<Fqdn>()
-            .ok_or_else(|| Error::from(ErrorKind::CustomDomainNotFound))?;
-        let project_name = self.project_name_for_custom_domain(&fqdn).await?;
+    // pub async fn route_fqdn(&self, req: Request<Body>) -> Result<Response<Body>, Error> {
```
brokad, line 206: I assume this was commented out because this calls `self.route` which needs a `ScopedUser`, and that we don't have one here (because this is meant for the proxy, not the control plane of the deployer). This should only be needed in the proxy though.

chesedo: Yes, but I also never used it in the proxy. It actually gets removed in #473

gateway/src/service.rs
```diff
@@ -223,9 +225,15 @@ impl GatewayService {
 
         debug!(target_url, "routing control");
 
+        let headers = req.headers_mut();
+        headers.append(
+            "X-Forwarded-Account-Name",
```
brokad, line 230: In #484 I have added `X-Shuttle-Project`. We should standardise these now. Which is better:
- `X-Forwarded-Project` or
- `X-Shuttle-Account-Name`?

chesedo: I have no opinion, just went with `X-Forwarded` since it seems well-known for proxies. But these are also very shuttle specific so the `X-Shuttle` makes sense :thinking: 

These will never appear in any web traffic so let's go with the `X-Shuttle` and keep these internal

PR #476, opened 2022-11-11
Don't specify tokio version in shuttle_service

snylonue:
It will prevent other packages depending on a newer version of tokio from building.


oddgrd:
Hey @snylonue, welcome and thanks for the PR! Unfortunately we have to lock down the tokio version currently, which does mean certain dependencies won't work with shuttle. Until we can fix this issue, you may be able to use an older version of the incompatible crate.

Context: https://github.com/shuttle-hq/shuttle/pull/348

chesedo:
Thanks for this @snylonue!! As Oddbjorn said, our current architecture needs to have the version pinned. However, this should be something we'll be able to get rid of in the future as we're improving our isolation model.

But still, thanks a lot for contributing to shuttle!!

PR #478, opened 2022-11-15
bug: deployer freezes

chesedo:
Reduce freezes in deployer by spawning less. This reduces complete lockups by deployer and makes deployer responsive to API calls during build jobs

Closes ENG-274


brokad (approved):
LGTM.

We should do something similar in gateway regarding custom domain certificate serde. We do spawn less ourselves over there, but we probably have a higher task count due to having more traffic.

PR #479, opened 2022-11-16
feat: find (soon to be) invalid project names

chesedo:
~~Waiting for some cleanups in #473~~

### DIFF
--- a/gateway/src/api/latest.rs
+++ b/gateway/src/api/latest.rs
@@ -1,4 +1,3 @@
-use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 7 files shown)


gateway/src/api/latest.rs
```diff
@@ -220,6 +221,87 @@ async fn request_acme_certificate(
     Ok("Certificate created".to_string())
 }
 
+async fn invalid_project_names(
```
brokad, line 224: Maybe a silly question: why not just create an API endpoint for listing projects and accounts (and their details), then consuming that on the admin CLI to do whatever we need? It would have a greater re-usability for the future, and we could tweak it further in between releases.

chesedo: Yip, this seems better. Thanks for the suggestion D!

PR #481, opened 2022-11-16
Feat(shuttle-next): stop runtime services

oddgrd:
Add a stop method to `runtime.proto`. We can't currently stop services on the legacy runtime, since we can't actually drop the legacy runtime. I implemented a stop method for it before discovering it was not possible, I left it there but with a comment. I'm happy to replace it with a `todo!()` if that's more tidy for now. 

I implemented stop for wasm serenity and axum, which is tested (manually) and works. We may want to setup integration tests for these runtimes with some helpers for the grpc requests, but maybe in a separate PR?

I opted for `oneshot` channels that are set on each `start`, but on second thought it may have been a mistake. I suppose the runtime should be able to load -> start several services, keeping them all alive until their ID/name receives a stop request. Meaning `broadcast` channels may be the correct choice here.

Note: This PR also changes the Dockerfile to install a newer protoc binary directly, rather than with `apt`, to allow us to use explicit optional fields in the protobufs. This is also done for the CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,6 +97,15 @@ commands:
           command: |
             curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
             sudo apt update && sudo apt install -y libssl1.1
+  install-protoc:
+    steps:
+      - run:
+          name: Install protoc
+          command: |
+            curl -OL https://github.com/protocolbuffers/protobuf/releases/download/v21.9/protoc-21.9-linux-x86_64.zip &&\
+              sudo unzip -o protoc-21.9-linux-x86_64.zip -d /usr bin/protoc &&\
+              sudo unzip -o protoc-21.9-linux-x86_64.zip -d /usr/ 'include/*' &&\
+              rm -f protoc-21.9-linux-x86_64.zip
 
 jobs:
   workspace-fmt:
@@ -104,7 +113,7 @@ jobs:
     steps:
       - checkout
       - restore-cargo-cache
-      - run: sudo apt install -y protobuf-compiler
+      - install-protoc
       - run: cargo fmt --all --check
       - run: cargo install cargo-sort
       - run: cargo sort --check --workspace
@@ -119,7 +128,7 @@ jobs:
     steps:
       - checkout
       - restore-cargo-cache
-      - run: sudo apt install -y protobuf-compiler
+      - install-protoc
       - run: |
           cargo clippy --tests \
                        --all-targets \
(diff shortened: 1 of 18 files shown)


.circleci/config.yml
```diff
@@ -97,14 +97,23 @@ commands:
...
             sudo apt update && sudo apt install -y libssl1.1
+  install-protoc:
+    steps:
+      - run:
+          name: Install protoc
+          command: |
+            curl -OL https://github.com/protocolbuffers/protobuf/releases/download/v21.9/protoc-21.9-linux-x86_64.zip &&\
+              sudo unzip -o protoc-21.9-linux-x86_64.zip -d /usr bin/protoc &&\
+              sudo unzip -o protoc-21.9-linux-x86_64.zip -d /usr/ 'include/*' &&\
+              rm -f protoc-21.9-linux-x86_64.zip
```
chesedo, line 108: Nice :star_struck:

runtime/src/axum/mod.rs
```diff
@@ -101,6 +104,30 @@ impl Runtime for AxumWasm {
...
+    ) -> Result<tonic::Response<StopResponse>, Status> {
+        let request = request.into_inner();
+
+        let service_name = ServiceName::from_str(request.service_name.as_str())
+            .map_err(|err| Status::from_error(Box::new(err)))?;
+
+        let kill_tx = self.kill_tx.lock().unwrap().deref_mut().take();
+
+        if let Some(kill_tx) = kill_tx {
+            tokio::spawn(async move {
```
chesedo, line 120: Can we do this without spawning? Else this might need to be a `spawn_blocking` instead.

runtime/src/legacy/mod.rs
```diff
@@ -140,14 +151,52 @@ impl Runtime for Legacy {
...
+    async fn stop(&self, request: Request<StopRequest>) -> Result<Response<StopResponse>, Status> {
+        let request = request.into_inner();
+
+        let service_name = ServiceName::from_str(request.service_name.as_str())
+            .map_err(|err| Status::from_error(Box::new(err)))?;
+
+        let kill_tx = self.kill_tx.lock().unwrap().deref_mut().take();
+
+        if let Some(kill_tx) = kill_tx {
+            tokio::spawn(async move {
```
chesedo, line 166: Same as above

chesedo (commented):
I'm happy with this. About the broadcast channel, yes, we'll eventually want this to be able to start and stop multiple deploys. But we can't currently start multiple services because of the hardcoded ports, so it' doesn't matter that stop does not use the deployment id / service name...

runtime/src/axum/mod.rs
oddgrd: I think we can do it without: https://docs.rs/tokio/latest/tokio/sync/oneshot/struct.Sender.html#method.send . I tested it now and it works. Thanks for the review!

PR #484, opened 2022-11-19
fix: custom domain routing

brokad:


deployer/src/proxy.rs
```diff
@@ -33,44 +34,58 @@ pub async fn handle(
...
+        trace!(?host, "proxy won't serve foreign domain");
+        return Ok(Response::builder()
+            .status(StatusCode::BAD_REQUEST)
+            .body(Body::from("this domain is not served by proxy"))
+            .unwrap());
+    }
 
+    // We only have one service per project, and its name coincides
+    // with that of the project
+    let service = match req.headers().get("X-Shuttle-Project") {
```
chesedo, line 63: In the future when we have multiple services per project support, what will `X-Shuttle-Project` be? I don't think we plan to determine the service using `gateway`. Ie `deployer` will figure out which service the request is for.

If deployer will eventually determine the service in the future, then it should be safe to remove this code now? Maybe for now get the project name from the `deployer` args :man_shrugging:

gateway/Cargo.toml
```diff
@@ -27,12 +24,17 @@ hyper = { version = "0.14.19", features = [ "stream" ] }
 # not great, but waiting for WebSocket changes to be merged
 hyper-reverse-proxy = { git = "https://github.com/chesedo/hyper-reverse-proxy", branch = "bug/host_header" }
 instant-acme = "0.1.0"
+
```
chesedo, line 27:
```suggestion
```

gateway/src/api/latest.rs
```diff
@@ -209,19 +202,50 @@ async fn request_acme_certificate(
...
     AxumJson(credentials): AxumJson<AccountCredentials<'_>>,
 ) -> Result<String, Error> {
     let fqdn: FQDN = fqdn
         .parse()
         .map_err(|_err| Error::from(ErrorKind::InvalidCustomDomain))?;
 
-    let (certs, private_key) = acme_client
-        .create_certificate(&fqdn.to_string(), ChallengeType::Http01, credentials)
-        .await?;
+    let (certs, private_key) = match service.project_details_for_custom_domain(&fqdn).await {
```
chesedo, line 213: Oh nice, I forgot about this case :star_struck:

chesedo (approved):
Seems good to me. LGTM

deployer/src/proxy.rs
brokad: > In the future when we have multiple services per project support, what will X-Shuttle-Project be? I don't think we plan to determine the service using gateway. Ie deployer will figure out which service the request is for.

Always the name of the project associated with this instance of deployer. Just the same as `X-Shuttle-Account-Name`.

> If deployer will eventually determine the service in the future, then it should be safe to remove this code now? Maybe for now get the project name from the deployer args

I see where you're coming from. But I think using custom headers is a better option than using CLI arguments. If anything, it makes tracing requests easier. Because this works and was tested and we're on a tight timeline I'll leave it for now, but we'll reopen this when we reopen the plan for multi-services support.

PR #485, opened 2022-11-20
feat: restore removed containers

brokad:
This
- Expands the default subnet to `/16`
- Has gateway re-create containers that have disappeared if they used to be ready, keeping the same image and fqdn as before


gateway/Cargo.toml
```diff
@@ -10,9 +10,6 @@ async-trait = "0.1.52"
 
 axum = { version = "0.5.8", features = [ "headers" ] }
 axum-server = { version = "0.4.4", features = [ "tls-rustls" ] }
-rustls = { version = "0.20.6" }
-rustls-pemfile = { version = "1.0.1" }
-pem = "1.1.0"
```
chesedo, line 13:
```suggestion
```

chesedo (approved):
LGTM, 64k projects here we come!!

PR #487, opened 2022-11-22
fix: backend bumps and hot fixes

brokad:


gateway/src/main.rs
```diff
@@ -108,14 +108,24 @@ async fn start(db: SqlitePool, fs: PathBuf, args: StartArgs) -> io::Result<()> {
         async move {
             loop {
                 tokio::time::sleep(Duration::from_secs(60)).await;
+                if sender.capacity() < WORKER_QUEUE_SIZE - SVC_DEGRADED_THRESHOLD {
+                    // if degraded, don't stack more health checks
+                    continue;
+                }
```
chesedo, line 114: Maybe a `tracing::warn!()` can help us here in the future to see why health checks weren't running?

chesedo (approved):
Seems good to me. Will there be a separate PR to bump the version of all the things?

gateway/src/main.rs
brokad: > Seems good to me. Will there be a separate PR to bump the version of all the things?

Indeed. The difference between all the things and this being client-side.

PR #488, opened 2022-11-23
feat(next): expand macro into axum routes

chesedo:
This is the first half of the codegen for our new framework. It just generates the same code that was created by hand for the axum wasi implementation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5726,6 +5726,7 @@ dependencies = [
 name = "shuttle-codegen"
 version = "0.7.0"
 dependencies = [
+ "http 0.2.8",
  "pretty_assertions",
  "proc-macro-error",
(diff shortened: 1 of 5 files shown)


oddgrd (approved):
This is really cool, this thing is really starting to take shape! :heart_eyes: Just one question, and this may be out of scope for this PR, but since axum routes can take several handlers per endpoint, should the `Endpoint` struct rather be something like this (or a new struct rather than tuple for handlers)?

```rust
struct Endpoint {
    route: LitStr,
    handlers: Vec<(Ident, Ident)>,
}
```

Edit: on second thought, I suppose the user will create each handler separately, and this will be built from those, so this case should probably be handled in the `App` instead :sweat_smile:

PR #490, opened 2022-11-24
Feat: parse shuttle::endpoint macro

oddgrd:
Create a functional macro in codegen that parses the entire codeblock of the macro and strips out every `#[shuttle::endpoint]` annotation, parsing them into `Endpoint`s that can be turned into axum routes.

I might think of a few more tests to add tomorrow, but this should be ready for review.

Thanks @chesedo for guiding me through this one!


codegen/src/next/mod.rs
```diff
@@ -1,13 +1,145 @@
...
+
+        let method = if let Some(method) = method {
+            method
+        } else {
+            emit_error!(
+                paren.span,
+                "no method provided";
+                hint = "Add a method to your endpoint: `#[shuttle_codegen::endpoint(method = get, route = \"/hello\")]`"
+            );
+            return None;
```
chesedo, line 132: I wonder if we can rather set a `has_error` variable to get all the errors to display in one go.

codegen/tests/ui/next/invalid-endpoint-syntax.stderr
```diff
@@ -0,0 +1,23 @@
+error: expected `,`
+
+         = help: The endpoint takes a comma-separated list of keys and values: `endpoint(method = get, route = "/hello")`
```
chesedo, line 3: I think we can drop this `help` as it might confuse users a bit

codegen/tests/ui/next/missing-endpoint-param.stderr
```diff
@@ -0,0 +1,32 @@
+error: no route provided
+
+         = help: Add a route to your endpoint: `#[shuttle_codegen::endpoint(method = get, route = "/hello")]`
```
chesedo, line 3: Maybe this help should only be the `route = "/hello"` part... same for the below errors

chesedo (commented):
This is looking exciiiitinng :partying_face:

codegen/src/lib.rs
```diff
@@ -1,11 +1,26 @@
...
 #[proc_macro_attribute]
 pub fn main(attr: TokenStream, item: TokenStream) -> TokenStream {
     main::r#impl(attr, item)
 }
+
+#[proc_macro_error]
+#[proc_macro]
+pub fn app(item: TokenStream) -> TokenStream {
+    let mut file = parse_macro_input!(item as File);
+    // todo: handle error
```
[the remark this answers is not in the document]
oddgrd: Woops!

codegen/tests/ui/next/invalid-endpoint-syntax.stderr
oddgrd: Maybe replace it with `The endpoint takes two arguments: `endpoint(method = get, route = "/hello")``, or would you prefer an empty  `= help:`?

oddgrd: I tried the above but it was still a bit confusing, I'll leave it blank unless we get a good idea for this message.

codegen/src/next/mod.rs
oddgrd: Good idea! I put it below the "no parameters" error, since that already explains that both method and route are needed in its hint. I also figured we may return early if the params syntax is invalid, or if there is no endpoint attribute. I can move it up though if we want to group the syntax errors in as well.

chesedo (approved):
LGTM, I'm excited for this! :partying_face:

codegen/src/next/mod.rs
```diff
@@ -1,13 +1,190 @@
...
+
+        // Parse the endpoint's parameters
+        let params: Params = match parse2(endpoint.tokens) {
+            Ok(params) => params,
+            Err(err) => {
+                // This will error on invalid parameter syntax
+                emit_error!(
+                    err.span(),
+                    err;
+                    hint = ""
```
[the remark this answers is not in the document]
oddgrd: Oh, nice!

codegen/src/next/mod.rs
```diff
@@ -1,13 +1,189 @@
...
+        // Find the index of an attribute that is an endpoint
+        for index in 0..item.attrs.len() {
+            // The endpoint ident should be the last segment in the path
+            if let Some(segment) = item.attrs[index].path.segments.last() {
+                if segment.ident.to_string().as_str() == "endpoint" {
+                    if endpoint_index.is_some() {
+                        emit_error!(
+                            item,
+                            "extra endpoint attribute";
+                            hint = "There should only be one endpoint annotation per handler function."
```
brokad, line 63: Not sure why the limitation? The same handler could be applied to multiple routes or methods.

Though TBF, since this PR only has the proc macro part of this, we can wait until we know how we use it in building routers to make a change to this.

brokad (approved):
Looks awesome 🤩 And I love how structured it is!

I wonder what we can do going forward to remove this top-level entrypoint `::app!`. I understand why it's needed here (or at all), but nesting everything inside a block is not ideal.

codegen/src/next/mod.rs
oddgrd: Regarding the `::app!`, I believe Pieter said we would use a build.rs for this down the line, but I'm not familiar with the details yet. 

I know that having several handlers with different methods for the same route is common, but is using the same handler for different routes common? Common or not though, it's certainly something we could handle, maybe by returning a Vec<Endpoint> from this and then validating that there aren't any invalid combinations. Doing this when we know more about the final setup sounds like a good plan :+1:

PR #495, opened 2022-11-28
feat: bump rust to 1.64, bump dependencies

oddgrd:
Bump the pinned rust version of containers to 1.64, and update dependencies. I wasn't able to bump `cargo` from 0.64 to 0.66, as it has some conflict over libgit2-sys with `cargo-edit`. I bumped everything else, most of it did not require code changes, except for configuring the timeout for the aws sdk in provisioner and general axum changes. These code changes are in separate commits.

I did not upgrade clap to 0.4 in provisioner, deployer and cargo-shuttle, this might be better to do in a separate PR.

This branch is set to use the updated examples from this branch: https://github.com/shuttle-hq/examples/pull/4, since quite a few of the frameworks will have incompatible version with the current ones in examples `main`. 

Another thing to note is that this PR upgrades our pinned tokio to 1.22, which works with all the updated frameworks. This, and updating the framework versions, means users will be forced to upgrade as well. 

Note: ~I'm not sure why the latest CI run segfaulted, it seems to only use tokio 1.22 and it passes locally.~ Edit: after ignoring the actix test (which is missing an example currently), the CI fails in a normal fashion.

Closes #425

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ orbs:
 executors:
   docker-rust:
     docker:
-      - image: cimg/rust:1.63.0
+      - image: cimg/rust:1.64.0
   image-ubuntu:
     machine:
       image: ubuntu-2204:2022.04.1
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "examples"]
 	path = examples
-	url = git@github.com:shuttle-hq/examples.git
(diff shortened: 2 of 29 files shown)


chesedo:
> for reasons I'm not quite sure I understand

:joy: The actix tests are not merged yet. Hence, they are also not in the updated PR for examples

chesedo (approved):

Issue #493, opened 2022-11-28
Make `cargo shuttle init` interactive


brokad:
Currently, to initialise a project with the right stub for different frameworks, we use `cargo shuttle init` with a set of mutually exclusive options (e.g. `--axum`, `--poem` etc).

We'd like an interactive route into this where `cargo shuttle init` prompts:
- The login flow, if the user is not logged in yet
- The name of the project to be deployed (indicating this will mean the project will be hosted at `${project_name}.shuttleapp.rs`)
- A multi-choice for all supported frameworks
- Asks if the users wants the environment for the project to be created on shuttle, and chain into it if yes

Procrat:
I'm happy to help out with this!

Did you still want to keep the non-interactive route around? If so, do we want it to only be triggered when all arguments are specified (inc. name and path)? If not, would there be any point in keeping those parameters?

brokad:
> Imo we can keep it around

I agree. What would be really nice IMO is if any missing option or parameter is prompted and nothing gets prompted if all required parameters are given. That way `cargo shuttle init --name="my-project"` would not prompt for project name but will prompt for a framework choice.

Procrat:
Thanks for the input! I've made a repo with a toy version of the interactive flow [here](https://github.com/Procrat/cargo-shuttle-init) in case you already wanted to give it a go and give some early feedback.

Just to make sure we're on the same page, the interactive and non-interactive versions behave quite differently here.  Whereas the non-interactive one (as it is now) just performs actions locally, the interactive one will also log in and create the environment on Shuttle. It might be a bit confusing that the scope of the command changes based on a flag, but I can see it's user-friendly to do everything in one go.

@oddgrd :I had already started with `dialoguer` instead of `inquire` although I did try them both out first. Happy to switch if you have a strong opinion about it though! I just chose it because it had fewer dependencies and seemed to have everything we needed.

oddgrd:
No opinion, feel free to use whichever crate fits best! 

That is a good point, originally the plan was to have just the boilerplate generation be interactive, so it would have been the same. I think tying it all together would be nice for the user though, like you said, so that `cargo shuttle init` is the only thing they need to know to get started with a hello-world app. Maybe we just continue with both, and see how we feel about it in the end?

brokad:
> Just to make sure we're on the same page, the interactive and non-interactive versions behave quite differently here.

[...]

PR #498, opened 2022-11-29
feat: interactive project initialization

Procrat:
PR for #493.

All the functionality should be there. However, when it came to writing tests, I noticed that the current integration test setup can't really test stdin/stdout. I think we'd have to call it as a standalone program to do so. Does anyone have any thoughts on that? ~I'm happy to try and change it to such a setup tomorrow.~ Edit: I've added some tests with that sort of setup now.

cc @oddgrd @brokad


cargo-shuttle/src/args.rs
```diff
@@ -198,3 +231,63 @@ fn parse_init_path(path: &OsStr) -> Result<PathBuf, io::Error> {
 
     parse_path(path)
 }
+
+#[cfg(test)]
```
Procrat, line 235: This test is just moved over from `init.rs`

cargo-shuttle/src/init.rs
```diff
@@ -1,15 +1,75 @@
...
+    Framework::Tower,
+    Framework::Poem,
+    Framework::Salvo,
+    Framework::Serenity,
+    Framework::Warp,
+    Framework::Thruster,
+];
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Framework {
```
Procrat, line 26: I'm not sure whether the distinction between `Framework` and `ShuttleInit` is really useful.

cargo-shuttle/tests/integration/init.rs
```diff
@@ -1,92 +1,28 @@
...
+use std::fs::read_to_string;
 
-use cargo_shuttle::{Args, Command, CommandOutcome, InitArgs, ProjectArgs, Shuttle};
+use cargo_shuttle::{Args, Shuttle};
+use clap::Parser;
 use indoc::indoc;
 use tempfile::Builder;
 
-/// creates a `cargo-shuttle` init instance with some reasonable defaults set.
-async fn cargo_shuttle_init(path: PathBuf) -> anyhow::Result<CommandOutcome> {
```
Procrat, line 11: It was only after I removed this test that I realised I effectively dropped support for initialising without any framework. We probably still want that, right? Should I add an extra option for that both in the interactive and non-interactive version?

Procrat:
I've written some tests for the interactive route that interact with `cargo-shuttle` as an executable now.

Procrat:
(Rebased.)

Procrat:
Hmmm, the test `run::actix_web_hello_world` seems to be failing, but I can't immediately see how it would be related to these changes. Could it be that it was already failing before?

cargo-shuttle/src/init.rs
chesedo: Nah, I like this since it is the framework subset of the init command

chesedo (commented):
This is really awesome, thanks @Procrat!

I like how this flows for users. Maybe we can refactor the code a bit to make its reading flow simpler.

Procrat:
Thanks for the thoughtful review, @chesedo! I'll have a look at it tomorrow :)

cargo-shuttle/tests/integration/init.rs
```diff
@@ -1,101 +1,152 @@
...
+
+    assert_valid_rocket_project(temp_dir_path.as_path(), "rocket-init");
+
+    Ok(())
 }
 
-#[tokio::test]
-async fn framework_init() {
+#[test]
+fn interactive_rocket_init_no_framework_prompt() -> Result<(), Box<dyn std::error::Error>> {
```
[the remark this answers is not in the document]
Procrat: Haha, yeah, that's fair!

cargo-shuttle/src/args.rs
```diff
@@ -176,6 +178,11 @@ pub struct InitArgs {
     /// Initialize with thruster framework
     #[clap(long, conflicts_with_all = &["actix-web","axum", "rocket", "tide", "tower", "poem", "warp", "salvo", "serenity"])]
     pub thruster: bool,
```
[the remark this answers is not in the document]
Procrat: I had a look at this as well when I came across it. There's an [open issue](https://github.com/clap-rs/clap/issues/2621) for modelling an option group as an enum. They mention [a workaround](https://github.com/clap-rs/clap/issues/2621#issuecomment-1074671496), but it only really deals with those `conflicts_with_all` arrays. It doesn't expose it as a, say, `framework` variable if you were to call the `ArgGroup` `framework`, so we'd still be stuck with the `if` checks :/

cargo-shuttle/src/lib.rs
```diff
@@ -103,14 +104,91 @@ impl Shuttle {
...
+            );
+            println!();
+        }
+
+        // 3. Confirm the project directory
+        let path = if interactive {
+            println!("Where should we create this project?");
+            let directory_str: String = Input::with_theme(&theme)
+                .with_prompt("Directory")
+                .default(".".to_owned())
```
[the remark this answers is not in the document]
Procrat: Ah, yeah, good catch!

Procrat:
Thanks for all the good suggestions, @chesedo! I've incorporated them now. Any thoughts on [this question](https://github.com/shuttle-hq/shuttle/pull/498#discussion_r1035734523)?

cargo-shuttle/tests/integration/init.rs
AlexCzar: While it's not strictly necessary, as it's easy enough to init some framework and clean it out, it is a quality of life feature which is welcome when working with an unsupported framework.
Since there is a possibility to implement custom services, there should probably be a way to bootstrap skeletons for them.

chesedo: Hmm, I agree with @AlexCzar about being able to bootstrap custom services. However, it should not form part of this PR as this PR only focuses on making the init interactive.

Since we will add a custom one eventually, it will be good to have the default back (which we'll then swap for the custom one once the custom one is ready). So we'll still want it, and it seems that an option (`--no-framework`?) will be the best for now

cargo-shuttle/src/args.rs
```diff
@@ -195,9 +230,81 @@ fn parse_path(path: &OsStr) -> Result<PathBuf, io::Error> {
...
+            "axum",
+            "rocket",
+            "tide",
+            "tower",
+            "poem",
+            "salvo",
+            "serenity",
+            "warp",
+            "thruster",
+        ];
```
chesedo: We should be able to use your enum to identify the framework instead right?

cargo-shuttle/src/args.rs
```diff
@@ -195,9 +230,81 @@ fn parse_path(path: &OsStr) -> Result<PathBuf, io::Error> {
...
+            Framework::Axum,
+            Framework::Rocket,
+            Framework::Tide,
+            Framework::Tower,
+            Framework::Poem,
+            Framework::Salvo,
+            Framework::Serenity,
+            Framework::Warp,
+            Framework::Thruster,
+        ];
```
chesedo: Can this be the new `init::Framework::iter()`?

cargo-shuttle/tests/integration/init.rs
Procrat: Yeah, makes perfect sense to me. I've added that option now.

Procrat:
Everything's good to go on my end 👍

chesedo (commented):
Just some small optimization suggestions :smile:

Procrat:
Good idea! Done!

chesedo (approved):
This turned out great, thanks @Procrat!

PR #500, opened 2022-11-30
feat(gateway,deployer): add more tracing events

akrantz01:


chesedo (approved):
Thanks for these @akrantz01

PR #503, opened 2022-11-30
Add spinner wait for `project new` and `project status --follow`

### LINKED ISSUE #492: `project new` subcommand waits for environment to be ready
Currently, running `cargo shuttle project new` returns immediately with a prompt stating the project is `creating`. 

However, the environment is not ready to be deployed to until `cargo shuttle project status` returns `ready`. Most times this is not a big deal, but under high load, environment creation can take a minute or so, which can be unexpectedly long and create a bad experience as users don't know when their environment is finally ready.

Maybe it'd be better to hold exiting from `project new` until the project is ready, continuously updating the prompt with the latest state, and including a spinner to show there is work in progress. This also requires covering the [entire state machine of environments](https://github.com/shuttle-hq/shuttle/blob/e3fb067b3e62b399b800f591df5404797f7581ce/gateway/src/project.rs#L147).

If users exit early from the prompt, they should be able to go back into the "blocking, continuously updating state" through `project status --wait`, which should be equivalent.

guerinoni:
- feat: Wait env ready after `project new`
- feat: Add --follow argument to `project status` command
- refactor: Remove unused reference
- refactor: Unify in same function the code for wait with spinner

This PR closes #492


cargo-shuttle/src/lib.rs
```diff
@@ -403,18 +406,78 @@ impl Shuttle {
...
         Ok(())
     }
 
-    async fn project_status(&self, client: &Client) -> Result<()> {
-        let project = client.get_project(self.ctx.project_name()).await?;
+    async fn wait_with_spinner<'a, F, Fut>(
+        &self,
+        states_to_check: &[project::State],
+        f: F,
+        prj_name: &'a ProjectName,
```
chesedo, line 444:
```suggestion
        project_name: &'a ProjectName,
```
Just to keep this aligned with the naming used else where

cargo-shuttle/src/lib.rs
```diff
@@ -403,18 +406,78 @@ impl Shuttle {
...
+        client: &'a Client,
+    ) -> Result<(), anyhow::Error>
+    where
+        F: Fn(&'a Client, &'a ProjectName) -> Fut,
+        Fut: std::future::Future<Output = Result<project::Response>> + 'a,
+    {
+        let mut project = f(client, prj_name).await?;
+        let pb = indicatif::ProgressBar::new_spinner();
+        pb.enable_steady_tick(std::time::Duration::from_millis(100));
```
chesedo, line 453:
```suggestion
        pb.enable_steady_tick(std::time::Duration::from_millis(350));
```
The current animation seems a little fast :smile:

chesedo (commented):
Thanks for this @guerinoni!!

I see you went with a pulling model which I'm happy to merge. However, I would prefer a pushing model from gateway if that is possible. Will you be able to check if gateway can work with a push approach instead? And maybe give a small write up if it is not possible?

chesedo (approved):
NIce, thanks @guerinoni!

cargo-shuttle/src/lib.rs
```diff
@@ -486,18 +489,78 @@ impl Shuttle {
...
 
-        println!("{project}");
+        Ok(())
+    }
+
+    async fn project_status(&self, client: &Client, follow: bool) -> Result<()> {
+        match follow {
+            true => {
+                self.wait_with_spinner(
+                    &[project::State::Ready, project::State::Destroyed],
```
brokad, line 507: Should include `Errored` as well, as one of the potential "terminal" states.

cargo-shuttle/src/lib.rs
```diff
@@ -486,18 +489,78 @@ impl Shuttle {
     }
 
     async fn project_create(&self, client: &Client) -> Result<()> {
-        let project = client.create_project(self.ctx.project_name()).await?;
+        self.wait_with_spinner(
+            &[project::State::Ready],
```
brokad, line 493: Should include `Errored` as well, as one of the potential "terminal" states. This could actually happen on create if, say, we run into resource constraints.

PR #504, opened 2022-12-01
feat: bump pinned rust version to 1.65

oddgrd:
Due to the error in CI on the first commit of this PR (`E: Unable to locate package protobuf-compiler`), I ended up backporting the updated protoc install command from shuttle-next.


codegen/src/lib.rs
```diff
@@ -1,10 +1,10 @@
-mod main;
+mod shuttle_main;
```
chesedo, line 1: Why was the rename needed? :smile:

chesedo (approved):
LGTM

codegen/src/lib.rs
oddgrd: Woops, forgot to write a note about this. It's because clippy wouldn't allow it anymore, and I couldn't find any way to allow it, there was no `#[allow(special-module-name)]` or similar  that I could find. 

Ref failed CI: https://app.circleci.com/pipelines/github/shuttle-hq/shuttle/1176/workflows/31f0e441-7778-462c-a4ce-2aca2b5eb2ba/jobs/19417

PR #506, opened 2022-12-01
feat: implement workspace inheritance

oddgrd:
This implements the new workspace inheritance feature from 1.64. For this PR I elected to just add some external crates to the workspace dependencies, since `cargo-chef` does not support local crate workspace deps yet. When it does, we can add more. I'm happy to add more external deps to this PR, but it should also be easy to do as part of other PRs as we see commonly used deps that aren't inherited.

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -1,26 +1,26 @@
 [package]
 name = "shuttle-common"
 version = "0.7.3"
-edition = "2021"
-license = "Apache-2.0"
+edition.workspace = true
+licence.workspace = true
 description = "Common library for the shuttle platform (https://www.shuttle.rs/)"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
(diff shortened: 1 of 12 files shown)


common/Cargo.toml
```diff
@@ -1,31 +1,29 @@
...
+serde = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true, optional = true }
 strum = { version = "0.24.1", features = ["derive"] }
-tracing = "0.1.37"
-uuid = { version = "1.2.2", features = ["v4", "serde"] }
+tracing = { workspace = true }
+uuid = { workspace = true, features = ["v4", "serde"] }
 
 [features]
-default = ["models"]
```
oddgrd, line 27: Dependencies inherited from the workspace cannot set the default-features flag, so I figured we may as well remove this. For reference: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace

PR #510, opened 2022-12-02
refactor: one store per request

chesedo:


oddgrd:
Nice, thanks P! This is a huge performance improvement. I'm wondering though, are we now writing to the same file descriptors concurrently, and is that safe? I may be misreading this though, and I suppose all our writes and reads are sync still. I also ran `wrk` against this with 4 threads and 200 conns, and it worked fine, so I'm not sure. :shrug: 

The test is broken for me, but I just had to clone the router for each request to fix it (like is done in the server). Axum has also auto-updated to 0.6.1 in `tmp/axum-wasm`, but the only change needed is to remove the `into_service()` call in `app`.

chesedo:
> are we now writing to the same file descriptors concurrently, and is that safe?

I'm not 100% sure yet. But it seems the file descriptors are linked to the context and every request also has its own context. So I don't think they are the same file descriptors

PR #517, opened 2022-12-08
fix(deployer): keep Cargo.lock between deployments

Procrat:
This way, we don't need to resolve dependencies between deployments each time.


chesedo (approved):

chesedo:
Thanks for this @Procrat

PR #524, opened 2022-12-10
bug: hacking static folders

chesedo:
Prevents use of absolute paths or a path that walk up in the directory tree for static folders


oddgrd (approved):
Nice, good catch! One question, and it may be out of scope for this PR, but should we also set a max for static file sizes? It looks like it should be doable with the `WalkBuilder`: https://docs.rs/ignore/latest/ignore/struct.WalkBuilder.html#method.max_filesize

PR #530, opened 2022-12-12
feat: 'clean' subcommand

chesedo:


cargo-shuttle/src/lib.rs
```diff
@@ -299,6 +301,16 @@ impl Shuttle {
         Ok(())
     }
 
+    async fn clean(&self, client: &Client) -> Result<()> {
+        let lines = client.clean_project(self.ctx.project_name()).await?;
+
+        for line in lines {
```
oddgrd, line 307: I'm not seeing any output after running the command, but it does clean up the `target` folder. What are we expecting these lines to be? Should we add a simple success `println`, perhaps?

oddgrd (approved):
Nice! :tada: I tested this locally on ubuntu and it works as expected, except that I'm not getting feedback from the CLI, I left a comment about that.

service/src/loader.rs
```diff
@@ -147,6 +147,51 @@ pub async fn build_crate(
...
+            profile_specified: true,
+            doc: false,
+        };
+
+        clean(&ws, &opts).unwrap();
+    });
+
+    let mut lines = Vec::new();
+
+    for message in Message::parse_stream(read) {
```
brokad, line 179: I suppose this thing knows when the stream ends? That is it is not relying on the `.read(...) == Ok(0)` behaviour to determine when the task is done, otherwise this could race with the `spawn_blocking` above and, if we get here before it starts, this could return too early?

brokad: Also, should we not put the whole thing under spawn_blocking and await completion?

brokad (approved):
LGTM

service/src/loader.rs
chesedo: Luckily this thing knows when the stream ends, else syncing the two would have been a nightmare :sweat_smile: 

About the `spawn_blocking`, in the `build_crate` function we actually spawn block this stream. But the axum handler complains about some trait not being implemented when I do that instead. I'm not sure what that is about though :man_shrugging:

PR #532, opened 2022-12-13
feat: build queue

chesedo:


deployer/src/deployment/queue.rs
```diff
@@ -55,13 +60,23 @@ pub async fn task(
             span.set_parent(parent_cx);
 
             async move {
+                match wait_for_queue(queue_client.clone(), id).await {
```
brokad, line 63: We probably want to wrap this in a timeout? Just in case gateway hangs and connections are not closed or a slot takes too long to become available.

PR #541, opened 2022-12-15
bug: no networks

chesedo:
After the VM upgrade a bunch of user containers are no longer connected to any network. This is potentially because they tried to start with the wrong network id (from the old VM). So this PR makes a change to connect to the user layer network when we start up deployers rather than when we create deployers. Moving the network attachment to start means containers that are restarted will also attach to the correct network correctly.

Lastly, the revive command is updated to detect and restart containers which are not on any networks.

### DIFF
--- a/gateway/src/project.rs
+++ b/gateway/src/project.rs
@@ -148,6 +148,7 @@ impl From<DockerError> for Error {
 #[serde(rename_all = "lowercase")]
 pub enum Project {
     Creating(ProjectCreating),
+    Attaching(ProjectAttaching),
     Starting(ProjectStarting),
     Started(ProjectStarted),
     Ready(ProjectReady),
@@ -160,6 +161,7 @@ pub enum Project {
 
 impl_from_variant!(Project:
                    ProjectCreating => Creating,
+                   ProjectAttaching => Attaching,
                    ProjectStarting => Starting,
                    ProjectStarted => Started,
                    ProjectReady => Ready,
@@ -222,6 +224,7 @@ impl Project {
             Self::Starting(_) => "starting",
             Self::Stopping(_) => "stopping",
             Self::Creating(_) => "creating",
+            Self::Attaching(_) => "attaching",
             Self::Destroying(_) => "destroying",
             Self::Destroyed(_) => "destroyed",
             Self::Errored(_) => "error",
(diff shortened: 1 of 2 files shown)


gateway/src/project.rs
```diff
@@ -300,6 +302,11 @@ where
         if let Ok(Self::Errored(errored)) = &mut new {
             errored.ctx = Some(Box::new(previous));
             error!(error = ?errored, "state for project errored");
+
+            if errored.kind == ProjectErrorKind::NoNetwork {
```
brokad, line 306: Adding this here means that any time a project ends up in an errored state from a previous state, as long as the error kind is `::NoNetwork`, we consider that, in fact, we had no `previous -> errored` state transition, but instead, we had `previous -> new`. This also means nothing on the outside sees the transition to `-> errored` because it is swallowed by this catch all.

This is potentially not what we want, but I'm not sure. Based on the description of the solution (i.e. moving the "connect to network" outside of container creation), I think what we actually want is a state inbetween started and ready (like "attaching" state) which runs the network connect steps. And then, just like if containers disappear they are recreated, this code should be in `refresh` so that it gets triggered by the health checks.

brokad (commented):
A few quick points as I glanced through it - but will spend more time with it when I can

gateway/src/project.rs
```diff
@@ -748,11 +785,11 @@ impl Service {
         let network = safe_unwrap!(container.network_settings.networks)
             .values()
             .next()
-            .ok_or_else(|| ProjectError::internal("project was not linked to a network"))?;
+            .ok_or_else(|| ProjectError::no_network("project was not linked to a network"))?;
 
         let target = safe_unwrap!(network.ip_address)
             .parse()
-            .map_err(|_| ProjectError::internal("project did not join the network"))?;
+            .map_err(|_| ProjectError::no_network("project did not join the network"))?;
```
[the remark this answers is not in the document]
chesedo: Nope, not from my testing. So when the network is not joined, then `container.network_settings.networks` will be an empty array (line 785). This would cause the `.next()` (line 787) to fail and we return the correct error in that case.

To truly harden this, we should probably rather loop through all the networks that were joined and find the one we know we created.

gateway/src/project.rs
chesedo: I ended up adding an `attaching` state between `creating` and `starting` because it produces less errors / edge cases when compared to attaching after `started`

brokad (approved):
LGTM 👌

PR #542, opened 2022-12-16
upgrade clap to 4.0 latest across whole project

bazylhorsey:
Hello, my name is Bazyl. I'm a very passionate Rust coder who happens to be applying for a role with Shuttle. I was able to successfully build docker with this code and updated packages.

I was not able to run ```make test``` successfully due to an error with protobuf, will try to get around it by properly installing the nix shell.

### DIFF
--- a/cargo-shuttle/Cargo.toml
+++ b/cargo-shuttle/Cargo.toml
@@ -15,7 +15,7 @@ cargo = "0.65.0"
 cargo-edit = { version = "0.11.6", features = ["cli"] }
 cargo_metadata = "0.15.2"
 chrono = { workspace = true }
-clap = { version = "3.2.17", features = ["derive", "env"] }
+clap = { version = "4.0.29", features = ["derive", "env"] }
 clap_complete = "3.2.5"
 crossbeam-channel = "0.5.6"
 crossterm = "0.25.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1378,7 +1378,7 @@ checksum = "e3a5eba325b274fc14e17df48888c3c45146e03be7331cdbd585c377a2bc8058"
 dependencies = [
  "anyhow",
  "cargo_metadata",
- "clap 4.0.27",
+ "clap 4.0.29",
  "concolor-control",
  "crates-index",
  "dirs-next",
@@ -1422,7 +1422,7 @@ dependencies = [
  "cargo-edit",
  "cargo_metadata",
  "chrono",
- "clap 3.2.23",
+ "clap 4.0.29",
  "clap_complete",
  "crossbeam-channel",
  "crossterm",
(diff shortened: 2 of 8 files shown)


bazylhorsey:
```
#[clap(
    version,
    about,
    // Cargo passes in the subcommand name to the invoked executable. Use a
    // hidden, optional positional argument to deal with it.
    arg(clap::Arg::with_name("dummy")
        .possible_value("shuttle")
        .required(false)
        .hidden(true))
)]
```
methods with_name and possible_value both seem to be deprecated. Using clap::Arg::new("dummy") could fix the first bit, but there doesn't seem to be a good alternative or any deprecation documentation I can find for .possible_value.

```
error[E0308]: mismatched types
   --> cargo-shuttle/src/lib.rs:281:37
    |
281 |             None => generate(shell, &mut app, name, &mut stdout()),
    |                     --------        ^^^^^^^^ expected struct `clap::builder::command::App`, found struct `clap::Command`
    |                     |
    |                     arguments to this function are incorrect
    |
    = note: expected mutable reference `&mut clap::builder::command::App<'_>`
               found mutable reference `&mut clap::Command`
note: function defined here
   --> /home/circleci/.cargo/registry/src/github.com-1ecc6299db9ec823/clap_complete-3.2.5/src/generator/mod.rs:225:8
    |
225 | pub fn generate<G, S>(gen: G, cmd: &mut clap::Command, bin_name: S, buf: &mut dyn Write)
    |        ^^^^^^^^
```
For this one we have ```let mut app = Command::command();``` which returns type clap::Command, they should be the same type. Open to ideas here and happy to implement with some input! :)

cargo-shuttle/Cargo.toml
```diff
@@ -15,7 +15,7 @@ cargo = "0.65.0"
 cargo-edit = { version = "0.11.6", features = ["cli"] }
 cargo_metadata = "0.15.2"
 chrono = { workspace = true }
-clap = { version = "3.2.17", features = ["derive", "env"] }
+clap = { version = "4.0.29", features = ["derive", "env"] }
 clap_complete = "3.2.5"
```
oddgrd, line 19: For the mismatched types error from the `generate` function, upgrading `clap_complete` to the latest should be enough.

oddgrd (commented):
Hey Bazyl, thanks for getting started on this one! It does indeed seem like `with_name` has been deprecated in favor of just using `new`. As for `possible_value`, I believe you can replace it with `.value_parser([PossibleValue::new("shuttle")])`, but I'm not a 100% sure it's the cleanest way, you can see it in the changelog for 3.2 here: https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#320---2022-06-13.

Other than that, you should also get a lot of deprecation warnings if you go through [these steps](https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#migrating), remember to run `cargo check --features clap/deprecated` (I missed that at first :sweat_smile:).

For the tests, `make test` runs the e2e tests, you can also run unit/integration tests, there is an explanation for this in our contributing.md document. 

And while `nix` is great, you shouldn't need it to run the e2e tests. Do you have `protoc` installed (check with `protoc --version`), and which error are you getting? For reference: https://grpc.io/docs/protoc-installation/

oddgrd:
Thanks for getting started Bazyl, but this was completed in https://github.com/shuttle-hq/shuttle/pull/570

PR #545, opened 2022-12-20
Fix: dependencies compiled with incompatible versions of rustc

oddgrd:
Crates with a `rust-toolchain.toml` override would forcibly install that toolchain and use it to compile these crates, leading to certain crates being compiled with different versions of rustc.

Setting the override with `RUSTUP_TOOLCHAIN` [trumps the .toml override](https://rust-lang.github.io/rustup/overrides.html#overrides). Tested by deploying an axum hello-world app with dashmap, as well as a serenity app with the `cache` feature (which pulls in dashmap).

### DIFF
--- a/gateway/Cargo.toml
+++ b/gateway/Cargo.toml
@@ -30,7 +30,6 @@ opentelemetry-http = "0.7.0"
 pem = "1.1.0"
 rand = "0.8.5"
 rcgen = "0.10.0"
-rustc_version_runtime = "0.2.1"
 rustls = "0.20.7"
 rustls-pemfile = "1.0.1"
 serde = { workspace = true, features = [ "derive" ] }
--- a/gateway/Cargo.toml
+++ b/gateway/Cargo.toml
@@ -30,6 +30,7 @@ opentelemetry-http = "0.7.0"
 pem = "1.1.0"
 rand = "0.8.5"
 rcgen = "0.10.0"
+rustc_version_runtime = "0.2.1"
 rustls = "0.20.7"
 rustls-pemfile = "1.0.1"
 serde = { workspace = true, features = [ "derive" ] }
--- a/gateway/src/project.rs
+++ b/gateway/src/project.rs
@@ -481,6 +481,14 @@ impl ProjectCreating {
             name: self.container_name(ctx),
         };
 
+        // Get the version of rustc gateway was compiled with, matching the deployer
+        // RUSTUP_TOOLCHAIN override with the version from our Containerfile
+        let version = rustc_version_runtime::version().to_string();
+        debug!(
+            "got installed rustc version for deployer toolchain override: {}",
+            version
+        );
+
         let container_config = self
             .from
             .as_ref()
@@ -515,10 +523,9 @@ impl ProjectCreating {
                     ],
                     "Env": [
                         "RUST_LOG=debug",
-                        // This should be set to the same version we pin Rust to in our Containerfile.
                         // If we don't set this, users' dependencies with a `rust-toolchain.toml`
                         // override will compile with incompatible versions of rust.
-                        "RUSTUP_TOOLCHAIN=1.65"
+                        format!("RUSTUP_TOOLCHAIN={version}")
                     ]
                 })
             });
--- a/gateway/src/project.rs
+++ b/gateway/src/project.rs
@@ -515,6 +515,10 @@ impl ProjectCreating {
                     ],
                     "Env": [
                         "RUST_LOG=debug",
+                        // This should be set to the same version we pin Rust to in our Containerfile.
+                        // If we don't set this, users' dependencies with a `rust-toolchain.toml`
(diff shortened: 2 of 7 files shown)


gateway/src/project.rs
```diff
@@ -515,6 +515,10 @@ impl ProjectCreating {
                     ],
                     "Env": [
                         "RUST_LOG=debug",
+                        // This should be set to the same version we pin Rust to in our Containerfile.
+                        // If we don't set this, users' dependencies with a `rust-toolchain.toml`
+                        // override will compile with incompatible versions of rust.
+                        "RUSTUP_TOOLCHAIN=1.65"
```
chesedo, line 521: Is there a way to get this version from the currently installed toolchain? Ie get it from the environment to keep it in sync with the Containerfile?

oddgrd: We can use [this crate](https://crates.io/crates/rustc_version_runtime) to get the version gateway is compiled with and pass it to the deployer container config. I'm not a 100% if this goes against the intentions of this crate, so alternatively we could set the `RUSTUP_TOOLCHAIN` env var from the `Makefile`, extract it with `std::env::var(RUSTUP_TOOLCHAIN)` and pass it to the config. What do you think is best here?

chesedo: I like the `Makefile` idea. We can go a step further and pass the value from `Makefile` in as a build arg to the `Containerfile`. In the `Containerfile` the build arg can then set both the image of rust being used as well as a `RUST_TOOLCHAIN` environment variable. Then we don't have to worry about setting it in gateway.

gateway/Cargo.toml
```diff
@@ -30,6 +30,7 @@ opentelemetry-http = "0.7.0"
 pem = "1.1.0"
 rand = "0.8.5"
 rcgen = "0.10.0"
+rustc_version_runtime = "0.2.1"
```
chesedo, line 33:
```suggestion
```
No longer being used

gateway/src/project.rs
```diff
@@ -515,6 +518,9 @@ impl ProjectCreating {
                     ],
                     "Env": [
                         "RUST_LOG=debug",
+                        // If we don't set this, users' dependencies with a `rust-toolchain.toml`
+                        // override will compile with incompatible versions of rust.
+                        format!("RUSTUP_TOOLCHAIN={rustup_toolchain}")
```
chesedo, line 523: Since this is now in the Containerfile is should be safe to remove from here, right?

chesedo (approved):
We can merge if you are happy with the question @oddgrd

gateway/Cargo.toml
oddgrd: Woops, thanks!

gateway/src/project.rs
oddgrd: I was sure I tried this and it didn't work, even though it obviously should, but I must have used an older image. I tested it again now and it works. Good looking out Pieter!

sd2k:
FYI I still experienced this issue with `axum-macros` and had to manually patch to a fork without `rust-toolchain` using `[patch.crates-io]` (i.e. [this commit](https://github.com/sd2k/recipes/commit/18a6aa82019858a60d25b1cb7a521550beb03c64) works (w.r.t this issue), but the previous commit doesn't).

oddgrd:
Hey! This fix hasn't been deployed to production yet, so you'll need the work-around for a little bit longer. By the way, looking at your cargo.toml comments, we do support path dependencies since `0.8.0`.

sd2k:
Ah gotcha, that makes sense.

It might be workspace dependencies, not path dependencies, that aren't working for me then 🤔 uncommenting the `{ workspace = true }` dependencies gives:

```
2023-01-01T20:40:19.707969953Z ERROR {error="Build error: failed to parse manifest at `/opt/shuttle/shuttle-builds/recipe-shuttle/Cargo.toml`"} shuttle_deployer::deployment::queue: service build encountered an error
```

if I change all of those `{ workspace = true }` to `{ path = "../whatever" }` I see:

```
2023-01-02T13:51:44.004967421Z ERROR {error="Build error: failed to get `recipe-db` as a dependency of package `recipe-shuttle-fd2affd4-0a12-43ae-9a8b-c8d5875887a3 v0.1.0 (/opt/shuttle/shuttle-builds/recipe-shuttle)`"} shuttle_deployer::deployment::queue: service build encountered an error
```

I suspect this should be in a separate issue though - lmk if these are unexpected issues and I'll create one 😄

oddgrd:
Hmm, it should work. :thinking: What is your `cargo shuttle --version`?

sd2k:
```
❯ cargo shuttle --version
cargo-shuttle 0.8.1
```

🤔

chesedo:
For the workspace dependencies, we aren't currently archiving the entire workspace but only the single crate. This is why the paths don't work and why the workspace toml can't be found (parsed). I've created a new issue for this #564

PR #546, opened 2022-12-20
ci: fix bin builds

oddgrd:
The `publish-github-release` workflow wasn't working since we can't use interpolated environment variables in circleci paths. To get around this and keep the git tag in the name, I move each binary archive to a subdirectory named after its target, and then flatten these sub-directories before running `ghr`.

Successful workflow run: https://app.circleci.com/pipelines/github/shuttle-hq/shuttle/1391/workflows/8ce94b4f-d14f-456f-bf10-b1fd15575082

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -358,8 +358,6 @@ jobs:
           at: artifacts
       - run:
           name: "Publish Release on GitHub"
-          environment:
-            GITHUB_TOKEN: $GITHUB_TOKEN
           command: |
             find artifacts -mindepth 2 -type f -exec mv -t artifacts {} +
             go install github.com/tcnksm/ghr@v0.16.0
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,7 +122,6 @@ commands:
           name: "Set git tag in the environment"
           command: |
             echo TAG=$(git describe --tags) >> $BASH_ENV
-            printenv TAG
       - run:
           name: Make artifact
           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -358,93 +367,93 @@ jobs:
           environment:
             GITHUB_TOKEN: $GITHUB_TOKEN
           command: |
+            mkdir binaries
+            find artifacts -type f -exec cp -t binaries {} +
```
oddgrd, line 371: I could just flatten artifacts by `mv`ing the bins out of the target dirs here, rather than copying to a new dir.

.circleci/config.yml
```diff
@@ -353,98 +361,94 @@ jobs:
...
           at: artifacts
+      - run:
+          name: "Set tag in environment"
+          command: |
+            cat artifacts/bash.env >> "$BASH_ENV"
+            rm artifacts/bash.env
       - run:
           name: "Publish Release on GitHub"
-          environment:
-            GITHUB_TOKEN: $GITHUB_TOKEN
```
oddgrd, line 359: There was also an issue with github authorization for publishing with `ghr`, removing this resolved it (adding `-t ${GITHUB_TOKEN}` to the `ghr` command was done in an earlier commit). :man_shrugging:

chesedo: Hmm, okay. I just didn't like the token being directly in the terminal which is why I removed the `-t ${GITHUB_TOKEN}`. Hopefully CircleCI never leaks it

chesedo (approved):
LGTM, thanks for all the effort here @oddgrd!

.circleci/config.yml
oddgrd: Yeah, this is how they do it in their tutorial too, so it *should* be okay. :crossed_fingers:

oddgrd:
No problem, I got a much needed crash-course in the intricacies of circleci :smile:

PR #548, opened 2022-12-21
bug: deployer drifting state

chesedo:
Make deployer cleanup invalid states on startup


oddgrd (approved):
LGTM! We have two unwraps in `start` now, but I suppose it's fine since the panic will happen at startup.

PR #550, opened 2022-12-22
Refactor: remove deprecated auth command

oddgrd:
Remove the auth command from the cargo-shuttle CLI, as it's no longer used.


chesedo (approved):
LGTM

PR #551, opened 2022-12-22
Feat: shuttle next bump deps

oddgrd:
Bump the wasmtime dependencies and fix the shuttle-axum test.

For reference: https://github.com/bytecodealliance/wasmtime/blob/main/RELEASES.md

### DIFF
--- a/runtime/src/axum/mod.rs
+++ b/runtime/src/axum/mod.rs
@@ -231,7 +231,7 @@ impl RouterInner {
             .unwrap()
             .into_func()
             .unwrap()
-            .typed::<(RawFd, RawFd), (), _>(&store)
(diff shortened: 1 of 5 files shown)


runtime/src/axum/mod.rs
```diff
@@ -231,7 +231,7 @@ impl RouterInner {
             .unwrap()
             .into_func()
             .unwrap()
-            .typed::<(RawFd, RawFd), (), _>(&store)
+            .typed::<(RawFd, RawFd), ()>(&store)
```
oddgrd, line 234: For reference: https://github.com/bytecodealliance/wasmtime/pull/5275

chesedo (approved):

PR #553, opened 2022-12-28
feat: create subcommand to list all projects of calling account

### LINKED ISSUE #547: List all projects belonging to an account
Currently there is no way to list all projects that belong to a single account. We'd like a subcommand of `cargo shuttle project` like `cargo shuttle project list` that prints out all the projects that belong to the calling account.

This will require extending [the `/projects` API endpoint][1] with a `GET` that returns an array of [project::Response][2] for all the projects that belong to this account. Then, a client-side change will be required to add the [`list` subcommand][3] that consumes that new endpoint via the client and prints out the result in a nicely formatted way (maybe the same tabular format we use for `deployment list`?).

[1]: https://github.com/shuttle-hq/shuttle/blob/998fff7ee056e35b5be000c50f96971f82f07b78/gateway/src/api/latest.rs#L457
[2]: https://github.com/shuttle-hq/shuttle/blob/998fff7ee056e35b5be000c50f96971f82f07b78/common/src/models/project.rs#L8
[3]: https://github.com/shuttle-hq/shuttle/blob/998fff7ee056e35b5be000c50f96971f82f07b78/cargo-shuttle/src/args.rs#L111

emmakuen:
This change creates a subcommand `cargo shuttle project list` that prints out all the projects that belong to the calling account in a tabular format.

resolves #547

### DIFF
--- a/gateway/src/api/latest.rs
+++ b/gateway/src/api/latest.rs
@@ -104,6 +104,23 @@ async fn get_project(
     Ok(AxumJson(response))
 }
 
+async fn get_projects_list(
+    State(RouterState { service, .. }): State<RouterState>,
+    _: User,
+) -> Result<AxumJson<Vec<project::Response>>, Error> {
+    let projects = service
+        .iter_projects_list()
+        .await?
+        .into_iter()
+        .map(|project| project::Response {
+            name: project.0.to_string(),
+            state: project.1.into(),
+        })
+        .collect();
+
+    Ok(AxumJson(projects))
+}
+
 #[instrument(skip_all, fields(%project))]
 async fn post_project(
     State(RouterState {
@@ -457,6 +474,7 @@ impl ApiBuilder {
         self.router = self
             .router
             .route("/", get(get_status))
+            .route("/projects", get(get_projects_list))
             .route(
                 "/projects/:project_name",
                 get(get_project).delete(delete_project).post(post_project),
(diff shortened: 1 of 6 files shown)


gateway/src/service.rs
```diff
@@ -276,6 +276,22 @@ impl GatewayService {
             .ok_or_else(|| Error::from_kind(ErrorKind::ProjectNotFound))
     }
 
+    pub async fn iter_projects_list(
+        &self,
+    ) -> Result<impl Iterator<Item = (ProjectName, Project)>, Error> {
+        let iter = query("SELECT project_name, project_state FROM projects")
```
chesedo, line 282: This is just missing a `WHERE` clause to only return a specific account's projects and not all projects. Might have to rename the method to something like `iter_users_projects` or something :man_shrugging:?

gateway/src/api/latest.rs
```diff
@@ -104,6 +104,23 @@ async fn get_project(
     Ok(AxumJson(response))
 }
 
+async fn get_projects_list(
+    State(RouterState { service, .. }): State<RouterState>,
+    _: User,
```
chesedo, line 109: Regarding the account comment, you should be able to get the account name from this guard.

chesedo (commented):
Wow, nice! Thanks for this @emmakuen!

Just a small comment to not return all the projects from all users.

gateway/src/service.rs
emmakuen: @chesedo Thanks for the feedback! I've updated the query to filter by account name and renamed the method.

gateway/src/api/latest.rs
emmakuen: @chesedo got it, thank you!

chesedo (commented):
This is great, thanks @emmakuen

chesedo (approved):

PR #555, opened 2022-12-28
ci: add Makefile command for Windows to convert .sh files to LF format

joshua-mo-143:
+ add makefile command to convert .sh files to Unix format
+ add relevant documentation in CONTRIBUTING.md

Explanation:
As noted in the CONTRIBUTING.md file, the Makefile currently does not play nicely with Windows and you'll have to set up using WSL/Bash to get anything to work. This isn't necessarily a problem in of itself as you currently need WSL anyway to run Docker Desktop; however, Windows seems to pull files from Git repos in CRLF format rather than LF format which is inconvenient as it breaks `make images`. The added script is intended to simply install dos2unix and then convert the .sh files to Unix format so that `make images` can properly read the .sh files. Relevant documentation has also been added to CONTRIBUTING.MD to explain all of this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -225,9 +225,8 @@ The rest are the following libraries:
 
 Lastly, the `user service` is not a folder in this repository, but is the user service that will be deployed by `deployer`.
 
-
 ## Windows Considerations
-Currently, if you try to use 'make images' on Windows, you may find that the shell files cannot be read by Bash/WSL. This is due to the fact that Windows may have pulled the files in CRLF format rather than LF, which causes problems with Bash as to run the commands, Linux needs the file in LF format. 
+Currently, if you try to use 'make images' on Windows, you may find that the shell files cannot be read by Bash/WSL. This is due to the fact that Windows may have pulled the files in CRLF format rather than LF[^1], which causes problems with Bash as to run the commands, Linux needs the file in LF format. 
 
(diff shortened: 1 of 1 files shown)


oddgrd:
Thanks for this Josh, it's about time we have someone look into our windows issues :smile:. I'm wondering though, have you explored tweaking this git setting: https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf? It might be an even easier solution. :crossed_fingers:

joshua-mo-143:
Oh, I didn't know that actually existed! That would probably be a more convenient solution, do you want me to add that instead then?

oddgrd:
I haven't tried it myself, but if it works I think that would be great, yeah. :pray:

joshua-mo-143:
It looks like the Git commands work, so I've made the required changes and squashed all commits into one final commit. If/when this gets approved, I'll make sure to add matching documentation updates on the docs repo to match this one.

oddgrd (approved):
This is great, thanks Josh, and thanks for offering to update the docs too! I'm wondering, should we also include a link to [the git configuration](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf)?

joshua-mo-143:
Retroactively added the link to the git config page in the documentation for traceability purposes. I'm personally undecided on the issue as we've already addressed the main bulk of the relevant information in our own docs but it doesn't really hurt to add it since having a citation to help back up information always helps.

PR #558, opened 2022-12-29
feat: return streaming body from wasm router

oddgrd:
This uses `futures::iter` to create a stream from a `BufReader::bytes()`. I needed to open another filedescriptor for body, since `.bytes()` consumes the FD. 

This improves performance (in simple benchmarks) by about ~30%~20%.

I've also experimented with async wasmtime, which wasn't so hard to set up, but making the cap_std unixstreams async proved difficult. There may not be too much to gain here either, since from what I can gather WASM async still has a ways to go: https://docs.wasmtime.dev/api/wasmtime/struct.Config.html#method.async_support.

### DIFF
--- a/runtime/src/axum/mod.rs
+++ b/runtime/src/axum/mod.rs
@@ -198,24 +198,22 @@ impl RouterInner {
             .unwrap();
 
         let (mut parts_stream, parts_client) = UnixStream::pair().unwrap();
-        let (body_read_stream, body_read_client) = UnixStream::pair().unwrap();
         let (mut body_write_stream, body_write_client) = UnixStream::pair().unwrap();
+        let (body_read_stream, body_read_client) = UnixStream::pair().unwrap();
 
         let parts_client = WasiUnixStream::from_cap_std(parts_client);
-        let body_read_client = WasiUnixStream::from_cap_std(body_read_client);
         let body_write_client = WasiUnixStream::from_cap_std(body_write_client);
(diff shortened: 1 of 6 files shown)


runtime/src/axum/mod.rs
```diff
@@ -219,21 +226,19 @@ impl RouterInner {
         parts_stream.write_all(&request_rmp).unwrap();
 
         // write body
-        body_stream
+        body_write_stream
             .write_all(hyper::body::to_bytes(body).await.unwrap().as_ref())
```
chesedo, line 230: I might be wrong, but won't `to_bytes()` consume the body? Ie if the service never uses the body, then this will still fully "read" it which is potentially not what we want

chesedo (approved):
Wow, this is exciting :star_struck:. Just wanna check one thing.

Also, remember to update the codegen with the final code.

runtime/src/axum/mod.rs
oddgrd: Yes, it's definitely not ideal that we read the full body for every request. Passing the body to hyper or axum is easy since they are async, but streaming it across the filedescriptors is tricky since they are sync. I'm trying to find a solution for this, but a different design may be required. 

As long as we do use `to_bytes()` we should probably do a size check, as suggested in the docs.

oddgrd: I wasn't able to find a way around `to_bytes` for now, so I implemented the request body size check.

PR #559, opened 2022-12-30
feat: embed runtime into client and deployer

chesedo:


tmp/axum-wasm-expanded/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+# via "hyper/tcp" which is not compatible with wasi
+axum = { version = "0.6.0", default-features = false }
+futures-executor = "0.3.21"
+http = "0.2.7"
+tower-service = "0.3.1"
+rmp-serde = { version = "1.1.1" }
+
+[dependencies.shuttle-common]
+path = "../../common"
+default-features = false
```
oddgrd, line 20: I removed the default feature from common btw, since we can't set `default-features = false` in inherited deps: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace

service/Cargo.toml
```diff
@@ -62,7 +62,9 @@ tokio = { version = "1.22.0", features = ["macros"] }
 uuid = { workspace = true, features = ["v4"] }
 
 [features]
-codegen = ["shuttle-codegen"]
+default = ["codegen"]
```
oddgrd, line 65: I removed this from default in my merge of main, which I think was a mistake since all dependents of shuttle-service should need it. Thanks! :sweat_smile:

oddgrd (commented):
~I noticed that on every `cargo shuttle run` we need to run a compilation step, I assume it's the runtime build.rs, which takes 12 seconds on my system. (tested with axum hello-world and poem postgres). I know you mentioned something about this, but do you think we can remedy it? Does it also happen for deploys?~ 

~That said, I love the idea and if we can figure out the re-compilation in a follow-up I think we can merge this. I need to read up on build scripts before I can offer advice on this though.~ See comment on cargo-shuttle build.rs, I found a way to solve it (for local runs on my system, at least).

There are also some minor CI errors (aside from the out-of-scope deployer ones).

deployer/build.rs
```diff
@@ -0,0 +1,17 @@
+use std::{env, process::Command};
+
+fn main() {
+    println!("cargo:rerun-if-changed=runtime");
```
oddgrd, line 4: I tried changing this to an absolute path to runtime now (`/home/oddgrd/dev/shuttle/runtime`), and then it doesn't recompile on every `cargo shuttle run` anymore.

cargo-shuttle/build.rs
```diff
@@ -0,0 +1,17 @@
+use std::{env, process::Command};
+
+fn main() {
+    println!("cargo:rerun-if-changed=runtime");
```
oddgrd, line 4: I tried changing this to an absolute path to runtime now (/home/oddgrd/dev/shuttle/runtime), and then it doesn't recompile on every cargo shuttle run anymore.

chesedo: Hmm, okay. I'll see how to make this update. Thanks!

Issue #350, opened 2022-09-22
Choose hostname when `cargo shuttle run`


tguichaoua:
It could be usefull when debuging to host the server not only on the `localhost` but also on network to allow other devices to send request.
For example [`vite`](https://vitejs.dev/) has a `--host` flag for this.

```
  VITE v3.1.3  ready in 399 ms

  ➜  Local:   http://localhost:5173/
  ➜  Network: http://192.168.1.100:5173/
```

trkohler:
hello! can i take this one as a part of hacktoberfest?

chesedo:
@0xpwnstar feel free to take it if you are still interested

joshua-mo-143:
I've been working on this over the past few days and believe I may have a solution to this problem - just need to do some testing and I will hopefully be able to push a PR if all goes well.

chesedo:
@joshua-mo-143 sounds exciting! :partying_face:

chesedo:
This is closed by #565

PR #565, opened 2023-01-04
feat: add flag for router IP local run

joshua-mo-143:
Reference: issue #350 

Summary:
+ adds `local_ip_address` crate to be able to grab local router IP
+ add new flag `--router_ip` to be able to use router IP instead of localhost, which allows other devices on the network to access the local deployment.
+ relevant testing has been added to account for new feature

Notes:
`local_ip_address` was added for this feature implementation as it looks to be a stable crate that doesn't appear to be going inactive any time soon and appears to work across OSes with no code change required (tested using Ubuntu and Win11). In the event that it does (or there's some breaking issue that renders it unusable), it looks like a method would have to be implemented per OS to get the router IP.

Windows users who are using this flag in Bash/WSL may have problems getting other devices to get this to work if they are not aware of the fact that there are extra steps[^1] required to be able to connect external devices to a locally deployed run in Bash/WSL due to the fact it uses the WSL virtual ethernet. Not a huge problem as it seems most people who use shuttle are currently running Linux, but it's something to keep in mind should any issues regarding this pop up.

The new test also uses the same example as one of the other tests (the "hello world" Rocket example), which may or may not break the tests. Not entirely sure because trying to run all integration tests is broken on my laptop due to memory issues, but although the test passes in isolation it's probably something to be aware of.

[^1]:(https://learn.microsoft.com/en-us/windows/wsl/networking)

### DIFF
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -143,9 +143,9 @@ pub struct RunArgs {
     /// port to start service on
     #[clap(long, default_value = "8000")]
     pub port: u16,
-    /// use router ip address instead of localhost
+    /// use 0.0.0.0 instead of localhost (for usage with local external devices)
     #[clap(long)]
-    pub router_ip: bool,
+    pub external: bool,
 }
 
 #[derive(Parser, Debug)]
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -145,7 +145,7 @@ pub struct RunArgs {
     pub port: u16,
     /// use router ip address instead of localhost
     #[clap(long)]
-    pub router_ip: bool
+    pub router_ip: bool,
 }
 
 #[derive(Parser, Debug)]
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -143,6 +143,9 @@ pub struct RunArgs {
     /// port to start service on
     #[clap(long, default_value = "8000")]
     pub port: u16,
+    /// use router ip address instead of localhost
+    #[clap(long)]
+    pub router_ip: bool
 }
 
 #[derive(Parser, Debug)]
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -406,7 +406,7 @@ impl Shuttle {
             working_directory.to_path_buf(),
         )?;
         let addr = if run_args.router_ip {
-            std::net::IpAddr::V4(Ipv4Addr::new(0,0,0,0))
+            std::net::IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0))
         } else {
             Ipv4Addr::LOCALHOST.into()
         };
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -27,7 +27,6 @@ use futures::StreamExt;
 use git2::{Repository, StatusOptions};
 use ignore::overrides::OverrideBuilder;
 use ignore::WalkBuilder;
-use local_ip_address::local_ip;
 use shuttle_common::models::{project, secret};
 use shuttle_service::loader::{build_crate, Loader};
 use shuttle_service::Logger;
@@ -406,8 +405,13 @@ impl Shuttle {
             secrets,
             working_directory.to_path_buf(),
         )?;
-        let addr = SocketAddr::new(if run_args.router_ip 
-                                    {local_ip().unwrap()} else {Ipv4Addr::LOCALHOST.into()}, run_args.port);
+        let addr = if run_args.router_ip {
+            std::net::IpAddr::V4(Ipv4Addr::new(0,0,0,0))
+        } else {
+            Ipv4Addr::LOCALHOST.into()
+        };
+
+        let addr = SocketAddr::new(addr, run_args.port);
 
         trace!("loading project");
         println!(
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -405,7 +405,7 @@ impl Shuttle {
             secrets,
             working_directory.to_path_buf(),
         )?;
-        let addr = if run_args.router_ip {
+        let addr = if run_args.external {
             std::net::IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0))
         } else {
             Ipv4Addr::LOCALHOST.into()
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -27,6 +27,7 @@ use futures::StreamExt;
 use git2::{Repository, StatusOptions};
 use ignore::overrides::OverrideBuilder;
 use ignore::WalkBuilder;
+use local_ip_address::local_ip;
 use shuttle_common::models::{project, secret};
 use shuttle_service::loader::{build_crate, Loader};
 use shuttle_service::Logger;
@@ -405,7 +406,8 @@ impl Shuttle {
             secrets,
             working_directory.to_path_buf(),
         )?;
-        let addr = SocketAddr::new(Ipv4Addr::LOCALHOST.into(), run_args.port);
+        let addr = SocketAddr::new(if run_args.router_ip 
+                                    {local_ip().unwrap()} else {Ipv4Addr::LOCALHOST.into()}, run_args.port);
 
         trace!("loading project");
         println!(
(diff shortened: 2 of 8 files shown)


cargo-shuttle/tests/integration/run.rs
```diff
@@ -359,3 +360,68 @@ async fn thruster_hello_world() {
 
     assert_eq!(request_text, "Hello, World!");
 }
+
+/// creates a `cargo-shuttle` run instance with some reasonable defaults set, but using the router IP instead.
+async fn cargo_shuttle_run_with_router_ip(working_directory: &str) -> u16 {
+    let working_directory = canonicalize(working_directory).unwrap();
+    let port = pick_unused_port().unwrap();
+    let ip = local_ip().unwrap();
+    let run_args = RunArgs { port, router_ip: true };
```
chesedo, line 369: Can you find a way to remove the duplicate code between this new function and `cargo_shuttle_run`?

Maybe a new function like:
```rust
async fn new_function(working_directory: &str, router_ip: bool) -> u16 {
...
}
```

That gets called by both `cargo_shuttle_run` and `cargo_shuttle_run_with_router_ip`.

chesedo (commented):
Nice, thanks for this @joshua-mo-143!

I'm wondering, won't binding to the address `0.0.0.0` be enough? Then the `local_ip` crate won't be needed and I would expect users to know how to get their address on the network if they want to test from other devices.

cargo-shuttle/src/lib.rs
```diff
@@ -405,7 +406,8 @@ impl Shuttle {
             secrets,
             working_directory.to_path_buf(),
         )?;
-        let addr = SocketAddr::new(Ipv4Addr::LOCALHOST.into(), run_args.port);
+        let addr = SocketAddr::new(if run_args.router_ip
```
chesedo, line 409: This is cool, but probably not quite the Rust way to use the `if`. It also reads a bit hard. Maybe do something along the lines of:

```rust
let addr = if run_args.router_ip {
         local_ip().unwrap()
      } else {
         Ipv4Addr::LOCALHOST.into()
      };
let addr = SocketAddr::new(addr, , run_args.port);
```

joshua-mo-143: Ah thanks for catching that, I have fixed as per requested - once the amendments are ready I'll push them through

joshua-mo-143:
Ah that makes a lot more sense, I didn't realise you could bind it to 0.0.0.0 - that would probably be more optimal since it eliminates the need to add a dependency and would eliminate WSL-specific issues which would be helpful for compatibility. Looks like I still have a lot to learn about using IPs and networking 😅

oddgrd:
This looks cool Josh! It looks like you need to do a `cargo fmt` to make the CI happy, and maybe clippy. And dont worry about squashing commits, we'll do that when we merge. 🙂

cargo-shuttle/src/args.rs
```diff
@@ -143,6 +143,9 @@ pub struct RunArgs {
     /// port to start service on
     #[clap(long, default_value = "8000")]
     pub port: u16,
+    /// use router ip address instead of localhost
+    #[clap(long)]
+    pub router_ip: bool,
```
chesedo, line 148: I'm not sold on the name, but can't think of anything better either :sweat_smile:

cargo-shuttle/tests/integration/run.rs
```diff
@@ -5,10 +5,24 @@ use std::{fs::canonicalize, process::exit, time::Duration};
 use tokio::time::sleep;
 
 /// creates a `cargo-shuttle` run instance with some reasonable defaults set.
-async fn cargo_shuttle_run(working_directory: &str) -> u16 {
+async fn cargo_shuttle_run(working_directory: &str, router_ip: bool) -> u16 {
     let working_directory = canonicalize(working_directory).unwrap();
+
+    let url = if !router_ip { "localhost" } else { "0.0.0.0" };
```
chesedo, line 11: Maybe also return the url (or full address) from this function now

cargo-shuttle/tests/integration/run.rs
```diff
@@ -359,3 +373,19 @@ async fn thruster_hello_world() {
 
     assert_eq!(request_text, "Hello, World!");
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn rocket_hello_world_with_router_ip() {
+    let port = cargo_shuttle_run("../examples/rocket/hello-world", true).await;
+
+    let request_text = reqwest::Client::new()
+        .get(format!("http://0.0.0.0:{port}/hello"))
```
chesedo, line 382: Reason for returning the url (or full address) from the helper function.

chesedo (commented):
So sorry for another round of changes, but it is small :smile:

cargo-shuttle/src/args.rs
joshua-mo-143: The only other things I can really think of are `network` or `local_ip`, but yeah to be honest I'm not huge on either of them and `router_ip` is not really that descriptive

cargo-shuttle/tests/integration/run.rs
joshua-mo-143: Ok that makes sense, I'll update and send the changes through. Should not be too much of an issue

joshua-mo-143:
It's ok, I appreciate the checks to make sure everything is optimised before pushing through - I'm personally more concerned about solving the circleCI cargo shuttle test failing since that seems to be a pressing issue. It looks like it's running into the same issue I had in local which is that it failed on trying to wait for start up

chesedo:
The "rocket/postgres" has intermittent issues starting up, so don't worry about it too much. If your new test works locally then I'm fine merging this even if the other tests fail on the CI

cargo-shuttle/src/args.rs
joshua-mo-143: Honestly tempted to use `network` or `lan` as the flag since `router_ip` or `use_router_ip` isn't technically accurate anymore, what do you think?

cargo-shuttle/tests/integration/run.rs
joshua-mo-143: Amended - please see new commit

cargo-shuttle/src/args.rs
chesedo: Hmm, I was thinking more like 'host' but don't like it. How about `external`?

cargo-shuttle/tests/integration/run.rs
joshua-mo-143: Amended - please see new commit

cargo-shuttle/src/args.rs
joshua-mo-143: Ok that sounds like a good idea, I'll add that in then

cargo-shuttle/tests/integration/run.rs
```diff
@@ -5,10 +5,28 @@ use std::{fs::canonicalize, process::exit, time::Duration};
...
+        RunArgs {
+            port,
+            router_ip: false,
+        }
+    } else {
+        RunArgs {
+            port,
+            router_ip: true,
+        }
+    };
```
chesedo, line 29:
```suggestion
    let run_args = RunArgs {
        port,
        router_ip,
    };
```

Just something small

cargo-shuttle/tests/integration/run.rs
chesedo: Nice :partying_face:

cargo-shuttle/tests/integration/run.rs
joshua-mo-143: Amended on new commit

cargo-shuttle/tests/integration/run.rs
```diff
@@ -85,11 +88,11 @@ async fn rocket_secrets() {
 // This example uses a shared Postgres. Thus local runs should create a docker container for it.
 #[tokio::test(flavor = "multi_thread")]
 async fn rocket_postgres() {
-    let port = cargo_shuttle_run("../examples/rocket/postgres").await;
+    let url = cargo_shuttle_run("../examples/rocket/secrets", false).await;
```
oddgrd, line 91: You probably did a find & replace for these, it looks like you forgot to change the paths back to the correct example (for this and the following tests).

joshua-mo-143: Oh whoops 🤦 thanks for catching that, I'll send a fix in quickly

oddgrd: No worries, been there :smile: Have a good weekend!

joshua-mo-143: Thanks, you too!

oddgrd (approved):
Thanks a lot Josh, this is a really useful feature! :rocket:

PR #566, opened 2023-01-06
feat: refactor router and router inner

oddgrd:
Refactor the axum-wasm router, and implement some error handling.

I have some doubts about the error handling in `handle_request`, currently I propagate the errors, `trace!` their context and return a `500` from hyper. But I'm thinking we should panic on a lot of these errors, since if we can't connect/write to the module in one request, we probably won't be able to in the next.

The wasmtime API's return `anyhow::Errors`, so I opted to use that and not create an `Error` enum. However since this refactor is meant to ease future development, we might want to do it now anyway? Another thing I'm wondering is if the `RouterBuilder` is worthwhile, since I feel it might as well just be a function currently, but here again having the builder might be worthwhile for future development when we might do more configuration.


common/src/wasm.rs
```diff
@@ -1,4 +1,4 @@
-use hyper::http::{HeaderMap, Method, Request, Response, StatusCode, Uri, Version};
+use hyper::http::{self, HeaderMap, Method, Request, Response, StatusCode, Uri, Version};
```
chesedo, line 1: `hyper::http` is actually just a re-export of the `http` crate. So we can make hyper a dev-dependency only

runtime/src/axum/mod.rs
```diff
@@ -140,67 +152,68 @@ struct RouterBuilder {
...
             module,
-        };
-        Router { inner }
+        })
     }
 }
 
 #[derive(Clone)]
-struct RouterInner {
+struct Router {
```
chesedo, line 190: Nice :partying_face:

runtime/src/axum/mod.rs
```diff
@@ -140,67 +152,68 @@ struct RouterBuilder {
...
 
-        let (mut parts_stream, parts_client) = UnixStream::pair().unwrap();
-        let (mut body_write_stream, body_write_client) = UnixStream::pair().unwrap();
-        let (body_read_stream, body_read_client) = UnixStream::pair().unwrap();
+        let (mut parts_stream, parts_client) =
+            UnixStream::pair().context("failed to open unixstream")?;
+        let (mut body_write_stream, body_write_client) =
+            UnixStream::pair().context("failed to open unixstream")?;
+        let (body_read_stream, body_read_client) =
+            UnixStream::pair().context("failed to open unixstream")?;
```
chesedo, line 216: Maybe we should rather have "failed to open unix stream for parts" else we won't know which one failed

runtime/src/axum/mod.rs
```diff
@@ -232,98 +247,98 @@ impl RouterInner {
...
+        async move {
+            Ok::<_, Infallible>(service_fn(move |req: Request<Body>| {
+                let mut router = router.clone();
+                async move {
+                    Ok::<_, Infallible>(match router.handle_request(req).await {
+                        Ok(res) => res,
+                        Err(err) => {
+                            error!("error sending request: {}", err);
+                            Response::builder()
+                                .status(hyper::http::StatusCode::INTERNAL_SERVER_ERROR)
```
chesedo, line 318: I love this! :heart_eyes:

chesedo (commented):
Wow, this is great, thanks @oddgrd!

I think going with anyhow is fine. This isn't a library and we can always upgrade in the future

runtime/src/axum/mod.rs
oddgrd: Good point.

PR #568, opened 2023-01-06
feat: hook in runtime logs

chesedo:


runtime/src/axum/mod.rs
```diff
@@ -32,15 +34,27 @@ use wasmtime_wasi::{WasiCtx, WasiCtxBuilder};
...
-    kill_tx: std::sync::Mutex<Option<oneshot::Sender<String>>>,
+    router: Mutex<Option<Router>>,
+    logs_rx: Mutex<Option<Receiver<Result<runtime::LogItem, Status>>>>,
+    logs_tx: Mutex<Sender<Result<runtime::LogItem, Status>>>,
+    kill_tx: Mutex<Option<oneshot::Sender<String>>>,
 }
 
 impl AxumWasm {
     pub fn new() -> Self {
+        // Allow about 2^15 = 32k logs of backpressure
```
oddgrd, line 45: Thanks for the detailed comments!

tmp/axum-wasm-expanded/src/lib.rs
```diff
@@ -44,15 +48,25 @@ async fn uppercase(body: BodyStream) -> impl IntoResponse {
...
     fd_5: std::os::wasi::prelude::RawFd,
 ) {
     use axum::body::HttpBody;
+    use shuttle_common::wasm::Logger;
     use std::io::{Read, Write};
     use std::os::wasi::io::FromRawFd;
+    use tracing_subscriber::prelude::*;
 
-    println!("inner handler awoken; interacting with fd={fd_3},{fd_4},{fd_5}");
+    println!("inner handler awoken; interacting with fd={fd_2},{fd_3},{fd_4},{fd_5}");
```
oddgrd, line 62: Should we change these to `trace!`? Here and in the codegen.

tmp/axum-wasm-expanded/src/lib.rs
```diff
@@ -44,15 +48,25 @@ async fn uppercase(body: BodyStream) -> impl IntoResponse {
...
+
+    // file descriptor 2 for writing logs to
+    let logs_fd = unsafe { std::fs::File::from_raw_fd(fd_2) };
+
+    tracing_subscriber::registry()
+        .with(Logger::new(logs_fd))
+        .init(); // this sets the subscriber as the global default and also adds a compatibility layer for capturing `log::Record`s
```
oddgrd, line 69: This turned out really neat and elegant, good job Pieter! :star_struck:

oddgrd (approved):
Looks great to me. Thanks for all the comments and tests!

tmp/axum-wasm-expanded/src/lib.rs
chesedo: I wanted to initially, but ended up deciding not to for this reason: if a user decides not to use a `tracing` call in a request path then the request has our current 10ms latency. But adding a `tracing` call makes the latency 12ms. So I figured it best to safe the 2ms for users who don't use `tracing` :smile:

PR #571, opened 2023-01-10
Respect $PORT environment variable

canac:
It is a common pattern for dev servers to let the `$PORT` environment variable dictate which port they will listen on. This change allows the port to be passed to `cargo shuttle run` via the `$PORT` environment variable in addition to the `--port` argument.


oddgrd (approved):
Apologies @canac, this went under the radar. I'm happy to merge this with the conflict resolved (if you've moved on, I can do it too). :slightly_smiling_face:

PR #574, opened 2023-01-11
feat: change log read to spawn blocking

oddgrd:
This improves throughput by about 10%, but the issue of tasks spawning perpetually after a stress test remains.

### DIFF
--- a/runtime/src/axum/mod.rs
+++ b/runtime/src/axum/mod.rs
@@ -268,14 +268,14 @@ impl Router {
             .data_mut()
             .insert_file(BODY_READ_FD, Box::new(body_read_client), FileCaps::all());
 
-        tokio::task::spawn(async move {
+        tokio::task::spawn_blocking(move || {
             let mut iter = logs_stream.bytes().filter_map(Result::ok);
 
             while let Some(log) = Log::from_bytes(&mut iter) {
                 let mut log: runtime::LogItem = log.into();
                 log.id = deployment_id.clone();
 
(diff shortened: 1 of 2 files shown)


runtime/src/axum/mod.rs
```diff
@@ -268,14 +268,14 @@ impl Router {
...
-        tokio::task::spawn(async move {
+        tokio::task::spawn_blocking(move || {
             let mut iter = logs_stream.bytes().filter_map(Result::ok);
 
             while let Some(log) = Log::from_bytes(&mut iter) {
                 let mut log: runtime::LogItem = log.into();
                 log.id = deployment_id.clone();
 
-                logs_tx.send(Ok(log)).await.unwrap();
+                logs_tx.blocking_send(Ok(log)).unwrap();
```
chesedo, line 278: What happens if we change `logs_tx` to a crossbeam channel? Or any other channel which is not async?

chesedo (commented):
Looks good. Small question

runtime/src/axum/mod.rs
oddgrd: Hmm, we'd have to make some changes to `subscribe_logs` then as well, since it uses `ReceiverStream` for the response and it takes a `tokio::sync::mpsc::Receiver`: https://github.com/shuttle-hq/shuttle/blob/shuttle-next/runtime/src/axum/mod.rs#L143

chesedo: Ahh, ok

chesedo (approved):
LGTM

PR #575, opened 2023-01-11
feat: deployer next

chesedo:
Still have to update the deployer tests. But will do that in a new PR


deployer/migrations/0001_next.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE deployments ADD is_next BOOLEAN DEFAULT 0 NOT NULL;
```
oddgrd, line 1: Does this need to be `ADD COLUMN`?

deployer/src/runtime_manager.rs
```diff
@@ -0,0 +1,109 @@
...
+        } else {
+            let (_runtime, mut runtime_client) = runtime::start(
+                &self.binary_bytes,
+                false,
+                runtime::StorageManagerType::Artifacts(self.artifacts_path.clone()),
+                &self.provisioner_address,
+                6001,
+            )
+            .await
+            .unwrap();
```
oddgrd, line 56: Should we handle these errors?

deployer/src/runtime_manager.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    pub async fn get_runtime_client(&mut self, is_next: bool) -> RuntimeClient<Channel> {
+        if is_next {
+            self.get_next_runtime_client().await
+        } else {
+            self.get_legacy_runtime_client().await
+        }
+    }
+
+    async fn get_legacy_runtime_client(&mut self) -> RuntimeClient<Channel> {
```
oddgrd, line 44: It would be nice to deduplicate these, but it might not be so easy. Something for a future PR maybe.

oddgrd (approved):
LGTM, just a few small comments.

deployer/src/runtime_manager.rs
chesedo: I thought about it but did not see an easy solution, let me try another swing though

deployer/migrations/0001_next.sql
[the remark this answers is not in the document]
chesedo: Yes, that will read better

deployer/src/runtime_manager.rs
```diff
@@ -0,0 +1,127 @@
...
+            legacy_process: None,
+            next: None,
+            next_process: None,
+            artifacts_path,
+            provisioner_address,
+            log_sender,
+        }))
+    }
+
+    pub async fn get_runtime_client(
```
oddgrd, line 39: Cool, you found a way! This'll be nice to have when we add more runtimes in the future as well.

oddgrd (approved):
Looks good, I think we can merge this (into shuttle-next). The CI errors are in shuttle-next as well, I'll try to figure them out.

PR #576, opened 2023-01-11
chore: opt out of panamax

brokad:


oddgrd (commented):
I forgot to mention this, but if the panamax container is live it will download the crates (to my roommates great frustration, I wasn't aware of this for a while). Can we also make the `USE_PANAMAX` flag toggle the spawning of the container? It may also be enough to toggle spawning `deck-chores` :thinking: 

I haven't done this before, but it looks like we can use [profiles](https://docs.docker.com/compose/profiles/) for this.

chesedo (commented):
LGTM

What Oddbjorn said will be helpful

chesedo (approved):

PR #577, opened 2023-01-12
feat: DX

chesedo:
Makes a new crate with the minimal dependencies for a next project

### DIFF
--- a/tmp/axum-wasm/src/lib.rs
+++ b/tmp/axum-wasm/src/lib.rs
@@ -1,6 +1,6 @@
 use tracing::debug;
 
-shuttle_codegen::app! {
+shuttle_next::app! {
     #[shuttle_codegen::endpoint(method = get, route = "/hello")]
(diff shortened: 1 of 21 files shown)


tmp/axum-wasm/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 use tracing::debug;
 
-shuttle_codegen::app! {
+shuttle_next::app! {
     #[shuttle_codegen::endpoint(method = get, route = "/hello")]
```
oddgrd, line 4: Should we change this to  `shuttle_next::endpoint`?

oddgrd (approved):
This is much cleaner, thanks!

tmp/axum-wasm/src/lib.rs
chesedo: :man_facepalming: Yes

PR #578, opened 2023-01-12
feat: get runtime binary from cargo install

oddgrd:


Containerfile
```diff
@@ -37,6 +37,10 @@ COPY --from=cache /build/ /usr/src/shuttle/
 FROM shuttle-common
 ARG folder
 COPY ${folder}/prepare.sh /prepare.sh
+RUN curl -OL https://github.com/protocolbuffers/protobuf/releases/download/v21.9/protoc-21.9-linux-x86_64.zip &&\
```
oddgrd, line 40: The prepare.sh install of shuttle-runtime complained errored on not having `protoc` installed, but it should be installed. Is there a better way to do this?

chesedo (approved):
LGTM, bar the comment :smile:

PR #579, opened 2023-01-16
feat: shuttle-next

chesedo:


runtime/src/axum/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+    ) -> Result<tonic::Response<StartResponse>, Status> {
+        let StartRequest {
+            deployment_id,
+            port,
+            ..
+        } = request.into_inner();
+
+        let address = SocketAddr::new(Ipv4Addr::LOCALHOST.into(), port as u16);
+
+        let logs_tx = self.logs_tx.lock().unwrap().clone();
```
chesedo, line 107: Comment missing for why `unwrap()` is safe to call

runtime/src/axum/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+        Ok(tonic::Response::new(message))
+    }
+
+    type SubscribeLogsStream = ReceiverStream<Result<runtime::LogItem, Status>>;
+
+    async fn subscribe_logs(
+        &self,
+        _request: tonic::Request<SubscribeLogsRequest>,
+    ) -> Result<tonic::Response<Self::SubscribeLogsStream>, Status> {
+        let logs_rx = self.logs_rx.lock().unwrap().deref_mut().take();
```
chesedo, line 140: Missing comment for why `unwrap` is safe. Do these even need to be behind a mutex?

runtime/src/axum/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+    async fn stop(
+        &self,
+        request: tonic::Request<StopRequest>,
+    ) -> Result<tonic::Response<StopResponse>, Status> {
+        let request = request.into_inner();
+
+        let service_name = ServiceName::from_str(request.service_name.as_str())
+            .map_err(|err| Status::from_error(Box::new(err)))?;
+
+        let kill_tx = self.kill_tx.lock().unwrap().deref_mut().take();
```
chesedo, line 158: Does the kill channel need to be behind a mutex? Why not just set it on `.new()`?

runtime/src/axum/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+        })
+    }
+
+    fn src<P: AsRef<Path>>(mut self, src: P) -> Self {
+        self.src = Some(src.as_ref().to_path_buf());
+        self
+    }
+
+    fn build(self) -> anyhow::Result<Router> {
+        let file = self.src.expect("module path should be set");
```
chesedo, line 204: `expect()` call

runtime/src/legacy/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+            deployment_id,
+            service_name,
+            port,
+        } = request.into_inner();
+        let service_address = SocketAddr::new(Ipv4Addr::LOCALHOST.into(), port as u16);
+
+        let service_name = ServiceName::from_str(service_name.as_str())
+            .map_err(|err| Status::from_error(Box::new(err)))?;
+
+        let deployment_id = Uuid::from_slice(&deployment_id).unwrap();
```
chesedo, line 138: Some `unwrap()` calls

runtime/src/legacy/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+
+        let mut factory = abstract_factory.get_factory(
+            service_name,
+            deployment_id,
+            secrets,
+            self.storage_manager.clone(),
+        );
+        trace!("got factory");
+
+        let logs_tx = self.logs_tx.lock().unwrap().clone();
```
chesedo, line 148: Can we remove the mutex on the logs and kill channels?

runtime/src/legacy/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+                }
+            });
+
+            Ok(Response::new(ReceiverStream::new(rx)))
+        } else {
+            Err(Status::internal("logs have already been subscribed to"))
+        }
+    }
+
+    // todo: this doesn't currently stop the service, since we can't stop the tokio runtime it
```
chesedo, line 193: Todo - I think it has been addressed though

runtime/src/provisioner_factory.rs
```diff
@@ -0,0 +1,156 @@
...
+use shuttle_proto::provisioner::{
+    database_request::DbType, provisioner_client::ProvisionerClient, DatabaseRequest,
+};
+use shuttle_service::{Factory, ServiceName};
+use tonic::{transport::Channel, Request};
+use tracing::{debug, info, trace};
+use uuid::Uuid;
+
+/// Trait to make it easy to get a factory (service locator) for each service being started
+pub trait AbstractFactory<S: StorageManager>: Send + 'static {
```
chesedo, line 14: I don't think the abstract factory is still needed (it's no longer being send across threads). We can just create the factory below directly

service/tests/integration/build_crate.rs
```diff
@@ -1,14 +1,18 @@
...
     let (tx, _) = crossbeam_channel::unbounded();
     let project_path = format!("{}/tests/resources/not-shuttle", env!("CARGO_MANIFEST_DIR"));
-    let so_path = build_crate(Default::default(), Path::new(&project_path), false, tx)
+    let so_path = match build_crate(Default::default(), Path::new(&project_path), false, tx)
         .await
-        .unwrap();
+        .unwrap()
+    {
+        Runtime::Legacy(path) => path,
+        _ => unreachable!(),
```
chesedo, line 14: Test changes in this file needs some updating

service/tests/resources/bind-panic/Cargo.toml
```diff
@@ -9,4 +9,4 @@ crate-type = ["cdylib"]
 [workspace]
 
 [dependencies]
-shuttle-service = { path = "../../../" }
+shuttle-service = { path = "../../../", features = ["codegen"] }
```
chesedo, line 12: I think this default feature was restored

service/tests/resources/build-panic/Cargo.toml
```diff
@@ -9,4 +9,4 @@ crate-type = ["cdylib"]
 [workspace]
 
 [dependencies]
-shuttle-service = { path = "../../../" }
+shuttle-service = { path = "../../../", features = ["codegen"] }
```
chesedo, line 12: I think this default feature was restored

service/tests/resources/is-cdylib/Cargo.toml
```diff
@@ -10,4 +10,4 @@ crate-type = ["cdylib", "staticlib"]
 
 [dependencies]
 rocket = "0.5.0-rc.2"
-shuttle-service = { path = "../../../", features = ["web-rocket"] }
+shuttle-service = { path = "../../../", features = ["codegen", "web-rocket"] }
```
chesedo, line 13: I think this default feature was restored

service/tests/resources/not-cdylib/Cargo.toml
```diff
@@ -10,4 +10,4 @@ crate-type = ["staticlib"]
 
 [dependencies]
 rocket = "0.5.0-rc.2"
-shuttle-service = { path = "../../../", features = ["web-rocket"] }
+shuttle-service = { path = "../../../", features = ["codegen", "web-rocket"] }
```
chesedo, line 13: I think this default feature was restored

service/tests/resources/sleep/Cargo.toml
```diff
@@ -9,4 +9,4 @@ crate-type = ["cdylib"]
 [workspace]
 
 [dependencies]
-shuttle-service = { path = "../../../" }
+shuttle-service = { path = "../../../", features = ["codegen"] }
```
chesedo, line 12: I think this default feature was restored

service/tests/resources/sqlx-pool/Cargo.toml
```diff
@@ -9,6 +9,6 @@ crate-type = ["cdylib"]
 [workspace]
 
 [dependencies]
-shuttle-service = { path = "../../../" }
+shuttle-service = { path = "../../../", features = ["codegen"] }
```
chesedo, line 12: I think this default feature was restored

tmp/axum-wasm/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    #[tokio::test]
+    async fn hello() {
+        let request = Request::builder()
+            .uri("http://local.test/hello")
+            .method(Method::GET)
+            .body(axum::body::boxed(axum::body::Body::empty()))
+            .unwrap();
+
+        let response = __app(request).await;
```
chesedo, line 31: We should confirm we are happy with this for testing from users

.circleci/config.yml
```diff
@@ -148,9 +148,9 @@ jobs:
       - restore-cargo-cache
       - install-protoc
       - run: cargo fmt --all --check
-      - run: cargo install cargo-sort
-      # TODO: this is incompatible with workspace inheritance, uncomment when 
+      # TODO: this is incompatible with workspace inheritance, uncomment when
       # https://github.com/DevinR528/cargo-sort/pull/29 is merged
+      # - run: cargo install cargo-sort
```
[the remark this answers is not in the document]
oddgrd: It's incompatible with workspace inheritance, but there is a suggested alternative in this issue: https://github.com/DevinR528/cargo-sort/issues/47

Cargo.toml
```diff
@@ -18,6 +20,7 @@ exclude = [
   "resources/persist",
   "resources/secrets",
   "resources/shared-db",
+  "tmp",
```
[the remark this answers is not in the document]
oddgrd: The test in `runtime/axum` relies on `axum-wasm-expanded`, at least the module it generates. Shall we move it to `runtime/tests/resources/...`? Keeping it in expanded form might be useful to have when we develop the module. We should also move the unexpanded version to examples of course.

codegen/src/next/mod.rs
```diff
@@ -0,0 +1,527 @@
...
+            // file descriptor 3 for reading and writing http parts
+            let mut parts_fd = unsafe { std::fs::File::from_raw_fd(parts_fd) };
+
+            let reader = std::io::BufReader::new(&mut parts_fd);
+
+            // deserialize request parts from rust messagepack
+            let wrapper: shuttle_next::RequestWrapper = shuttle_next::from_read(reader).unwrap();
+
+            // file descriptor 4 for reading http body into wasm
+            let mut body_read_stream = unsafe { std::fs::File::from_raw_fd(body_read_fd) };
```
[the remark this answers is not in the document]
oddgrd: It was because streaming the bytes from the inner read_stream to the axum router consumed the stream, and so couldn't be used to write after. That didn't actually end up working, so it should be possible to use only one stream for the body now. On the runtime side I did also change it to drop the body write stream after it completes to signal EOF, but if we're reusing it we'd need to change it back to writing a `0`. I can look into this.

deployer/tests/resources/bind-panic/Cargo.toml
```diff
@@ -11,4 +11,4 @@ crate-type = ["cdylib"]
 [workspace]
 
 [dependencies]
-shuttle-service = { path = "../../../../service" }
+shuttle-service = { path = "../../../../service", features = ["codegen"] }
```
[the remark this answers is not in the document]
oddgrd: Yup

runtime/Makefile
```diff
@@ -0,0 +1,13 @@
+.PHONY: axum
+
+all: axum
+
+axum:
```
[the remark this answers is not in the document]
oddgrd: We can compile the module in a test helper I reckon.

runtime/src/axum/mod.rs
oddgrd: I might be mistaken but I think there is some consensus on `unwrap` of `.lock` being okay, since it will only panic if something else panicked. We could also expect I suppose.

runtime/src/axum/mod.rs
oddgrd: Good question :smile:

Cargo.toml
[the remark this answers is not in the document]
chesedo: Yip, I agree. This will work perfectly

codegen/src/next/mod.rs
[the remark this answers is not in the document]
chesedo: There is also some method call that allows you to split the stream into a reader and writer. You can then safely drop the writer without needing to write the `0`

common/src/wasm.rs
```diff
@@ -0,0 +1,521 @@
...
+use rmps::Serializer;
+use serde::{Deserialize, Serialize};
+use tracing::Subscriber;
+use tracing_subscriber::Layer;
+
+use crate::tracing::JsonVisitor;
+
+extern crate rmp_serde as rmps;
+
+// todo: add http extensions field
```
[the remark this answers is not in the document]
chesedo: I don't know if support for extensions makes sense even. Maybe we can add it once there is a feature request for it?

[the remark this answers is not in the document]
oddgrd: Yeah, I agree.

Cargo.toml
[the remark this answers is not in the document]
oddgrd: Done in #586

codegen/src/next/mod.rs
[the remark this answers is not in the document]
oddgrd: Thanks for the tip! I found a way in #586

codegen/src/next/mod.rs
```diff
@@ -0,0 +1,527 @@
...
+                    if route.is_some() {
+                        emit_error!(
+                            key_ident,
+                            "duplicate endpoint route";
+                            hint = "The endpoint `route` should only be set once."
+                        );
+                        has_err = true;
+                    }
+                    if let Expr::Lit(literal) = value {
+                        if let Some(Lit::Str(literal)) = Some(literal.lit) {
```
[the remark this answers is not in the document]
oddgrd: Done in #588

codegen/src/next/mod.rs
```diff
@@ -0,0 +1,527 @@
...
+        match method.to_string().as_str() {
+            "get" | "post" | "delete" | "put" | "options" | "head" | "trace" | "patch" => {}
+            _ => {
+                emit_error!(
+                    method,
+                    "method is not supported";
+                    hint = "Try one of the following: `get`, `post`, `delete`, `put`, `options`, `head`, `trace` or `patch`"
+                )
+            }
+        };
```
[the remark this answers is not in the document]
oddgrd: I think that makes sense, I implemented it in #588

PR #580, opened 2023-01-16
ci: resolve CI errors in shuttle-next

oddgrd:
Todo: 
- Figure out how to communicate to the deployer that a service has stopped or crashed after startup.
- Run the correct cleanup functions in `run::task`to fix the deploy layer tests.
- Figure out why the sqlx-pool test in service is hanging.


deployer/src/deployment/run.rs
```diff
@@ -292,25 +305,34 @@ async fn run(
...
 
-            break;
+            cleanup(response);
         }
-    }
+        Err(ref status) => {
+            error!("failed to start service, status code: {}", status);
 
-    cleanup(response);
+            cleanup(RuntimeResponse::Start(response));
```
chesedo, line 333: I think this can call `start_crashed_cleanup()` directly. Then you won't have to worry about the enum

oddgrd: My thinking is that tests (like `run::panic_in_main`) rely on the closure being called since they pass a sender into it. But there may very well be a better way to handle this.

runtime/src/legacy/mod.rs
```diff
@@ -255,3 +268,21 @@ async fn load_service(
 
     Ok(loader.load(factory, addr, logger).await?)
 }
+
+#[instrument(skip(_id), fields(id = %_id, state = %State::Stopped))]
```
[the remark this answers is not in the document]
oddgrd: Ah, thanks! I didn't realize that.

deployer/src/deployment/run.rs
```diff
@@ -470,33 +514,36 @@ mod tests {
             _ = sleep(Duration::from_secs(5)) => panic!("cleanup should have been called as service stopped on its own"),
             Ok(()) = cleanup_recv => {},
         }
+
+        // Prevent the runtime manager from dropping earlier, which will kill the processes it manages
+        drop(runtime_manager);
```
oddgrd, line 519: Clever :brain:

oddgrd (commented):
Nice, thanks for tackling this beast Pieter! The changes in #696 might make it easier to test this too, so I'm happy to merge this when you're ready, and then we can merge it into that. :+1: CI is complaining about the service integrations, I don't know why it's suddenly upset after all this time, but luckily they are going away in #702

deployer/src/deployment/run.rs
chesedo: Took me a while to figure why the test kept failing :sweat_smile:

oddgrd: I can imagine! :smile:

PR #583, opened 2023-01-17
feat: build our images with the release profile

oddgrd:
Currently our containerfile builds crates with the default debug profile, which means when `get_runtime_executable` is called it will try to install from a checked out local version of `shuttle-runtime`. This will work when making images, but fail when the deployer containers start a runtime, since the local version wont be available (even though for deployers the runtime will be installed during make images).

Changing to build the crates will release profile will also not work currently, since it looks for the runtime on the production branch.

Tldr: using debug assertions might not be enough for this, I will investigate further on wednesday.


oddgrd:
Yes, I agree, I'll implement that.

proto/src/lib.rs
```diff
@@ -287,36 +287,50 @@ pub mod runtime {
...
-                .arg("shuttle-runtime")
-                .arg("--git")
-                .arg("https://github.com/shuttle-hq/shuttle")
-                .arg("--branch")
-                .arg("production")
-                .output()
-                .expect("failed to install the shuttle runtime");
+        let current_crate = env!("CARGO_PKG_NAME");
+
+        if current_crate == "cargo-shuttle" {
```
chesedo, line 302: Shouldn't it rather check if the runtime is already installed and just skip if it is?

chesedo (approved):
LGTM if you are happy with the comment

Containerfile
```diff
@@ -25,7 +25,9 @@ COPY --from=planner /build/recipe.json recipe.json
 RUN cargo chef cook --recipe-path recipe.json
```
brokad, line 25: Don't you want to also specify a target here since this builds dependencies?

proto/src/lib.rs
oddgrd: We'll just have to make sure to install the shuttle-runtime then when making changes to it in local dev, but that may be better than it always using the latest by force anyway. I'll implement this.

chesedo: Hmm, maybe we should then only check when in the `release` profile :man_shrugging:

Containerfile
oddgrd: Wont that mean `shuttle-${folder}` will be compiled to wasm32-wasi and put in `target/wasm32-wasi`, which we don't actually want? Only the project, like the axum-wasm example, should be compiled to wasm32-wasi, and all the wasi stuff will be in there hidden by the codegen. But the deployer needs the target added so it can compile wasm32-wasi.

proto/src/lib.rs
oddgrd: Yes, this will also prevent the deployer from updating the runtime unexpectedly. To update it now with the check, the user will need to `rm` and `new` their project, so it will use the new image.

oddgrd: Perhaps what we should do is check that it is the same version as cargo-shuttle, and if it isn't then try to install? Because currently it will always try to install when in release mode, which may cause runtime to be updated before cargo-shuttle. If we change it to not install if it already is installed, the user will have to delete shuttle-runtime to update it.

Containerfile
brokad: Sorry my bad, I meant profile (`--release`), not target.

oddgrd: Ah, of course, that makes sense. Yes, we should do that, I'll add it there too.

PR #588, opened 2023-01-18
Feature/eng 378 axum wasm multiple handlers per endpoint

oddgrd:


codegen/src/next/mod.rs
```diff
@@ -339,46 +416,89 @@ mod tests {
         };
 
         let actual = quote!(#endpoint);
-        let expected = quote!(.route("/hello", shuttle_next::routing::get(hello)));
```
chesedo, line 342: Won't this error on `get` being unknown?

codegen/src/next/mod.rs
```diff
@@ -339,46 +416,89 @@ mod tests {
...
+                        request: shuttle_next::Request<shuttle_next::body::BoxBody>,
+                    ) -> shuttle_next::response::Response {
+                        use shuttle_next::routing::*;
+                        use shuttle_next::Service;
 
-                response
-            }
-        );
+                        let mut router = shuttle_next::Router::new()
+                            .route("/goodbye", get(get_goodbye).post(post_goodbye))
```
chesedo, line 487: Nice :partying_face:

codegen/src/next/mod.rs
```diff
@@ -237,13 +279,46 @@ impl ToTokens for App {
...
+            })
+            .into_iter()
+            .map(|(key, value)| EndpointChain {
+                route: key,
+                handlers: value,
+            })
+            .collect::<Vec<EndpointChain>>();
+
+        // Sort endpoints by route alphabetically to ensure test consistency
+        endpoint_chains.sort_by(|a, b| a.route.value().cmp(&b.route.value()));
```
chesedo, line 312: Swapping the `HashMap` for a `BTreeMap` should remove the need for this sort

codegen/src/next/mod.rs
```diff
@@ -124,7 +126,23 @@ impl Endpoint {
...
                     if let Expr::Path(path) = value {
-                        method = Some(path.path.segments[0].ident.clone());
+                        let method_ident = Some(path.path.segments[0].ident.clone());
+
+                        // Unwrap is safe because we just set the value
+                        match method_ident.as_ref().unwrap().to_string().as_str() {
+                            "get" | "post" | "delete" | "put" | "options" | "head" | "trace"
+                            | "patch" => {
+                                method = method_ident;
+                            }
```
chesedo, line 136:
```suggestion
                        let method_ident = path.path.segments[0].ident.clone();

                        match method_ident.as_ref().to_string().as_str() {
                            "get" | "post" | "delete" | "put" | "options" | "head" | "trace"
                            | "patch" => {
                                method = Some(method_ident);
                            }
```
So that we don't wrap to unwrap again - ie moving the wrap down

chesedo (commented):
Looking good @oddgrd!

codegen/src/next/mod.rs
oddgrd: I tried this, but unfortunately `LitStr` doesn't implement `Ord`. On second thought I could implement `Ord` for it, and compare the `.value()` in the `cmp` methods.

codegen/src/next/mod.rs
oddgrd: The reason I removed the full paths is it was causing a bug in chaining the handlers when running the example. In the tests it worked and gave the correct output: 
```rust
let mut router = shuttle_next::Router::new()
        .route("/goodbye", shuttle_next::routing::get(get_goodbye).shuttle_next::routing::post(post_goodbye))
        .route("/hello", shuttle_next::routing::get(hello));
```
But when compiling the example it generated this:
```rust
let mut router = shuttle_next::Router::new()
        .route(
            "/goodbye",
            shuttle_next::routing::get(get_goodbye).shuttle_next,
            ::routing::post(post_goodbye),
        )
        .route("/hello", shuttle_next::routing::get(hello))
```

When removing the full path it works, and I make sure it's imported by adding it to the `app` in the `App` `ToTokens` impl with  `use shuttle_next::routing::*`. I am not very happy with that solution though, and the bug doesn't make sense to me, so I'm sure I've done something wrong here.

chesedo: Ah, I see. I think the use will be fine :smile: 

Fyi, this will be the line that you will need to edit correctly to not have the `use`
https://github.com/shuttle-hq/shuttle/pull/588/files#diff-ef21a60b2fde7fdab29d9510e71bb56dc2c9f4e85f06bf0a4b38a8c1ceef00e1R247

codegen/src/next/mod.rs
chesedo: Okay, this is simpler then :smile:

codegen/src/next/mod.rs
oddgrd: Yeah, I tried it like this: 
```rust
let route = quote!(.route(#route, #(shuttle_next::routing::#handlers).*));
```
I still get the same bug, but I'm pretty sure this is correct

chesedo: It's almost correct - you want the namespace outside the repeat:

```rust
let route = quote!(.route(#route, shuttle_next::routing::#(#handlers).*));
```

oddgrd: Ah, of course :facepalm: Thanks, that works!

PR #591, opened 2023-01-24
feat: initial commit of elasticache integration

oddgrd:


iulianbarbu:
@oddgrd , I am curious if we should keep this open as a draft for now, as opposed to closing it and focusing on it when we'll prioritize it's implementation. Let me know what you think.

PR #592, opened 2023-01-25
added feedback

### LINKED ISSUE #589: Add feedback command
We should add a command to the cargo-shuttle CLI that opens the "create new issue" page on the shuttle github. It should be possible to implement this with the [webbrowser](https://crates.io/crates/webbrowser) crate, which we already use for the `login` command.

gautamprikshit1:
Fixed #589


cargo-shuttle/src/lib.rs
```diff
@@ -225,6 +226,15 @@ impl Shuttle {
         self.ctx.load_local(project_args)
     }
 
+    /// Provide feedback on GitHub.
+    async fn feedback(&mut self) -> Result<()> {
```
oddgrd, line 230:
```suggestion
    async fn feedback(&self) -> Result<()> {
```

oddgrd (approved):
This looks great, thanks @gautamprikshit1!

PR #593, opened 2023-01-25
refactor: hide some pg catalogs

### LINKED ISSUE #557: Other DB names not from my project visible in pgAdmin
1. Create a project that uses `shuttle_shared_db::Postgres` and deploy it.
2. You get back a string like this after deployment: `postgres://user-punchcards-server:<redacted>@db.shuttle.rs:5432/db-punchcards-server`. 
3. Use that string which includes a username, password, hostname, and db name to connect to the server in pgAdmin.
4. View the databases drop-down. It lists all databases under that host name, including those from other users. I should not be able to view the names of the databases from other users sharing the same DB server.

chesedo:
Closes #557

However, this breaks some PG clients like PgAdmin. Other clients like beekeeper-studio still work though. For anyone that needs fuller PG support, there is still the RDS option that will work with all clients.

### DIFF
--- a/provisioner/src/lib.rs
+++ b/provisioner/src/lib.rs
@@ -11,7 +11,7 @@ pub use shuttle_proto::provisioner::provisioner_server::ProvisionerServer;
 use shuttle_proto::provisioner::{
     aws_rds, database_request::DbType, shared, AwsRds, DatabaseRequest, DatabaseResponse, Shared,
 };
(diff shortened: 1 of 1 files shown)


oddgrd (approved):
Nice! I know you had some more statements in your write up for this, in the end were only these three necessary?

chesedo:
Yip, I want to be as conservative as possible. If there are more things we want to block in the future then we can just add them to the vec :smile:

Issue #594, opened 2023-01-25
Add logout command


joshua-mo-143:
The CLI doesn't currently offer (as far as I can see) a way to be able to log out of the platform. 

I'm currently working on a way to allow users to log out. This function at the moment would be arbitrary more than anything, but it's always useful to have as users may be required to log out because of unforeseen circumstances (security reasons, new computer... etc) and mostly for completeness since if users are able to log in, there should be a way to log out.

joshua-mo-143:
Closing this issue because the PR that mentions this issue has been merged to main.

PR #595, opened 2023-01-25
feat: add logout function

joshua-mo-143:
* add logout function for CLI

Reference: issue #594 

Current queries:
- api_key is currently defined as a String type, which means I've had to set the API key in the logout function as an empty string. This technically works, but I'm wondering if it's more correct to set it as None?

### DIFF
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -87,6 +87,8 @@ pub enum Command {
     Secrets,
     /// login to the shuttle platform
     Login(LoginArgs),
+    /// log out of the shuttle platform
+    Logout,
     /// run a shuttle service locally
     Run(RunArgs),
     /// manage a project on shuttle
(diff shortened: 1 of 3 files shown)


oddgrd:
Nice! While it does work, I think your hunch is correct, it would be better to not just set it to an empty string. I'm thinking we could either change the `set_api_key` function to take an `Option<ApiKey>`, or we could add a new method, something like `clear_api_key`. I'm not quite sure what I would prefer :smile: Maybe you can try it out and see what feels best?

joshua-mo-143:
> Nice! While it does work, I think your hunch is correct, it would be better to not just set it to an empty string. I'm thinking we could either change the `set_api_key` function to take an `Option<ApiKey>`, or we could add a new method, something like `clear_api_key`. I'm not quite sure what I would prefer 😄 Maybe you can try it out and see what feels best?

I decided to add a `clear_api_key` function in the end since I think that makes more sense for the purpose of modularity, in case we need to use it anywhere else.

oddgrd (approved):
LGTM, thanks!

PR #596, opened 2023-01-25
feat: local uri

chesedo:
Also need this for:
- [x] RDS
- [x] Shared mongo

PR #597, opened 2023-01-26
feat: allow string interpolation on resource options

### LINKED ISSUE #497: Override provisioned resources on local runs
Running `shuttle run` will start a local postgres container if the service's main has a database parameter.

We'd like a way to override this behavior to let the service connect to an existing (user managed) postgres instance.

I could think of two ways of exposing the choice to the user:

- Adding a parameter annotation specifying the postgres URI, e.g.

```rust
#[shuttle_service::main]
async fn main(#[local(uri = "...")] pg: PgPool) {
  ...
}
```

- Adding an entry in `Shuttle.toml` specifying local deployment configuration, e.g.

```toml
[local]
postgres = "..."
```

I am not sure which is best. But going with option 2 may also help with #350.

chesedo:
This PR allows options passed to resources to be string interpolated. For example in this `local_uri` option, the password is being read from the `Secrets.toml` file

```rust
#[shuttle_service::main]
async fn main(
    #[shuttle_shared_db::Postgres(
        local_uri = "postgres://postgres:{secrets.password}@localhost:16695/postgres"
    )]
    pool: PgPool,
) -> shuttle_service::ShuttlePoem<impl poem::Endpoint> {
    ...
}
```

With #596, this closes #497

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6015,6 +6015,7 @@ dependencies = [
  "shuttle-codegen",
  "shuttle-common",
  "sqlx",
+ "strfmt",
  "sync_wrapper",
  "thiserror",
  "thruster",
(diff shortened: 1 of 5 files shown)


oddgrd (approved):
This is really cool! :star_struck: This may be obvious, but what will the error be if it tries to interpolate a secret and it doesn't exist? It looks like it will always be interpolation error, should there also be a not found error if we check and it doesn't exist? Or we could just add a bit about making sure the secrets is added to Secrets.toml.

chesedo:
@oddgrd This is the error message:
```
Error: Failed to interpolate string: KeyError(Invalid key: secrets.passwor)
```

chesedo:
Now the error is :smile: 

```
Error: Failed to interpolate string. Is your Secrets.toml correct?

Caused by:
    KeyError(Invalid key: secrets.Password)
```

Issue #564, opened 2023-01-03
Add workspace support


chesedo:
When I deploy a crate that is within a workspace and that has a `{ workspace = true }` dependency, then I get an error that the manifest cannot be parsed.

The reason for this is because shuttle did not archive the entire workspace, but only the single crate.

gautamprikshit1:
Any help on how to achieve this

chesedo:
I can think of the following changes that will be needed (each in a separate PR):
1. Our way of finding the `Cargo.toml` file is wrong since it currently stops at the first one found. Rather it should do whatever `cargo` is doing to continue looking in the parent tree for another `Cargo.toml` that might be the workspace one. When shuttle collects the correct `Cargo.toml` (for the workspace) then I believe everything is already in place for shuttle to correctly archive the entire workspace. Identifying the correct `Cargo.toml` is a change that needs to happen in `cargo-shuttle` somewhere.
2. We have a `loader.rs` file (in service) which builds an archive, but then only gets the first `cdylib` from the build. Since the archive is now the entire workspace there might be multiple `cdylib`s so this build method needs to be updated to return a `Vec<_>`. The caller might then also need a small update, but I believe shuttle has everything else in place to support running multiple services / crates in a workspace.

brokad:
Just to confirm here @chesedo, I suppose this also requires a change in deployer somewhere since a workspace could contain multiple services, each of which needs recognition in the deployer's state?

chesedo:
@brokad the deployer state can already start multiple services. For services that depend on one another, starting them in the correct order is just missing.

chesedo:
Change number 2 for the above comment is implemented in #767

PR #599, opened 2023-01-27
workspace-support

gautamprikshit1:
Made an attempt at #564 please check

### DIFF
--- a/cargo-shuttle/src/init.rs
+++ b/cargo-shuttle/src/init.rs
@@ -3,6 +3,7 @@ use std::io::Write;
 use std::path::{Path, PathBuf};
 
 use anyhow::Result;
+use cargo::core::find_workspace_root;
 use cargo::ops::NewOptions;
 use cargo_edit::{find, get_latest_dependency, registry_url};
 use indoc::indoc;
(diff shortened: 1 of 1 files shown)


cargo-shuttle/src/init.rs
```diff
@@ -669,7 +670,12 @@ pub fn cargo_shuttle_init(path: PathBuf, framework: Framework) -> Result<()> {
     let mut dependencies = Table::new();
 
     // Set "shuttle-service" version to `[dependencies]` table
-    let manifest_path = find(Some(path.as_path())).unwrap();
+    let config = cargo::util::config::Config::default().unwrap();
```
chesedo, line 673: Nice!

I'm just not sure if `cargo_shuttle_init` is the correct function to update though. I think this function needs to be updated or removed (which is better)

https://github.com/shuttle-hq/shuttle/blob/79ff57e1a2e712748803d20e7aa51b1f9dbf72af/cargo-shuttle/src/lib.rs#L210-L214

chesedo (commented):
Wow, that was easy. Thanks @gautamprikshit1 

Just confirm this is updating the correct function. The easiest would be to create a workspace with two crates:
1. A crate with a library function like `get_name()`
2. A crate that uses shuttle and crate 1 to get a name

Then from inside crate 1 run `cargo shuttle run`. If it fails then the wrong function might have been updated here

cargo-shuttle/src/init.rs
gautamprikshit1: Have you run it already or shall I do it??

gautamprikshit1: https://docs.rs/project-root/latest/src/project_root/lib.rs.html#19-34

I think replacing existing function with this one will do.

chesedo: I'll leave the testing to you :smile:

Issue #598, opened 2023-01-27
Switch tempdir to tempfile


chesedo:
[Tempdir is deprecated](https://github.com/shuttle-hq/shuttle/pull/570) and got merged into [tempfile](https://github.com/Stebalien/tempfile). We should update all references to `tempdir` to rather use `tempfile`

Basically search all `Cargo.toml` files for `tempdir`

oddgrd:
Closed by #603

PR #603, opened 2023-01-27
Tempdir

gautamprikshit1:
Worked on #598

### DIFF
--- a/e2e/Cargo.toml
+++ b/e2e/Cargo.toml
@@ -11,4 +11,4 @@ portpicker = "0.1.1"
 rand = "0.8.5"
 reqwest = { version = "0.11.10", features = ["blocking", "json"] }
 which = "4.2.5"
-tempdir = "0.3.7"
+tempfile = "3.3.0"
--- a/e2e/Cargo.toml
+++ b/e2e/Cargo.toml
@@ -11,3 +11,4 @@ portpicker = "0.1.1"
 rand = "0.8.5"
 reqwest = { version = "0.11.10", features = ["blocking", "json"] }
 which = "4.2.5"
+tempfile = "3.3.0"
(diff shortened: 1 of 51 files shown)


chesedo (approved):
Great, thanks for this @gautamprikshit1

chesedo:
Just have to check the failing test @gautamprikshit1

e2e/Cargo.toml
```diff
@@ -11,4 +11,4 @@ portpicker = "0.1.1"
 rand = "0.8.5"
 reqwest = { version = "0.11.10", features = ["blocking", "json"] }
 which = "4.2.5"
-tempdir = "0.3.7"
+tempfile = "3.3.0"
```
chesedo, line 14: Since there are no other updates for `e2e`, does that mean `e2e` does not use `tempdir`?

gautamprikshit1: apologies did that with fuzzy search, didn't check that

chesedo: Okay, if we can remove this line then lets remove it

oddgrd: We do use tempdir in the e2e tests @gautamprikshit1, specifically in `tests/integration/helpers`, but it might not be used when you run locally. So we should also update tempdir to tempfile here. By the way, if you haven't tried it yet, there are instructions for how to run the e2e tests in the e2e directory README.

e2e/tests/integration/helpers/mod.rs
```diff
@@ -11,7 +11,7 @@ use crossterm::style::{Color, Stylize};
 use reqwest::blocking::RequestBuilder;
 
 use lazy_static::lazy_static;
-use tempdir::TempDir;
+use tempfile::{Builder, TempDir};
```
oddgrd, line 14: Is `TempDir` an unused import now?

oddgrd (approved):
LGTM, thanks! Just one final comment.

e2e/tests/integration/helpers/mod.rs
gautamprikshit1: No it's used as a type

oddgrd: Ah, indeed! :smile:

PR #608, opened 2023-01-30
fix: add repo url to cargo-shuttle for binstall

oddgrd:
[`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md#support-for-cargo-binstall) needs the crate to have a repository set in its manifest for it to find the binary download. We should only need to set it for `cargo-shuttle`, but it's nice to have for all our published crates. 

There was also an issue with the windows bin build where the tag was not included in the archive name, due to that build running the `make-artifact` command with powershell. I resolved this by changing the shell for the windows build to bash, and using a powershell override for the other commands.

I also changed the name of the directory inside the archive to be [compatible](https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md#examples) with `cargo-binstall`.

And last but not least, I added the `x86_64-unknown-linux-gnu` target to the linux binary builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -339,7 +339,7 @@ jobs:
           command: |
             # From https://github.com/alexcrichton/openssl-src-rs/issues/45
             # Because of https://github.com/openssl/openssl/issues/9048
-            OPENSSL_SRC_PERL="C:\Strawberry\perl\bin\perl.exe"
+            export OPENSSL_SRC_PERL="C:\Strawberry\perl\bin\perl.exe"
             # "vendored-openssl" is from https://github.com/cross-rs/cross/issues/229#issuecomment-597898074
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -379,7 +386,6 @@ jobs:
             ghr -t ${GITHUB_TOKEN} -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -c ${CIRCLE_SHA1} -delete -draft ${TAG} ./artifacts/
 
 workflows:
-  version: 2
```
oddgrd, line 382: This field is deprecated as of circleci 2.1

chesedo (approved):
LGTM

Issue #563, opened 2023-01-03
Prod and dev secrets


chesedo:
Currently prod and dev use the same `Secrets.toml` file. It would be great if there can be a prod and/or dev overwrite. Related to #497 

See discussion at https://discord.com/channels/803236282088161321/953005742842069013/1058463999718588476

Issue #609, opened 2023-01-31
Add -r / --release flag to `cargo-shuttle run`


jonaro00:
Add an argument to `cargo-shuttle` that sets the flag at
https://github.com/shuttle-hq/shuttle/blob/main/cargo-shuttle/src/lib.rs#L393
to `true` to allow local runs to use release build.

Since `cargo-shuttle run` is basically a wrapper around `cargo run`, there could also be other flags that are of interest to implement. But since the release mode flag is already supported by the loader, it should be an easy addition.

oddgrd:
Good idea, and it should indeed be straightforward to implement. Thanks!

PR #610, opened 2023-02-02
feat: add dev/prod secrets functionality

joshua-mo-143:
* Local deployments now use `Secrets.dev.toml` by default

* Added a flag to `run_args` that allows usage of prod secrets in a local deployment should the user need to do so

Ref: Issue #563 

Notes:
I was thinking about using `Secrets.prod.toml` as a filename for live deployment but it will probably break any and all deployments currently using `Secrets.toml`, as well as there being an example project in testing that currently uses `Secrets.toml`, so until there's further consideration about how to deal with live deployments that currently use secrets it's probably a good idea to leave this unchanged until a discussion on how to solve this has been made.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,26 +119,31 @@ commands:
         default: ""
     steps:
       - run:
-          name: Set git tag in the environment"
+          name: Set git tag in the environment
           command: |
             echo TAG=$(git describe --tags) >> $BASH_ENV
       - run:
+          name: Set binary directory in the environment
+          command: |
+            echo BIN_DIR=cargo-shuttle-<< parameters.target >>-$TAG >> $BASH_ENV
+      - run:
           name: Make artifact
           command: |
-            mkdir shuttle
-            mv target/<< parameters.target >>/release/cargo-shuttle<< parameters.suffix >> shuttle/cargo-shuttle<< parameters.suffix >>
-            mv LICENSE shuttle/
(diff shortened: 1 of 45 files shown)


jonaro00:
I suggest that the flag should instead be called -r / --release (and therefore at the same time implement #609), so that a local release build will be run when using local production secrets. (This makes the most sense at least, but perhaps the prod_secrets flag could be separate from --release, and implied when using --release)

joshua-mo-143:
> I suggest that the flag should instead be called -r / --release (and therefore at the same time implement #609), so that a local release build will be run when using local production secrets. (This makes the most sense at least, but perhaps the prod_secrets flag could be separate from --release, and implied when using --release)

I think `prod_secrets` should probably be kept separate from `--release` as in my mind they're two entirely separate things and I'd personally like to keep modularity in mind. Maybe we could create an entirely new flag called `simulate_prod` that combines both of these flags together?

oddgrd:
We'll go with just defaulting to the `Secrets.dev.toml` file if it exists for now, without the CLI arg to enable prod secrets. If people want to use prod secrets locally it won't be difficult anyway. We can look at this again down the line as we keep improving the Secrets feature.

joshua-mo-143:
Ok no worries, that's fine then - we can proceed as is

oddgrd (approved):
Great, thanks!

PR #611, opened 2023-02-03
Release flag

### LINKED ISSUE #609: Add -r / --release flag to `cargo-shuttle run`

gautamprikshit1:
Fixes #609

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -386,7 +386,7 @@ workflows:
              - workspace-fmt
            matrix:
              parameters:
-               framework: ["web-actix-web", "web-axum", "web-rocket", "web-poem", "web-thruster", "web-tide", "web-tower","web-warp", "web-salvo", "bot-serenity"]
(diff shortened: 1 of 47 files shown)


oddgrd:
This looks good @gautamprikshit1, but do you think you could clean up those extra commits? Generally when you start a new PR, you should make sure `main` [is updated](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork#syncing-a-fork-branch-from-the-command-line), then you create your new branch from the updated main branch. Here it looks like you started a new PR from a previous PR branch.

oddgrd (approved):
Thanks!

PR #616, opened 2023-02-06
fix: delete the deployments before deleting a service

### LINKED ISSUE #561: Error: foreign key constraint failed when running `cargo shuttle delete`
I'm getting the following error when trying to run `cargo shuttle delete`:

```
❯ cargo shuttle delete
message: Persistence failure: Database error: error returned from database: (code: 787) FOREIGN KEY constraint failed
```

I haven't managed to successfully deploy anything yet (due to #527, I think?) so this might be an invalid command, but the error message could be nicer 🙂

jdrouet:
Right now, the database doesn't have a defined strategy when a service gets deleted in the database and therefore, where the is a foreign key, the request fails.
This change makes sure to remove the deployments from a service after killing them allowing the service to be deleted properly.

For later: the `Persistence` struct should be updated to be able to start a transaction so that we can rollback if something goes wrong.

This should fix #561

Signed-off-by: Jérémie Drouet <jeremie.drouet@gmail.com>

### DIFF
--- a/deployer/src/handlers/mod.rs
+++ b/deployer/src/handlers/mod.rs
@@ -277,6 +277,10 @@ async fn delete_service(
             .map(Into::into)
             .collect();
 
+        // TODO wrap this in a transaction
+        persistence
+            .delete_deployments_by_service_id(&service.id)
+            .await?;
         persistence.delete_service(&service.id).await?;
 
         let response = shuttle_common::models::service::Detailed {


oddgrd:
Hey @jdrouet! Apologies for the quite unclear issue description :sweat_smile:. If you'll see my latest comment in the issue, we don't really want to delete anything from the db here. We should rather refactor the `cargo shuttle delete` to `cargo shuttle stop`, and simply stop the service via the deployment manager. There are some issues with the actual stopping of the service under the hood, but that's a separate issue.

If I'm not mistaken, stopping the service will also change the state in persistence to `Stopped` via the cleanup closure in `deployer/deployment/run.rs`, but I'd have to double-check that.

jdrouet:
Ok, so I'll close this PR and open a new one, thanks 😉

PR #619, opened 2023-02-08
feat: rename delete to stop

### LINKED ISSUE #561: Error: foreign key constraint failed when running `cargo shuttle delete`

jdrouet:
According to [this comment](https://github.com/shuttle-hq/shuttle/pull/616#issuecomment-1421255649), it's not necessary to remove the service from the database and we only need to stop it.
So, in the deployer, we can rename the `delete_service` function to `stop_service` and just remove the persistence line for dropping the service in database.

Considering this endpoint no longer deletes the service but just stops it, we can rename the delete command to stop in cargo-shuttle.

This should fix #561


oddgrd:
Thanks @jdrouet! This looks like it's nearly there, but won't we still get an FK error from calling `delete_deployments_by_service_id`?

jdrouet:
@oddgrd good point, thank you for noticing it 😉

jdrouet:
> LGTM, thanks! One last thing: I think we can safely remove the delete_deployments_by_service_id function, I don't think it's used anywhere else.

Sure, I'm used to have a clippy job that would break on the CI for this kind of issues so I tend to forget to check 😅

Issue #615, opened 2023-02-06
Allow for resource deletion


brokad:
Currently there is no way to delete a resource attached to a deployment (such as a database that was provisioned from an annotation to the `shuttle_service::main` function). We'd like to expose such functionality so that resources attached to services can be removed imperatively from the CLI with a command such as `cargo shuttle resource delete {resource-name}`.

For this we need a change to [provisioner](https://github.com/shuttle-hq/shuttle/tree/main/provisioner/) to write the deletion logic and expose the methjod through gRPC, then a change to [deployer](https://github.com/shuttle-hq/shuttle/tree/main/deployer/) to write the deletion logic (calling provisioner) and checking for the sanity of the operation (is a running service using it?) and expose the API through handlers, and then finally a change to [cargo-shuttle](https://github.com/shuttle-hq/shuttle/tree/main/cargo-shuttle/) which exposes the command to the CLI.

MrCoolTheCucumber:
Hey, I think I can try to tackle this! I have an initial question though:

> resources attached to services can be removed imperatively from the CLI with a command such as `cargo shuttle resource delete {resource-name}`

Looking at the schema in `deployer/migrations/...` it doesn't seem like resources currently have names? Is this something that will need to be added? Or will the resource name be some sort of "composed" name based on other information such as for example `"{service_name}-{resource_type}"`? (Though I guess that would only work if its assumed that only one of each resourse type can exist)

brokad:
> For this we need a change to [provisioner](https://github.com/shuttle-hq/shuttle/tree/main/provisioner/) to write the deletion logic and expose the methjod through gRPC, then a change to [deployer](https://github.com/shuttle-hq/shuttle/tree/main/deployer/) to write the deletion logic (calling provisioner) and checking for the sanity of the operation (is a running service using it?) and expose the API through handlers, and then finally a change to [cargo-shuttle](https://github.com/shuttle-hq/shuttle/tree/main/cargo-shuttle/) which exposes the command to the CLI.

For whoever is stumbling upon this, #622 implemented the required provisioner changes, but left out the needed deployer and client-side changes.

kierendavies:
@MrCoolTheCucumber Is it cool with you if I take the remaining work here? Do you have any info or WIP to share before I do?

MrCoolTheCucumber:
> @MrCoolTheCucumber Is it cool with you if I take the remaining work here? Do you have any info or WIP to share before I do?

[...]

PR #622, opened 2023-02-09
feat: implement rpc method in provisioner that allows for the deletion of resources

MrCoolTheCucumber:
Addresses the first step of https://github.com/shuttle-hq/shuttle/issues/615 by implementing an rpc method in the provisioner that allows for the deletion of resources.


chesedo (approved):
This is looking great, thanks @MrCoolTheCucumber !

PR #623, opened 2023-02-10
Feat: optimize sqlite db in deployer and gateway

oddgrd:
Up until sqlx 0.6.1 sqlite databases were created in [WAL mode](https://www.sqlite.org/wal.html) by default, which should be more performant. Our gateway DB was created before 0.6.1 and WAL mode is a permanent setting, so it's still in WAL mode. Since we're now on 0.6.2 we need to explicitly set it to WAL mode so new DBs are created in it. 

On top of that, using the `NORMAL` [synchronous setting](https://www.sqlite.org/pragma.html#pragma_synchronous) is durable and corruption-proof in WAL mode, and it is more performant then the default `FULL` setting.

Sources: sqlite docs linked above, https://phiresky.github.io/blog/2020/sqlite-performance-tuning/ and https://activesphere.com/blog/2018/12/24/understanding-sqlite-busy

TODO: stress test this and see if it resolves the deadlock.

### DIFF
--- a/deployer/src/persistence/mod.rs
+++ b/deployer/src/persistence/mod.rs
@@ -61,7 +61,13 @@ impl Persistence {
             std::fs::canonicalize(path).unwrap().to_string_lossy()
         );
 
-        let pool = SqlitePool::connect(path).await.unwrap();
+        let sqlite_options = SqliteConnectOptions::from_str(path)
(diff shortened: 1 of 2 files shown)


chesedo (approved):
LGTM :heart:

PR #628, opened 2023-02-13
feat: show progress bar on stopping service

oddgrd:
In the previous implementation stop service would just return a list of the deployments fetched before stopping the service, so it would show as running. This changes it to return a summary of the active deployment and a progress bar until it's stopped (which is almost immediate).

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -4,6 +4,7 @@ pub mod config;
 mod factory;
 mod init;
 
+use indicatif::ProgressBar;
 use shuttle_common::project::ProjectName;
 use std::collections::BTreeMap;
 use std::ffi::OsString;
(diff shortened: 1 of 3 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -267,7 +268,22 @@ impl Shuttle {
...
+            };
+
+            if let shuttle_common::deployment::State::Stopped = deployment.state {
+                break;
+            }
+
+            progress_bar.set_message(format!("Stopping {}", deployment.id));
+            service = client.get_service_summary(self.ctx.project_name()).await?;
+        }
+        progress_bar.finish_and_clear();
```
chesedo, line 286: Any reason for not using the current `wait_with_spinner()` helper?

chesedo (approved):
Seems good to me, LGTM

cargo-shuttle/src/lib.rs
oddgrd: Yes, I tried to do that for a little while, but it ended up getting quite complicated. While it is generic, it's generic for project commands. So making it generic for deployer as well meant adding another generic function arg to fetch the latest state, making it generic over which type of `states_to_check` as well as response type. So we could do that, but this felt a lot simpler. I'm not sure what the best choice is, to be honest, so I'm open to changing it.

chesedo: Let's go with what you currently have. We can always add another generic version for deployer when we need it a second time

PR #632, opened 2023-02-15
[auth] feat: axum with routes

chesedo:


gateway/src/api/latest.rs
```diff
@@ -436,36 +436,16 @@ impl ApiBuilder {
...
-                        debug!(
-                            latency = format_args!("{} ns", latency.as_nanos()),
-                            "finished processing request"
-                        );
-                    },
-                ),
+            TraceLayer::new(|request| {
+                request_span!(
+                    request,
+                    account_name = field::Empty,
```
chesedo, line 442:
```suggestion
                    account.name = field::Empty,
```

common/src/backends/metrics.rs
```diff
@@ -33,3 +41,279 @@ where
...
+            path.as_str()
+        } else {
+            ""
+        };
+
+        tracing::debug_span!(
+            "request",
+            http.uri = %$request.uri(),
+            http.method = %$request.method(),
+            http.status_code = tracing::field::Empty,
```
oddgrd, line 186: Doesn't the comment for this macro indicate we will record the http status?

common/src/backends/metrics.rs
```diff
@@ -33,3 +41,279 @@ where
...
+
+impl TraceLayer<MakeSpanSimple> {
+    /// Switch to the span maker which does not add propagation details from the request headers
+    pub fn without_propagation(self) -> Self {
+        self
+    }
+}
+
+impl TraceLayer<MakeSpanPropagation> {
+    /// Switch to the span maker which adds propagation details from the request headers
```
oddgrd, line 98: Tracing noob question: it means propagation from request headers to response headers, right?

oddgrd (approved):
LGTM! I need to work on it for a bit to get a complete understanding, but I think this is good to go. I added a few questions I had.

The failing test works for me locally, maybe it needs a longer sleep in CI? :shrug:

common/src/backends/metrics.rs
chesedo: Yip, however we'll only know the status once the request is complete. This happens on the [OnResponseStatusCode](https://github.com/shuttle-hq/shuttle/pull/632/files#diff-f4cf651c8b021e33d0f88c58c9b6648ff6db38f596dbb9130eaa5163a8fd4880R159)

common/src/backends/metrics.rs
chesedo: Only from the request headers. If we have an HTTP client that calls another service (like gateway calling the deployers), then we have to manually set these propagation details on the headers (currently)

chesedo: Here's an example of the headers being set
https://github.com/shuttle-hq/shuttle/blob/b3e11b5ec0eb129dfbe06f90b99c785ea5334126/gateway/src/proxy.rs#L157-L159

chesedo: So `MakeSpanPropagation` will set it from the request headers to the tracing context (for the parent span) to be more precise

oddgrd: Thanks!

PR #633, opened 2023-02-15
feat: add a users layer

oddgrd:


auth/migrations/0000_init.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS users (
+  account_name TEXT PRIMARY KEY,
+  secret TEXT UNIQUE,
```
chesedo, line 3: Is `secret` the old `key`?

auth/migrations/0000_init.sql
```diff
@@ -0,0 +1,6 @@
...
+  super_user BOOLEAN DEFAULT FALSE,
+  account_tier TEXT DEFAULT "basic" NOT NULL
```
chesedo, line 5: I'm thinking we can merge these two and only have `account_tier`?

auth/src/api/mod.rs
```diff
@@ -0,0 +1,15 @@
...
+mod handlers;
+
+pub use builder::ApiBuilder;
+
+pub async fn serve(router: Router, address: SocketAddr) {
+    Server::bind(&address)
+        .serve(router.into_make_service())
+        .await
+        .unwrap_or_else(|_| panic!("Failed to bind to address: {}", address));
+}
```
chesedo, line 15: Because of how small this function is, I'm wondering if the refactor into its own file is worth it :man_shrugging:

If we want to reuse it, maybe it should be in common and gateway and deployer made to call it?

auth/src/api/handlers.rs
```diff
@@ -0,0 +1,43 @@
...
+
+#[instrument(skip(user_manager))]
+pub(crate) async fn post_user(
+    State(RouterState { user_manager }): State<RouterState>,
+    Path(account_name): Path<AccountName>,
+) -> Result<Json<User>, Error> {
+    let user = user_manager.create_user(account_name).await?;
+    // TODO: use the user response struct in common, or create a new one
+    Ok(Json(user))
+}
```
chesedo, line 31: We should also add a guard so that only admins can call these two endpoints.

auth/src/user.rs
```diff
@@ -0,0 +1,170 @@
...
+        query("INSERT INTO users (account_name, secret) VALUES (?1, ?2)")
+            .bind(&name)
+            .bind(&secret)
+            .execute(&self.pool)
+            .await?;
+
+        Ok(User::new_with_defaults(name, secret))
+    }
+
+    // TODO: get from token?
```
chesedo, line 35: Nah, the account should be known at this level. Only the handlers should know about tokens.

For the bigger use case of adding a user the details are as follow: some admin account will post the account to add to `/user/:user_name`

chesedo (commented):
This is a very good start. Thanks O!

auth/migrations/0000_init.sql
oddgrd: I was a bit confused about if this would be the same as gateway and I was thinking we would use an encryption secret here, but the issue did say it would be the same :smile: I'll change it to be the same for now.

auth/migrations/0000_init.sql
oddgrd: Sure, that makes sense.

auth/src/api/mod.rs
oddgrd: Yeah, I agree, I'll look into refactoring this.

auth/src/user.rs
oddgrd: Cool, thanks!

auth/src/api/mod.rs
oddgrd: I made it a method on ApiBuilder for now, like is done in gateway.

auth/src/api/handlers.rs
```diff
@@ -0,0 +1,45 @@
...
+};
+
+#[instrument(skip(user_manager))]
+pub(crate) async fn get_user(
+    _: Admin,
+    State(RouterState { user_manager }): State<RouterState>,
+    Path(account_name): Path<AccountName>,
+) -> Result<Json<User>, Error> {
+    let user = user_manager.get_user(account_name).await?;
+    // TODO: use the user response struct in common, or create a new one
```
oddgrd, line 20: Not quite sure what the desired response is here, should we leave it as User for now?

auth/src/main.rs
```diff
@@ -30,19 +31,20 @@ async fn main() {
...
-    //         .to_string_lossy()
-    // );
+    let db_uri = db_path.to_str().unwrap();
 
-    start(args).await;
+    // If the start command is called, start the auth server with given DB path and address.
+    // If the init command is called, do the same but insert an admin user as well.
+    match args.command {
+        Commands::Start => start(db_uri, args.address, None).await,
+        Commands::Init(init_args) => start(db_uri, args.address, Some(init_args)).await,
```
chesedo, line 48: Maybe init should only edit the DB and exit. And not start up the whole API?

auth/src/user.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    pub fn new_with_defaults(name: AccountName, key: Key) -> Self {
+        Self {
+            name,
+            key,
+            permissions: Permissions::default(),
+        }
+    }
+
+    async fn retrieve_from_key(user_manager: &UserManager, key: Key) -> Result<User, Error> {
```
chesedo, line 103: We'll have to invert this pattern so that nothing relies on `UserManager`. Rather everything should use `UserManagement`

auth/src/api/builder.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    pub async fn serve(self, address: SocketAddr) {
+        let router = self.into_router();
+
+        Server::bind(&address)
+            .serve(router.into_make_service())
+            .await
+            .unwrap_or_else(|_| panic!("Failed to bind to address: {}", address));
+    }
```
chesedo, line 121: If we keep these two separate then maybe the test can call router (from `into_router`) directly and therefore don't need to start up on a port? You can then also reach into the router state to set an admin user

auth/src/user.rs
```diff
@@ -0,0 +1,287 @@
...
+use rand::distributions::{Alphanumeric, DistString};
+use serde::{Deserialize, Deserializer, Serialize};
+use sqlx::{query, Pool, Row, Sqlite};
+use tracing::{trace, Span};
+
+use crate::{api::RouterState, error::Error};
+
+#[async_trait]
+pub(crate) trait UserManagement {
+    async fn create_user(&self, name: AccountName) -> Result<User, Error>;
```
chesedo, line 19: Maybe we should add the role to this interface?

auth/tests/api/users.rs
```diff
@@ -0,0 +1,96 @@
...
+async fn post_user() {
+    let app = spawn_app().await;
+
+    // POST user without bearer token.
+    let response = app
+        .api_client
+        .post(&format!("{}/user/{}", app.address, "test-user"))
+        .send()
+        .await
+        .expect("Failed to execute request.");
```
chesedo, line 14: Love this one :heart_eyes:

auth/tests/api/users.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    assert_eq!(response.status().as_u16(), 400);
```
chesedo, line 16: You should be able to compare against `StatusCode::FORBIDDEN`(?) which will be easier that the code

chesedo (commented):
It seems like this is almost done

auth/src/api/handlers.rs
chesedo: We might want to interact with this in the future from `shuttle-admin`, so let's pull the response from common

auth/src/api/builder.rs
oddgrd: Good idea, that is a lot simpler, thanks!

auth/tests/api/users.rs
oddgrd: Indeed!

auth/src/main.rs
oddgrd: Yup, with your router suggestion for tests this was doable. Thanks!

auth/src/user.rs
oddgrd: I ended up removing this method since it was redundant. Does that resolve this, or was there more to it?

chesedo: Nope, making this method redundant was the idea :grin:

auth/src/user.rs
oddgrd: I did it with path params, a bit unsure if it should have been with json instead..

auth/src/api/builder.rs
```diff
@@ -0,0 +1,79 @@
...
+            .route("/auth/session", post(convert_cookie))
+            .route("/auth/key", post(convert_key))
+            .route("/auth/refresh", post(refresh_token))
+            .route("/public-key", get(get_public_key))
+            .route("/user/:account_name", get(get_user))
+            .route("/user/:account_name/:account_tier", post(post_user))
+            .route_layer(from_extractor::<Metrics>())
+            .layer(
+                TraceLayer::new(|request| {
+                    request_span!(request, request.params.account_name = field::Empty)
```
chesedo, line 51: We should add an `, account.name = field::Empty` for the guard to record correctly.

auth/src/user.rs
```diff
@@ -0,0 +1,258 @@
...
+#[derive(Clone, Debug, sqlx::Type, PartialEq, Hash, Eq, Serialize, Deserialize)]
+#[serde(transparent)]
+#[sqlx(transparent)]
+pub struct Key(String);
+
+// impl Key {
+//     pub fn as_str(&self) -> &str {
+//         &self.0
+//     }
+// }
```
chesedo, line 140: Comment :smile:

auth/src/user.rs
```diff
@@ -0,0 +1,258 @@
...
+impl Display for AccountTier {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            AccountTier::Basic => write!(f, "basic"),
+            AccountTier::Pro => write!(f, "pro"),
+            AccountTier::Team => write!(f, "team"),
+            AccountTier::Admin => write!(f, "admin"),
+        }
+    }
+}
```
chesedo, line 206: You might want to replace this with the `strum` crate

chesedo (approved):
LGTM

auth/src/user.rs
chesedo: We could also have gone with Post (form) data. But this will be fine for now, thanks

auth/src/api/builder.rs
oddgrd: I assume you meant add `account_tier`, which I did :smile:

Issue #634, opened 2023-02-15
Recreating a project with a custom domain makes it lose it


brokad:
Currently projects that have a registered custom domain with a TLS termination at the `gateway` service lose their custom domain status when you run `cargo shuttle project rm && cargo shuttle project new`. We'd like the custom domain status to remain even after such operations.

PR #637, opened 2023-02-16
(#634) restore custom domain when recreating project

angelorendina:
feat: restore previous custom domain (if any) when recreating a project (#634)

### DIFF
--- a/gateway/src/project.rs
+++ b/gateway/src/project.rs
@@ -201,10 +201,6 @@ impl Project {
         }
     }
 
-    pub fn create(project_name: ProjectName) -> Self {
-        Self::Creating(ProjectCreating::new_with_random_initial_key(project_name))
-    }
-
     pub fn destroy(self) -> Result<Self, Error> {
         if let Some(container) = self.container() {
             Ok(Self::Destroying(ProjectDestroying { container }))
@@ -613,6 +609,11 @@ impl ProjectCreating {
 
     pub fn initial_key(&self) -> &str {
         &self.initial_key
+    }
+
+    #[cfg(test)]
+    pub fn fqdn(&self) -> &Option<String> {
+        &self.fqdn
     }
 
     fn container_name<C: DockerContext>(&self, ctx: &C) -> String {
--- a/gateway/src/project.rs
+++ b/gateway/src/project.rs
@@ -611,7 +611,6 @@ impl ProjectCreating {
         &self.initial_key
     }
 
-    #[cfg(test)]
     pub fn fqdn(&self) -> &Option<String> {
         &self.fqdn
     }
--- a/gateway/src/service.rs
+++ b/gateway/src/service.rs
@@ -25,7 +25,7 @@ use tracing_opentelemetry::OpenTelemetrySpanExt;
 use crate::acme::CustomDomain;
 use crate::args::ContextArgs;
 use crate::auth::{Key, Permissions, ScopedUser, User};
-use crate::project::Project;
+use crate::project::{Project, ProjectCreating};
 use crate::task::{BoxedTask, TaskBuilder};
 use crate::worker::TaskRouter;
 use crate::{AccountName, DockerContext, Error, ErrorKind, ProjectDetails, ProjectName};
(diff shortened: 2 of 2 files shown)


gateway/src/service.rs
```diff
@@ -515,6 +526,26 @@ impl GatewayService {
             .map_err(|_| Error::from_kind(ErrorKind::Internal))
     }
 
+    pub async fn find_custom_domain_for_project(
```
angelorendina, line 529: Utility method for `project->domain` lookup. Inverse of the method below which performs `domain->project`.

gateway/src/service.rs
```diff
@@ -428,7 +428,18 @@ impl GatewayService {
             let project = row.get::<SqlxJson<Project>, _>("project_state").0;
             if project.is_destroyed() {
                 // But is in `::Destroyed` state, recreate it
-                let project = Project::create(project_name.clone());
```
angelorendina, line 431: `Project::create` does not allow to customise the `ProjectCreating` payload, so manually creating the latter using the existing builder pattern. Perhaps there could be an ad-hoc builder for Project itself.

gateway/src/service.rs
```diff
@@ -835,4 +866,51 @@ pub mod tests {
...
+        while let TaskResult::Pending(_) = work.poll(()).await {}
+        assert!(matches!(work.poll(()).await, TaskResult::Done(())));
+
+        let recreated_project = svc
+            .create_project(project_name.clone(), account.clone())
+            .await
+            .unwrap();
+
+        assert!(matches!(recreated_project, Project::Creating(_)));
+        // TODO: check `recreated_project->fqdn == domain`
```
angelorendina, line 912: Possibly related to the issue above on how to instantiate a project. There is no way to check that the `fqdn` field in ProjectCreating is as expected since it's private. Could add a `#[cfg(test)]` getter but I haven't seen it done elsewhere in the project,

chesedo: Making the field `pub(crate)` will also be acceptable

angelorendina: I ended up making it a `cfg(test)` getter, I reckon it's more suggestive of its expected usage (and also ends up behaving as `pub(crate)` anyway).

gateway/src/project.rs
```diff
@@ -201,10 +201,6 @@ impl Project {
         }
     }
 
-    pub fn create(project_name: ProjectName) -> Self {
-        Self::Creating(ProjectCreating::new_with_random_initial_key(project_name))
-    }
-
```
angelorendina, line 207: (3rd commit proposal, can be removed if not liked)
Removing this public method as there is already a builder pattern for the inner `ProjectCreating`.
Alternatively one could change this method to accept a `fqdn: Option<String>` as well, to reflect the current usage of the method - I prefer the former though.

angelorendina (commented):
Rebased and pushed more commits, also waiting to see if the CI is happy this time so still keeping this a draft

oddgrd (approved):
LGTM, thanks Angelo!

gateway/src/service.rs
oddgrd: That makes sense, it shouldn't be more public than it needs to be. We can always make it pub(crate) in the future if we need it outside of tests. :ok_hand:

angelorendina:
@oddgrd thanks for the review. I have just rebased from latest master and squashed the commits into one, so it should be good to go as per your guidelines 👍

gateway/src/project.rs
```diff
@@ -615,6 +611,11 @@ impl ProjectCreating {
         &self.initial_key
     }
 
+    #[cfg(test)]
+    pub fn fqdn(&self) -> &Option<String> {
+        &self.fqdn
+    }
+
```
chesedo, line 618:
```suggestion
    pub fn fqdn(&self) -> &Option<String> {
        &self.fqdn
    }

```

chesedo (approved):

chesedo:
Thanks for this @angelorendina!

PR #642, opened 2023-02-20
[auth] refactor: update gateway deployer and provisioner

chesedo:


common/src/backends/auth.rs
```diff
@@ -3,17 +3,90 @@ use std::{future::Future, ops::Add, pin::Pin};
...
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Request<Body>) -> Self::Future {
+        let error = match req.headers().typed_try_get::<XShuttleAdminSecret>() {
+            Ok(Some(secret)) if secret.0 == self.secret => None,
+            Ok(_) => Some(StatusCode::UNAUTHORIZED),
+            Err(_) => Some(StatusCode::BAD_REQUEST),
```
oddgrd, line 71: Maybe we should use unauthorized if the header is missing too? I looked for a consensus on this myself when doing the auth layer, and it seems to be the convention. Auth0 says this: https://auth0.com/blog/forbidden-unauthorized-http-status-codes/#When-to-Use-401-Unauthorized

common/src/backends/auth.rs
```diff
@@ -141,6 +236,16 @@ impl Claim {
     }
 }
 
+pub async fn public_key_from_auth(auth_uri: Uri) -> impl Fn() -> Vec<u8> + Clone {
+    let client = Client::new();
+    let uri = format!("{auth_uri}public-key").parse().unwrap();
+    let res = client.get(uri).await.unwrap();
```
oddgrd, line 242: Should we handle this error?

common/src/backends/auth.rs
```diff
@@ -223,6 +328,88 @@ where
...
+        } else {
+            error!("claim extension is not set");
+
+            return Box::pin(async move {
+                Ok(Response::builder()
+                    .status(StatusCode::INTERNAL_SERVER_ERROR)
+                    .body(Default::default())
+                    .unwrap())
+            });
+        };
```
oddgrd, line 392: Can we write this with let else?
```suggestion
        let Some(claim) = req.extensions().get::<Claim>() else {
            error!("claim extension is not set");
            return Box::pin(async move {
                Ok(Response::builder()
                    .status(StatusCode::INTERNAL_SERVER_ERROR)
                    .body(Default::default())
                    .unwrap());
        };
```

deployer/src/handlers/mod.rs
```diff
@@ -27,57 +33,70 @@ pub use {self::error::Error, self::error::Result};
...
         .route("/projects/:project_name/status", get(get_status))
         .route_layer(from_extractor::<Metrics>())
         .layer(
             TraceLayer::new(|request| {
                 let account_name = request
                     .headers()
-                    .get("X-Shuttle-Account-Name")
-                    .map(|value| value.to_str().unwrap_or_default().to_string());
+                    .typed_get::<XShuttleAccountName>()
+                    .unwrap_or_default();
```
oddgrd, line 95: Nice :star_struck:

oddgrd (approved):
This looks great Pieter! :partying_face: I left a few comments, but nothing major.

common/src/backends/auth.rs
chesedo: Ah, nice. Yes, I agree. So the `Ok(None)` variant (second arm) means the header is missing. The `Err(_)` arm is when it failed to read the value in the header.

common/src/backends/auth.rs
chesedo: :heart:

deployer/src/deployment/provisioner_factory.rs
```diff
@@ -85,82 +97,103 @@ pub enum ProvisionerError {
...
+                    })?,
+                })
+                .await
+                .map_err(|err| {
+                    shuttle_service::Error::Database(format!("failed to store resource: {err}"))
+                })?;
+
+            info
+        } else {
+            info!("Getting a {db_type} from a previous provision");
```
oddgrd, line 165: I think this is okay, but in the future there may be a problem since resources will likely be scoped, if a users scope is reduced the old resource should not be restarted. If this used a refresh token, that problem may be easier to solve (unless I misunderstood the purpose of refresh tokens). However, we don't know when resources will be scoped, so I don't think we should worry about this yet.

chesedo: Yip, I agree

PR #643, opened 2023-02-20
feat: auth cache

oddgrd:
This PR adds a cache in front of the reverse proxy in the `ShuttleAuthLayer` that tries to upgrade a cookie or api-key bearer token to a JWT. The cache is first checked for the key (cookie or api-key), if it is there set it on the request in the `ShuttleAuthLayer`. If it isn't, request an upgrade from the `auth` server through the proxy. If it succeeds, extract the token from the proxy response and set it in the cache, before setting it on the request and proceeding.

It would have been more elegant to have the cache be a layer on the proxy client, but I had some difficulty implementing this. It's certainly something we should consider in the future.

All tests pass except for one assertion in the `api_create_get_delete_projects` test, which is commented out. It was indeed broken, but it was testing functionality we don't have.


auth/src/api/handlers.rs
```diff
@@ -51,12 +52,29 @@ pub(crate) async fn login(
...
+        cookie.get("shuttle.sid").map(|id| id.to_string())
+    } else {
+        None
+    };
+
+    // If there was an id in the cookie, clear it from the cache.
+    if let Some(key) = cache_key {
+        // Clear the session's associated JWT from the cache.
+        cache_manager.invalidate(&key);
+    }
```
chesedo, line 76: We should be able tto combine the two ifs here - ie removing the `cache_key` variable.

chesedo (commented):
Progress seems good to me :smile:

auth/src/api/handlers.rs
oddgrd: Ah, indeed! This part of logout was removed in the move to gateway though, we'll need to invalidate it there instead, either in its own layer or in the auth layer.

auth/Cargo.toml
```diff
@@ -9,28 +9,30 @@ anyhow = { workspace = true }
 async-trait = { workspace = true }
```
chesedo, line 9: I might be wrong, but we should be able to undo all of the changes in this file right? Since the layer moved to gateway.

gateway/src/api/cache_layer.rs
```diff
@@ -0,0 +1,275 @@
...
+                                error = &error as &dyn std::error::Error,
+                                "failed to get public key"
+                            );
+
+                            return Ok(Response::builder()
+                                .status(StatusCode::SERVICE_UNAVAILABLE)
+                                .body(Default::default())
+                                .unwrap());
+                        }
+                    };
```
chesedo, line 210: I wonder if we can get the middle part of the jwt here and then base64 decode it. Then finally convert it to a `serde_json::Value` to read the `exp`. This way we won't have to convert it to a claim and won't need the public key function.

An example of how we could get the `exp`
https://github.com/shuttle-hq/shuttle/blob/e8536e881e33a93491750e085615b71665da988d/common/src/backends/auth.rs#L779-L785

chesedo: Because of all the matches this will produce, you might want to make a function that takes the body (line 156) as an argument and will calculate the duration (line 237). Then have this function's return type as `Result<Duration, StatusCode>` and you'll be able to replace the `match`es with `?`

chesedo (commented):
This is exciting, thanks Oddbjorn!!

chesedo (approved):
LGTM

gateway/src/api/cache_layer.rs
oddgrd: That's a good idea, I'll do that. Thanks!

auth/Cargo.toml
oddgrd: Indeed! The remaining ones are from cargo-sort.

gateway/src/api/latest.rs
```diff
@@ -659,30 +672,32 @@ pub mod tests {
...
-
-        // delete returns 404 for project that doesn't exist
-        router
-            .call(delete_project("resurrections").with_header(&authorization))
-            .map_ok(|resp| {
-                assert_eq!(resp.status(), StatusCode::NOT_FOUND);
-            })
-            .await
-            .unwrap();
+        // TODO: setting the user to admin here doesn't update the cached token, so the
```
oddgrd, line 675: We don't actually have an API for upgrading a user to the admin `AccountTier`, so in a way this test is testing functionality we don't have. We should either add this functionality before releasing the `auth` crate and then fix this test, or make a note of this and make sure we invalidate the cached token when we add the endpoint for changing account tiers.

gateway/src/api/auth_layer.rs
```diff
@@ -225,6 +289,35 @@ where
...
+    let exp = claim["exp"]
+        .as_i64()
+        .ok_or(StatusCode::INTERNAL_SERVER_ERROR)?;
+
+    let expiration_timestamp = Utc
+        .timestamp_opt(exp, 0)
+        .single()
+        .ok_or(StatusCode::INTERNAL_SERVER_ERROR)?;
+
+    let duration = expiration_timestamp - Utc::now();
```
chesedo, line 316: I've been thinking whether we want to subtract an extra 30 seconds or minute from the duration so that the cache expires a little bit before the token?

I'm worried about the case where a user might do a `cargo shuttle deploy` when their token cache might expire in 30 seconds or so. In this case the token will reach deployer which might take longer than 30 seconds to build the project before sending it on to provisioner which will fail saying the token is invalid.

gateway/src/api/latest.rs
```diff
@@ -489,10 +490,22 @@ impl ApiBuilder {
 
     pub fn with_auth_service(mut self, auth_uri: Uri) -> Self {
         let auth_public_key = AuthPublicKey::new(auth_uri.clone());
+
+        let cache = Arc::new(RwLock::new(TtlCache::new(1000)));
```
chesedo, line 494: Can we move this line to a `CacheManager::new()`?

gateway/src/api/latest.rs
```diff
@@ -489,10 +490,22 @@ impl ApiBuilder {
...
             .layer(JwtAuthenticationLayer::new(auth_public_key))
-            .layer(ShuttleAuthLayer::new(auth_uri));
+            .layer(ShuttleAuthLayer::new(
+                auth_uri,
+                Arc::new(Box::new(cache_manager)),
+            ));
+        // .layer(CacheLayer {
+        //     cache_manager: Arc::new(Box::new(cache_manager)),
+        //     public_key_fn: auth_public_key,
+        // });
```
chesedo, line 508: Comments :smiley:

gateway/src/api/latest.rs
chesedo: Commenting out the test is good for now. Thanks O

chesedo (approved):
Nice :partying_face:, thanks @oddgrd 

LGTM

gateway/src/api/auth_layer.rs
oddgrd: I've had the same thought, I agree, we should do that. We can start with 30 seconds and then test that it's enough of a buffer on monday.

gateway/src/api/latest.rs
oddgrd: Yup!

gateway/src/api/latest.rs
oddgrd: Woops :smile:

gateway/src/api/auth_layer.rs
```diff
@@ -139,74 +161,116 @@ where
...
                 let mut auth_details = None;
+                let mut cache_key = None;
 
                 if let Some(bearer) = req.headers().typed_get::<Authorization<Bearer>>() {
+                    cache_key = Some(bearer.token().trim().to_string());
                     auth_details = Some(make_token_request("/auth/key", bearer));
                 }
 
                 if let Some(cookie) = req.headers().typed_get::<Cookie>() {
+                    if let Some(id) = cookie.get("shuttle.sid") {
```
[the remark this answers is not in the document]
oddgrd: I'll look into this :+1:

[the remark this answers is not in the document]
oddgrd: Yes, I suppose we were only checking if there was *any* cookie, and if so create the upgrade request. We should indeed rather make sure there is a cookie and that it is the shuttle cookie.

gateway/src/api/auth_layer.rs
oddgrd: Subtracting 60 seconds for now.

PR #650, opened 2023-02-24
Fix: remove target from bin build name

oddgrd:
Remove the target from the binary name to be compatible with binstall, and also switch to using only the short-form git tag.

See first commit's workflow run for successful run, and see the new draft release in releases for a preview of the binaries.


chesedo (approved):
LGTM

PR #651, opened 2023-02-24
feat: allow admin scoped user to recreate any project

oddgrd:


gateway/src/service.rs
```diff
@@ -652,7 +663,38 @@ pub mod tests {
 
         // If recreated by the same user
         assert!(matches!(
-            svc.create_project(matrix, neo).await,
+            svc.create_project(matrix.clone(), neo, false).await,
+            Ok(Project::Creating(_))
+        ));
+
+        // If running project is created a different user which is admin
```
chesedo, line 670:
```suggestion
        // If running project is created by a different user which is not admin
```

gateway/src/service.rs
```diff
@@ -361,26 +361,36 @@ impl GatewayService {
...
-            .bind(&project_name)
-            .bind(&account_name)
-            .fetch_optional(&self.db)
-            .await?
+        if let Some(row) = query(
+            r#"
+        SELECT project_name, account_name, initial_key, project_state 
+        FROM projects 
+        WHERE (project_name = ?1) 
+        AND (account_name = ?2 OR ?3 = true)
```
chesedo, line 371:
```suggestion
        AND (account_name = ?2 OR ?3)
```
Might work :man_shrugging:

chesedo (approved):
Nice, LGTM

gateway/src/service.rs
oddgrd: Ah, woops, this was supposed to test that if an admin tried to `project new` with the name of running project, it would still return `ProjectAlreadyExists`. It's a bit redundant, perhaps... 

There is an assertion above this that tests if a different user that isn't admin tries to create a project with a taken name.

chesedo: Haha, okay. Maybe we can remove this one then?

oddgrd: Yeah, I'll do that.

gateway/src/service.rs
oddgrd: It does, TIL :partying_face:

PR #652, opened 2023-02-27
refactor: update opentelemetry in all our crates

chesedo:


oddgrd (approved):
Looks great! Nice and clean. I think getting rid of the box in the new tracing layers will be quite straight-forward, but we should probably not prioritize that today.

PR #653, opened 2023-02-27
Feat: unbox the ClaimService and Scoped futures

oddgrd:
We can unbox the future type in tower layers if we implement future ourselves. For how I landed on the implementation for ClaimService, see: https://github.com/tower-rs/tower/blob/master/guides/building-a-middleware-from-scratch.md

For what inspired the enum solution for the Scoped future, see: https://github.com/tower-rs/tower/blob/master/tower/src/load_shed/future.rs#L13-L63

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6088,6 +6088,7 @@ dependencies = [
  "opentelemetry-datadog",
  "opentelemetry-http",
  "pem",
+ "pin-project",
  "portpicker",
  "rand 0.8.5",
  "rcgen",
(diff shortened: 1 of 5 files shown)


chesedo (approved):
This is awesome, thanks @oddgrd :heart: 

It also seems small enough to merge. Btw, do you have rewrk stats for this maybe?

oddgrd:
Thanks! Yes, the difference is quite small, so small that I would need to do more tests to verify its significance. One could argue that this is a bit overkill, but I learned a lot about how to do this and I think we can see more gains in layers like ShuttleAuthLayer. So I'm happy whether we merge this or not. :+1:

chesedo:
Nah, I would definitely want to convert all of them eventually. At a scale of thousands this won't be an overkill as these are called for almost every request

oddgrd:
That's good because I had fun converting them :smile:

PR #655, opened 2023-02-27
Feat: cache public key

oddgrd:


chesedo (commented):
This is great! I think we can make this even simpler by implementing the cache in this function. Then we don't have to pass around a cache manager since we can initiate the cache inside `AuthPublicKey::new()`

https://github.com/shuttle-hq/shuttle/blob/6686657df487c7222b981fdcf953fb0ca2270d73/common/src/backends/auth.rs#L286-L292

chesedo (approved):
LGTM

PR #661, opened 2023-02-28
feat: add bug report and PR templates

### LINKED ISSUE #602: Add github issue templates
We should have some templates for creating issues, not only to so issues are clearer and bug reports have more information, but also so it will be easy to open a good issue. We're open to suggestions on what they would look like, and we can definitely take inspiration from other open-source projects here. There is also [this format](https://github.com/stevemao/github-issue-templates/blob/master/.github/ISSUE_TEMPLATE/BUG-REPORT.yml) which might be worth exploring.

We could start with one template for bug requests, and one for feature requests.

oddgrd:
This adds a bug report template that users can choose when opening an issue, that requires the user to include some information that is usually needed. To see how it looks rendered, click view file. Or see this link: https://github.com/shuttle-hq/shuttle/blob/f0a5066244b414cec93dabbacf2a4ab0c389dead/.github/ISSUE_TEMPLATE/BUG-REPORT.yml

The user will be able to choose the bug report template or a blank issue (when we add more templates we can disable blank templates), or they can choose to follow a link to our discord.

It also adds a PR template to encourage contributors to self-review and add docs/tests.

Closes #602

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG-REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG-REPORT.yml
@@ -1,0 +1,68 @@
+name: Bug Report
+description: File a bug report
+title: "[Bug]: "
+labels: ["bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this bug report!
+  - type: textarea
+    id: what-happened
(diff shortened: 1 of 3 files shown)


chesedo (approved):
:partying_face: LGTM

PR #662, opened 2023-03-01
refactor: unboxing AdminSecret

chesedo:
Removes the `Box::pin` from the AdminSecret. Also allows more reuse of the future.

### DIFF
--- a/common/src/backends/mod.rs
+++ b/common/src/backends/mod.rs
@@ -1,6 +1,61 @@
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use axum::response::Response;
+use http::StatusCode;
+use pin_project::pin_project;
+
 pub mod auth;
 pub mod cache;
 pub mod future;
 pub mod headers;
 pub mod metrics;
 pub mod tracing;
+
+/// Future for layers that might return a different status code
+#[pin_project]
+pub struct StatusCodeFuture<F> {
+    #[pin]
+    state: ResponseState<F>,
+}
+
+#[pin_project(project = ResponseStateProj)]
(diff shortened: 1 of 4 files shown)


common/src/backends/mod.rs
```diff
@@ -1,5 +1,44 @@
...
 pub mod tracing;
+
+/// Future for layers that might return a different status code
+#[pin_project(project = StatusCodeProj)]
+pub enum StatusCodeFuture<F> {
+    // A future that should be polled
+    Poll(#[pin] F),
+
+    // A status code to return
+    Code(StatusCode),
```
oddgrd, line 24: This make a lot more sense :smile: I created a `future.rs` dir in common in my PR since I'm reusing the simple ResponseFuture: https://github.com/shuttle-hq/shuttle/pull/663/files#diff-ff58cb873c6b133b246865d0399cf4a2900caf908be6bc3859ffeeb3081236e4 , maybe we can have all the reusable ones there? Or in mod, no big deal tbh.

oddgrd (approved):
LGTM! Nice cleanup of the ScopedFuture as well. :+1: 

Btw, I created a `future.rs` dir in common in my PR since I'm reusing the simple ResponseFuture: https://github.com/shuttle-hq/shuttle/pull/663/files#diff-ff58cb873c6b133b246865d0399cf4a2900caf908be6bc3859ffeeb3081236e4 , maybe we can have all the reusable ones there? Or in mod.rs, either is fine.

common/src/backends/mod.rs
chesedo: `futures.rs` sounds better for me, so I'll wait for your PR :smile:

PR #663, opened 2023-03-01
feat: unbox InjectPropagation and ExtractPropagation

oddgrd:
I also pulled out the ResponseFuture type, it's very simple and we can reuse it.

### DIFF
--- a/common/src/backends/tracing.rs
+++ b/common/src/backends/tracing.rs
@@ -15,7 +15,7 @@ use opentelemetry_http::{HeaderExtractor, HeaderInjector};
 use opentelemetry_otlp::WithExportConfig;
 use pin_project::pin_project;
 use tower::{Layer, Service};
-use tracing::{debug_span, Span, Subscriber};
+use tracing::{debug_span, instrument::Instrumented, Instrument, Span, Subscriber};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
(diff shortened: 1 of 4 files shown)


common/src/backends/tracing.rs
```diff
@@ -67,15 +74,47 @@ pub struct ExtractPropagation<S> {
...
+impl<F, Body, Error> Future for ExtractPropagationFuture<F>
+where
+    F: Future<Output = Result<Response<Body>, Error>>,
+{
+    type Output = Result<Response<Body>, Error>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+
+        let _guard = this.span.enter();
```
chesedo, line 93: I'm wondering if it's possible to enter a span multiple times, or if it will crash here?

chesedo (approved):
Just on question, else LGTM

common/src/backends/tracing.rs
oddgrd: My thinking is that since this is not an `async` block, when this scope is exited the EnterGuard will get dropped and the span exited. And then it will enter again when it's polled again. I'm not a 100% on this though, I'll try to find a definitive answer.

oddgrd: Went for instrumenting the future instead, so it will be entered each time it's polled, without needing to pass the span in the future struct.

PR #664, opened 2023-03-01
Remove time v0.1 dependency from main crates

jonaro00:
- Mongodb v2.4.0 got rid of its time v0.1 dependency. This plus removing default features on chrono results in all of shuttle's main  crates not depending on it (by default, shuttle-service with thruster still does due to thruster's direct dependency on it).
- Removed Cargo.lock from .gitignore since it is being tracked by git.


chesedo (approved):
Nice, thanks for this @jonaro00

oddgrd (approved):
Thanks @jonaro00! This is good to go, but we may need to update the mongodb example too then? We can do it before the next release, but if you'd like to do it that would be great :slightly_smiling_face:

PR #667, opened 2023-03-02
bug: ws not getting logs from broadcast channel

chesedo:
This fixes the WebSocket connection returning early which made it seem like deployments are done, when in reality they were not.

We have no idea why the sqlite setting is affecting the broadcast channel however. I suspect it is causing some panic inside the thread, which is in persistence, that would have send the log to the broadcast channel.

### DIFF
--- a/deployer/src/persistence/mod.rs
+++ b/deployer/src/persistence/mod.rs
@@ -20,9 +20,7 @@ use chrono::Utc;
 use serde_json::json;
 use shuttle_common::STATE_MESSAGE;
 use sqlx::migrate::{MigrateDatabase, Migrator};
-use sqlx::sqlite::{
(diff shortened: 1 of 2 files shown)


deployer/src/persistence/mod.rs
```diff
@@ -63,10 +61,17 @@ impl Persistence {
             std::fs::canonicalize(path).unwrap().to_string_lossy()
         );
 
+        // We have found in the past that setting synchronous to anything other than the default (normal) breaks the
```
oddgrd, line 64:
```suggestion
        // We have found in the past that setting synchronous to anything other than the default (full) breaks the
```

oddgrd (approved):
LGTM!

Issue #665, opened 2023-03-01
`cargo shuttle init` does not handle directory argument properly


jonaro00:
![image](https://user-images.githubusercontent.com/54029719/222243019-97cd0e69-b144-4236-ab90-6d476d6f060e.png)

First invocation: The init command fails when the directory does not exist.

Second invocation: When directory exists, the command still asks for directory, even though it was provided. I manually typed it again.

---

Env: WSL 2, Ubuntu 22.04
Shuttle: 0.11.0
Rust and cargo: 1.66.1

h-dav:
I’ll sort this

d2weber:
Also, I noticed that, when I create a new project, the name in Cargo.toml seems to be wrong:
```
/tmp/pwd$ cargo shuttle init --axum
How do you want to name your project? It will be hosted at ${project_name}.shuttleapp.rs.
✔ Project name · myname

Where should we create this project?
✔ Directory · .

     Created library (shuttle) package

✔ Do you want to create the project environment on Shuttle? · no

/tmp/pwd$ rg name Cargo.toml 
2:name = "pwd"
```
Should I create a new issue?

oddgrd:
As of https://github.com/shuttle-hq/shuttle/pull/706, which will be released on monday, the directory argument will now be properly handled. It will not error, it will create the dir and present the path to it in the `where would you like to create this project`. I'll not close this issue yet, since we might want to improve it further. Do we want to prompt the user for a path at all if they gave one in the command? Also, should we make the command require `--path <path>` so it's more clear what the command does?

Regarding the naming @d2weber, yes, I think an issue for that is good. Did it really name your project "pwd"? :joy:

oddgrd:
This is great, thanks @jonaro00! So to summarize the name fixes: 
- Generating a shuttle.toml with the project name 
- Suggesting to the user to pick a unique name
- If the name is taken, show it in the error message when creating a problem (this may be made redundant in the init command if we don't error on name taken, but can be useful for other calls to project new)
- If the name is taken, don't error, rather prompt to enter a new name and try to create a project again
- And for the path issue, I agree with you, we can remove the directory prompt if they entered a path to the init command

Would you mind opening a new issue for this? And of course, if you'd like to work on this yourself we'd be happy to answer any questions you may have :slightly_smiling_face:

PR #671, opened 2023-03-02
cargo shuttle init does not handle directory argument properly

h-dav:
Issue #665

## Description of change

Skip over prompting for directory name if directory name is specified in init command

Note: Not solved issue of file path being created properly as don't have time right now but will look into this later

closes issue #665


oddgrd:
Hey @h-dav, thanks for this :pray: There is a note in your comment about looking into the file path being created correctly, are you still doing that or is this ready for review? No rush of course, just curious!

h-dav:
Hi @oddgrd, I’m happy to do the file path being created correctly issue but currently don’t have time. The pull request is ready to be reviewed but it only fixes one of the issues detailed on the linked issue. However, I can create another pull request for the file path issue once it’s ready.

cargo-shuttle/src/lib.rs
```diff
@@ -152,12 +152,21 @@ impl Shuttle {
...
+                    .default(".".to_owned())
+                    .interact()?;
+                println!();
+            } else {
+                directory_str = project_args
+                    .working_directory
+                    .into_os_string()
+                    .into_string()
+                    .unwrap();
+            }
```
oddgrd, line 169: So I tested this now, and it seems that if you pass in the path now it will always create the project in the current directory. However the way it was, if you passed the path in the interactive prompt it would actually create the project in that directory if it exists. 

Would it be better to simply pass in the `args.path`? At least then it will show up in the prompt, and not be ignored. Though I think we should fix both problems before we merge this, but please take your time. We can also look into it if you don't think you will be able to for a while.
```suggestion
            let path = args.path.to_str().context("path arg should always be set")?;

            println!("Where should we create this project?");
            let directory_str: String = Input::with_theme(&theme)
                .with_prompt("Directory")
                .default(path.to_owned())
                .interact()?;
            println!();
```

oddgrd: In addition to the above suggestion, I think all we would need to do to fix the second issue (the directory not being created if path is passed in to init command) is to change this: https://github.com/shuttle-hq/shuttle/blob/9f7a48287ea452c578e98d72d4267c62a5208807/cargo-shuttle/src/args.rs#L206 to use the `parse_init_path` function rather than the `parse_path` function.

oddgrd:
@h-dav thanks for the effort and thanks for getting started on this! :pray:

Since I was already updating init for our upcoming release, I included my suggestions from above in this PR: https://github.com/shuttle-hq/shuttle/pull/706. Feel free to open a new PR if you find any other issues with init (or anything else)!

PR #674, opened 2023-03-03
Fix/move docker profiles to dev

oddgrd:
## Description of change

Docker swarm does not support docker profiles, so to be able to run the deploy command we had to remove these from `docker-compose.yml`. We fix this by moving them into `docker-compose.dev.yml` so they are only applied to the local runs.

Verified that the local behavior is the same, and we deployed yesterday after removing the profiles from the main compose file.


chesedo (approved):
LGTM

PR #675, opened 2023-03-03
refactor: find code to wrap legacy runtime

chesedo:
## Description of change
This finds the new code that we will need to codegen to start a legacy service with our new runtime.

With this change we will also be able to do the following:
- Get rid of the `Sync` requirement for services
- Get rid of the `runtime` in resources
- Correctly stop deploys
- Get rid of the FFI :partying_face: :tada: :confetti_ball: 

Manually using `grpcurl`


oddgrd (approved):
I love it!

PR #677, opened 2023-03-03
feat: remove box self from services, remove syncwrapper from axum service

oddgrd:
## Description of change

Removes box self from services and syncwrapper from axum service since they are no longer needed.

This PR is based on https://github.com/shuttle-hq/shuttle/pull/675

### DIFF
--- a/service/src/lib.rs
+++ b/service/src/lib.rs
@@ -540,18 +540,19 @@ pub type ShuttleWarp<T> = Result<warp::filters::BoxedFilter<T>, Error>;
 
 #[cfg(feature = "web-axum")]
 #[async_trait]
-impl Service for sync_wrapper::SyncWrapper<axum::Router> {
+impl Service for axum::Router {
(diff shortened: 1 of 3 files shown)


service/src/lib.rs
```diff
@@ -389,7 +389,7 @@ pub trait Factory: Send + Sync {
 #[async_trait]
 pub trait ResourceBuilder<T> {
     fn new() -> Self;
-    async fn build(self, factory: &mut dyn Factory, runtime: &Runtime) -> Result<T, crate::Error>;
```
chesedo, line 392: I wonder if anything else in `shuttle_service` still needs tokio now :thinking:

chesedo (approved):
LGTM

service/src/lib.rs
oddgrd: I've started on removing the runtime from resources, I'll can check in that one :+1:

PR #680, opened 2023-03-03
Feature: remove tokio runtime from all resources

oddgrd:
## Description of change

We no longer need to pass a runtime to the resourcebuilder, so we can remove it and tokio from all the resources.

This has been tested by making sure they all compile and that the tests pass for the resources that have tests.

### DIFF
--- a/resources/aws-rds/src/lib.rs
+++ b/resources/aws-rds/src/lib.rs
@@ -28,7 +28,6 @@ macro_rules! aws_engine {
                         .get_db_connection_string(Type::AwsRds(AwsRdsEngine::$struct_ident))
                         .await?;
 
-                    // A sqlx Pool cannot cross runtime boundaries, so make sure to create the Pool on the service end
                     let pool = $options_path::new()
                         .min_connections(1)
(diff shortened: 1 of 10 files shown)


resources/aws-rds/src/lib.rs
```diff
@@ -24,22 +23,17 @@ macro_rules! aws_engine {
                     Self {}
                 }
 
-                async fn build(self, factory: &mut dyn Factory, runtime: &Runtime) -> Result<$pool_path, shuttle_service::Error> {
+                async fn build(self, factory: &mut dyn Factory) -> Result<$pool_path, shuttle_service::Error> {
                     let connection_string = factory
                         .get_db_connection_string(Type::AwsRds(AwsRdsEngine::$struct_ident))
                         .await?;
 
                     // A sqlx Pool cannot cross runtime boundaries, so make sure to create the Pool on the service end
```
chesedo, line 31: We can remove the comment too :smile:

chesedo (approved):
Thanks O!! LGTM

PR #681, opened 2023-03-03
feat: trim the service loader, unpin tokio

oddgrd:
## Description of change

This start the trim of the service loader, bootstrapper and other code related to the FFI. We still use tokio for building and cleaning crates, which I'll need to think about whether we will still use. But I did unpin tokio in this PR. 


service/src/lib.rs
```diff
@@ -392,87 +383,17 @@ pub trait ResourceBuilder<T> {
...
-
 /// The core trait of the shuttle platform. Every crate deployed to shuttle needs to implement this trait.
 ///
 /// Use the [main][main] macro to expose your implementation to the deployment backend.
-//
-// TODO: our current state machine in the api crate stores this service and can move it across
-// threads (handlers) causing `Service` to need `Sync`. We should remove this restriction
 #[async_trait]
-pub trait Service: Send + Sync {
+pub trait Service: Send {
```
chesedo, line 390: :partying_face:

service/src/loader.rs
```diff
@@ -1,106 +1,18 @@
-use std::any::Any;
-use std::ffi::OsStr;
-use std::net::SocketAddr;
-use std::panic::AssertUnwindSafe;
 use std::path::{Path, PathBuf};
```
chesedo, line 1: What remains in this file? Should we possibly rename it and rename the feature?

chesedo (approved):
Woohoo, thanks O! LGTM

service/src/loader.rs
oddgrd: Indeed. We still need to build the crate, we still need `check_version` annd `check_no_panic`. I removed `ensure_cdylib`, but we should probably have a `ensure_not_cdylib` function now. Perhaps we can rename it to `builder.rs`/`build.rs`? :thinking:

chesedo: `build.rs` (or something along that) sounds good :+1:

PR #686, opened 2023-03-06
Feature: eng 465 update all the codegens

oddgrd:
## Description of change

Rewrite the codegen of shuttle_main to use shuttle_runtime to start the users project binary. Also refactored the `local_run` function to work with this change. Refactoring deployer remains, but I think that should be done in a follow-up PR.

TODO:

- [x] Clean up the codegen, I essentially just unwrapped the wrapper currently. 
- [x] Refactor unwrap of return type.
- [x] Fix the tests.

Local runs of two axum examples, one hello_world and one with the static files resource.


runtime/src/legacy/mod.rs
```diff
@@ -100,21 +100,29 @@ where
 {
     type Service: Service;
 
-    async fn load(self, factory: Fac) -> Result<Self::Service, shuttle_service::Error>;
+    async fn load(
+        self,
+        factory: Fac,
+        logger: shuttle_service::Logger,
```
chesedo, line 106: Can we move `Logger` down to the runtime?

runtime/src/legacy/mod.rs
```diff
@@ -154,9 +162,12 @@ where
         );
         trace!("got factory");
 
+        let logs_tx = self.logs_tx.clone();
+        let logger = shuttle_service::Logger::new(logs_tx, deployment_id);
```
chesedo, line 166: Thanks for the hookup O!

service/src/lib.rs
```diff
@@ -227,6 +227,7 @@ mod logger;
 pub use logger::Logger;
 
 pub use shuttle_common::database;
+pub use shuttle_common::storage_manager::StorageManager;
```
chesedo, line 230: Can we move this down to runtime too?

chesedo (approved):
Looking good :partying_face: LGTM

runtime/src/legacy/mod.rs
oddgrd: Sure!

service/src/lib.rs
oddgrd: I assume you mean move the re-export, which yes, we can :smile: The storage manager itself is used in a lot of crates that I'm not sure yet if will depend on shuttle-runtime. Maybe we can revisit this when we know.

Issue #689, opened 2023-03-07
[Bug]: Deployment crashes when using static folder and the project awakes


robertohuertasm:
### What happened?

This has been already reported in Discord over [here](https://discord.com/channels/803236282088161321/953005742842069013/1082428521835663411) and [here](https://discord.com/channels/803236282088161321/1076126506323030017/1081574721189068831).

After the introduction of [cargo-shuttle 0.11.2](https://github.com/shuttle-hq/shuttle/releases/tag/v0.11.2), with the idle-to-sleep changes, I've seen that all my projects that have a dependency of `shuttle-static-folder` will crash when they try to recover from sleep.

The error I'm getting is this one: `2023-03-06T16:48:13.792356279Z ERROR {error="Run error: Custom error: failed to provision shuttle_static_folder :: StaticFolder"} shuttle_deployer::deployment::run: service startup encountered an error`.

For now, the only alternative is to redeploy everything with `--idle-minutes 0` flag.

### Version

v0.11.2

### Which operating systems are you seeing the problem on?

Linux

### Which CPU architectures are you seeing the problem on?

x86_64

### Duplicate declaration

- [X] I have searched the issues and there are none like this.

oddgrd:
This was fixed in #692 and the fix has been released. An announcement will also be posted in Discord soon.

For your deployer to use the fixed version you'll need to first run cargo shuttle project new and then cargo shuttle project rm. Then you need to bump your dependency on shuttle-static-folder to 0.11.3, before re-deploying. You do not need to upgrade your cargo-shuttle cli or other dependencies.

robertohuertasm:
That was quick! Thanks to the team for the fast fix! ❤️

PR #692, opened 2023-03-07
bug: idle static folder

chesedo:
## Description of change

This fixes static folder projects not starting up correctly after they went into idle mode. This would happen because the static folder resource would move the folder and therefore could not move it a second time when an idle project is woken up.

This change causes the static folder resource to rather copy the folder.

Closes issue #689 

Locally, with a full running setup and the axum websocket example


oddgrd (approved):
LGTM!

PR #693, opened 2023-03-07
Feature: eng 483 trim and fix the tests in shuttle-service

oddgrd:
## Description of change

Fix the tests we should keep and remove the loader tests.

TODO: remove loader tests.

Ran the updated tests.


service/src/builder.rs
```diff
@@ -173,6 +177,42 @@ fn is_next(summary: &Summary) -> bool {
         .any(|dependency| dependency.package_name() == NEXT_NAME)
 }
 
+/// Make sure the project is a binary for legacy projects.
+fn ensure_binary(manifest: &Manifest) -> anyhow::Result<()> {
+    if manifest
+        .targets()
+        .iter()
+        .find(|target| target.is_bin())
+        .is_some()
```
chesedo, line 186:
```suggestion
        .any(|target| target.is_bin())
```
A bit of a nicer way to check it :smile:

service/tests/integration/build_crate.rs
```diff
@@ -1,64 +1,39 @@
 use std::path::{Path, PathBuf};
 
-use shuttle_service::loader::{build_crate, Runtime};
+use shuttle_service::builder::{build_crate, Runtime};
 
-#[tokio::test(flavor = "multi_thread")]
+#[tokio::test]
+#[should_panic]
```
chesedo, line 6: Let's add the expected error to this panic to make sure it is (and stays) correct

service/tests/resources/bind-panic/src/lib.rs
```diff
@@ -1,18 +0,0 @@
...
-
-struct MyService;
-
-#[shuttle_service::async_trait]
-impl Service for MyService {
-    async fn bind(
-        mut self: Box<Self>,
-        _: std::net::SocketAddr,
-    ) -> Result<(), shuttle_service::Error> {
-        panic!("panic in bind");
```
chesedo, line 11: We should make a new test for this case

chesedo (commented):
I wonder how many of the tests we would like to keep in some form

service/tests/integration/build_crate.rs
oddgrd: It just gives a '1 job failed' error from the cargo compilation, I tried to check the error chain if there was more info but couldn't find any (I assume the compilation error is from not having a main function). So I added '1 job failed' to expected.

runtime/tests/resources/bind-panic/src/main.rs
```diff
@@ -1,13 +1,13 @@
...
         panic!("panic in bind");
     }
 }
 
-#[shuttle_service::main]
-async fn bind_panic() -> Result<MyService, shuttle_service::Error> {
+#[shuttle_runtime::main]
+async fn bind_panic() -> Result<MyService, shuttle_runtime::Error> {
     Ok(MyService)
 }
```
oddgrd, line 13: This works here since we're not using any service integration from shuttle-service, but for those that do this won't be great. We'd have to add feature flags to runtime that adds a flag to runtime's dependency on shuttle-service. So it's probably best to pull out the integrations, like you suggested. Will we do that by making them individual crates, like we do for the resources?

chesedo: Yes, we will at least try to make them individual crates

chesedo (approved):

chesedo:
LGTM

PR #696, opened 2023-03-08
Feature/eng 486 update deployer with runtime changes

oddgrd:
## Description of change

This updates deployer to work with the changes we made to the runtime. So for legacy it starts the users project with the embedded runtime (and sends the path to the same runtime in the load request, but it isn't used in the legacy runtime load function), and for shuttle-next it installs and starts the shuttle-next runtime binary, and then sends the users compiled .wasm file in the load request.

TODO:

- [ ] Can probably clean up some of the code in deployer.
- [ ] Figure out if we fix the tests here or after merging/backporting changes from https://github.com/shuttle-hq/shuttle/pull/580

Most of the tests pass, but the ones that deal with a project crashing or stopping don't. This is likely solved in https://github.com/shuttle-hq/shuttle/pull/580, so I'm not sure if we need to do it in this PR.


common/src/storage_manager.rs
```diff
@@ -2,14 +2,14 @@ use std::{fs, io, path::PathBuf};
...
+pub trait StorageManager: Sync + Send {
     /// Path for a specific service build files
-    fn service_build_path<S: AsRef<str>>(&self, service_name: S) -> Result<PathBuf, io::Error>;
+    fn service_build_path(&self, service_name: String) -> Result<PathBuf, io::Error>;
 
     /// Path to folder for storing deployment files
-    fn deployment_storage_path<S: AsRef<str>>(
+    fn deployment_storage_path(
         &self,
-        service_name: S,
```
chesedo, line 12: I assume this gave some issues? Maybe we rather want to use `&str`?

runtime/src/legacy/mod.rs
```diff
@@ -47,21 +47,28 @@ use self::args::Args;
...
-) {
+pub async fn start(loader: impl Loader<ProvisionerFactory> + Send + 'static) {
     let args = Args::parse();
     let addr = SocketAddr::new(Ipv4Addr::LOCALHOST.into(), args.port);
 
     let provisioner_address = args.provisioner_address;
     let mut server_builder =
         Server::builder().http2_keepalive_interval(Some(Duration::from_secs(60)));
 
+    let storage_manager: Arc<dyn StorageManager> = match args.storage_manager_type {
```
chesedo, line 58: Wow, nice!!

chesedo (commented):
Nice solve O!

runtime/src/legacy/mod.rs
oddgrd: Thanks!

common/src/storage_manager.rs
oddgrd: Yes, I believe having generics in the methods is not object safe, but we can probably use &str, let me check!

oddgrd: Yeah, generic types in methods are not object safe: https://doc.rust-lang.org/reference/items/traits.html#object-safety

chesedo (approved):
Thanks for all the effort on this one @oddgrd!!

LGTM

PR #698, opened 2023-03-08
Update README.md

d2weber:
## Description of change

Be sure to reference any related issues by adding `closes issue #`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ And to deploy it, write:
 
 ```bash
 cargo shuttle project new
-cargo shuttle project status // until the project is "ready"
+cargo shuttle project status # until the project is "ready"
 cargo shuttle deploy
 ```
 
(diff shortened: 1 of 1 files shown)


oddgrd (commented):
Hey and thanks for this! I think we can take this one step further and remove the `cargo shuttle project status` command, project new will no longer return before the project is ready. But, we should also add a comment after `cargo shuttle project new` since the project might have been created in the init flow above. Something like `# if you didn't create your environment on shuttle already`.

d2weber:
Yes I like that idea. I also mentioned that steps should be performed within the project directory as it makes the example more complete.

oddgrd (approved):
Thanks! This is perfect, but I think we should hold off on merging in until we merge #671, since currently if you pass a path that doesn't exist to the init command it will just error.

PR #700, opened 2023-03-09
fix: Respect `Cargo.lock` when building containers

kierendavies:
Building the shuttle containers was failing during the `cargo chef cook` step because of a cyclic dependency introduced by an upstream crate. This did not happen when building outside of a container, because the versions pinned by `Cargo.lock` did not contain any cycles.

See https://github.com/tkaitchuck/aHash/issues/95


chesedo (approved):

chesedo:
Thansk for this @kierendavies!

PR #702, opened 2023-03-10
feat: extract service integrations into separate crates

oddgrd:
## Description of change

This PR extracts service integrations into their own crates. The user will need to import `shuttle_runtime` as well as the integration they want, e.g `ShuttleAxum`.

This is what an example would look like currently:
```rust 
use axum::{routing::get, Router};
use shuttle_axum::{ShuttleAxum};

async fn hello_world() -> &'static str {
    "Hello, world!"
}

#[shuttle_runtime::main]
async fn axum() -> ShuttleAxum {
    let router = Router::new().route("/hello", get(hello_world));

    Ok(router.into())
}
```

TODO: 

- [ ] in this PR or another, clean up the `shuttle_service/lib.rs` docs.
- [ ] update all the test resource crates

Refactoring the examples on this branch: https://github.com/shuttle-hq/examples/pull/26 All examples have been ran with `cargo run --bin cargo-shuttle` locally and all of them work. The ones with databases may fail on the first run, and then succeed on the second, but I believe this is an old bug.

I also upgraded some dependencies, not a lot, but I ran `make images` to make sure it doesn't make us hit the cyclic dependency bug.


runtime/src/lib.rs
```diff
@@ -4,10 +4,16 @@ mod logger;
...
 pub use next::{AxumWasm, NextArgs};
 pub use provisioner_factory::ProvisionerFactory;
 pub use shuttle_common::storage_manager::StorageManager;
-pub use shuttle_service::{main, Error, Service};
+pub use shuttle_service::{main, Error, ResourceBuilder, Service};
+
+pub type CustomError = anyhow::Error;
+pub use anyhow::Context;
+pub use tracing;
+pub use tracing_subscriber;
```
chesedo, line 19: Do we remove these from service now?

chesedo (commented):
Just some early comments

runtime/src/lib.rs
oddgrd: They aren't used for anything else, so it makes sense to me :+1:

integrations/shuttle-axum/src/lib.rs
```diff
@@ -0,0 +1,41 @@
+//! Shuttle service integration for the Axum web framework.
+//! ## Example
+//! ```rust,no_run
+//! use shuttle_axum::AxumService;
+//!
+//! async fn hello_world() -> &'static str {
+//!     "Hello, world!"
+//! }
+//!
+//! #[shuttle_axum::main]
```
chesedo, line 10:
```suggestion
//! #[shuttle_runtime::main]
```
I might be wrong about this :see_no_evil:

integrations/shuttle-axum/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+//! async fn axum() -> shuttle_service::ShuttleAxum {
```
chesedo, line 11:
```suggestion
//! async fn axum() -> impl shuttle_service::Service {
```
Not sure if we can do this?

service/src/builder.rs
```diff
@@ -208,28 +208,29 @@ fn ensure_cdylib(manifest: &mut Manifest) -> anyhow::Result<()> {
...
-
-    if version_req.matches(&valid_version) {
-        Ok(())
-    } else {
-        Err(anyhow!(
-            "the version of `shuttle-service` specified as a dependency to this service ({version_req}) is not supported by this project instance ({valid_version}); try updating `shuttle-service` to '{valid_version}' or update the project instance using `cargo shuttle project rm` and `cargo shuttle project new`"
-        ))
-    }
-}
+// TODO: do we still need this? If so we need to check the runtime version, since that's what the user depends on.
```
chesedo, line 211: Nope, should be safe to remove :partying_face:

chesedo (commented):
Me likes :smiley:

integrations/shuttle-axum/src/lib.rs
oddgrd: Either could work, but I think this is better. The dependency on shuttle_runtime is actually only needed for the codegen, but the user doesn't know that, so this makes it clear that it's not a redundant dependency.

integrations/shuttle-axum/src/lib.rs
oddgrd: As discussed probably not, but I'll leave this open in case we have any bright ideas.

service/Cargo.toml
```diff
@@ -10,39 +10,18 @@ homepage = "https://www.shuttle.rs"
...
 tracing = { workspace = true }
-tracing-subscriber = { workspace = true, features = ["env-filter"] }
-warp = { version = "0.3.3", optional = true }
-
-# Tide does not have tokio support. So make sure async-std is compatible with tokio
-# https://github.com/http-rs/tide/issues/791
-[dependencies.async-std]
-version = "1.12.0"
-optional = true
-features = ["tokio1"]
```
chesedo, line 45: This is just beautiful :cry:

service/src/builder.rs
```diff
@@ -57,7 +57,7 @@ pub async fn build_crate(
     let is_next = is_next(summary);
 
     if !is_next {
-        check_version(summary)?;
+        // check_version(summary)?;
```
chesedo, line 60: Comment?

Cargo.toml
```diff
@@ -20,7 +20,8 @@ exclude = [
   "resources/persist",
   "resources/secrets",
   "resources/shared-db",
-  "resources/static-folder"
+  "resources/static-folder",
+  "integrations"
```
chesedo, line 24: I'm not sure if we want to call these `services` instead?

chesedo (approved):
LGTM :partying_face: 

Can we also add a step to circle CI to check fmt for all the integrations?

service/Cargo.toml
oddgrd: Yes, so many good deletes in this release :heart:

Cargo.toml
oddgrd: That makes sense, I'll rename it. :+1:

chesedo (approved):
LGTM

PR #704, opened 2023-03-13
[next] tests: CI go green

chesedo:
## Description of change
Fixes up the last few tests in the next branch to make sure they will all pass.

Using `cargo test`

### DIFF
--- a/runtime/src/legacy/mod.rs
+++ b/runtime/src/legacy/mod.rs
@@ -13,7 +13,7 @@ use async_trait::async_trait;
 use clap::Parser;
 use core::future::Future;
 use shuttle_common::{
-    backends::{auth::ClaimLayer, tracing::InjectPropagationLayer},
+    claims::{ClaimLayer, InjectPropagationLayer},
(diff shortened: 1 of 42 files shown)


runtime/src/legacy/mod.rs
```diff
@@ -221,13 +221,56 @@ where
...
+                                let panic = error.into_panic();
+                                let msg = panic.downcast_ref::<&str>()
+                                    .map(|x| x.to_string())
+                                    .unwrap_or_else(|| "<no panic message>".to_string());
+
+                                error!(error = msg, "service panicked");
+
+                                stopped_tx
+                                    .send((StopReason::Crash, msg))
+                                    .unwrap();
```
oddgrd, line 250: Ah, so that's how you can get the panic message from a `JoinHandle`, cool!

oddgrd: Ah, indeed!

oddgrd (approved):
Looks great, thanks! I'm not sure what's up with those mongodb tests in provisioner, they pass locally for me. :thinking:

PR #711, opened 2023-03-14
[next] refactor: update runtime manager

chesedo:
## Description of change

Update the runtime manager to run multiple deploys at the same time. This is to allow a new deploy to be loaded while an old deploy stays up running

By running deloyer locally

### DIFF
--- a/deployer/src/deployment/run.rs
+++ b/deployer/src/deployment/run.rs
@@ -208,7 +208,7 @@ impl Built {
         let runtime_client = runtime_manager
             .lock()
             .await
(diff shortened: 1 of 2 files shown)


oddgrd (approved):
LGTM, a lot cleaner too! I was wondering if `get_runtime_client` should try to get the client from the `runtimes` map first?

chesedo:
> I was wondering if `get_runtime_client` should try to get the client from the `runtimes` map first?

If the code tries to get a runtime for a Uuid that is already in the map then we have a serious bug :sweat_smile:

PR #715, opened 2023-03-15
feat: embed protoc in shuttle-proto

oddgrd:
## Description of change

This embeds protoc in shuttle-proto, so it doesn't need to be installed by the user to install cargo-shuttle.

Seems to work on linux, about to test on windows.


chesedo (approved):
LGTM

PR #716, opened 2023-03-15
[next] bug: communicating resources

chesedo:
## Description of change
This makes a runtime use the resources given to it by deployer (when waking from idling). It also makes runtime communicate back to deployer which resources where provisioned.

Also closes ENG-252

Using local runners

### DIFF
--- a/proto/src/lib.rs
+++ b/proto/src/lib.rs
@@ -210,6 +210,7 @@ pub mod runtime {
         wasm: bool,
         storage_manager_type: StorageManagerType,
         provisioner_address: &str,
+        auth_uri: Option<&String>,
         port: u16,
         get_runtime_executable: impl FnOnce() -> PathBuf,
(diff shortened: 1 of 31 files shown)


proto/src/lib.rs
```diff
@@ -247,9 +254,12 @@ pub mod runtime {
...
 
-        let runtime_client = runtime_client::RuntimeClient::connect(conn)
-            .await
-            .context("connecting runtime client")?;
+        let channel = conn.connect().await.context("connecting runtime client")?;
+        let channel = ServiceBuilder::new()
+            .layer(ClaimLayer)
+            .layer(InjectPropagationLayer)
+            .service(channel);
+        let runtime_client = runtime_client::RuntimeClient::new(channel);
```
oddgrd, line 262: Nice!

oddgrd (approved):
Looks great!

PR #717, opened 2023-03-15
fix: static folder local run clearing file contents, add missing tests in cargo-shuttle

oddgrd:
## Description of change

This fixes the bug where `cargo shuttle run` deletes the contents of files in the static assets folder. I also added tests to cargo shuttle for static folder and shuttle-next projects.

Additionally this PR exports `Factory` from `shuttle_runtime` and imports it in the codegen in the `extra_imports` variable, to fix a bug where the codegen was missing the `Factory` trait for static file projects.

Tested with local run, and added a test that runs `axum/static-files` locally.

### DIFF
--- a/cargo-shuttle/tests/integration/run.rs
+++ b/cargo-shuttle/tests/integration/run.rs
@@ -136,14 +136,7 @@ async fn axum_static_files() {
 
     assert_eq!(request_text, "Hello, world!");
 
-    let request_text = client
-        .get(format!("{url}"))
-        .send()
-        .await
-        .unwrap()
-        .text()
-        .await
-        .unwrap();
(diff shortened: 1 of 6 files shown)


cargo-shuttle/tests/integration/run.rs
```diff
@@ -120,6 +120,65 @@ async fn rocket_postgres() {
...
+        .post(format!("{url}/uppercase"))
+        .body("uppercase this")
+        .send()
+        .await
+        .unwrap()
+        .text()
+        .await
+        .unwrap();
+
+    assert_eq!(post_text, "UPPERCASE THIS");
```
chesedo, line 179: :heart_eyes:

chesedo (approved):
LGTM

PR #722, opened 2023-03-16
chore: Update bollard to v0.14.0

kierendavies:
This is in preparation for Podman rootless support, which requires https://github.com/fussybeaver/bollard/pull/298


chesedo (approved):
Thanks for all these PRs @kierendavies

PR #723, opened 2023-03-16
Reference container images by full name

kierendavies:
This is useful for users of Podman, which isn't configured out-of-the-box to use the `docker.io` registry. It's also a good security practice, in case of a misconfigured default registry.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,6 +1,6 @@
 #syntax=docker/dockerfile-upstream:1.4.0-rc1
 ARG RUSTUP_TOOLCHAIN
-FROM rust:${RUSTUP_TOOLCHAIN}-buster as shuttle-build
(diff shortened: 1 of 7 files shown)


chesedo (approved):
Nice, thanks for this @kierendavies

kierendavies:
I don't see a merge button, presumably because of the timeouts in the `platform-test-cargo-shuttle` CI job. Is there anything I can do about that?

kierendavies:
I missed some hard-coded image names. Branch updated. (And also rebased onto latest `main`.)

chesedo:
> I don't see a merge button, presumably because of the timeouts in the `platform-test-cargo-shuttle` CI job. Is there anything I can do about that?

Oh, I was just waiting to get the latest release out before merging all these. Should happen sometime today

PR #725, opened 2023-03-17
[next] bug: misc fixes

chesedo:
## Description of change
This fixes logging being broken in serenity since a span would fail to drop when using a thread local subscriber.

Also fixes tokens expiring too soon since builds now seem to take longer than 5 minutes.

Logging has been tested locally. ~Still need to test the new expiry status code though~

### DIFF
--- a/gateway/src/api/auth_layer.rs
+++ b/gateway/src/api/auth_layer.rs
@@ -5,7 +5,6 @@ use axum::{
     headers::{authorization::Bearer, Authorization, Cookie, Header, HeaderMapExt},
     response::Response,
 };
-use chrono::{TimeZone, Utc};
 use futures::future::BoxFuture;
 use http::{Request, StatusCode, Uri};
 use hyper::{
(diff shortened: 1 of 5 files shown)


gateway/src/api/auth_layer.rs
```diff
@@ -290,46 +280,6 @@ where
     }
 }
 
-fn extract_token_expiration(token: String) -> Result<Duration, StatusCode> {
```
oddgrd, line 293: I assume the reasoning here is that since we have such a large buffer in the cache, we don't need to know the exact expiration time of the token? I suppose if the time it takes for gateway to receive the token from auth  is significant we have a bigger problem.

runtime/src/alpha/mod.rs
```diff
@@ -54,8 +54,6 @@ pub async fn start(loader: impl Loader<ProvisionerFactory> + Send + 'static) {
     let args = Args::parse();
     let addr = SocketAddr::new(Ipv4Addr::LOCALHOST.into(), args.port);
 
-    setup_tracing(tracing_subscriber::registry(), "shuttle-alpha");
```
oddgrd, line 57: Is this not needed because we have a subscriber setup in the alpha codegen?

oddgrd (approved):
LGTM! Just a couple of questions.

gateway/src/api/auth_layer.rs
chesedo: Yes, with the bigger buffer it feels like doing calculations becomes more complex and might have weird and unexpected knock on effects. But then again, setting `EXP_MINUTES` to less than 5 min will also have weird knock on effects here :sweat_smile:

PR #728, opened 2023-03-20
bug: next runtime not sending stop signal

chesedo:
## Description of change
Fixes the next runtime that missed an implementation for the stop subscriber

Fails locally currently.

### DIFF
--- a/deployer/src/deployment/run.rs
+++ b/deployer/src/deployment/run.rs
@@ -60,15 +60,25 @@ pub async fn task(
             active_deployment_getter.clone(),
             runtime_manager.clone(),
         );
-        let cleanup = move |response: SubscribeStopResponse| {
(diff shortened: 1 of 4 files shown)


deployer/src/deployment/run.rs
```diff
@@ -369,15 +379,15 @@ async fn run(
             info!(response = ?response.into_inner(),  "start client response: ");
 
             // Wait for stop reason
-            let reason = stream.message().await.unwrap().unwrap();
+            let reason = stream.message().await.unwrap();
```
oddgrd, line 382: Should we handle this error, or change it to expect to make it clear why it shouldn't fail?

oddgrd (approved):
LGTM! One small comment.

PR #736, opened 2023-03-20
chore: update workspace dependencies

jonaro00:
## Description of change

Update workspace dependencies.

`cargo check` passes, no other tests done (let's see if CI likes it).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,7 @@ version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "453e534d4f46dcdddd7aa8619e9a664e153f34383d14710db0b0d76c2964db89"
 dependencies = [
- "base64",
+ "base64 0.13.1",
  "hyper",
  "openssl",
  "reqwest",
@@ -48,9 +48,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
+version = "0.7.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
+checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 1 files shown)


oddgrd:
Awesome @jonaro00, thanks for taking this on! So is this from a `cargo update`? If that works it should confirm we avoided the ahash cyclical dependency bug, because we hit that the last time we ran `cargo update`. :partying_face: 

For this PR I think we will also need to update the various `Cargo.toml`'s, however, which makes this a larger undertaking. The main reason is that lockfiles will be ignored by depenendants on library crates: https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries Would you be interested in doing that as well?

By the way I am updating rust and cargo separately so we can get that released on monday: https://github.com/shuttle-hq/shuttle/pull/738

jonaro00:
Yes it is `cargo update`.

Yeah, the lock file currently only affects binaries such as `cargo-shuttle` (are there others?), and dependencies while developing. When a user depends on the libraries, they will pick the highest available version compatible with `Cargo.toml`, so there is not really a big need to bump non-major versions in `Cargo.toml`s since updates to such dependencies are handled by the dependant. Bumping those versions too much could even cause incompatibilities downstream (kind of like the Tokio pin was doing). I can go through and check for major updates in libs and bump minor ones in binaries.

Also, while I was trying the `scripts/upgrade-deps.sh`, I realized that it was (1) broken and (2) doesn't make sense: The cargo update at the root level makes sense, but going through all examples etc where there are no lock files checked in and running `cargo update` only affects the developers' (ignored) lock files, while "checking that the sub-crate builds with the latest deps, which is something that I hope CI is already doing. The update script could be modified to only branch, update root, and commit, but that makes it kind of trivial. I don't see a reason to keep it if lock files in examples etc are not checked in.

I also noticed that the new `runtime` crate has a lot of unnecessary dependencies, such as `axum` (from common/backend IIRC). This looks like it would only be of interest if the `next` feature is used. (Trimming deps in library crates is good if possible)

oddgrd:
> Yeah, the lock file currently only affects binaries such as cargo-shuttle (are there others?), and dependencies while developing. When a user depends on the libraries, they will pick the highest available version compatible with Cargo.toml, so there is not really a big need to bump non-major versions in Cargo.tomls since updates to such dependencies are handled by the dependant. Bumping those versions too much could even cause incompatibilities downstream (kind of like the Tokio pin was doing). I can go through and check for major updates in libs and bump minor ones in binaries.

I see what you mean, for example if we bump tokio to "1.26" and a users dependency is pinned to a lower version, that could be a problem (are there other scenarios I am missing here?). Although I don't think it's likely to happen often, we could also just bump these dependencies when we need to. Bumping all of them could be simpler, though, and it's what we've done in the past. If you'd rather do it conservatively (not bump everything), that's fine too. 

> Also, while I was trying the scripts/upgrade-deps.sh, I realized that it was (1) broken and (2) doesn't make sense: The cargo update at the root level makes sense, but going through all examples etc where there are no lock files checked in and running cargo update only affects the developers' (ignored) lock files, while "checking that the sub-crate builds with the latest deps, which is something that I hope CI is already doing. The update script could be modified to only branch, update root, and commit, but that makes it kind of trivial. I don't see a reason to keep it if lock files in examples etc are not checked in.

I don't think anyone has touched that for a while :joy: It actually runs `cargo upgrade` which upgrades deps in `Cargo.toml`. So it could be useful with some tweaks, but I don't think I used it myself last time I did this.

> I also noticed that the new runtime crate has a lot of unnecessary dependencies, such as axum (from common/backend IIRC). This looks like it would only be of interest if the next feature is used. (Trimming deps in library crates is good if possible)

Yes, good catch! This is a result of us developing a little bit fast-and-loose to finish up this release. There is certainly room for trimming dependencies, probably by quite a bit, which should also make significant improvements to compile times. The last time we did this thoroughly, mainly trimming the common crate deps and feature flagging, we reduced compile times by almost 1/3. So if you see some easy wins here, it would be terrific to have that as well for this PR. If you see a lot of this, it might also make sense to do it in a follow-up PR.

The common crate has also grown increasingly complex and might be due for a refactor as well, but that might be best to do in a separate PR.

gateway/Cargo.toml
```diff
@@ -6,44 +6,48 @@ license.workspace = true
 publish = false
 
 [dependencies]
-acme2 = "0.5.1"
```
oddgrd, line 9: Good catch!

oddgrd (approved):
LGTM! Thank you so much @jonaro00 for this cleanup and all these bumps, it was overdue. :pray:

iulianbarbu (approved):
Thanks for upgrading the deps! 👍

PR #737, opened 2023-03-20
Add feature suggestion issue template

jonaro00:


.github/ISSUE_TEMPLATE/FEATURE-SUGGESTION.yml
```diff
@@ -0,0 +1,30 @@
+0name: Feature suggestion
```
oddgrd, line 1: "0name" is not a valid key. I'm also wondering if it should be called "Feature Request", but that's a bit of a nit and either is fine.
```suggestion
name: Feature suggestion
```

.github/ISSUE_TEMPLATE/FEATURE-SUGGESTION.yml
```diff
@@ -0,0 +1,30 @@
...
+    attributes:
+      label: Suggestion or Example of how the feature would be used
+      description: What should the feature look like?
+  - type: checkboxes
+    id: duplicate
+    attributes:
+      label: Duplicate declaration
+      description: Please confirm that you are not creating a duplicate issue.
+      options:
+        - label: I have searched the issues and there are none like this.
```
oddgrd, line 29: Should this be something like "... and this feature has not been requested before."?

oddgrd (approved):
Thanks so much @jonaro00, this is great. Just two very small comments and this is good to merge :partying_face:

.github/ISSUE_TEMPLATE/FEATURE-SUGGESTION.yml
jonaro00: Not sure how that got there 😄

ivancernja:
lgtm!

PR #738, opened 2023-03-21
chore: bump rust and cargo to 1.68

oddgrd:
## Description of change

This bumps our pinned rust version for deployer to 1.68, as well as bumping our `cargo` deps to v1.68. There were some changes to how cargo handles test errors in 1.66, I need to double check to make sure I have implemented these changes correctly on our end. For reference: https://github.com/rust-lang/cargo/pull/11028

Ran deploy tests locally and made images. 

TODO: test on unstable.

### DIFF
--- a/deployer/src/error.rs
+++ b/deployer/src/error.rs
@@ -38,7 +38,7 @@ pub enum Error {
 
 #[derive(Error, Debug)]
 pub enum TestError {
-    #[error("Tests failed")]
+    #[error("The deployment's tests failed.")]
     Failed(CliError),
     #[error("Failed to setup tests run: {0}")]
     Setup(#[from] anyhow::Error),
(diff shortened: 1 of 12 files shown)


deployer/src/error.rs
```diff
@@ -38,8 +38,8 @@ pub enum Error {
 
 #[derive(Error, Debug)]
 pub enum TestError {
-    #[error("Tests failed: {0}")]
-    Failed(#[from] CargoTestError),
```
oddgrd, line 42: CargoTestError was removed in the PR I referenced in the PR description, and CliError does not impl Display. I'll need to double-check that how I handled these changes is correct.

chesedo (approved):
LGTM

iulianbarbu (approved):

jonaro00:
@oddgrd By the way, after this is releases, wont the recompilation issue just switch over so that it happens for users on Rust <=1.67?
On that note, would calling the user's installed `cargo` instead of bundling cargo be a goal worth pursuing? Or is the usage of cargo too integrated?

PR #740, opened 2023-03-21
feat: resources endpoint

chesedo:
## Description of change
Prepares a resources endpoint for the upcoming console :speak_no_evil: 

NA

### DIFF
--- a/common/src/resource.rs
+++ b/common/src/resource.rs
@@ -5,9 +5,13 @@ use serde_json::Value;
 
 use crate::{database, DatabaseReadyInfo};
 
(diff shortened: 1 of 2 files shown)


oddgrd (approved):
Looks great, I look forward to seeing it in action! One thing I am wondering, and this is a bit of a brainstorm, should we have a consistent strategy for documenting the endpoints intended to be consumed by the console (or all of them :sweat_smile:)? Perhaps something like openapi down the line (with which I only have experience consuming). It may not be so clear to our console developers what this returns. :thinking: 

Either way looks good to me! Just something we can keep in mind as we develop the console api.

PR #742, opened 2023-03-22
Cargo.lock: remove patch unused occurrences

iulianbarbu:
Introduced previously in a PR and this doesn't need to live in the repo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,6 +83,11 @@ commands:
   apply-patches:
     steps:
       - run:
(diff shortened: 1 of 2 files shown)


oddgrd (approved):
LGTM, thanks Iulian! Just one thought, perhaps this should be a step in the workspace job instead?

iulianbarbu:
> LGTM, thanks Iulian! Just one thought, perhaps this should be a step in the workspace job instead?

I think it makes sense to move it to the workspace job. I am testing it rn.

PR #744, opened 2023-03-22
fix: Timing of deployment status and local run printouts

jonaro00:
## Description of change

1. ~Prints the "Starting ... on localhost:8000" string in local run earlier so that is always prints before any output from the run.~ Didn't make sense.
2. Wait 500ms before calling project summary after deploy. This delay fixes "Deployment has not entered running state" showing on normal deploys since the deploy now has enough time to enter Running/Crashed state.
3. Fixed alignment of some printouts in local run.
4. Other minor cleanup.

Tested on local run and deploy against prod.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -12,6 +12,7 @@ use shuttle_common::models::resource::get_resources_table;
 use shuttle_common::project::ProjectName;
 use shuttle_common::resource;
 use shuttle_proto::runtime::{self, LoadRequest, StartRequest, SubscribeLogsRequest};
+use tokio::task::JoinSet;
 
 use std::collections::HashMap;
 use std::ffi::OsString;
@@ -37,7 +38,7 @@ use git2::{Repository, StatusOptions};
 use ignore::overrides::OverrideBuilder;
 use ignore::WalkBuilder;
 use shuttle_common::models::{project, secret};
-use shuttle_service::builder::{build_workspace, Runtime};
+use shuttle_service::builder::{build_workspace, BuiltService};
 use std::fmt::Write;
 use strum::IntoEnumIterator;
 use tar::Builder;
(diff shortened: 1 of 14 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -439,12 +436,13 @@ impl Shuttle {
...
 
+        let service_name = self.ctx.project_name();
         let working_directory = self.ctx.working_directory();
 
         trace!("building project");
         println!(
-            "{:>12} {}",
-            "Building".bold().green(),
+            "{} {}",
+            "    Building".bold().green(),
```
chesedo, line 445: I might be a bit confused, why is this change needed?

cargo-shuttle/src/lib.rs
```diff
@@ -603,15 +590,27 @@ impl Shuttle {
...
         };
+        let addr = SocketAddr::new(addr.into(), run_args.port);
 
-        let addr = SocketAddr::new(addr, run_args.port);
+        println!(
+            "    {} {} on http://{}\n",
+            "Starting".bold().green(),
+            service_name,
+            addr
+        );
```
chesedo, line 613:
```suggestion
        println!(
            "{:>12} {} on http://{}\n",
            "Starting".bold().green(),
            service_name,
            addr
        );
```
This just makes it easier to keep things aligned without having to count spaces.

chesedo (commented):
Nice, thanks for the touch ups @jonaro00! I just have one question :smile:

cargo-shuttle/src/lib.rs
jonaro00: The bold and green modifiers add control characters before and after the word, meaning it is already more than 12 bytes/chars long when given to the formatter, meaning no spaces are added. (Try it yourself :D)

You can solve this by either adding the correct amount of space, or to add the space alignment before bold+green, resulting in 2 format-like calls.

I chose the former.

cargo-shuttle/src/lib.rs
jonaro00: Same as other comment

cargo-shuttle/src/lib.rs
chesedo: Ahh, of course :man_facepalming:

chesedo (approved):

PR #746, opened 2023-03-23
feat: record resources in codegen

chesedo:
## Description of change
This updates the codegen to correctly record all types of resources being created... including secrets and static folders.

Using the local runner.

### DIFF
--- a/resources/secrets/src/lib.rs
+++ b/resources/secrets/src/lib.rs
@@ -3,7 +3,7 @@ use std::collections::BTreeMap;
 use async_trait::async_trait;
 
 use serde::{Deserialize, Serialize};
-use shuttle_service::{Error, Factory, ResourceBuilder};
+use shuttle_service::{Error, Factory, ResourceBuilder, Type};
 
 #[derive(Serialize)]
 pub struct Secrets;
@@ -11,6 +11,8 @@ pub struct Secrets;
 /// Get a store with all the secrets available to a deployment
 #[async_trait]
 impl ResourceBuilder<SecretStore> for Secrets {
+    const TYPE: Type = Type::Secrets;
+
     type Output = SecretStore;
 
     fn new() -> Self {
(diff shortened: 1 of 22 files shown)


resources/secrets/src/lib.rs
```diff
@@ -2,25 +2,36 @@ use std::collections::BTreeMap;
...
+
+    async fn output(self, factory: &mut dyn Factory) -> Result<Self::Output, crate::Error> {
         let secrets = factory.get_secrets().await?;
 
         Ok(SecretStore { secrets })
     }
 }
 
 /// Store that holds all the secrets available to a deployment
+#[derive(Deserialize, Serialize, Clone)]
```
chesedo, line 34: Since this is also the `Output` we'll end up storing the secret values in the DB. I don't know if we want to do that though :man_shrugging:

resources/secrets/src/lib.rs
```diff
@@ -2,25 +2,36 @@ use std::collections::BTreeMap;
...
         Self {}
     }
 
-    async fn build(self, factory: &mut dyn Factory) -> Result<SecretStore, Error> {
+    async fn build(build_data: &Self::Output) -> Result<SecretStore, crate::Error> {
+        Ok(build_data.clone())
+    }
+
+    async fn output(self, factory: &mut dyn Factory) -> Result<Self::Output, crate::Error> {
         let secrets = factory.get_secrets().await?;
```
chesedo, line 27: Maybe this can now read the secrets from the filesystem rather than getting them from factory? This means we can remove the secrets from the factory.

chesedo: But maybe we would still want to access historical secrets?

oddgrd: I think storing them on the filesystem rather than the database makes sense, that way they reflect the latest build. But I'm not sure if that will be an issue when we allow setting secrets from the console, I feel like in that scenario having them in the DB would make more sense. :thinking:

chesedo: I'm also thinking maybe they should not be a mutable resource

oddgrd (approved):
LGTM, thanks!

PR #747, opened 2023-03-23
Feat: update resources with resourcebuilder changes

oddgrd:
## Description of change

This updates the resources with the changes in https://github.com/shuttle-hq/shuttle/pull/746, which needs to be merged before this.

Local runs.

### DIFF
--- a/common/src/models/service.rs
+++ b/common/src/models/service.rs
@@ -10,7 +10,7 @@ use comfy_table::{
 };
 use crossterm::style::Stylize;
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, fmt::Display};
+use std::{collections::HashMap, fmt::Display, path::PathBuf};
 use uuid::Uuid;
 
 #[derive(Deserialize, Serialize)]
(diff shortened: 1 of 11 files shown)


chesedo (approved):
Seems all good to me. Thanks O

resources/static-folder/src/lib.rs
```diff
@@ -28,11 +30,18 @@ impl<'a> StaticFolder<'a> {
 
 #[async_trait]
 impl<'a> ResourceBuilder<PathBuf> for StaticFolder<'a> {
+    const TYPE: Type = Type::Secrets;
```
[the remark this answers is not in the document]
oddgrd: Woops, thanks!

common/src/models/service.rs
```diff
@@ -200,3 +210,52 @@ fn get_secrets_table(secrets: &[&resource::Response]) -> String {
...
+    for (i, _) in persist_instances.iter().enumerate() {
+        table.add_row(vec![format!("Persist Instance {i}")]);
+    }
+
+    format!(
+        r#"These instances are linked to this service
+{table}
+"#,
+    )
+}
```
oddgrd: This one I'm not so sure about how we should handle, since the output just contains the project name. :thinking:

common/src/models/service.rs
```diff
@@ -121,31 +122,140 @@ Most recent deploys for {}
...
+    for _ in persist_instances {
+        table.add_row(vec!["Instance"]);
+    }
+
+    format!(
+        r#"These instances are linked to this service
+{table}
+"#,
+    )
 }
```
oddgrd, line 261: This one I'm not so sure about how we should handle, since the output just contains the project name. :thinking:

chesedo (approved):
LGTM, we can fix all the tables in ENG-518

PR #749, opened 2023-03-25
fix: unknown resource type

oddgrd:
## Description of change

Fixes a bug where resources other than databases were not recognized due to missing match arms in the `FromStr` impl for `persistence::resource::Type`. Error message below:

```bash
message: Persistence failure: Database error: error occurred while decoding column "type": 'secrets' is an unknown resource type
```

Local deploy of the postgres poem example after adding `shuttle_secrets` to it, and also tested with all resources in https://github.com/shuttle-hq/shuttle/pull/750 with the fix from this branch.


chesedo (approved):
Thanks O :heart:

PR #750, opened 2023-03-25
feat: polish CLI after 0.13 updates

oddgrd:
## Description of change

This polishes the resource tables by:
- Adding the service name in bold to the table description
- Making the table headers bold
- Making the static folder display only the name of the folder, not the full path
- Changing the small tables to not use full width (making them consistent with the secrets table).
- Updating the existing tables (deployment list, secrets) with the above changes to be consistent.
- Making the service name bold in `cargo shuttle status` output.

What I did not do in this PR was remove the `cargo shuttle secrets` command, since I don't want to make this last-minute PR too large. I suggest we do it for the next release.

Tested locally on this branch: https://github.com/shuttle-hq/shuttle/pull/749, with the `postgres` poem example, but with all resources (database, static, persist, secrets). Here is what it looks like:

image

image

image

### DIFF
--- a/common/src/models/resource.rs
+++ b/common/src/models/resource.rs
@@ -78,10 +78,11 @@ fn get_databases_table(databases: &Vec<&Response>, service_name: &str) -> String
     }
 
     format!(
-        r#"These databases are linked to {}
+        r#"These {} are linked to {}
 {table}
 "#,
-        service_name.bold()
+        "databases".bold(),
+        service_name
     )
 }
 
@@ -102,10 +103,11 @@ fn get_secrets_table(secrets: &[&Response], service_name: &str) -> String {
     }
 
     format!(
-        r#"These secrets can be accessed by {}
+        r#"These {} can be accessed by {}
 {table}
 "#,
-        service_name.bold()
+        "secrets".bold(),
+        service_name
     )
 }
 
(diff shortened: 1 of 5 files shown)


common/src/models/resource.rs
```diff
@@ -74,19 +78,22 @@ fn get_databases_table(databases: &Vec<&Response>) -> String {
     }
 
     format!(
-        r#"These databases are linked to this service
+        r#"These databases are linked to {}
 {table}
 "#,
+        service_name.bold()
```
chesedo, line 84: I was thinking more of making "databases" bold  and not the service name. Not sure how you feel?

common/src/models/resource.rs
```diff
@@ -95,57 +102,61 @@ fn get_secrets_table(secrets: &[&Response]) -> String {
...
     for folder in static_folders {
         let path = serde_json::from_value::<PathBuf>(folder.data.clone())
             .unwrap()
-            .display()
-            .to_string();
+            .file_name()
+            .expect("static folder path should have a final component")
+            .to_str()
+            .expect("static folder file name should be valid unicode")
+            .to_owned();
```
chesedo, line 129: Can't we rather get it from `folder.config` somehow?

chesedo (approved):
I happy to merge this as is.

common/src/models/resource.rs
oddgrd: Hmm, for me it's 50/50, so if you have a preference then we can go with that. :+1:

common/src/models/resource.rs
oddgrd: Ah, I didn't think of that! Yes, that should be possible. I likely won't have time to address these today, by the way, but I can get to them in the morning.

common/src/models/resource.rs
oddgrd: Implemented now and it does make it clearer.

PR #753, opened 2023-03-26
feat: commit generated proto files

oddgrd:
## Description of change

This commit uses a test to compile and generate the `.proto` files, which we then commit, and with the test running in CI we ensure that the generated files are up-to-date if any changes are made to the `.proto` files. This means the protobuf compiler does not need to be installed to compile `shuttle-proto`, which also means it won't be required to install the `cargo-shuttle` CLI.

Protoc will still be required to run/build shuttle projects until we update `opentelemetry-otlp` to 0.19, which has been released, we just need to wait for `tracing-opentelemtry` to be updated as well (or do it ourselves). After upgrading these crates `protoc` will no longer be required for shuttle users, it will only be required to make changes to the `.proto` files in `shuttle-proto`.

With this PR we should also be able to remove the protoc install from the binary builds.

Big thanks to @sd2k for suggesting this solution, this PR is largely based on their PR to tonic-health: https://github.com/hyperium/tonic/pull/1065/

Added a test in `shuttle-proto` and ran `PROTOC="" cargo build -p cargo-shuttle`.

### DIFF
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -20,5 +20,5 @@ tracing = { workspace = true }
 workspace = true
 features = ["claims", "error", "service", "wasm"]
 
-[build-dependencies]
+[dev-dependencies]
 tonic-build = { workspace = true }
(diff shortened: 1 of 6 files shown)


chesedo (approved):
LGTM :heart:

PR #762, opened 2023-03-27
fix: windows path canonicalization bug in static folder

oddgrd:
## Description of change

The static folder will fail to be provisioned on windows due to path canonicalization converting the input path to UNC. This PR fixes that by using the `dunce` crate, like we did in cargo-shuttle.

Tested local run of `examples/axum/static-files` with and without this fix on windows. Without this fix, it will fail to provision the static folder for local runs, but deployments work.


chesedo (approved):
Nice, thanks O :partying_face:

PR #765, opened 2023-03-29
Remove cargo from cargo shuttle

gautamprikshit1:
## Description of change

Removed cargo dependency from cargo shuttle workspace member.


cargo-shuttle/src/lib.rs
```diff
@@ -4,7 +4,7 @@ pub mod config;
 mod init;
 mod provisioner_server;
 
-use cargo::util::ToSemver;
+//use cargo::util::ToSemver;
```
chesedo, line 7: Just an extra comment :smile:

chesedo (approved):
Nice, thanks for the dependency trim @gautamprikshit1

cargo-shuttle/src/lib.rs
chesedo, line 7:
```suggestion
```

PR #768, opened 2023-03-29
refactor: Enable exhaustiveness check of command matching

kierendavies:


chesedo (approved):
Nice catch @kierendavies . No idea why I never though of this solution way back when :man_facepalming:

PR #769, opened 2023-03-29
feat: resource delete

chesedo:
## Description of change
This creates a CLI command to delete resources. It is still a WIP as it only allows listing resources first to know which ones to delete.

...

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -195,6 +195,13 @@ impl SecretStore {
     }
 }
 
+pub(crate) fn split_first_component(s: &str) -> (&str, Option<&str>) {
+    match s.split_once("::") {
+        Some((first, rest)) => (first, Some(rest)),
+        None => (s, None),
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use proptest::prelude::*;
--- a/common/src/resource.rs
+++ b/common/src/resource.rs
@@ -1,11 +1,11 @@
-use std::fmt::Display;
+use std::{fmt::Display, str::FromStr};
 
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 #[cfg(feature = "openapi")]
 use utoipa::ToSchema;
 
-use crate::database;
+use crate::{database, split_first_component};
 
 /// Common type to hold all the information we need for a generic resource
 #[derive(Clone, Deserialize, Serialize)]
@@ -25,7 +25,7 @@ pub struct Response {
     pub data: Value,
 }
 
-#[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq)]
+#[derive(Clone, Copy, Debug, Deserialize, Serialize, Eq, PartialEq)]
 #[serde(rename_all = "lowercase")]
 #[cfg_attr(feature = "openapi", derive(ToSchema))]
 #[cfg_attr(feature = "openapi", schema(as = shuttle_common::resource::Type))]
@@ -61,3 +61,30 @@ impl Display for Type {
         }
     }
 }
+
+impl FromStr for Type {
+    type Err = ParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match split_first_component(s) {
+            ("database", Some(db_type)) => {
+                Ok(Type::Database(db_type.parse().map_err(|_| ParseError)?))
+            }
+            ("secrets", None) => Ok(Type::Secrets),
+            ("static_folder", None) => Ok(Type::StaticFolder),
+            ("persist", None) => Ok(Type::Persist),
+            _ => Err(ParseError),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct ParseError;
+
+impl Display for ParseError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "invalid resource type")
+    }
+}
+
+impl std::error::Error for ParseError {}
--- a/deployer/src/handlers/mod.rs
+++ b/deployer/src/handlers/mod.rs
@@ -5,7 +5,7 @@ use axum::extract::{Extension, Path, Query};
 use axum::handler::Handler;
 use axum::headers::HeaderMapExt;
 use axum::middleware::{self, from_extractor};
-use axum::routing::{get, post, Router};
+use axum::routing::{delete, get, post, Router};
 use axum::{extract::BodyStream, Json};
 use bytes::BufMut;
 use chrono::{TimeZone, Utc};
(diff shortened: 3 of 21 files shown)


kierendavies:
@chesedo Oops, I didn't notice before that you created a PR for me.  This branch is now ready for review.

I didn't write any automated tests for the new `resource delete` command.  The test harness could do with a bit of love, so that it's possible to write maintainable tests for this, and I'm not able to put in the time for that.  AFAICT not much of the CLI is auto-tested anyway.

I also deliberately avoided some obvious candidates for refactoring, to keep the scope reasonable.

kierendavies:
CI caught an inconsistency when the `errors` feature is disabled. Can you help me understand the intended behaviour so I can fix it?

oddgrd:
Thanks Kieren, this is looking good! The reason this errors is because the `ParseError` has some dependencies we don't want exposed to all dependents of `common` (which includes user projects), so it's behind a feature flag. Now it doesn't error in `deployer` which does not enable the `errors` feature of `common`, but I believe this is by accident, since deployer depends on shuttle-proto which does have the `errors` feature enabled.

Since your additions to the `ParseError` does not deal with parsing of prost generated types, perhaps it should be a separate error? Or perhaps we should keep it like it was (type Err = String)?

kierendavies:
Thanks @oddgrd, that makes sense.

Using `String` as an error type is not great, because then consumers can't use the usual Rust mechanisms to inspect the error (and it wastes a few bytes of memory :stuck_out_tongue:). You're right that there ideally should be a separate error type, but the common models need a bit of a rework before it can be done sensibly. So for now I'm kinda hacking it and making it unit.

deployer/src/persistence/resource/database.rs
```diff
@@ -88,6 +88,7 @@ impl Display for Type {
     }
 }
 
+// TODO: Unify with shuttle_common::resource::Type
 impl FromStr for Type {
```
chesedo, line 92: I think this `FromStr` can now be removed right?

deployer/src/persistence/resource/mod.rs
```diff
@@ -81,18 +87,9 @@ impl FromStr for Type {
...
-            match s {
-                "secrets" => Ok(Self::Secrets),
-                "static_folder" => Ok(Self::StaticFolder),
-                "persist" => Ok(Self::Persist),
-                _ => Err(format!("'{s}' is an unknown resource type")),
-            }
+        match shuttle_common::resource::Type::from_str(s) {
+            Ok(t) => Ok(t.into()),
+            Err(e) => Err(e.to_string()),
         }
```
chesedo, line 93: Nice :star_struck:

common/src/lib.rs
```diff
@@ -195,6 +195,13 @@ impl SecretStore {
     }
 }
 
+pub(crate) fn split_first_component(s: &str) -> (&str, Option<&str>) {
+    match s.split_once("::") {
+        Some((first, rest)) => (first, Some(rest)),
+        None => (s, None),
+    }
+}
```
chesedo, line 203: Nice helper here too!

deployer/src/handlers/mod.rs
```diff
@@ -287,6 +296,54 @@ pub async fn get_service_resources(
...
+        // TODO: replace persistence::ResourceType with shuttle_common::resource::Type
+        // or implement From<&shuttle_common::resource::Type> for persistence::ResourceType
+        .get_resource(&service.id, resource_type.clone().into())
+        .await?
+        .ok_or_else(|| Error::NotFound("resource not found".to_string()))?;
+
+    resource_manager
+        .delete_resource(&project_name, &resource_type)
+        .await
+        .map_err(|e| Error::Custom(e.into()))?;
```
chesedo, line 342: Does this also eventually delete a resource in persistence?

chesedo (commented):
Looking good, thanks @kierendavies. I left a few comments

common/src/resource.rs
```diff
@@ -61,3 +61,31 @@ impl Display for Type {
         }
     }
 }
+
+// TODO: Unify with shuttle_deployer::persistence::resource::Type
```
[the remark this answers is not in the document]
kierendavies: Good catch. It's not.

deployer/src/persistence/resource/database.rs
kierendavies: Indeed

deployer/src/handlers/mod.rs
kierendavies: OMG how did I miss that :facepalm:. Fixing...

oddgrd (approved):
Apologies for the delayed review Kieren.This looks great to me! :heart: I just have two questions, not blocking:
- Should we make the CLI command interactive and prompt for confirmation, since this is destructive? We'll also have some other routes that may warrant confirmation soon, so we could also add something that works for all of them in a follow-up issue/PR.
- I know the deletion logic in provisioner was written before this PR, just curious if you tested the deletions while working on this?

kierendavies:
1. To be quite honest, I think the fundamental workflow here is wrong, and the "right" way looks closer to [how reconciliation works in Kubernetes](https://speakerdeck.com/thockin/kubernetes-what-is-reconciliation), but with a retention period before automatic deletion - and then `resource delete` just immediately ends the retention period. But that would need pretty deep changes to track exactly what resources were declared by every deployment. Right now, it's outright impossible, because you can't even know what resources are declared by a deployment until you successfully build and run it.

    Still, you're right, a prompt is probably a good idea. If other commands need it to, I'm in favour of leaving it out of scope of this PR.

2. I was naughty. The otel collector config broke all the locally-running services (I forgot how I fixed it last time), and I gave up and pushed before I successfully tested my changes. I just got it running now, and found a bug. I could use some help fixing it.

oddgrd:
1. Thanks for the link, we're planning to use Kubernetes and I have a lot to learn about it so that's interesting! I think a retention period is a great idea and something we'll need to have in the future. About tracking resources, Pieter actually just submitted a PR for this on our feature branch for a big horizontal scaling refactor we're working on here: https://github.com/shuttle-hq/shuttle/pull/906 (he also made a diagram for the new architecture, we're going to put it in the contribution docs when we update them).

2. Haha, no worries! I'll take a look at that tomorrow.

oddgrd:
On 2., could you elaborate a bit on the bug you found that caused this error? Which command did you try to run?

Also does the otel collector break locally running services for you? For me it just logs some errors, but I've just ignored them. We should of course disable it for local development like we do for panamax, but it should be fine to ignore.

jonaro00:
We are soon adding this feature, but a lot has changed since this PR. We might re-use parts of it. Thanks for the good work.

PR #770, opened 2023-03-29
tests: make provisioner test deterministic

chesedo:
## Description of change
The first time this test would run, it would output nothing since the DB did not exist (and therefore the user did not exist either). This fixes it so that the first run is the same as the subsequent runs by making sure the DB exists.

By running the tests locally

### DIFF
--- a/provisioner/tests/provisioner.rs
+++ b/provisioner/tests/provisioner.rs
@@ -167,8 +167,19 @@ async fn shared_mongodb_role_does_not_exist() {
     .await
     .unwrap();
 
+    exec_mongosh(
+        r#"db.createUser({
+            user: "user-exist",
+            pwd: "secure_password",
+            roles: [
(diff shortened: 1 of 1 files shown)


provisioner/tests/helpers.rs
```diff
@@ -51,6 +51,10 @@ impl DockerInstance {
 
         Self::wait_ready(Duration::from_secs(120), &is_ready_cmd);
 
+        // Mongo starts up twice. So wait for the first one to finish
+        sleep(Duration::from_millis(350));
+        Self::wait_ready(Duration::from_secs(120), &is_ready_cmd);
+
```
oddgrd, line 57: If it only starts twice for mongo, should check the db type before this second wait?

oddgrd (approved):
Nice, thanks!

PR #771, opened 2023-03-29
feat: Rename `project new/rm` to `start/stop`, add `restart` + other args fixes

jonaro00:
## Description of change

- Capitalize and clarify all argument descriptions
- Rename `new/rm` to `start/stop`: *Why? The "rm" implies that something is deleted, which confuses some*. Other suggestions for names are accepted. *"create/destroy" instead? recreate doesn't sound as good as restart :/*
- Add `cargo shuttle project restart` (stop+start), since this is and has been a very commonly recommended operation (We will save so many keystrokes in Discord 🤠)
- Let `cargo shuttle project list` be run whether in a Cargo project or not

Tested against production. The restart command works as expected from both running and stopped states.


cargo-shuttle/src/lib.rs
```diff
@@ -68,7 +68,13 @@ impl Shuttle {
             Command::Deploy(..)
                 | Command::Deployment(..)
                 | Command::Resource(..)
-                | Command::Project(..)
+                | Command::Project(
+                    // ProjectCommand::List does not need to know which project we are in
```
chesedo, line 72: Nice one :heart_eyes:

cargo-shuttle/src/lib.rs
```diff
@@ -366,7 +375,7 @@ impl Shuttle {
             if let Some(deployment) = summary.deployment {
                 deployment.id
             } else {
-                return Err(anyhow!("could not automatically find a running deployment for '{}'. Try passing a deployment ID manually", self.ctx.project_name()));
+                bail!("Could not automatically find a running deployment for '{}'. Try passing a deployment ID manually", self.ctx.project_name());
```
chesedo, line 378: I can't quite remember, but doesn't `return Err(...)` make the format of the error better by catching the `Err` somewhere?

chesedo (commented):
Thanks for these @jonaro00! :partying_face: I just have one comment about the use of `bail!`

cargo-shuttle/src/lib.rs
jonaro00: From https://docs.rs/anyhow/latest/anyhow/macro.bail.html

> This macro is equivalent to `return Err(anyhow!($args...))`.

I'd be surprised if that actually affected formatting 🧐

chesedo (approved):

oddgrd:
Terrific, thanks @jonaro00! :heart:

PR #772, opened 2023-03-30
feat: start all services in a workspace for local runs

oddgrd:
## Description of change

This makes it so all services in a workspace are started for local runs. This required making some changes to build_crate and refactoring the runtime enum so we can get the working directory of each crate, as well as the service name.

TODO:

- [x] properly get the service_name for each crate (check for shuttle.toml before using package name)
- [ ] consider making abstractions for the runtime that would work for both deployer and local run, perhaps a trait with a start method? 

Tested by running updated tests and local runs of the `service/tests/resources/workspace` project.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -457,8 +457,6 @@ impl Shuttle {
 
         let runtimes = build_workspace(working_directory, run_args.release, tx).await?;
 
-        let mut service_count = 0;
-
         let mut runtime_handles = JoinSet::new();
 
         for Runtime {
@@ -468,8 +466,6 @@ impl Shuttle {
             working_directory,
         } in runtimes
         {
-            service_count += 1;
-
             trace!("loading secrets");
 
             let secrets_path = if working_directory.join("Secrets.dev.toml").exists() {
@@ -497,7 +493,7 @@ impl Shuttle {
                         .expect("failed to find cargo home dir")
                         .join("bin/shuttle-next");
 
-                    println!("Installing shuttle runtime. This can take a while...");
+                    println!("Installing shuttle-next runtime. This can take a while...");
 
                     if cfg!(debug_assertions) {
                         // Canonicalized path to shuttle-runtime for dev to work on windows
@@ -548,7 +544,7 @@ impl Shuttle {
                 runtime::StorageManagerType::WorkingDir(working_directory.to_path_buf()),
                 &format!("http://localhost:{}", run_args.port + 1),
                 None,
-                run_args.port + 2 + service_count, // TODO: use portpicker?
+                portpicker::pick_unused_port().unwrap(),
                 runtime_path,
             )
             .await
@@ -657,7 +653,7 @@ impl Shuttle {
             println!(
                 "a service future completed with exit status: {:?}",
                 res.unwrap().unwrap().code()
-            )
+            );
         }
 
         Ok(())
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -38,7 +38,7 @@ use git2::{Repository, StatusOptions};
 use ignore::overrides::OverrideBuilder;
 use ignore::WalkBuilder;
 use shuttle_common::models::{project, secret};
-use shuttle_service::builder::{build_workspace, Runtime};
+use shuttle_service::builder::{build_workspace, BuiltService};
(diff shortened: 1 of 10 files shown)


service/tests/resources/workspace/Shuttle.toml
```diff
@@ -0,0 +1 @@
+name = "workspace-test"
```
oddgrd, line 1: Without this a local run of a workspace project will fail, since it will not be able to find a project name anywhere else. I can improve the error message for this as well.

cargo-shuttle/src/lib.rs
```diff
@@ -448,197 +449,212 @@ impl Shuttle {
...
 
-        trace!(?start_request, "starting service");
-        let response = runtime_client
-            .start(tonic::Request::new(start_request))
-            .or_else(|err| async {
-                provisioner_server.abort();
-                runtime.kill().await?;
+            let addr = SocketAddr::new(
+                addr,
+                portpicker::pick_unused_port().expect("unable to find available port"),
```
oddgrd, line 619: I opted for using portpicker here, but then I am ignoring the port arg the user passed in to start the service. But they can only pass in one port, and just using that +=1 for each service seems less reliable. I'll think about this one.

jonaro00:
Feels like the idea of starting everything at once will turn into chaos one way or another (imagine the combined log output of just 3 services 🤯). I'm more fond of the idea of specifying crate `-p abc` to run one service, and just use multiple terminals (at least this feels like an easier starting point). A `--all` option might work (mainly for deploy I suppose), but also for deploying I think defaulting to one at a time is the better way.

jonaro00:
> If you just want to start a single service, that can be done in the same way it currently is, one service per project.

This works assuming you isolate them in their own freestanding crates. If they are connected in a workspace with a virtual manifest, you would want to be able to run any one of them individually from the workspace root with `-p service1`.
In this scenario, the services don't share any shuttle resource directly, but are meant to be deployable in parallel and use each other's APIs for example, and take advantage of being in a shared workspace.

Correct me if I am misunderstanding your scenarios.

cargo-shuttle/src/lib.rs
chesedo: Yip, once we have the router, that will again take care of only the one port the user wants to bind to. So +=1 will only be a temp thing if we go with that.

service/src/builder.rs
```diff
@@ -75,7 +130,15 @@ pub async fn build_workspace(
...
             .binaries
             .iter()
-            .map(|binary| Runtime::Alpha(binary.path.clone()))
+            .map(|binary| {
+                BuiltService::new(
+                    binary.path.clone(),
+                    false,
+                    binary.unit.pkg.name().to_string(),
+                    binary.unit.pkg.root().to_path_buf(),
+                    binary.unit.pkg.manifest_path().to_path_buf(),
```
chesedo, line 139: Maybe we should store the entire `binary` or `binary.unit` and thereby have access to anything in the future?

service/tests/resources/workspace/Shuttle.toml
chesedo: This is fixed in #774

chesedo (approved):
Looking good, thanks O! :heart: 

In terms of making it generic, feel free to do that in the deployer PR. We also don't have to solve the case where project `a` requires project `b` and therefore we have to start `b` before `a` now.

chesedo:
> Feels like the idea of starting everything at once will turn into chaos one way or another (imagine the combined log output of just 3 services exploding_head)

Fair point, but docker compose does the same thing and that has never really bothered me personally. I can see us having something like your suggested `-p` option to start a single service (and all its dependencies) in the future. That's just not part of the plan right now since no one has 20 services in their project today. And the default seems best to start them all at this time.

I like your concerns here however - they keep us on our toes to not create more problems in the future.

service/src/builder.rs
oddgrd: Sure! I didn't do it because I was trying to be less dependent on `cargo` since we're planning to remove it eventually, but that will require a bit of work here anyway to get all the required data.

oddgrd: Ah, Unit is also !Send and !Sync :thinking:

oddgrd: We'd have to put in a mutex or something, so I may just stick with the current implementation now and see if I can find a smarter way when working on deployer.

PR #774, opened 2023-03-31
refactor: getting project name

chesedo:
## Description of change
This PR improves how we get the name of the project and how we identify the root `Cargo.toml` file in a project.

This is the preferred order for a project name:
1. Name given on command line                           
2. Name from Shuttle.toml file                          
3. Name from Cargo.toml package if it's a crate         
3. Name from the workspace directory if it's a workspace

By running the tests locally and adding new tests. Note, the tests require shuttle-hq/examples#35

### DIFF
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -5,6 +5,8 @@ use std::{
     path::PathBuf,
 };
 
+use anyhow::Context;
+use cargo_metadata::MetadataCommand;
 use clap::builder::{OsStringValueParser, PossibleValue, TypedValueParser};
 use clap::Parser;
 use clap_complete::Shell;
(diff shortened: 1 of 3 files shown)


cargo-shuttle/src/args.rs
```diff
@@ -46,6 +48,41 @@ pub struct ProjectArgs {
...
+            .unwrap();
+        let package_name = if let Some(root_package) = meta.root_package() {
+            root_package.name.clone().parse()?
+        } else {
+            workspace_path
+                .file_name()
+                .context("failed to get project name from workspace path")?
+                .to_os_string()
+                .into_string()
+                .expect("to turn workspace name into path")
```
oddgrd, line 78: Should this be something like this?
```suggestion
                .expect("workspace file name should be valid unicode")
```

cargo-shuttle/src/lib.rs
```diff
@@ -1004,10 +991,12 @@ mod tests {
     use std::path::PathBuf;
     use std::str::FromStr;
 
-    fn path_from_workspace_root(path: &str) -> PathBuf {
+    pub fn path_from_workspace_root(path: &str) -> PathBuf {
         PathBuf::from(std::env::var("CARGO_MANIFEST_DIR").unwrap())
             .join("..")
             .join(path)
+            .canonicalize()
```
oddgrd, line 998: I'm not certain this would be an issue, but should we use `dunce::canonicalize` to be safe on windows?

oddgrd (approved):
LGTM! Good idea getting the project name from the workspace file_name, I'll add a note about updating the docs with this as well.

PR #780, opened 2023-04-03
refactor: get the static folder name from the

chesedo:
## Description of change
This change causes `ResourceBuilder`s to output the build config they were constructed with. This allows for richer info when building the resource tables.

Using the local runner


oddgrd (approved):
Looks great to me. :partying_face: Thanks for the docs and comment updates in `service` as well!

PR #781, opened 2023-04-04
Feature/eng 563 start all services in a project deployer

oddgrd:
## Description of change

This PR will enable deploying all services in a workspace.

TODO


oddgrd:
This does not make sense to do until we implement a router service for projects with multiple services, since each project is currently assigned one address, when starting multiple services we won't be able to route traffic to them.

Issue #775, opened 2023-03-31
[Bug]: stripping repository path fails on Windows


jonaro00:
### What happened?

Original from this thread: https://discord.com/channels/803236282088161321/1087505717776359586

```
Error: stripping repository path from working directory

Caused by:
    prefix not found
```

The `Path::strip_prefix` in `cargo-shuttle/src/lib.rs -> Shuttle::is_dirty` method seems to fail on Windows.
There are also other calls to that method which might need to be double checked.

### Version

v0.13.0

### Which operating systems are you seeing the problem on?

Windows

### Which CPU architectures are you seeing the problem on?

x86_64

### Duplicate declaration

- [X] I have searched the issues and there are none like this.

b1scoito:
Getting the same error, coincidentally, has just been fixed!

Waiting for the next release so I can test shuttle :D

PR #783, opened 2023-04-05
fix: is_dirty path bug on windows

oddgrd:
## Description of change

This fixes the bug where the deploy command would error if the working tree is dirty, rather than display the warning about the working tree being dirty and the suggestion to us --allow-dirty.

Fixes https://github.com/shuttle-hq/shuttle/issues/775

Tested by deploying to unstable (on windows).


chesedo (approved):
Thanks O

PR #787, opened 2023-04-11
fix: revert use of portpicker for local run

oddgrd:
## Description of change

Revert the use of portpicker to get the port to start a service on for local runs.

Local runs.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -436,9 +436,11 @@ impl Shuttle {
         );
 
         let provisioner = LocalProvisioner::new()?;
+        let provisioner_port =
+            portpicker::pick_unused_port().expect("unable to find available port");
         let provisioner_server = provisioner.start(SocketAddr::new(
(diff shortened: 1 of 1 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -530,9 +532,9 @@ impl Shuttle {
             let (mut runtime, mut runtime_client) = runtime::start(
                 is_wasm,
                 runtime::StorageManagerType::WorkingDir(working_directory.to_path_buf()),
-                &format!("http://localhost:{}", run_args.port + 1),
+                &format!("http://localhost:{provisioner_port}"),
                 None,
-                portpicker::pick_unused_port().expect("unable to find available port"),
+                run_args.port + 1 + i as u16,
```
chesedo, line 537:
```suggestion
                run_args.port - (1 + i) as u16,
```
Not sure if this is correct

chesedo (approved):
LGTM

PR #788, opened 2023-04-12
chore: v0.14.0

oddgrd:
## Description of change

Bump version to 0.14 ahead of release, and update examples submodule.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -885,10 +885,7 @@ impl Shuttle {
 
         let secrets_path = self.ctx.working_directory().join("Secrets.toml");
         if secrets_path.exists() {
-            entries.insert(
-                secrets_path.clone(),
-                Path::new("shuttle").join("Secrets.toml"),
-            );
+            entries.insert(secrets_path, Path::new("shuttle").join("Secrets.toml"));
         }
 
         // Append all the entries to the archive.
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -855,18 +855,18 @@ impl Shuttle {
         {
             let dir_entry = dir_entry.context("get directory entry")?;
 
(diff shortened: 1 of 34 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -855,18 +855,15 @@ impl Shuttle {
...
-                    let path = secrets_path
-                        .strip_prefix(base_directory)
-                        .context("strip the base of the archive entry")?;
-                    tar.append_path_with_name(secrets_path.clone(), path)?;
-                }
+            let secrets_path = dir_entry.path().join("Secrets.toml");
+
+            // Make sure to add any `Secrets.toml` files.
+            if secrets_path.exists() {
+                tar.append_path_with_name(secrets_path, Path::new("shuttle").join("Secrets.toml"))?;
```
chesedo, line 862: Won't this override the same `Secret.toml` the whole time? I guess it's also fine since we are not supporting multiple services yet

chesedo (approved):
LGTM

cargo-shuttle/src/lib.rs
```diff
@@ -879,12 +876,6 @@ impl Shuttle {
                 .context("archive entry")?;
         }
 
-        // Make sure to add any `Secrets.toml` files in the root of the workspace.
-        let secrets_path = self.ctx.working_directory().join("Secrets.toml");
-        if secrets_path.exists() {
-            tar.append_path_with_name(secrets_path, Path::new("shuttle").join("Secrets.toml"))?;
-        }
```
chesedo, line 886: This might be needed for non-workspace projects

oddgrd: Yes, I think so, the problem is it is causing duplicate entries: https://app.circleci.com/pipelines/github/shuttle-hq/shuttle/2287/workflows/ef63bbba-4f8f-4f8e-9871-df8cafb33b63/jobs/41573 :thinking:

Issue #779, opened 2023-04-03
[Bug]: Cargo shuttle `Error: failed to make get request`


oxabz:
### What happened?

When I try to use any command of cargo shuttle that tries to access shuttle, I get an error.

I took a look at wireshark and it doesn't even get to the point of sending an HTTP request. So the error might be in cargo shuttle.

### Version

v0.13.0

### Which operating systems are you seeing the problem on?

Linux

### Which CPU architectures are you seeing the problem on?

x86_64

### Relevant log output

```sh
Error: failed to make get request

Caused by:
    0: Request error: builder error: failed to parse header value
    1: builder error: failed to parse header value
    2: failed to parse header value
```

### Duplicate declaration

- [X] I have searched the issues and there are none like this.

oxabz:
Okay my bad (sort of). I seem to have used `ctrl+p` instead of `ctrl+shift+p` to paste my API key while logging in. 
After login in again correctly, the error is gone. 

I'm still leaving the bug open because it might be useful to add a check on the input of `cargo shuttle login`  or a more expressive error message in a future version.

oddgrd:
Yeah, we should look into improving this error message. Thanks @oxabz!

PR #797, opened 2023-04-14
feat: ensure API key is valid

oddgrd:
## Description of change

This creates a newtype for the api-key on the client side, so we can maintain that it is always a valid api-key. We should extend this type to also be usable in the `auth` crate in the future.

Closes https://github.com/shuttle-hq/shuttle/issues/779

Added tests for the `parse` function, and tested variations of login and other commands with and without the `SHUTTLE_API_KEY` env against unstable.

### DIFF
--- a/cargo-shuttle/Cargo.toml
+++ b/cargo-shuttle/Cargo.toml
@@ -18,7 +18,7 @@ clap = { workspace = true, features = ["env"] }
 clap_complete = "4.1.5"
 crossbeam-channel = { workspace = true }
 crossterm = { workspace = true }
-dialoguer = { version = "0.10.3", features = ["fuzzy-select"] }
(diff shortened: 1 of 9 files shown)


oddgrd:
On second thought, this will break the flow of setting a jwt as apikey for local deployer runs. I may scope this down to validating the login input. 🤔

cargo-shuttle/Cargo.toml
```diff
@@ -18,7 +18,7 @@ clap = { workspace = true, features = ["env"] }
 clap_complete = "4.1.5"
 crossbeam-channel = { workspace = true }
 crossterm = { workspace = true }
-dialoguer = { version = "0.10.3", features = ["fuzzy-select"] }
+dialoguer = { version = "0.10.4", features = ["fuzzy-select"] }
```
oddgrd, line 21: This is to get the `validate_with` method.

PR #799, opened 2023-04-15
feat: Add `cargo shuttle logs --latest`

### LINKED ISSUE #743: [Feature]: Show latest log when no deployment is active
### Describe the feature

Currently, when a deployment is not active:

```
$ cargo shuttle logs
Error: could not automatically find a running deployment for 'jonaro00-test'. Try passing a deployment ID manually
```

Instead of having to copy+paste the latest deployment id, simply show the log of the latest deployment. If no deployments, then show the error.

Alternatively, add `cargo shuttle logs --latest`.

_No response_

### Duplicate declaration

jonaro00:
## Description of change

Closes #743 
/claim #743

Adds the described functionality.

Tested against my projects in production. Works as expected.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -360,17 +360,18 @@ impl Shuttle {
             if latest {
                 // Find latest deployment (not always an active one)
                 let deployments = client.get_deployments(proj_name).await?;
-                if deployments.is_empty() {
-                    bail!("Could not find any deployments for '{}'. Try passing a deployment ID manually", proj_name);
-                }
-                let most_recent = deployments.last().unwrap(); // length 0 is checked above
+                let most_recent = deployments.last().context(format!(
+                    "Could not find any deployments for '{proj_name}'. Try passing a deployment ID manually",
+                ))?;
 
                 most_recent.id
             } else if let Some(deployment) = client.get_service(proj_name).await?.deployment {
                 // Active deployment
                 deployment.id
             } else {
-                bail!("Could not find a running deployment for '{}'. Try with '--latest', or pass a deployment ID manually", proj_name);
+                bail!(
+                    "Could not find a running deployment for '{proj_name}'. Try with '--latest', or pass a deployment ID manually"
+                );
             }
         };
 
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -364,7 +364,7 @@ impl Shuttle {
             } else if latest {
                 // Find latest non-active deployment
                 let deployments = client.get_deployments(proj_name).await?;
-                if deployments.len() == 0 {
+                if deployments.is_empty() {
                     bail!("Could not find any deployments for '{}'. Try passing a deployment ID manually", proj_name);
                 }
                 let most_recent = deployments
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -98,7 +98,9 @@ impl Shuttle {
                 return self.deploy(deploy_args, &self.client()?).await;
             }
             Command::Status => self.status(&self.client()?).await,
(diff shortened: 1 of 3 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -343,16 +345,42 @@ impl Shuttle {
...
             id
         } else {
-            let summary = client.get_service(self.ctx.project_name()).await?;
+            let proj_name = self.ctx.project_name();
+            let summary = client.get_service(proj_name).await?;
 
             if let Some(deployment) = summary.deployment {
+                // Active deployment
                 deployment.id
+            } else if latest {
```
oddgrd, line 364: Should we perhaps always get the latest if the latest flag is passed in? So with no flag we try to get the current running one, with the flag we try to get the latest one. Because a running deployment might not always be the last deployment.

oddgrd (commented):
Thanks @jonaro00! I added a couple of comments/questions, but this looks good!

cargo-shuttle/src/lib.rs
```diff
@@ -343,16 +345,42 @@ impl Shuttle {
...
+                let most_recent = deployments
+                    .into_iter()
+                    .reduce(|acc, dep| {
+                        if acc.last_update > dep.last_update {
+                            acc
+                        } else {
+                            dep
+                        }
+                    })
+                    .unwrap(); // length 0 is checked above
```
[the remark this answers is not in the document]
jonaro00: I've seen the list be out of order, so I wanted to be sure.
image

cargo-shuttle/src/lib.rs
jonaro00: Sounds good

cargo-shuttle/src/lib.rs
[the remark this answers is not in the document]
oddgrd: Ah, I see! Then perhaps the best solution would be to add an `order_by` clause to the `get_deployments` query, this way `deployment --list` should also be fixed.

cargo-shuttle/src/lib.rs
```diff
@@ -343,16 +345,32 @@ impl Shuttle {
...
-            let summary = client.get_service(self.ctx.project_name()).await?;
+            let proj_name = self.ctx.project_name();
+
+            if latest {
+                // Find latest deployment (not always an active one)
+                let deployments = client.get_deployments(proj_name).await?;
+                if deployments.is_empty() {
+                    bail!("Could not find any deployments for '{}'. Try passing a deployment ID manually", proj_name);
+                }
+                let most_recent = deployments.last().unwrap(); // length 0 is checked above
```
oddgrd, line 366: This is a bit of a nitpick since `get_deployments` will almost always 404 if there are no deployments, because if there are no deployments there is most likely no service (although it is possible).
```suggestion
                let most_recent = deployments.last().context(format!(
                    "Could not find any deployments for '{proj_name}'. Try passing a deployment ID manually",
                ))?;
```

deployer/src/persistence/mod.rs
```diff
@@ -604,75 +604,126 @@ mod tests {
...
             address: Some(SocketAddr::new(Ipv4Addr::LOCALHOST.into(), 9876)),
             is_next: false,
         };
         let deployment_queued = Deployment {
-            id: queued_id,
+            id: Uuid::new_v4(),
             service_id,
             state: State::Queued,
-            last_update: Utc::now(),
+            last_update: time.checked_add_signed(Duration::seconds(3)).unwrap(),
```
oddgrd, line 702: The next one here is also 3 seconds, maybe this will fail sporadically? :thinking:

oddgrd (approved):
LGTM @jonaro00!  Thanks for adding a test as well. And congratulations on submitting the first approved bounty PR, *very* well deserved!

PR #801, opened 2023-04-17
ci: add mac qa

chesedo:
## Description of change
QA tests to ensure everything works on a mac machine before a release.

Yip, using the CI


oddgrd:
In case you didn't see this P, might be worth switching to this/having this as well for mac qa: https://circleci.com/blog/m1-mac-resource-class/

PR #805, opened 2023-04-19
cargo-shuttle: fix address in use error when service panicked in a previous run

iulianbarbu:
## Description of change

Running user services locally through `cargo shuttle run` will run a runtime that will load/start a service as a tokio task. If the task panics, the shuttle-runtime will live on because of tokio behavior: https://github.com/tokio-rs/tokio/issues/2002.

This is not the case anymore. I haven't seen this making any difference for our use cases. Removing it to have the `bind_panic` test succeed.
~~Extra: set to true the `kill_on_drop` flag for started shuttle-runtimes to make sure we either `.wait()` or `.kill()` on them or if dropped, the `kill` is automatically issued: https://docs.rs/tokio/latest/tokio/process/struct.Command.html#method.kill_on_drop.~~

Run locally the `hello-world-rocket-app` example with panic in it. Consequent local runs without the fix hang with the `Address in use` error. The fix ensures the runtime and provisioner are killed in case the runtime response is not successful.

Also, if `cargo-shuttle` process receives `SIGINT` or `SIGTERM` it's successfully stopping the existing runtimes.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -657,7 +657,7 @@ impl Shuttle {
             .stop(tonic::Request::new(stop_request))
             .or_else(|err| async {
                 runtime.kill().await?;
-                error!(status = ?err, "killed the runtime by force because stopping it errored out");
+                trace!(status = ?err, "killed the runtime by force because stopping it errored out");
                 Err(err)
             })
             .await?
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -5,14 +5,18 @@ mod init;
 mod provisioner_server;
 
 use indicatif::ProgressBar;
+use shuttle_common::claims::{ClaimService, InjectPropagation};
 use shuttle_common::models::deployment::get_deployments_table;
 use shuttle_common::models::project::{State, IDLE_MINUTES};
 use shuttle_common::models::resource::get_resources_table;
 use shuttle_common::project::ProjectName;
 use shuttle_common::resource;
-use shuttle_proto::runtime::{self, LoadRequest, StartRequest, SubscribeLogsRequest};
+use shuttle_proto::runtime::runtime_client::RuntimeClient;
+use shuttle_proto::runtime::{self, LoadRequest, StartRequest, StopRequest, SubscribeLogsRequest};
 
-use tokio::task::JoinSet;
+use tokio::process::Child;
+use tokio::task::JoinHandle;
+use tonic::transport::Channel;
 
 use std::collections::HashMap;
 use std::ffi::OsString;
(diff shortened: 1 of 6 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -588,6 +586,10 @@ impl Shuttle {
 
             if !response.success {
                 error!(error = response.message, "failed to load your service");
+                // TODO: we must kill the rest of the runtimes when we'll start multiple services from a workspace,
+                // but now this is not concerning given we're not supporting more than one services.
+                provisioner_server.abort();
+                runtime.kill().await?;
```
oddgrd, line 592: For local runs we do actually start all the services we find currently. But perhaps we can just abort all the other handles in the joinset when one fails? :thinking:

oddgrd (commented):
Thanks @iulianbarbu! I'm wondering if this test failure in runtime is spurious or due to the change in proto, did you see it fail locally too?

cargo-shuttle/src/lib.rs
iulianbarbu: Addressed this here: https://github.com/shuttle-hq/shuttle/pull/805/commits/949cccb1d4e6756b29443293237640772d3754a5. Testing works well, all the previous successfully started runtimes are aborted and the processes reaped when a runtime's service panicks, including the current runtime.

iulianbarbu:
> Thanks @iulianbarbu! I'm wondering if this test failure in runtime is spurious or due to the change in proto, did you see it fail locally too?

I am curious if they will fail again now. I am running them now locally as well.

iulianbarbu:
It seems like the `cargo-shuttle` and `shuttle-runtime` tests fail again on the CI. 

However, on my machine, the `cargo-shuttle` related one passes. It's worth mentioning that running the tests as the CI does ends with the below message for `cargo-shuttle`.  However, I don't see any failed tests in the output.
```
error: test failed, to rerun pass `-p cargo-shuttle --test integration`

Caused by:
  process didn't exit successfully: `/Users/iulianbarbu/repos/shuttle/target/debug/deps/integration-9fe6044d5f6f7b77 --nocapture` (exit status: 1)
  ```
  
The `shuttle-runtime` `bind_panic` fails it seems because of the `kill_on_drop` flag I added. I'll investigate if removing it is not doing a regression for the fix, because we might not need it.

**Comment by iulianbarbu** [comment]
@oddgrd @chesedo , it seems like the previously failing tests pass now but this time `shuttle-deployer` tests failed. I think this is a spurious one because for the previous commit it was successful and the last commit is only comments related. I think this is mergeable unless improving a bit how the code looks.

File: runtime/src/alpha/mod.rs [hunk 2]
```diff
@@ -315,6 +315,21 @@ where
         tokio::spawn(async move {
             let mut background = handle.spawn(service.bind(service_address));
 
+            let (_sigterm_notif, _sigint_notif) = if cfg!(target_family = "unix") {
```
chesedo, line 318: Why do we have these?

chesedo (commented):
Seems about good to go.

cargo-shuttle/src/lib.rs
```diff
@@ -486,203 +684,173 @@ impl Shuttle {
...
+                        rt.1.stop(tonic::Request::new(stop_request))
+                            .or_else(|err| async {
+                                provisioner_server.abort();
+                                rt.0.kill().await?;
+                                Err(err)
+                            })
+                            .await?
+                            .into_inner();
+                    continue;
                 }
```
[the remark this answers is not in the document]
iulianbarbu: The above handling refers to receiving a signal during the runtimes initializations. If `signal_received` is `true` it should exit the `local_run` returning `Ok(())`. Otherwise, it passes it and moves to the logic of waiting on each `runtime` in a for loop. However, while waiting on these runtimes we must handle signals as well, hence we're checking the `signal_received` before actually waiting on a runtime in a new iteration loop. If it's true, then we've received a signal in a previous iteration and we must stop all the runtimes which we haven't waited on instead on continuing to wait on them.

runtime/src/alpha/mod.rs
iulianbarbu: Leftovers, will remove them.

cargo-shuttle/src/lib.rs
```diff
@@ -486,203 +684,173 @@ impl Shuttle {
...
-                .await?
-                .into_inner();
-
-            if !response.success {
-                error!(error = response.message, "failed to load your service");
-                exit(1);
+        // If prior signal received is set to true we must stop all the existing runtimes and
+        // exit the `local_run`.
+        if signal_received {
+            for mut rt in runtimes {
```
oddgrd, line 753: Perhaps de-structuring here would be cleaner, something like this?
```suggestion
            for mut (runtime, client) in runtimes {
```

oddgrd (approved):
Thanks for figuring this out Iulian, I understand why they have a warning about this being tricky in the tokio::signal docs :smile: I left a few comments but mostly nits, and a note about us considering doing this for windows too in the future.

~~We might also want to test this on windows to see that it doesn't cause any regressions, IIRC importing something that's only available on unix may be an issue. But I can check this quickly tomorrow.~~ Edit: It looks like all uses of unix signals is inside cfg(unix), so this should be fine.

cargo-shuttle/src/lib.rs
iulianbarbu: Done.

cargo-shuttle/src/lib.rs
```diff
@@ -486,203 +684,173 @@ impl Shuttle {
...
+                );
+                let _response =
+                    rt.1.stop(tonic::Request::new(stop_request))
+                        .or_else(|err| async {
+                            provisioner_server.abort();
+                            rt.0.kill().await?;
+                            Err(err)
+                        })
+                        .await?
+                        .into_inner();
```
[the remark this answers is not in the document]
iulianbarbu: Done.

cargo-shuttle/src/lib.rs
```diff
@@ -486,203 +684,173 @@ impl Shuttle {
...
-
-                    Err(err)
-                })
-                .await?
-                .into_inner();
-
-            if !response.success {
-                error!(error = response.message, "failed to load your service");
-                exit(1);
+        // If prior signal received is set to true we must stop all the existing runtimes and
```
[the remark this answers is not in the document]
iulianbarbu: An addendum for the above comment: I had to adjust a bit this logic. Now, if a service panics during initialization then all the workspace runtimes must be killed and `cargo-shuttle` exit with code 1. Because we forcefully exit, the destructors for the runtime are not called and we must forcefully kill the runtime besides suggesting it to stop in a graceful way.

cargo-shuttle/src/lib.rs
```diff
@@ -670,6 +666,34 @@ impl Shuttle {
...
+        )>,
+        provisioner_server: &JoinHandle<Result<(), tonic::transport::Error>>,
+    ) -> Result<(), Status> {
+        match runtime {
+            Some(inner) => existing_runtimes.push(inner),
+            None => {
+                // Stopping all runtimes gracefully and then forcefully because we must follow up with a `std::process:exit`
+                // which doesn't guarantee destructors run, which means dropped runtimes don't result in killed runtimes (per
+                // https://docs.rs/tokio/latest/tokio/process/struct.Child.html#caveats).
+                for rt_info in existing_runtimes {
```
iulianbarbu, line 686: I had to do a mutable borrow here because:
1) We can not do destructuring because we can not destructure and mutable borrow at the same time the values within the tuple, e.g. this breaks compilation:
```rust
for (&mut rt, &mut rt_client) in existing_runtimes {
...
}
```

while simply moving out the tuple contents from a mutable reference (existing runtimes is a `&Vec<(Child, RuntimeClient)>`) breaks the compilation again:
```
for (mut rt, mut rt_client) in existing_runtimes {
...
}
```

cargo-shuttle/src/lib.rs
```diff
@@ -657,6 +657,7 @@ impl Shuttle {
             .stop(tonic::Request::new(stop_request))
             .or_else(|err| async {
                 runtime.kill().await?;
+                error!(status = ?err, "killed the runtime by force because stopping it errored out");
```
oddgrd, line 660: What if we do `trace!` here, then the user won't see an error (which won't be relevant to them in the event of `CTRL C` since as far as they are concerned it was a success, not an error, right?)?

PR #806, opened 2023-04-20
fix: minor development bug fixes

oddgrd:
## Description of change

This fixes some minor development bug fixes, namely:
- When releasing a new version some old projects without `idle_minutes` set will be deserialized without it, leading to panics. Added a serde default to `ProjectCreating` to remedy this.
- Removed the unused shuttle-secrets dependency from cargo-shuttle that was causing the new version of cargo-chef to fail to cook.

The idle-minutes change is hard to test since we're not entirely sure which projects caused it, but we should not see the panics anymore. The cargo-chef fix is tested by running `make images`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -410,7 +410,7 @@ jobs:
   linux-qa:
     machine:
       image: ubuntu-2004:current
-      resource_class: xlarge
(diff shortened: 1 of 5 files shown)


iulianbarbu (approved):
LGTM!

PR #817, opened 2023-04-25
fix(common): feature flagged utoipa dependency

iulianbarbu:
## Description of change

User projects have a dependency on `shuttle-common`, which depends on `utoipa`. Howerver, `utoipa` dependency is not relevant to the user projects so this PR is feature gating the use of `utoipa` dependency and makes it optional when consuming `shuttle-common`.

Created a shuttle-rocket project, built it, and verified that `utoipa` is not appearing as a dependency in the `target` directory.


oddgrd (approved):
Nice, thanks Iulian!

PR #818, opened 2023-04-26
fix: apply admin layer to admin routes only

oddgrd:
## Description of change

The admin layer was being applied to all default routes, with this change it will be applied only to the routes in the nested admin router.

TODO


iulianbarbu (approved):
Thanks @oddgrd for the fix ! LGTM!

PR #823, opened 2023-04-27
cargo-shuttle: separate local runs per target family

iulianbarbu:
## Description of change

Separated `local_runs` per target family. Runtime checking the target family will not compile on all platforms because we use platform specific APIs which can not be compiled on all platforms.

Same as for #805 .


oddgrd (approved):
LGTM, thanks!

PR #832, opened 2023-04-27
fix: Remove unused project list filtering

jonaro00:
## Description of change

I noticed that this endpoint is not active, so I clarify that to the user.

### DIFF
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -176,11 +176,7 @@ pub enum ProjectCommand {
         idle_minutes: u64,
     },
     /// List all projects belonging to the calling account
-    List {
-        #[arg(long)]
-        /// Return projects filtered by a given project status
-        filter: Option<String>,
-    },
+    List,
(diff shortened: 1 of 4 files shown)


oddgrd (commented):
Good catch! This has never worked in a released version (we forgot to remove it client-side :see_no_evil:), so while it is technically a breaking change to remove it, arg and all, no-one will be depending on it. So I think we can just remove it for now.

We are planning to have projects for this in the batch as well, pagination and/or filtering of some of the GET-many endpoints, but it may be a little bit before we get started on that.

jonaro00:
Since the boilerplate is already there, the commented code could be kept until the feature is picked up again... 🤔 
But sure, I can just remove it. Your call.

oddgrd:
I'm a bit torn, but I say lets just get rid of it for now when you get a chance, there is some boilerplate but not much and it may change in the future as well.

oddgrd (approved):
Thanks! :heart:

PR #835, opened 2023-05-01
feat: ApiKey newtype to ensure key is always valid format

oddgrd:
## Description of change

Implements an `ApiKey` newtype in common that maintains the invariant that a key must be a 16 char alphanumeric string.

This breaks the local deployer flow until #810 is merged, so it should not be merged before that.

Tested with local setup, setting the key with login, env or manipulating config file directly.


cargo-shuttle/src/config.rs
```diff
@@ -122,17 +122,17 @@ impl ConfigManager for LocalConfigManager {
 /// Global client config for things like API keys.
 #[derive(Deserialize, Serialize, Default)]
 pub struct GlobalConfig {
-    pub api_key: Option<ApiKey>,
+    api_key: Option<String>,
```
iulianbarbu, line 125: Why is this needed?

cargo-shuttle/src/config.rs
```diff
@@ -122,17 +122,17 @@ impl ConfigManager for LocalConfigManager {
...
     pub api_url: Option<ApiUrl>,
 }
 
 impl GlobalConfig {
-    pub fn api_key(&self) -> Option<&ApiKey> {
-        self.api_key.as_ref()
+    pub fn api_key(&self) -> Option<Result<ApiKey>> {
+        self.api_key.as_ref().map(|key| ApiKey::parse(key))
```
iulianbarbu, line 131: I am just thinking that would've been best to avoid the parsing each time we call `api_key`, but only if not needed.

common/Cargo.toml
```diff
@@ -7,7 +7,7 @@ repository.workspace = true
 description = "Common library for the shuttle platform (https://www.shuttle.rs/)"
 
 [dependencies]
-anyhow = { workspace = true, optional = true }
+anyhow = { workspace = true }
```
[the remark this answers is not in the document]
iulianbarbu: I am not sure though why we need it in this PR.

common/src/lib.rs
```diff
@@ -21,27 +21,67 @@ pub mod wasm;
...
-pub type ApiKey = String;
 pub type ApiUrl = String;
 pub type Host = String;
 #[cfg(feature = "service")]
 pub type DeploymentId = Uuid;
 
+#[derive(Clone, Serialize, Deserialize, Debug)]
+#[cfg_attr(feature = "persist", derive(sqlx::Type, PartialEq, Hash, Eq))]
+#[cfg_attr(feature = "persist", serde(transparent))]
+#[cfg_attr(feature = "persist", sqlx(transparent))]
```
iulianbarbu, line 47: A naming question, haha: why is the feature called `persist`? Maybe because this `ApiKey` will persist in the `.config` dir?

cargo-shuttle/src/config.rs
oddgrd: My thinking here is that since the API-key will be stored as a string in the config file, and this can be freely modified manually, every time we read it from the config it needs to be parsed. There may well be a better way to do this though :thinking:

common/Cargo.toml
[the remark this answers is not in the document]
oddgrd: It's because I use it for the `ApiKey::parse` error, and it's just a bit simpler. I am still a bit torn on if we should do this though

cargo-shuttle/src/config.rs
oddgrd: So the ApiKey was just a type alias for `String`. By also making it not pub you can't set it (unless you manually edit the config file) without using the `set_api_key` method, which takes an `ApiKey` (the newtype that can only be created by `ApiKey::parse`, so it will always be a valid format).

common/src/lib.rs
oddgrd: Good question! :smile: I struggled a bit with this name, it's persist now because it enables sqlx, pretty much. We might use it elsewhere in common as well in the future. Very open for naming suggestions here :joy:

oddgrd: We might also want to not derive debug for this, so it can't be unintentionally logged :thinking:

oddgrd: That's a good point, I believe this is what the secrecy crate does as well. Perhaps we should use that here.

PR #850, opened 2023-05-03
[orchestrator] refactor: prepare feature branch

chesedo:
## Description of change
Prepare the feature branch for the new deployer crate.

Using `cargo check`


oddgrd (approved):
Thanks!

iulianbarbu (commented):
I am wondering if introducing the new deployer as `deployer-beta` or `deployer-next` would be better suited rather than modifying the current deployer into `deployer-alpha`. This is just because if we merge this the documentation still points to shuttle-deployer. Either this or we we update the docs as well?

PR #852, opened 2023-05-03
build(docker): Change default provisioner port to 3000

### LINKED ISSUE #851: [Improvement]: Change default Provisioner port
On MacOs, the AirPlay receiver is listening on localhost:5000 by default, which leads to Mac users not being able to run e.g. `cargo shuttle run` as described in CONTRIBUTING.md and may take some time to figure out. Let's change the port we are using to 3000 to avoid complications for Mac users.

### Duplicate declaration

morlinbrot:
Closes #851

On MacOs, port 5000 is used by AirPlay receiver which will make the 'make up' command described in CONTRIBUTING.md fail for Mac users

## Description of change

All occurrences of using port 5000 for the provisioner were changed to use port 3000 instead.

Tested by activating the AirPort receiver (on port 5000) and running 'make up' - works. `docker ps` shows that the provisioner is running on 3000.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -181,7 +181,7 @@ cargo shuttle login --api-key test-key
 We're now ready to start a local run of the deployer:
 
 ```bash
-cargo run -p shuttle-deployer -- --provisioner-address http://localhost:5000 --auth-uri http://localhost:8008 --proxy-fqdn local.rs --admin-secret test-key --local --project <project_name>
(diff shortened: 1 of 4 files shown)


oddgrd (approved):
Thanks, LGTM! :partying_face: We'll also need to change the port here (this is on the latest main version of contributing.md): https://github.com/shuttle-hq/shuttle/blob/d81c201952bfbbeaad9b021166a4a392a1936392/CONTRIBUTING.md?plain=1#L184

morlinbrot:
I've rebased my branch on the newest main and added the requested change. I amended the original commit (thought that was cleaner) so there's no new commit to see but the original hash has changed.

Issue #846, opened 2023-05-02
[Improvement]: Improve errors in the event of failed deployment


forrestpatwalker:
### Description
When the logs stream in the cargo-shuttle::Shuttle::deploy method is broken, and the reason is not that a log_item with the Running state was returned, we will just print the Deployment has not entered the running state to the user and return. This is because after the logs stream stops, we check if the deployment we tried to start is the currently running deployment for this project.

We should improve this by re-connecting the logs stream if the deployment is in a state that makes sense to retry, like if the stream ended abruptly but the deployment is still building. In the case of non-retryable states like Stopped and Completed we could expand the error message to reflect these states.

There are also cases where the deployment crashes, and this will be apparent in the logs but we will not get the expected Deployment crashed output in the cargo-shuttle output. We should figure out why this is, and try to ensure that the reason for a crash (or at least what information we currently have) should always be returned.

### Requirements
- Implement some retry-logic for the websocket stream in cargo-shuttle::Shuttle::deploy
  - This is basically a loop which subscribes the logs and then gets the deployment status (new call to deployer) and decide passed on the deployment status if the loop should end.
- Ensure the reason for a crash is always included in the cargo-shuttle cli output, not just Deployment has not entered the running state
- Improve error message in the even of Completed/Stopped states, can we also have the reason here?

### Duplicate declaration

- [X] I have searched the issues and this feature has not been requested before.

PR #853, opened 2023-05-03
feat(cargo-shuttle): log reconnects and improved error messages

iamwacko:
## Description of change

Partially addresses #846, we still need to reproduce getting the incorrect output from cargo-shuttle. The main change is that the websockets client will try to reconnect if disconnected in the queuing, building, or loading phase of the deploy process. I also changed the error outputs for the running, completed, stopped, and unknown states.

`cargo fmt` and `cargo clippy` were run. I could try to reproduce a web-sockets disconnect, but I think that would be unnecessary.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -902,33 +902,50 @@ impl Shuttle {
             .get_logs_ws(self.ctx.project_name(), &deployment.id)
             .await?;
 
-        while let Some(Ok(msg)) = stream.next().await {
-            if let tokio_tungstenite::tungstenite::Message::Text(line) = msg {
-                let log_item: shuttle_common::LogItem =
-                    serde_json::from_str(&line).expect("to parse log line");
+        let mut last_state: shuttle_common::deployment::State = shuttle_common::deployment::State::Unknown;
 
-                match log_item.state {
-                    shuttle_common::deployment::State::Queued
-                    | shuttle_common::deployment::State::Building
-                    | shuttle_common::deployment::State::Built
-                    | shuttle_common::deployment::State::Loading => {
-                        println!("{log_item}");
-                    }
-                    shuttle_common::deployment::State::Crashed => {
-                        println!();
-                        println!("{}", "Deployment crashed".red());
-                        println!("Run the following for more details");
-                        println!();
-                        print!("cargo shuttle logs {}", deployment.id);
-                        println!();
+        loop {
+            let mesg = stream.next().await;
+            if let Some(Ok(msg)) = mesg {
+                if let tokio_tungstenite::tungstenite::Message::Text(line) = msg {
+                    let log_item: shuttle_common::LogItem =
+                        serde_json::from_str(&line).expect("to parse log line");
 
-                        return Ok(CommandOutcome::DeploymentFailure);
-                    }
-                    shuttle_common::deployment::State::Running
-                    | shuttle_common::deployment::State::Completed
-                    | shuttle_common::deployment::State::Stopped
-                    | shuttle_common::deployment::State::Unknown => break,
+                    match log_item.state.clone() {
+                        shuttle_common::deployment::State::Queued
+                        | shuttle_common::deployment::State::Building
+                        | shuttle_common::deployment::State::Built
+                        | shuttle_common::deployment::State::Loading => {
+                            last_state = log_item.state.clone();
+                            println!("{log_item}");
+                        }
+                        shuttle_common::deployment::State::Crashed => {
+                            println!();
+                            println!("{}", "Deployment crashed".red());
+                            println!();
+                            println!("Run the following for more details");
+                            println!();
+                            print!("cargo shuttle logs {}", &deployment.id);
+                            println!();
+                            if last_state == shuttle_common::deployment::State::Unknown {
+                                println!("Note: Deploy failed immediately");
+                            };
+
+                            return Ok(CommandOutcome::DeploymentFailure);
+                        }
+                        shuttle_common::deployment::State::Running
+                        | shuttle_common::deployment::State::Completed
+                        | shuttle_common::deployment::State::Stopped
+                        | shuttle_common::deployment::State::Unknown => {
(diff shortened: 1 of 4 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -902,33 +902,55 @@ impl Shuttle {
...
-                    shuttle_common::deployment::State::Running
-                    | shuttle_common::deployment::State::Completed
-                    | shuttle_common::deployment::State::Stopped
-                    | shuttle_common::deployment::State::Unknown => break,
+        let mut last_state: shuttle_common::deployment::State =
+            shuttle_common::deployment::State::Unknown;
+
+        loop {
+            let mesg = stream.next().await;
+            if let Some(Ok(msg)) = mesg {
```
oddgrd, line 910: Naming nit:
```suggestion
            let message = stream.next().await;
            if let Some(Ok(msg)) = message {
```

cargo-shuttle/src/lib.rs
```diff
@@ -902,33 +902,55 @@ impl Shuttle {
...
-                        println!();
-
-                        return Ok(CommandOutcome::DeploymentFailure);
-                    }
-                    shuttle_common::deployment::State::Running
-                    | shuttle_common::deployment::State::Completed
-                    | shuttle_common::deployment::State::Stopped
-                    | shuttle_common::deployment::State::Unknown => break,
+        let mut last_state: shuttle_common::deployment::State =
+            shuttle_common::deployment::State::Unknown;
```
oddgrd, line 906: I'm not sure we should start in the `Unknown` state, maybe this should rather be an Option?

oddgrd:
Nice! Hmm, no I am not certain how to reproduce this, but your changes may also have fixed it :thinking:

cargo-shuttle/src/lib.rs
```diff
@@ -952,7 +973,29 @@ impl Shuttle {
...
+            println!();
+            match last_state {
+                Some(shuttle_common::deployment::State::Stopped) => {
+                    println!("State: Stopped - Deployment was running, but has been stopped by the user.")
+                }
+                Some(shuttle_common::deployment::State::Completed) => {
+                    println!("State: Completed - Deployment was running, but stopped running all by itself.")
+                }
+                Some(shuttle_common::deployment::State::Unknown) => {
+                    println!("State: Unknown - This may be because deployment was in an unknown stateWe never expect this state and entering this state should be considered a bug.")
```
iulianbarbu, line 986:
```suggestion
                    println!("State: Unknown - This may be because deployment was in an unknown state. We never expect this state and entering this state should be considered a bug.")
```

cargo-shuttle/src/lib.rs
```diff
@@ -902,33 +902,54 @@ impl Shuttle {
...
+        let mut last_state: Option<shuttle_common::deployment::State> = None;
+
+        loop {
+            let message = stream.next().await;
+            if let Some(Ok(msg)) = message {
+                if let tokio_tungstenite::tungstenite::Message::Text(line) = msg {
+                    let log_item: shuttle_common::LogItem =
+                        serde_json::from_str(&line).expect("to parse log line");
+
+                    match log_item.state.clone() {
```
iulianbarbu, line 914: nit: I would move the clone to the place it's really needed.

cargo-shuttle/src/lib.rs
```diff
@@ -902,33 +902,54 @@ impl Shuttle {
...
+                                println!("Note: Deploy failed immediately");
+                            };
+
+                            return Ok(CommandOutcome::DeploymentFailure);
+                        }
+                        shuttle_common::deployment::State::Running
+                        | shuttle_common::deployment::State::Completed
+                        | shuttle_common::deployment::State::Stopped
+                        | shuttle_common::deployment::State::Unknown => {
+                            last_state = Some(log_item.state);
```
iulianbarbu, line 940: nit: I would `clone` it here instead.

iulianbarbu (approved):
LGTM! 🥳  Thanks. Testing would've been great but the code is simple enough.

cargo-shuttle/src/lib.rs
iulianbarbu: Ah, this is me being silly. We'll not be able to do this I think, because this is a use after move case. I'll resolve this.

iulianbarbu:
@iamwacko , how's everything going? From my side, we'll just need to update the left comment from my review and we're good.

cargo-shuttle/src/lib.rs
```diff
@@ -952,7 +973,29 @@ impl Shuttle {
...
+                    println!("State: Stopped - Deployment was running, but has been stopped by the user.")
+                }
+                Some(shuttle_common::deployment::State::Completed) => {
+                    println!("State: Completed - Deployment was running, but stopped running all by itself.")
+                }
+                Some(shuttle_common::deployment::State::Unknown) => {
+                    println!("State: Unknown - This may be because deployment was in an unknown state. We never expect this state and entering this state should be considered a bug.")
+                }
+                Some(shuttle_common::deployment::State::Running) => {
+                    println!("State: Running - Deployment is running ie. its thread is active.")
```
oddgrd, line 989: This should not be possible, and if it were to happen it would conflict with the default println at the top here (`Deployment has not entered the running state`). The reason it should not be possible is if we enter this branch, it's because the `client.get_service` call above did not find any running deployment for this project. 

I'm also wondering if we should rather than calling `get_service` and seeing if it finds a running deployment, we should call `get_deployment` with the deployment id of the deployment we started in this `deploy` call. This guarantees that we are checking the state of the deployment we started here, although the current method should also work because when a deployment is promoted to `Running` it should first stop any other running deployment. TLDR: consider this second paragraph optional, unless you have any thoughts on the matter! :smile:

oddgrd (commented):
I have one more small comment that I missed the first time around and then we're good to go. Sorry for dragging this out :sweat_smile:

oddgrd (approved):
Awesome, thanks!

Issue #843, opened 2023-05-02
[Feature]: Improve errors for panics in the main function


piewhat:
### Describe the feature

Currently a few panics in the main function will return a log error of no panic message, which is the default if no panic message is received in the alpha runtime: https://github.com/shuttle-hq/shuttle/blob/main/runtime/src/alpha/mod.rs#L330

Requirements
- Reproduce this on a local environment
- Investigate why it is happening
- Propose a solution to fix it

### Duplicate declaration

- [x] I have searched the issues and this feature has not been requested before.

piewhat:
Team: Clawed Crusaders

pedromfedricci:
If anyone has any leads on reproducing this it would be great! :)

oddgrd:
We have had a few reports of this on discord, a search for `no panic message` should yield some results. Thanks for picking this up!

PR #854, opened 2023-05-03
fix: some panic messages get lost

piewhat:
## Description of change
Downcast panic message to `String` or `&str`.

closes issue #843 

Causing panics with Option and Result types.

### DIFF
--- a/runtime/src/alpha/mod.rs
+++ b/runtime/src/alpha/mod.rs
@@ -247,10 +247,13 @@ where
 
                 if error.is_panic() {
                     let panic = error.into_panic();
-                    let msg = panic
-                        .downcast_ref::<&str>()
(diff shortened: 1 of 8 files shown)


oddgrd (commented):
Oh, cool, thanks! It would be good to have a test that covers this as well in `runtime`. A test that tries to run a project like [this](https://github.com/shuttle-hq/shuttle/blob/main/deployer/tests/resources/main-panic/src/main.rs), but rather than a panic in bind, it has an unwrap/expect in `shuttle_runtime::main` and we test that we get the error message not `<no panic message>`. Looking at the other test in `runtime/tests/integration/loader` should be useful. :+1:

oddgrd (approved):
Great job, thank you! :heart_eyes:

PR #855, opened 2023-05-03
chore: add `.editorconfig`

AlphaKeks:
## Description of change

This adds a simple `.editorconfig` file to the root of the repository so that different editors/IDEs may pick up the settings. This makes it easier to have consistent formatting while editing, not just after running `cargo fmt`. See [here](https://editorconfig.org) for more details.

My neovim likes it.


oddgrd (approved):
Ah, I see Thanks, I look forward to trying this out!

PR #857, opened 2023-05-03
feat: allow resetting a user's API-key

### LINKED ISSUE #838: [Feature]: allow resetting a users API-key
### Describe the feature

A user should be able to reset their api-key, this is important functionality to have in case users leak their key. This should generate a new api-key and persist it in the users table.

### Questions

- A JWT will be cached with the api-key or a cookie sid as the key for 10 minutes, should we also clear this? Can we?
- Should this only be possible through the console?

### Tasks

- [x] Add a new reset endpoint to auth
- [x] Update gateway to forward reset requests to auth
- [x] Clean the cache in gateway
- [ ] Add a subcommand to the cli

_No response_

### Duplicate declaration

hseeberger:
A user should be able to reset their api-key, this is important functionality to have in case users leak their key. This should generate a new api-key and persist it in the users table.

Fixes #838.

### DIFF
--- a/auth/src/api/handlers.rs
+++ b/auth/src/api/handlers.rs
@@ -40,9 +40,19 @@ pub(crate) async fn post_user(
 }
 
 pub(crate) async fn put_user_reset_key(
+    session: ReadableSession,
     State(user_manager): State<UserManagerState>,
-    Path(account_name): Path<AccountName>,
+    key: Option<Key>,
 ) -> Result<(), Error> {
+    let account_name = match session.get::<String>("account_name") {
+        Some(account_name) => account_name.into(),
+
+        None => match key {
+            Some(key) => user_manager.get_user_by_key(key.into()).await?.name,
+            None => return Err(Error::Unauthorized),
+        },
+    };
+
     user_manager.reset_key(account_name).await
 }
 
@@ -71,12 +81,12 @@ pub(crate) async fn convert_cookie(
     session: ReadableSession,
     State(key_manager): State<KeyManagerState>,
 ) -> Result<Json<shuttle_common::backends::auth::ConvertResponse>, StatusCode> {
-    let account_name: String = session
-        .get("account_name")
+    let account_name = session
+        .get::<String>("account_name")
         .ok_or(StatusCode::UNAUTHORIZED)?;
 
-    let account_tier: AccountTier = session
-        .get("account_tier")
+    let account_tier = session
+        .get::<AccountTier>("account_tier")
         .ok_or(StatusCode::UNAUTHORIZED)?;
 
     let claim = Claim::new(account_name, account_tier.into());
@@ -99,7 +109,7 @@ pub(crate) async fn convert_key(
     let User {
         name, account_tier, ..
     } = user_manager
-        .get_user_by_key(key.as_ref().clone())
+        .get_user_by_key(key.into())
         .await
         .map_err(|_| StatusCode::UNAUTHORIZED)?;
 
--- a/auth/src/api/handlers.rs
+++ b/auth/src/api/handlers.rs
@@ -39,6 +39,13 @@ pub(crate) async fn post_user(
     Ok(Json(user.into()))
 }
 
+pub(crate) async fn put_user_reset_key(
+    State(user_manager): State<UserManagerState>,
+    Path(account_name): Path<AccountName>,
+) -> Result<(), Error> {
+    user_manager.reset_key(account_name).await
+}
+
 pub(crate) async fn login(
     mut session: WritableSession,
     State(user_manager): State<UserManagerState>,
--- a/auth/src/user.rs
+++ b/auth/src/user.rs
@@ -124,7 +124,7 @@ where
         let user_manager: UserManagerState = UserManagerState::from_ref(state);
 
         let user = user_manager
-            .get_user_by_key(key.as_ref().clone())
+            .get_user_by_key(key.into())
             .await
             // Absorb any error into `Unauthorized`
             .map_err(|_| Error::Unauthorized)?;
@@ -146,13 +146,12 @@ impl From<User> for shuttle_common::models::user::Response {
     }
 }
 
-/// A wrapper around [ApiKey] so we can implement [FromRequestParts]
-/// for it.
+/// A wrapper around [ApiKey] so we can implement [FromRequestParts] for it.
 pub struct Key(ApiKey);
 
-impl AsRef<ApiKey> for Key {
-    fn as_ref(&self) -> &ApiKey {
-        &self.0
+impl From<Key> for ApiKey {
+    fn from(key: Key) -> Self {
+        key.0
     }
 }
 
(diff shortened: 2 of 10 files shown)


auth/src/api/handlers.rs
```diff
@@ -39,6 +39,13 @@ pub(crate) async fn post_user(
     Ok(Json(user.into()))
 }
 
+pub(crate) async fn put_user_reset_key(
+    State(user_manager): State<UserManagerState>,
+    Path(account_name): Path<AccountName>,
+) -> Result<(), Error> {
+    user_manager.reset_key(account_name).await
+}
```
oddgrd, line 47: Should we rather try to get the account to reset from either the session or the api-key bearer token? This way a user can only reset their own key, and you won't need to first get your account name (which is the github-id).

oddgrd (commented):
I see this is a draft, just wanted to say it's looking great so far! Just one small suggestion.

auth/src/api/handlers.rs
hseeberger: I still have to figure out the whole picture including gateway, authentication and authorization; but I was assuming that the route leading to this function would only be accessible for the users themselves (and maybe admins).

Wouldn't using the api key to reset the api key be a huge security issue? What if the api key was leaked and now someone else uses it to reset it?

oddgrd: There is a question if we should only allow this through the console, but the console isn't quite ready yet. The reason I think we could allow doing it with the api-key auth is that if you steal a key and reset it, the only way to get the new key will be to login with github on the website, and you'll also not be able to make new requests with the old key if it is invalidated in the cache. But still, there could be more problems with this I am missing, so if this isn't secure we could start with only allowing reset via console (which is wip) or by admin (then using the admin guard in `auth` on the route).

So if a user is logged in now through the CLI, their api-key will be set as a Bearer token on their requests to gateway. The gateway will then proxy this request to `auth` (if cache is missed in gateway auth layer), where the api-key bearer token will be converted to a JWT with the scopes that account has. 

The JWT is returned to the `gateway` auth layer, and then the request proceeds through the gateway router with the JWT as it's Bearer token. So if your request was to deploy a service, the request will go to `deployer` which will decode this token, and the claim's scopes will be used to determine which endpoints you have permissions to call on the `deployer` router. In the `deployer` the claim will also be set as an extension on the requests to the `runtime` to load and start your service, and when your service runtime requests a database from the `provisioner` the claim scopes will again be checked to make sure you have permissions to provision a database of that type (currently anyone can provision any resource).

I realize this became a bit of a wall of text and I likely missed some details, we should probably do a proper write-up of this and add it to contributing.md/in a new "project layout" document. :sweat_smile:

oddgrd: You're absolutely right about the current dashboard issues, we've been aware of them but we have been prioritizing work on the console, thinking it would be a duplicated effort since the console will arrive soon. In hindsight we should have fixed the dashboard first. :see_no_evil: 

Currently the reset key endpoint on `auth` is under `/users` so requests to it will be forward directly from `gateway` to `auth` (see `forward_to_auth` in `gateway/api/auth_layer`). My thinking is we can check for a cookie *or* an api-key in the new handler, and then get the user from that and reset the key. This way both console and CLI can use it. We'd also need to try to get the cookie sid or api-key (either can be a cache key) in the auth layer and invalidate the token if it's there in the auth layer. Since we do almost the same thing in the case of `!forward_to_auth`, we might do some refactoring of the auth layer logic to make this cleaner.

Unless you guys can see any other issues with the CLI being able to reset the token, the way I see it an intruder resetting the token will only reduce their access. With the token reset and invalidated in the cache, they won't be able to send new requests, but any jobs they started previously may complete since the token lives until expiry. The reason the JWT lives so long (15 min) is architectural constraints, so a deployer might take very long to compile a project, and then the JWT needs to live long enough after to provision resources. We should be able to improve this with the new architecture where the builder is it's own service, I think.

hseeberger: Thanks, @oddgrd, for the detailed explanations. I have changed the PR and followed your advice wrt `shuttle-auth`. `gateway` still needs to be done.

oddgrd: Great, the changes look good! Let us know if you need help with the rest :slightly_smiling_face:

oddgrd (commented):
This looks good to me guys, thanks a lot! I left a few minor comments, and I'm also wondering if we should make the reset command interactive with a confirmation prompt, since it's destructive. What do you guys think? We'll also have other destructive paths soon, like resource deletion, so we should be able to re-use the functionality there.

cargo-shuttle/src/args.rs
```diff
@@ -188,6 +188,12 @@ pub struct LoginArgs {
     pub api_key: Option<String>,
 }
 
+#[derive(Parser, Clone, Debug)]
+pub struct LogoutArgs {
+    /// Invalidate the API key before logging out
+    #[arg(long)]
+    pub invalidate_api_key: bool,
```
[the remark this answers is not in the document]
jonaro00: We were discussing these things internally.
I agree that 'reset' sounds a bit better.
We decided to go for a flag instead of a subcommand in order to not flood the list of subcommands too much. Another idea we had was to have some sort of `cargo shuttle keys` command, with potential for expanding to more than 1 key, different types of keys, etc.
Do you know if the team has future considerations like this?

cargo-shuttle/src/lib.rs
```diff
@@ -280,13 +283,30 @@ impl Shuttle {
...
         Ok(())
     }
 
+    async fn invalidate_api_key(&self, client: &Client) -> Result<()> {
+        client
+            .invalidate_api_key()
+            .await
+            .and_then(|res| {
+                if res.status().is_success() {
+                    Ok(())
```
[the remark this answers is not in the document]
jonaro00: Good idea.

auth/src/user.rs
```diff
@@ -128,13 +146,12 @@ impl From<User> for shuttle_common::models::user::Response {
...
 pub struct Key(ApiKey);
 
-impl AsRef<ApiKey> for Key {
-    fn as_ref(&self) -> &ApiKey {
-        &self.0
+impl From<Key> for ApiKey {
+    fn from(key: Key) -> Self {
+        key.0
     }
 }
```
chesedo, line 156: Nice one :star_struck:

auth/src/api/handlers.rs
```diff
@@ -39,6 +39,23 @@ pub(crate) async fn post_user(
...
+        Some(account_name) => account_name.into(),
+
+        None => match key {
+            Some(key) => user_manager.get_user_by_key(key.into()).await?.name,
+            None => return Err(Error::Unauthorized),
+        },
+    };
+
+    user_manager.reset_key(account_name).await
+}
```
chesedo, line 57: How do users get their new key? Maybe this should return the key?

chesedo (commented):
Just a suggestion about returning the new key when resetting it

auth/src/api/handlers.rs
hseeberger: Hmm ... when you login via the browser, the API key is shown. The CLI command to reset the key is via `logout`, hence the user has to login again and will see the new key.

jonaro00: @chesedo If a key hijacker can get the new key, they can create an infinite loop of key regeneration, locking the account owner out. You should only be able to get the key from the web.

oddgrd (approved):
Looks good to me! :partying_face:  Thanks for all your work on this one @jonaro00 and @hseeberger, and thanks for the discussion!

PR #859, opened 2023-05-04
ci: download sccache instead of compiling it

iamwacko:
## Description of change

sccache takes up a significant amount of CI time compiling, reducing the effectiveness of the feedback loop.

I ran the commands locally, but CI itself should check this.


oddgrd (approved):
Thanks @iamwacko!

PR #862, opened 2023-05-04
feat(gateway, cargo-shuttle): implement pagination for project list

### LINKED ISSUE #845: [Feature]: Refactor GET-many endpoints in deployer to allow pagination
### Describe the feature

We should implement pagination for the endpoints that get a list of items from the deployer and gateway. This will be useful for CLI users so they can get only a subset of e.g. projects, rather than all of them. It will also be useful for the console, so we can fetch a subset at a time, rather than fetch them all and then paginate them in the browser.

- [ ] Implement pagination for project list (”/projects” endpoint on the router in gateway/api/latest)
- [ ] Implement pagination for deployment list
- [ ] Consider whether we should follow this up with some filtering, e.g. should we be able to filter project list by project state?

Are there more endpoints we should implement this for? Should we consider using [cursor pagination](https://dev.to/jackmarchant/offset-and-cursor-pagination-explained-b89) for this?

### Duplicate declaration

AlphaKeks:

Closes #845. This adds pagination functionality to the `project list` and `deployment list` subcommands for `cargo-shuttle` and changes the API endpoints to accommodate that.

Unit tests in `gateway/src/service.rs` and `deployer/src/persistence/mod.rs`


Kazy:
Regarding the number of results returned, by default the API will keep returning every single result, since we believe the endpoint is also used in the web console. And to go with it, there is also no limitation of the `limit` parameter as well. 

The CLI though does pagination by default, returning the first 10 items, which is a breaking change. We wanted to also improve the experience on that on two fronts but didn't to first get your opinion on how we should solve them (maybe in an other PR). First there is no mention that more pages are available. Second in the case of a limit of 0 or of a page that goes beyond the actual number of pages; there is a mention that no deployment/project exists, which isn't strictly correct.

The easy solution is just to rework the messages. We can for instance say that there might be more results available in the next page, even if that isn't the case. The other solution would be to fetch the total number of items, using either a second SQL request, or a single one but with a `count(*) over()` to get the total count. A less expensive call could be done by just fetching if there is a new row or not (`lead() over()`).

AlphaKeks:
As I mentioned in our private conversation, we could very well make the `--limit` flag default to `u32::MAX` which would effectively keep the old behavior.

As for the potential issue with supplying a `--page` value that's too high, I would probably run 2 queries, returning early if the specified page is too high. But I'm open to do something else as well.

deployer/src/persistence/mod.rs
```diff
@@ -549,6 +566,36 @@ mod tests {
         );
     }
 
+    #[tokio::test(flavor = "multi_thread")]
+    async fn get_deployments() {
```
iulianbarbu, line 570: Thanks for adding a test. What would happen if we pass an out-of-bounds `offset`?

gateway/src/api/latest.rs
```diff
@@ -84,6 +84,14 @@ impl StatusResponse {
     }
 }
 
+#[derive(Debug, Clone, Copy, Deserialize, IntoParams)]
+struct PaginationDetails {
```
iulianbarbu, line 88: Seeing it again here makes me think this can be as well something declared under `shuttle-common`. What do you think?

gateway/src/api/latest.rs
```diff
@@ -115,14 +123,21 @@ async fn get_project(
     responses(
         (status = 200, description = "Successfully got the projects list.", body = [shuttle_common::models::project::Response]),
         (status = 500, description = "Server internal error.")
+    ),
+    params(
+        PaginationDetails
```
iulianbarbu, line 128: Do we need it here given we've derived for it the `IntoParams` trait?

gateway/src/service.rs
```diff
@@ -271,20 +271,31 @@ impl GatewayService {
...
-                    (
-                        row.get("project_name"),
-                        row.get::<SqlxJson<Project>, _>("project_state").0,
-                    )
-                });
+        let mut query = QueryBuilder::new(
+            "SELECT project_name, project_state FROM projects WHERE account_name = ",
+        );
+        query.push_bind(account_name);
+        query.push(" LIMIT ").push_bind(limit);
```
iulianbarbu, line 282: nit: I guess these two can be chained into the expression above.

gateway/src/service.rs
```diff
@@ -772,14 +783,49 @@ pub mod tests {
...
+        assert_eq!(
+            svc.iter_user_projects_detailed(&neo, 0, 20)
+                .await
+                .unwrap()
+                .map(|item| item.0)
+                .collect::<Vec<_>>(),
+            all_projects[..20]
+        );
+        assert_eq!(
+            svc.iter_user_projects_detailed(&neo, 20, 20)
```
iulianbarbu, line 821: Same as before, what would happen if we pass an out-of-bounds offset?

iulianbarbu (commented):
Thanks @AlphaKeks ! The PR looks really good, it's almost there. Just left a few questions that are unclear on my side.

gateway/src/service.rs
```diff
@@ -271,20 +271,31 @@ impl GatewayService {
...
-                    )
-                });
+        let mut query = QueryBuilder::new(
+            "SELECT project_name, project_state FROM projects WHERE account_name = ",
+        );
+        query.push_bind(account_name);
+        query.push(" LIMIT ").push_bind(limit);
+
+        if offset > 0 {
+            query.push(" OFFSET ").push_bind(offset);
```
iulianbarbu, line 285: I just realized that we don't have an ORDER BY clause on this one, but this might be because there isn't something timestamp-wise we can order by. Do you know if we have guarantees that the query will return the rows in the same order though (and which one that might be)? I am thinking this can be an issue only if we query for a specific page based on a projects order and then query for the next page but we'll be basing it on another projects order.

cargo-shuttle/src/args.rs
```diff
@@ -141,7 +141,15 @@ pub enum Command {
 #[derive(Parser)]
 pub enum DeploymentCommand {
     /// List all the deployments for a service
-    List,
+    List {
+        #[arg(long = "page", default_value = "1")]
```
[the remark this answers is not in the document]
AlphaKeks: No, we don't actually. I named the field `offset` initially and forgot the remove the `long` attribute again after changing it.

cargo-shuttle/src/lib.rs
```diff
@@ -984,8 +986,13 @@ impl Shuttle {
         Ok(())
     }
 
-    async fn projects_list(&self, client: &Client) -> Result<()> {
-        let projects = client.get_projects_list().await?;
+    async fn projects_list(&self, client: &Client, page: u32, limit: u32) -> Result<()> {
+        if limit == 0 {
+            println!();
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
AlphaKeks: Yep, we should 👍

gateway/src/api/latest.rs
AlphaKeks: Yeah sounds good. We'll look for a good place to put it and remove the duplication.

gateway/src/api/latest.rs
AlphaKeks: From my understanding the `IntoParams` trait is necessary to put it there in the first place? The `utoipa` macro is used for generating documentation so I would assume every parameter needs that trait so utoipa can generate docs for it.

common/src/models/project.rs
```diff
@@ -17,6 +18,14 @@ pub const fn idle_minutes() -> u64 {
     IDLE_MINUTES
 }
 
+#[derive(Debug, Clone, Copy, Deserialize, IntoParams)]
+pub struct PaginationDetails {
```
iulianbarbu, line 22: Ah, yep. Close to what I meant but I was thinking about having a single common struct that can be shared between deployments and projects list endpoints. The thing is that I am not really sure where it can be put, here is one thought thread:
1) Create `common/src/api_query`.
2) Create the `PaginationDetails` struct under `common/src/backends/api_query.rs` and consume it from there.
3) Curious if you have thoughts on creating `common/src/api_query/mod.rs` instead of `common/src/backends/api_query.rs`. I would go with 2) now rather than 3) because 3) makes sense only if there are other `api_query` functionality that can be split between multiple sub-module parts that can target specificities with `api_query` (e.g. for deployments, projects, and others).
4) Maybe another idea?

Let me know what you think.

deployer/src/persistence/mod.rs
AlphaKeks: An out-of-bounds `offset` where? If someone makes an API request with a value that can't be parsed into a `u32`, the handler won't be called and the user will receive an error from Axum. I'm also pretty sure that the database will accept any `u32` as an `OFFSET` for a query, so that should be fine.

gateway/src/api/latest.rs
iulianbarbu: Ah indeed, it seems we still need to mention it, my bad: https://docs.rs/utoipa/latest/utoipa/attr.path.html#intoparams-type. So should we add it to the `get_deployments` route from the deployer as well?

common/src/models/project.rs
AlphaKeks: The issue with a shared struct is that if we end up adding filtering capabilities to `project list` (the third bullet point [here](https://github.com/shuttle-hq/shuttle/issues/845#issue-1692726868), the filters would also be shared between both commands, which I think is not ideal. We talked about having a common struct that contains _only_ the pagination data, and then `#[serde(flatten)]` any additional parameters, but we decided that it's not worth the additional layer of abstraction and that it's better to just keep 2 separate structs around. What do you think? For reference, we are working on the filters on [this branch](https://github.com/Kazy/shuttle/tree/feat/project-filtering).

gateway/src/service.rs
Kazy: Indeed we shouldn't [rely on that](https://www.postgresql.org/docs/current/queries-order.html). Most of the time this should not be an issue, but let's not bet on that :)

> If sorting is not chosen, the rows will be returned in an unspecified order. The actual order in that case will depend on the scan and join plan types and the order on disk, but it must not be relied on. 

The best way would be to add a `created_at` column to the project table. Should we go with that solution ?

gateway/src/api/latest.rs
Kazy: Indeed that's an oversight !

common/src/models/project.rs
iulianbarbu: Let me know if this is accurate: your suggestion is for having two separate structs called `PaginationDetails`, one for project and one for deployment, and the project filters would be added as additional struct fields on the project's `PaginationDetails` struct. This should work as well. 

Your suggestion for `#[serde(flatten)]` opened me up for an example regarding pagination, present in the official docs. My thinking was along the below lines: 

```rust
#[derive(Serialize, Deserialize)]
struct PaginationDetails {
    limit: u64,
    offset: u64
}

#[derive(Serialize, Deserialize)]
struct ProjectFilter {
    state: State,

    #[serde(flatten)]
    pagination: PaginationDetails,
}
```

So instead of having the shared pagination struct with the pagination fields + additional `#[serde(flatten)]` parameters, we can have the queries for each resource defined in their own struct adding the `PaginationDetails` with `#[serde(flatten)]`.

Using such a layout for the struct can be easily mapped to my second point [here](https://github.com/shuttle-hq/shuttle/pull/862#discussion_r1185996913). Either way, I do not have a strong preference right now about the `PaginationDetails` being a single struct versus two separate structs.

I am curious how everything will look like once you add project filtering. 😄

common/src/models/deployment.rs
```diff
@@ -7,12 +7,21 @@ use comfy_table::{
...
 use serde::{Deserialize, Serialize};
+use utoipa::IntoParams;
 #[cfg(feature = "openapi")]
 use utoipa::ToSchema;
 use uuid::Uuid;
 
 use crate::deployment::State;
 
+#[derive(Debug, Clone, Copy, Deserialize, IntoParams)]
+pub struct PaginationDetails {
```
iulianbarbu, line 18: Late review comment: I am not sure what this looks like from the perspective of a client who wants to find how many pages to display in a browser page, listing the deployments (same for the projects). It will not be able to display a certain number of pages because it's not knowing how many it can display. I think this is a separate API that has to be done by the client. I also think that another PR can address this. What do you guys think?

common/src/models/project.rs
AlphaKeks: Yes, this was pretty much what I had in mind as well. However, I think it's an unnecessary abstraction and unless we intend to add this functionality to other commands as well (such as `cargo shuttle resource`, I don't think it's worth doing.

gateway/src/service.rs
Kazy: Hey no worries ! 

The idea was to actually backfill the `created_at` field with the current timestamp. This won't be correct but since we got to start tracking it at some point. The issue with that solution is that we still need to solve ties between old projects. For that I actually found a way to generate a datetime for `created_at` that follows the disk order, while still having a different `created_at` value between each project. This should be the closest we can get from the actual insertion order, and we shouldn't get any tie after that. I'll go ahead and implement this, but please let me know if you see any problem with this solution.

common/src/models/deployment.rs
iulianbarbu: I just had a sync with @Kazy and we can omit resolving the issue I've described because the functionality can be still used like it is. The reason is that the pagination API returns an empty response list for an out-of-bounds page. This means that the web UIs who want to display page numbers will be able to understand where to stop. Assuming the UIs display a limited interval of pages, it will need to do requests for for all of them and if a request for a page returns an empty list, the client knows it reached the end. I will resolve this since it's a non-issue for now and will be relevant only for cases where the clients prefer to do some kind of lazy fetching which will require a pre-fetching of the projects count or this total count added with each page response.

common/src/models/project.rs
iulianbarbu: Sure, there is not a mandatory requirement on this one given the struct is duplicated only for two places. We can refactor this in the future if needed. Resolving this as well.

deployer/src/persistence/mod.rs
iulianbarbu: Yeah, the concern was more toward the fact that the OFFSET can be bigger than the total number of rows. However, @Kazy assured me the result will still be an empty row list and error-free. I will keep this comment open still because we agreed to add in the tests a case for querying for an offset bigger than the rows count.

gateway/src/service.rs
iulianbarbu: We've agreed to go with adding the `created_at` column for the next projects and sort only by the `project_name` for the old projects where we can not derive the `created_at` column in a sane manner. I'll keep this comment open to track this effort as well.

gateway/src/service.rs
iulianbarbu: I'll keep this open as well because would be good to include a case where we query for an offset greater than the rows count.

iulianbarbu:
Also FYI, this project was just changed to a `medium` one, because of the changed requirements. Let us know if you have any questions.

Kazy:
Hey @iulianbarbu ! I've updated the branch with some of the changes we discussed.

Regarding the `created_at` column, I assumed everything was using Postgres, but this isn't the case when developing locally where instead Sqlite is used. I don't know if you actually support both at the same time (PSQL in prod, Sqlite in dev) or if you also use Sqlite in the context of the `gateway` and the `deployer` in production ? 

In the meantime I've made the changes to use `project_name` for ordering in the `gateway` and `last_updated_at` (latest updated first, as it makes more sense when using the CLI) for the `deployer`. I've also added a text that an another page might be available: happy to rewrite it in any way.

iulianbarbu:
> Hey @iulianbarbu ! I've updated the branch with some of the changes we discussed.
> 
> Regarding the `created_at` column, I assumed everything was using Postgres, but this isn't the case when developing locally where instead Sqlite is used. I don't know if you actually support both at the same time (PSQL in prod, Sqlite in dev) or if you also use Sqlite in the context of the `gateway` and the `deployer` in production ?

We use sqlite in prod as well. We use it for both `gateway` and `deployer` at this moment.

> In the meantime I've made the changes to use project_name for ordering in the gateway and last_updated_at (latest updated first, as it makes more sense when using the CLI) for the deployer. I've also added a text that an another page might be available: happy to rewrite it in any way.

Thanks, taking a look rn.

iulianbarbu (commented):
Looks good so far @Kazy ! I think the only thing left is to add a migration for the projects table, where we alter the table by adding the `created_at` column. This column should be changed when creating a new project, info persisted in the `gateway` sqlite. And then, when iterating through the account projects, we should `ORDER BY created_at DESC, project_name`. What do you think?

Kazy:
Hey @iulianbarbu ! Thank you for the review ! 

I've added the `created_at` column to the `gateway` DB. I didn't use a `DEFAULT` statement when updating the column because with SQLite you cannot add a non constant default value for a column (in our case `CURRENT_TIMESTAMP`).

Kazy:
Hey @iulianbarbu, I've fixed the tests, everything should be good for this PR now. Let me know if there is anything left to do :)

PR #868, opened 2023-05-05
Reimplemented JwtAuthentication with struct-based Future.

arturaz:


oddgrd (approved):
This looks great, thanks @arturaz!

PR #869, opened 2023-05-05
feat: use `include_str!` instead of duplicating entire source files.

### LINKED ISSUE #867: [Feature]: Refactor init to bootstrap examples
### Describe the feature

We currently bootstrap projects for various service integrations with hard-coded boilerplate code. The current implementation also generates the `Cargo.toml` with the latest available versions, which can be a problem: ‣

So for this project we should:

1) Investigate the viability of bootstrapping projects from examples in our examples repository: https://github.com/shuttle-hq/examples
2) Open a draft PR with a proof of concept
3) Update the `cargo-shuttle` cli to be able to init any example, not just hello-world for each integration.
4) Implement the refactor for all integrations

_No response_

### Duplicate declaration

paulotten:
## Description of change

Draft for #867.

(Eventually will) close #867.

The macros compile. The macros reference code that compiles.

TODO: manually run at least one of the sample projects created by this code.

### DIFF
--- a/cargo-shuttle/src/init.rs
+++ b/cargo-shuttle/src/init.rs
@@ -779,6 +779,7 @@ pub fn write_main_file(boilerplate: &'static str, main_path: &Path) -> Result<()
 #[cfg(test)]
 mod shuttle_init_tests {
     use super::*;
+    use indoc::indoc;
 
     fn cargo_toml_factory() -> Document {
         indoc! {r#"
--- a/cargo-shuttle/src/init.rs
+++ b/cargo-shuttle/src/init.rs
@@ -4,7 +4,6 @@ use std::path::{Path, PathBuf};
 
 use anyhow::{bail, Result};
 use cargo_edit::{find, get_latest_dependency, registry_url};
-use indoc::indoc;
 use shuttle_common::project::ProjectName;
 use toml_edit::{value, Array, Document, Table};
 use url::Url;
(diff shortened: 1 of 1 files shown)


paulotten:
I see the failed CI task. It's my fault. I'm fixing it.

paulotten:
Looks like here: https://github.com/shuttle-hq/shuttle/blob/main/.circleci/config.yml#L192

@oddgrd If I add a command to initialize the git submodule before this, will it effect this PR, or would it have to merged to main first to take effect? (If the later I'll probably split that to it's own PR.)

oddgrd:
Hey @paulotten, thanks for getting started on this one! So I don't think the `include_str!` strategy would work, since a user won't get the examples submodule when they install `cargo-shuttle`. It would also limit us to the projects in the examples repo.

What we had in mind here (and this was unclear in the ticket, apologies) was a solution where we could pass in a URL to a git repo (like this https://github.com/shuttle-hq/shuttle-examples/tree/main/axum/hello-world), and this would bootstrap a project with that project source code. I'll think on this a bit and see if I can suggest some next steps here, but I assume we'll use the `git2` crate for this which we already have as a dependency in `cargo-shuttle`. I'll get back to you here if I have any bright ideas!

paulotten:
Ok, so I'm thinking something like [cargo-generate](https://crates.io/crates/cargo-generate).

@oddgrd Do we want to copy (some of) this functionality, or should we just direct the user to `cargo-generate` in the documentation?

oddgrd:
I think we can close this one in favor of the `cargo-generate` one, right @paulotten? Either way, thanks for exploring this strategy!

Issue #690, opened 2023-03-07
Consider making the TLS implementation used in `shuttle-shared-db` configurable


aumetra:
(Opening this issue based on a question asked [on Discord](https://discord.com/channels/803236282088161321/1078827844651720775/1082063956618772622))

Currently the SQLx connection pool provisioned by `shuttle-shared-db` uses `native-tls` to establish a secure connection to the database server.  
Would it be thinkable to gate these behind features like `rustls` and `native-tls`, where one of them is set as the default? (this would be a breaking change)

---

Looking at the SQLx source code, they use the `webpki-roots` certificates for their trust anchors.  
So if Shuttle installs their own CA certificates, using the `rustls` feature will break it.

moritzheiber:
This is especially relevant if you're using `sqlx` in your application already (e.g. to run migrations or query the database) and don't wish to rely on `native-tls` as a feature of the crate (i.e. you cannot have the `sqlx` dependency of `shuttle-shared-db` built with `native-tls` and your own version of `sqlx` with e.g. `rustls`)

oddgrd:
The thing I'm not sure about here is that since we use `native-tls` everywhere else (which we might also be able to change, but that needs investigation), will using `rustls` here then conflict? If someone wants to explore this in a PR and prove that it works, I think we can add this feature flag to `shuttle-shared-db`.

oddgrd:
@utterstep awesome! I sent you an invite now to get the `triager` role on this repo, this should allow you to add labels and set assignments

aumetra:
Closed by #870

PR #870, opened 2023-05-05
add option to use rustls instead of native-tls in `shuttle-shared-db`

utterstep:
## Description of change

Adds feature `postgres-rustls` to `shuttle-shared-db`, allowing to use Postgres without relying on bindings to native SSL implementation.

Closes issue #690.

Created new example in https://github.com/shuttle-hq/shuttle-examples/pull/43, local run with both `postgres` and `postgres-rustls` feature is working:

`postgres`:
image

`postgres-rustls`:
image
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -509,6 +509,8 @@ workflows:
                 - services/shuttle-tower
                 - services/shuttle-warp
       - check-standalone:
+        # shuttle-shared-db has mutually exclusive features
+        # so we run checks for each feature separately
           name: "resources/shared-db: << matrix.features >>"
           matrix:
             alias: check-standalone-shared-db
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -509,7 +509,7 @@ workflows:
                 - services/shuttle-tower
                 - services/shuttle-warp
       - check-standalone:
-          name: resources/shared-db
+          name: "resources/shared-db: << matrix.features >>"
           matrix:
             alias: check-standalone-shared-db
             parameters:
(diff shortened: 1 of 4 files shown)


utterstep:
image
Now `test --all-feature` fails, as it includes mutually exclusive features..
Will think about possible workarounds

utterstep:
Hmm, to be honest, I don't see an obvious way to solve this problem. This PR introduces feature `postgres-rustls` which is mutually exclusive with `postgres` bc the `sqlx` crate intentionaly fails to compile if both `rustls` and `native-tls` are present. 

And `cargo test --all-features` enables all features available.
Maybe consider splitting `shuttle-shared-db` tests in CI in two batches: with `native-tls` and with `rustls`?

@oddgrd what do you think about it?

utterstep:
Split CI job for `shuttle-shared-db` into separate jobs for each feature.

Before (with mutually incompatible features from SQLx being selected by `--all-features`):
image

After:
image

Also it seems like `shuttle-deployer` tests are flaky?
Here are two consequent runs on my local machine without any diff between them:
image
image

.circleci/config.yml
```diff
@@ -505,6 +508,16 @@ workflows:
                 - services/shuttle-tide
                 - services/shuttle-tower
                 - services/shuttle-warp
+      - check-standalone:
```
oddgrd, line 511: Maybe a short comment here about why we need this extra job could be useful?

oddgrd (approved):
Great, thanks for this @utterstep! :heart_eyes: Regarding the CI issue, I think your solution is good :ok_hand:

.circleci/config.yml
utterstep: Sure, did that :)

Issue #840, opened 2023-05-02
[Feature]: Add a health endpoint to all services


Cyril-Cf:
### Details

We want to eventually send a notification to users when their service goes down. But for this we need to know a service is still healthy.

### Requirements

- Create a `/healthz` endpoint on every service under `/services`
    - Have return just a status code of 200

### Duplicate declaration

- [X] I have searched the issues and this feature has not been requested before.

frectonz:
I will also be working this

jonaro00:
The team is evaluating a new approach to this problem.

jonaro00:
Closing this in favour of #1557

PR #876, opened 2023-05-06
feat(services): add `/healthz` route that responds with `200 OK` to `services`

frectonz:
This commit includes `/healthz` route implementations for `warp` and `axum`. This might conflict with user defined routes if they have a `/healthz` defined in their code.

## Description of change
This is still a work in progress (WIP) there are services in which this feature is not implemented in yet.

By running the `examples` and requesting the `/healthz` route of the app.

Issue #840


frectonz:
Implemented services

services/shuttle-axum/src/lib.rs
```diff
@@ -25,6 +25,13 @@ impl shuttle_runtime::Service for AxumService {
     /// Takes the router that is returned by the user in their [shuttle_runtime::main] function
     /// and binds to an address passed in by shuttle.
     async fn bind(mut self, addr: SocketAddr) -> Result<(), Error> {
+        let healthz_router = axum::Router::new().route(
+            "/healthz",
+            axum::routing::get(|| async { axum::http::StatusCode::OK }),
+        );
+
+        self.0 = (self.0).merge(healthz_router);
```
oddgrd, line 33: Could we just do something like `self.0.route("/healthz", get(|| async {}))` here?

services/shuttle-salvo/src/lib.rs
```diff
@@ -27,8 +27,19 @@ impl shuttle_runtime::Service for SalvoService {
...
+        use salvo::prelude::*;
+
+        #[handler]
+        async fn healthz_handler(res: &mut Response) {
+            res.set_status_code(StatusCode::OK);
+        }
+
+        let healthz_router = salvo::Router::with_path("healthz").get(healthz_handler);
+
+        let app = salvo::Router::new().push(healthz_router).push(self.0);
```
oddgrd, line 39: Will this make the healthz route the innermost route (so middleware might be applied to it)? Is it possible with salvo to just add the healthz route to the router in `self.0`?

services/shuttle-tide/src/lib.rs
```diff
@@ -25,8 +25,10 @@ where
     /// Takes the router that is returned by the user in their [shuttle_runtime::main] function
     /// and binds to an address passed in by shuttle.
     async fn bind(mut self, addr: SocketAddr) -> Result<(), Error> {
+        self.0
+            .at("/healthz")
+            .get(|_| async { Ok(tide::StatusCode::Ok) });
```
oddgrd, line 30: Nice!

services/shuttle-tower/src/lib.rs
```diff
@@ -58,7 +58,8 @@ where
     /// Takes the service that is returned by the user in their [shuttle_runtime::main] function
     /// and binds to an address passed in by shuttle.
     async fn bind(mut self, addr: SocketAddr) -> Result<(), Error> {
-        let shared = tower::make::Shared::new(self.0);
+        let shared = tower::make::Shared::new(HealthCheck::new(self.0));
```
oddgrd, line 61: This is great!

services/shuttle-warp/src/lib.rs
```diff
@@ -20,20 +20,28 @@ pub struct WarpService<T>(pub T);
...
     /// and binds to an address passed in by shuttle.
     async fn bind(mut self, addr: SocketAddr) -> Result<(), Error> {
-        warp::serve((*self).clone()).run(addr).await;
+        use warp::Filter;
+
+        let app_filter = (*self).clone();
+        let health = warp::path("healthz")
+            .and(warp::get())
+            .and(warp::path::end())
+            .map(warp::reply);
```
oddgrd, line 35: Can we do this?
```suggestion
        let health = warp::path("healthz")
            .map(warp::reply);
```

services/shuttle-poem/src/lib.rs
```diff
@@ -21,14 +21,23 @@
...
+
 #[shuttle_runtime::async_trait]
 impl<T> shuttle_runtime::Service for PoemService<T>
 where
     T: poem::Endpoint + Send + 'static,
 {
     async fn bind(mut self, addr: std::net::SocketAddr) -> Result<(), shuttle_runtime::Error> {
+        let app = poem::Route::new()
+            .at("/healthz", poem::get(healthz))
+            .nest("/", self.0);
```
oddgrd, line 37: Maybe we should swap this so `/healthz` is the outermost route?

services/shuttle-rocket/src/lib.rs
```diff
@@ -18,11 +18,18 @@
...
+use rocket::response::status;
 use std::net::SocketAddr;
 
 /// A wrapper type for [rocket::Rocket<rocket::Build>] so we can implement [shuttle_runtime::Service] for it.
 pub struct RocketService(pub rocket::Rocket<rocket::Build>);
 
+#[rocket::get("/healthz")]
+fn health_check() -> status::Custom<()> {
+    status::Custom(Status::Ok, ())
+}
```
oddgrd, line 31: Nit: could we just return an empty tuple here? We'll just be interested in the response status, which I believe will default to 200 OK

oddgrd (commented):
Hey @frectonz, thanks for this! I see you are still working on this, I just wanted to leave an initial review. This looks great, I left a few comments and questions, but I'm not deeply familiar with all of these frameworks, so I may well be wrong! :smile:

services/shuttle-axum/src/lib.rs
frectonz: Yeah we probably should do that, made the change in the latest commit. 

btw sorry for taking a long time to respond, i was having my final exams so i couldn't make time to respond

services/shuttle-salvo/src/lib.rs
frectonz: Unfortunately, `salvo` doesn't provide a way to attach a nested path for an existing router. The only way to nest a route under an existing `Router`  is using `.push()`. [here is an example form the salvo repo](https://github.com/salvo-rs/salvo/blob/42677bfe9fa2dcb7a8f88b378875875b50626f26/examples/todos/src/main.rs#L26)

As for the middlewares, i don't think the middlewares the user defines are going to run on the `healthz_router` since we are pushing it first. Because the `.hoop()` method is how middlewares are added and the docs say middlewares are only applied to the router's descendants [here](https://docs.rs/salvo/latest/salvo/prelude/struct.Router.html#method.hoop)

services/shuttle-tide/src/lib.rs
frectonz: thanks :pray:

services/shuttle-tower/src/lib.rs
frectonz: thanks :pray: btw this was my first time implementing a `tower::Service`,  I watched this [youtube stream](https://youtu.be/16sU1q8OeeI) from **David Pedersen** to learn how to do it

services/shuttle-warp/src/lib.rs
frectonz: we can do that but each of the filters have their own use

- `warp::get()` matches that the request is `GET` request, if we don't add this `/healthz` route will respond to any http verb
- `warp::path::end()` ensures that the current request handler *(`warp::reply` in this case)*, doesn't handle nested routes under `/healthz` like `/healthz/cat`, here it is in the  [warp docs](https://docs.rs/warp/latest/warp/filters/path/fn.end.html)

services/shuttle-poem/src/lib.rs
frectonz: If you mean doing this
```rust
self.0 = self.0.at("/healthz", poem::get(healthz));
```
we can't do this because the trait bound on `PoemService` expects the trait [`poem::Endpoint`](https://docs.rs/poem/1.3.55/poem/trait.Endpoint.html) which doesn't have a why to chain requests so we need to create our own `poem::Route` or change the trait bound.

services/shuttle-rocket/src/lib.rs
frectonz: yeah we could that like this
```rust
#[rocket::get("/healthz")]
fn health_check() {}
```
and it does work wouldn't it better to explicitly show that we are responding with `Status:Ok`

services/shuttle-poem/src/lib.rs
iulianbarbu: Hmm, maybe @oddgrd  referred to something like this? Haven't tried it myself so not sure if it works.
```suggestion
        let app = poem::Route::new()
            .at("/", self.0),
            .at("/healthz", poem::get(healthz));
```

services/shuttle-rocket/src/lib.rs
iulianbarbu: This is only a nit. Feel free to do it as you wish. I prefer the explicit aspect, but I have no strong opinion about it.

services/shuttle-salvo/src/lib.rs
iulianbarbu: Cool @frectonz ! I think your response resolves O comment.

services/shuttle-warp/src/lib.rs
iulianbarbu: Very nice @frectonz ! This should be resolved as well.

iulianbarbu (commented):
Apart from the comments, I think a general aspect of it all would be about how the `/healthz` behaves is handled by the platform and mentioning a specification for it. Do you think it's possible to easily provide tests along these two dimensions?

1) Making sure `/healthz` route is not affected by middlewares defined by the user for each service. We can write a user middleware for each of the crates, apply it to a user route and then `assert!(false)` in it. Then call the `/healthz` route which shouldn't make the test fail.

2) Making sure we handle the cases where users define their own `/healthz` for each of the service.

chesedo:
> This is however something that we'll be required for doing liveliness probes in an accurate way, so we either to this [1], or let user define their own health check endpoint [2], or involve other means for checking if the service is managed by a `shuttle-runtime` is still alive [3]

I think we can start with [1] (being opinionated). As users raise issues with this, we can evolve this. But currently we don't really know what all the possible issues could be and which issues will appear more often.

services/shuttle-axum/src/lib.rs
```diff
@@ -25,6 +25,11 @@ impl shuttle_runtime::Service for AxumService {
     /// Takes the router that is returned by the user in their [shuttle_runtime::main] function
     /// and binds to an address passed in by shuttle.
     async fn bind(mut self, addr: SocketAddr) -> Result<(), Error> {
+        self.0 = self.0.route(
+            "/healthz",
+            axum::routing::get(|| async { axum::http::StatusCode::OK }),
+        );
```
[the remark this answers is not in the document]
frectonz: user defined middlewares are probably applied to this one to solve this we can create a new router and nest the `healthz` route and the user route `self.0`.

services/shuttle-poem/src/lib.rs
frectonz: that works, converted to this in the latest commit

services/shuttle-axum/src/lib.rs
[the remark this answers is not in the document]
iulianbarbu: Hmm, I am not sure though. The documentation states this: https://docs.rs/axum/latest/axum/middleware/index.html#ordering. We add the `/healthz` route at the end though, after `self.0` is created by the user.

In either case, would be great to confirm the behavior with tests. We want to spot early if our expectation will change in the future (either by us breaking this contract or by the axum library).

oddgrd:
Hey @frectonz, I changed this to an M since it seems to be a bit more extensive than we anticipated. I saw that you are busy with exams, but when you have time to get back to it and if you need help with the remaining services feel free to reach out here or on Discord :slightly_smiling_face:

jonaro00:
@oddgrd I could continue and finish this PR. Having this enables many bugfixes + improvements.

jonaro00:
Implementation question: What if this did not need to be baked into every single framework? What if `Service` had another bind method (default implemented in trait) that on a different port than the main bind, hosts a lightest-possible hyper server that serves this endpoint? (also, make sure internal checks of `/healthz` does not spam user logs)

+ Hyper is already in runtime dep tree, so no added deps.
+ No interference with user code.

oddgrd:
> @oddgrd I could continue and finish this PR. Having this enables many bugfixes + improvements.

That would be terrific! 

Regarding your suggestion, I've had the same thought, but if we're starting them asynchronously there may be a tiny delay before the user service is ready. A server will return Pending until it completes/crashes (which means we will not pass the .await), so I'm not sure how we would be able to ensure the users service is running before we start ours. We could spawn it and sleep, but that may not be bullet-proof. There may also be a clever solution to this problem I don't know about, so if you have any thoughts on these points I think this is definitely worth exploring :smile:

On the feature branch we have implemented a ping endpoint for the runtime, but currently it technically just verifies that the runtime is running, which does not mean the service is started and ready. I'm wondering if we could track the state in the runtime, e.g. set it to running if start succeeds, and set it to stopped if subscribe_stop gets a stop. Then in the runtime ping endpoint we return 200 if the state is running (if the runtime is crashed, the service will also be down and it will return !200 anyway).

Maybe I overthought this. :joy: 

cc @chesedo @iulianbarbu

jonaro00:
Is there a way to check if a port is in use? Maybe polling that (exp backoff?) until bind has bound to addr?

> ... set it to running if start succeeds ...

sounds nice

iulianbarbu:
I think implementing a health endpoint separate from the user service is a partial solution and might open up other 
 issues. I see also two scenarios here:
1) Shuttle needs to check the user service health through the runtime, for core logic reasons. We can work our way through this with a `runtime` state, stored in memory, that will follow through the `loading`/`starting` phases and will respond with the corresponding state on a `ping` RPC.
2) Shuttle needs to check the user service health through the user service (if possible), to show it somewhere (console maybe?). In this case, my thought is that this health check should be implemented by the user and declared/specified through our `shuttle_runtime::main`. This will eventually turn out to be something used by Shuttle to bring more information up to the users and potentially automate any actions based on the health check output.

I think we're interested to solve first 1) and then think about how to structure 2) (which is important but not urgent IMO).

iulianbarbu:
> When the user's Router (etc) is done and performs the bind call to addr, it should be completely initialized and ready to serve, right?

I don't think we can assume this. The bind call can carry long IO ops which might delay serving directly.

> Feels like this needs some clear definitions of how Shuttle should be able to interact with a user service. What kind of "commands" and "requests" should Shuttle be able to send to the user service? (looking into the future)
If there is more than a health check, giving the user the burden to implement these will cause a lot of friction. There should at least be an easy way to get a "default" impl.

We need to clarify this internally and we're planning to come up with some thoughts by EOW 🤞

iulianbarbu:
Also probably we'll need to move the discussions on the associated ticket #840. If anyone has thoughts please share them there. We'll probably come up with a summary and way forward by EOW there as well.

PR #878, opened 2023-05-07
feat(Makefile): add option to disable --detach on make up

iamwacko:
## Description of change

SHUTTLE_DETACH can be set to disable, so when `make up` is run docker-compose stays attached. This was discussed in batch-tech.

I ran `make up` with SHUTTLE_DETACH set and without it set.


oddgrd (approved):
Thanks @iamwacko, looks good to me!

Issue #873, opened 2023-05-06
[Feature]: Switch shuttle to Rustls


iamwacko:
### Describe the feature

We currently use `native-tls` everywhere (openssl). This can cause issues for certain users, for example for users with out of date installations of openssl. We should investigate if we can switch to rustls everywhere, and if there are any drawbacks to doing it. If not, we should do it. #100 is similar, but it just wants a feature gate. If this gets resolved native-tls will just be gone.

This is S SB2, and already moved to assigned on the Notion. @oddgrd I need perms for labels and assignments.

### Suggestion or Example of how the feature would be used

- [x] Check if it is possible and works for everyone
- [x] Implement the changes
  - [x] cargo-edit depends on native-tls
  - [x] reqwest depends on native-tls
  - [x] axum depends on native-tls
  - [x] sqlx depends on native-tls

### Duplicate declaration

- [X] I have searched the issues and this feature has not been requested before.

iamwacko:
cargo-edit seems to use rustls by default, but native-tls is still in the dependency tree because it will use native-tls on platforms rustls doesn't support.

iamwacko:
It looks like we can move away completely from rustls, but bringing in cargo as a dependency means we still have to use openssl

Issue #821, opened 2023-04-26
[Bug]: cargo shuttle init: Could not query the latest version of shuttle-runtime


0atman:
### What happened?

While running `cargo shuttle init`, after selecting the web framework it panics with `Could not query the latest version of shuttle-runtime`

### Version

cargo-shuttle 0.14.0 (but I have also tested 0.13.0 with the same problem)

### Which operating systems are you seeing the problem on?

Linux

### Which CPU architectures are you seeing the problem on?

ARM64
(specifically I'm running asahi linux on mac mini)
`6.2.0-asahi-11-1-edge-ARCH  aarch64 GNU/Linux`

### Relevant log output

```sh
# Short backtrace

thread 'main' panicked at 'Could not query the latest version of shuttle-runtime', cargo-shuttle/src/init.rs:985:33
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2c[...]74/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/2c[...]74/library/core/src/panicking.rs:64:14
   2: cargo_shuttle::init::get_latest_dependency_version
   3: cargo_shuttle::init::cargo_shuttle_init
   4: cargo_shuttle::Shuttle::run::{{closure}}
   5: cargo_shuttle::main::{{closure}}
   6: tokio::runtime::park::CachedParkThread::block_on
   7: tokio::runtime::runtime::Runtime::block_on
   8: cargo_shuttle::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

# Full backtrace and context

❯ RUST_BACKTRACE=full cargo shuttle init
How do you want to name your project? It will be hosted at ${project_name}.shuttleapp.rs.
✔ Project name · tststststs

Where should we create this project?
✔ Directory · tstststs

Shuttle works with a range of web frameworks. Which one do you want to use?
· axum

thread 'main' panicked at 'Could not query the latest version of shuttle-runtime', /home/oatman/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-shuttle-0.14.0/src/init.rs:985:33
stack backtrace:
   0:     0xaaab01446a1c - std::backtrace_rs::backtrace::libunwind::trace::ha0deed5dc23ee505
                               at /rustc/c609da59d9fc05b1c7dc879d79700ccd8140b5fc/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
[...]

PR #879, opened 2023-05-07
feat: switch from native-tls to rustls

iamwacko:
## Description of change

This PR changes` native-tls` to `rustl`s on almost all of our dependencies. `cargo-edit` still technically depends on `native-tls`, but that is because it uses `native-tls` on architectures unsupported by `rustls`. Thanks to our dependency on `cargo` OpenSSL is still in the dependency tree, so I will hold off on actually merging this until the cargo library is completely removed, by me or someone else. Closes issue #873.

It should just work, but I did compile and run it successfully. To check that native-tls wasn't anywhere in the dependency tree I ran `cargo tree --target all -i native-tls`.


oddgrd (commented):
Cool! I'm curious if we will have similar issues with other crates in the future if we don't use native-tls :thinking: 

By the way we use the `vendored-openssl` feature flag in our CI in a few places, this should also be updated. :pray:

iamwacko:
#821 strikes again.

e2e/Cargo.toml
```diff
@@ -8,6 +8,6 @@ crossterm = "0.26.1"
 lazy_static = "1.4.0"
 portpicker = "0.1.1"
 rand = "0.8.5"
-reqwest = { version = "0.11.10", features = ["blocking", "json"] }
+reqwest = { version = "0.11.10", features = ["blocking", "json", "rustls-tls"] }
```
oddgrd, line 11: I think we need to disable default features here, right?

oddgrd (commented):
This LGTM except for one small comment!

e2e/Cargo.toml
iamwacko: `cargo tree` says otherwise. Reqwest isn't bringing in native-tls or openssl with how it is now.

```
├── reqwest v0.11.18
│   ├── base64 v0.21.1
│   ├── bytes v1.4.0
│   ├── encoding_rs v0.8.32
│   │   └── cfg-if v1.0.0
│   ├── futures-core v0.3.28
│   ├── futures-util v0.3.28 (*)
│   ├── h2 v0.3.19 (*)
│   ├── http v0.2.9 (*)
│   ├── http-body v0.4.5 (*)
│   ├── hyper v0.14.26 (*)
│   ├── hyper-rustls v0.24.0
│   │   ├── http v0.2.9 (*)
│   │   ├── hyper v0.14.26 (*)
│   │   ├── rustls v0.21.1
│   │   │   ├── log v0.4.17 (*)
│   │   │   ├── ring v0.16.20 (*)
│   │   │   ├── rustls-webpki v0.100.1
│   │   │   │   ├── ring v0.16.20 (*)
│   │   │   │   └── untrusted v0.7.1
│   │   │   └── sct v0.7.0 (*)
│   │   ├── rustls-native-certs v0.6.2
│   │   │   ├── openssl-probe v0.1.5
│   │   │   └── rustls-pemfile v1.0.2
│   │   │       └── base64 v0.21.1
│   │   ├── tokio v1.28.1 (*)
│   │   └── tokio-rustls v0.24.0
│   │       ├── rustls v0.21.1 (*)
│   │       └── tokio v1.28.1 (*)
│   ├── ipnet v2.7.2
│   ├── log v0.4.17 (*)
│   ├── mime v0.3.17
│   ├── mime_guess v2.0.4
│   │   ├── mime v0.3.17
│   │   └── unicase v2.6.0
│   │       [build-dependencies]
│   │       └── version_check v0.9.4
│   │   [build-dependencies]
│   │   └── unicase v2.6.0 (*)
│   ├── once_cell v1.17.1
│   ├── percent-encoding v2.2.0
│   ├── pin-project-lite v0.2.9
│   ├── rustls v0.21.1 (*)
│   ├── rustls-pemfile v1.0.2 (*)
│   ├── serde v1.0.163 (*)
│   ├── serde_json v1.0.96 (*)
│   ├── serde_urlencoded v0.7.1 (*)
│   ├── tokio v1.28.1 (*)
│   ├── tokio-rustls v0.24.0 (*)
│   ├── tower-service v0.3.2
│   ├── url v2.3.1 (*)
│   └── webpki-roots v0.22.6 (*)
```

oddgrd: Oh, interesting. Is that from running inside the e2e directory?

iamwacko: I added `default-features = false` anyway, as it brings in fewer dependencies.

jonaro00:
Still waiting for #888 to remove cargo-edit for this to complete, right?

e2e/Cargo.toml
jonaro00: e2e is not in the same workspace, so that's why it does not show up in the tree. Regardless, due to not being in the tree, it doesn't really matter whether it uses native-tls.

iamwacko:
> Still waiting for https://github.com/shuttle-hq/shuttle/pull/888 to remove cargo-edit for this to complete, right?

Sort of, but not really. native-tls will still show up in the `cargo tree`, but it won't be compiled unless 
```
[target.'cfg(not(any(target_arch = "x86_64", target_arch = "arm", target_arch = "x86", target_arch = "aarch64")))'.dependencies]`
```

See [line 89 of this.](https://github.com/killercup/cargo-edit/blob/e591c61f7ea58e65c6ac2a99f3b55f422774ca8a/Cargo.toml#L89) What this means is that in practice native-tls is gone and OpenSSL with it.

oddgrd (approved):
Thanks!

Issue #875, opened 2023-05-06
[Feature]: upgrade shuttle-codegen to use syn 2.0


SyedFasiuddin:
### Describe the feature

The current implementation of shuttle-codegen package depends on version 1.0.19 of syn while some of the dependencies of shuttle use v2.0 of syn, this issue (feature) tries to upgrade shuttle-codegen to use v2.0.

### Duplicate declaration

- [X] I have searched the issues and this feature has not been requested before.

PR #880, opened 2023-05-08
Update/syn 2.0

SyedFasiuddin:
## Description of change

This PR upgrades shuttle dependency of syn to v2
closes issue #875 

using `cargo test —-package shuttle-codegen --all-features`


codegen/src/next/mod.rs
```diff
@@ -95,11 +92,17 @@ impl Endpoint {
         };
 
         // We'll use the paren span for errors later
-        let paren = params.paren_token;
+        let endpoint_delim_span = &endpoint
+            .meta
+            .require_list()
+            .expect("Params Delimiter")
```
oddgrd, line 98: I think this expect is okay since the params parsing above will error if the endpoint [meta](https://docs.rs/syn/latest/syn/enum.Meta.html) is not a `MetaList`, but perhaps we should rephrase it?
```suggestion
            .expect("endpoint meta should be a list")
```

oddgrd (approved):
Thanks a lot @SyedFasiuddin! :star_struck:  This looks good to me, and it's also cool to see some of the improvments/cleanups that came with syn v2. I left one comment but it's just a rephrasing suggestion.

Issue #756, opened 2023-03-27
[Enhancement]: Get rid of `cargo` dependency


jonaro00:
Current uses of bundled cargo (three subtasks):
- ~~**cargo-shuttle**~~ - Done #765 
  - ~~When using `shuttle init`. This can be replaced with a Command call.~~
  - ~~ToSemver when checking versions of cargo-shuttle and runtime. This is not strictly needed, a == is probably sufficient. (or just semver crate)~~
- ~~**deployer**~~ -  Done #907
  - ~~CliError is used when pre deploy testing. (can be replaced)~~
  - ~~Various enums and config struct used to construct the test command. All options used are easily used from CLI args.~~
- **service part 1** - Pending #922
  - is\_next and ensure\_{bin,cdylib}: can be done through `cargo_metadata`.
  - clean_crate: should be easily done with a command.
- **service part 2** - In progress by iamawacko #922
  - get_config: gets cargo config, but is never modified before use, so it is not needed if bundled cargo is not used.
  - get_compile_options: uses standard CLI options. (minor TODO here is to not limit cpus to 4 on local release run)
  - check_no_panic: ~~could be done by parsing toml i guess... (main challenge here is that `cargo config` is nightly)~~ -- can probably be dropped instead https://github.com/shuttle-hq/shuttle/issues/756#issuecomment-1488168471
  - build_workspace: figure out another way to find the crates of the workspace, as well as getting the data we need for the BuiltService struct.

oddgrd:
Thanks again for this writeup @jonaro00, this looks good to me! For anyone that wants to pick this one up, it would be preferable to split this up into separate PRs, one for each crate outlined in the issue.

As for your additional suggestion, I'll need to think more about that one.

jonaro00:
Yeah, at this point it is mainly a thing that could be kept in mind when designing this rewrite.

oddgrd:
Before anyone takes this one further, we are currently working on supporting workspaces, so it may be good to hold off on this until we can verify that we can do that without the cargo library.

chesedo:
The workspace support changes are in #767

chesedo:
With the new runtime, I think we can also drop `check_no_panic` when making this change

gautamprikshit1:
How do I test the deployer?

oddgrd:
[...]

PR #888, opened 2023-05-08
feat: use `cargo generate` instead of hardcoding examples source code

paulotten:
## Description of change

Use `cargo-generate` (as a library) to bootstrap copies of `shuttle-examples` apps.

Closes https://github.com/shuttle-hq/shuttle/issues/867.

Ran locally.


cargo-shuttle/src/init.rs
```diff
@@ -50,963 +44,178 @@ impl Template {
...
-        _get_dependency_version_fn: GetDependencyVersionFn,
-    ) {
+    fn get_repo_url(&self) -> &'static str {
+        "http://github.com/shuttle-hq/shuttle-examples.git"
     }
 
-    fn get_boilerplate_code_for_framework(&self) -> &'static str {
-        ""
+    fn get_sub_path(&self) -> Option<&'static str> {
+        todo!()
```
paulotten, line 190: I'll need to create an empty project for this in `shuttle-examples`

paulotten:
@oddgrd what do you think of this approach using `cargo generate`?

It requires the user to have `cargo generate` installed, which I'll have to add documentation for.

In theory it's possible to include `cargo-generate` as a library instead. In practice I haven't been able to make the dependencies compile. (cargo-generate=0.18 requires git2=0.18, which conflicts with other things we use. cargo-generate=0.17 requires git2=0.16 which we use, but then fails to compile. cargo-generate=0.16 requires git2=0.14 which we is too low.)

oddgrd:
The dependency conflict should also be easier to resolve after we remove the `cargo` dependency, which is in progress. The `cargo` dependency currently ties down the `git2` version.

iulianbarbu:
@paulotten out of curiosity. The `cargo-shuttle` dependency on cargo was removed here: https://github.com/shuttle-hq/shuttle/pull/765/files. Would you still be blocked by the `cargo` dependency that we have in the other places, when trying to fix the conflicts coming from `git2` dependency?

iulianbarbu:
Oh, that makes sense. Ok, thanks for the clarification! 🙏

iulianbarbu:
#756  was resolved by #922 . @paulotten I am curious if we can continue with this PR. Do you have the time to take another stab at it? The idea is that we want to close #821 pretty soon and we're interested in getting this PR over the line. Let us know what you think

cargo-shuttle/src/init.rs
paulotten: Please see https://github.com/shuttle-hq/shuttle-examples/pull/57

jonaro00: I can fix a better (useful) version of https://github.com/shuttle-hq/shuttle-examples/pull/57 so that this is resolved.

jonaro00: https://github.com/shuttle-hq/shuttle-examples/pull/59

cargo-shuttle/src/lib.rs
```diff
@@ -217,8 +217,7 @@ impl Shuttle {
         };
 
         // 5. Initialize locally
-        init::cargo_init(path.clone(), project_args.name.clone().unwrap())?;
-        init::cargo_shuttle_init(path.clone(), framework)?;
+        init::cargo_generate(path.clone(), project_args.name.clone().unwrap(), framework)?;
```
iulianbarbu, line 220:
```suggestion
        init::cargo_generate(
            path.clone(),
            project_args
                .name
                .as_ref()
                .expect("to have a project name provided"),
            framework,
        )?;
```

cargo-shuttle/src/init.rs
```diff
@@ -50,1465 +45,170 @@ impl Template {
...
-    // Get `[dependencies]` table
-    let dependencies = cargo_doc["dependencies"]
-        .as_table_mut()
-        .expect("manifest to have a dependencies table");
-
-    let manifest_path = find(Some(path.as_path())).unwrap();
-    let url = registry_url(manifest_path.as_path(), None).expect("Could not find registry URL");
-
-    let init_config = framework.init_config();
+pub fn cargo_generate(path: PathBuf, name: ProjectName, framework: Template) -> Result<()> {
```
iulianbarbu, line 195:
```suggestion
pub fn cargo_generate(path: PathBuf, name: &ProjectName, framework: Template) -> Result<()> {
```

iulianbarbu (commented):
For some reason it doesn't work well on my side. Doing an init for an `axum` project brings over the entire shuttle-examples repo. Can you reproduce it?

jonaro00:
I also get the entire repo

cargo-shuttle/src/init.rs
paulotten: change applied manually

cargo-shuttle/src/lib.rs
paulotten: change applied manually

jonaro00:
You can bump the examples submodule to the latest on main :)

paulotten:
> For some reason it doesn't work well on my side. Doing an init for an `axum` project brings over the entire shuttle-examples repo. Can you reproduce it?

Should be fixed now.

Now I'm noticing that the crate name isn't getting set correctly. Looking into it.

paulotten:
> You can bump the examples submodule to the latest on main :)

Fixed, I think.

cargo-shuttle/src/lib.rs
```diff
@@ -217,8 +217,14 @@ impl Shuttle {
         };
 
         // 5. Initialize locally
-        init::cargo_init(path.clone(), project_args.name.clone().unwrap())?;
-        init::cargo_shuttle_init(path.clone(), framework)?;
```
jonaro00, line 221: @paulotten This `cargo_shuttle_init` used to do the project name change. You need to restore that part of it.

paulotten:
> Now I'm noticing that the crate name isn't getting set correctly. Looking into it.

Should be fixed. (Ended up using `toml-edit` to set it.)

jonaro00 (approved):
LGTM! Maybe just a tiny nitpick: the toml_edit imports could be at top of file.
I'm looking at if we can fix the failing rocket init tests and kill two birds with one stone.

jonaro00:
> Looks like just a new line at end of file difference.

Yeah, but there are also some tests that fail sporadically. I think I've managed to reproduce it.

This can be merged if the team wants, and I'll add the fixes to the subsequent PR.

cargo-shuttle/Cargo.toml
```diff
@@ -23,9 +22,7 @@ dirs = { workspace = true }
 dunce = "1.0.3"
 flate2 = { workspace = true }
 futures = { workspace = true }
-# This needs to be 0.16.0 for its dependency on libgit2-sys to be compatible
-# with cargo 0.69.0's dependency on libgit2-sys.
-git2 = "0.16.0"
+git2 = "0.17.2"
```
chesedo, line 25: Nice, to see this resolved!! :partying_face:

chesedo (commented):
Always love to see a PR that removes code.

Small nit: maybe the `Shuttle.toml` should be deleted since it will override the project to something that already exists.

paulotten:
> Ya, I can do that.

There should be an update to this PR for this commit https://github.com/shuttle-hq/shuttle/commit/65ab862983616adb17b9c59c4947ff2aee95bb7a

GitHub is having issues though. https://www.githubstatus.com/

> [Incident with Actions, Pull Requests and Webhooks](https://www.githubstatus.com/incidents/1g1gkh0qpyvs)

paulotten:
> GitHub is having issues though.

Appears to be fixed. I see the expected commit now.

iulianbarbu (approved):
LGTM! Thanks a lot for this. 🙏

jonaro00:
I saw now that two things are missing! The `init` flag to generate does not initialize a git repo like `cargo init` does. One more thing that cargo does is add a .gitignore with `target/`. We should probably add this aswell but we can also include `Secrets*.toml`!

oddgrd:
> I saw now that two things are missing! The `init` flag to generate does not initialize a git repo like `cargo init` does. One more thing that cargo does is add a .gitignore with `target/`. We should probably add this aswell but we can also include `Secrets*.toml`!

Good catch! Should we include that here, or have you started on these in your new PR @jonaro00?

jonaro00:
I have not started on those. You can do them here if you want.

paulotten:
> One more thing that cargo does is add a .gitignore with `target/`.

`cargo_generate` has a `vcs` option. Setting it to `Git` however doesn't seem to create a `.gitignore`.

> We should probably add this aswell but we can also include `Secrets*.toml`!

`cargo_generate` isn't going to know to add that. I'll just write a `.gitignore` file manually.

jonaro00:
Great.
The template could have a gitignore already. Should we append to the file if it exists? Or parse it and only add the lines if they are missing?

paulotten:
> Good point! Another thought, can we use the git2 library for this?

It depends how complex we need to make this.

If we need something like this, I'd probably go reaching for a `gix` crate (Rust rewrite of git functionality). Maybe [gix_ignore](https://docs.rs/gix-ignore/latest/gix_ignore/#).

jonaro00:
So shall we add a .gitignore with `target/` and `Secrets*.toml` to all examples? It seems reasonable that they have this as a standard now that cloning will be the default way of starting off.

jonaro00:
> Setting it to Git however doesn't seem to create a .gitignore.

We can still pass that option so that a git repo is init'ed.

oddgrd (approved):
LGTM, thanks for finding the `cargo-generate` solution and for finishing this up so quickly after the cargo removal! :heart:

Issue #891, opened 2023-05-09
[Feature]: Updated documentation in the shuttle crate


AliSajid:
### Describe the feature

Right now the shuttle documentation is incomplete for several components. This makes intellisense less useful.

It would be a good idea to update the docs.

### Duplicate declaration

- [X] I have searched the issues and this feature has not been requested before.

PR #892, opened 2023-05-09
docs: updates for return types for examples actix-web, tide, serenity, tower

ethanholz:
## Description of change

Updated docs for better intellisense.

Related to #891 

n/a


resources/shared-db/src/lib.rs
```diff
@@ -1,4 +1,9 @@
 #![doc = include_str!("../README.md")]
+//! # [Shuttle Shared Databases](https://docs.shuttle.rs/resources/shuttle-shared-db)
+//!
+//! This plugin manages databases that are shared with other services on shuttle.
+//! Your database will share a postgres server with other users,
+//! but it will not be accessible by other users.
```
iulianbarbu, line 6:
```suggestion
//! Your database will share the server with other users, but it will not 
//! be accessible by other users.
```

iulianbarbu (commented):
How have you decided between choosing what Shuttle services to document in terms of their return types? I think the rest can benefit too from such docs, but I am wondering if you thought about stopping this at a few and letting the rest for another PR, or you want to continue adding, for this PR to be complete. Let me know your thoughts.

Also, resources/persist can benefit too from a lib level docs.

ethanholz:
These are just the ones in the example repo to start per the batch board. My partner has a separate PR for the other docs in the examples. The purpose of this PR is to help provide a better intellisense for new users and provide a framework for how we might implement these into other parts in the future.

resources/secrets/src/lib.rs
```diff
@@ -1,9 +1,35 @@
...
+//!     };
+//!
+//!     let state = MyState { secret };
+//!     let rocket = rocket::build().mount("/", routes![secret]).manage(state);
+//!
+//!     Ok(rocket.into())
+//! }
+//!
+//! ```
+
```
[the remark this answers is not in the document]
iulianbarbu: It's still useful this way, mirroring the public docs, so let's have it like this for the moment, even if it might bring some confusion down the line. This should be addressed as soon as we revamp our public docs. Thanks, resolving this!

iulianbarbu (approved):
Thanks @ethanholz !

codegen/src/lib.rs
```diff
@@ -6,13 +6,65 @@ mod shuttle_main;
...
+/// ```
 #[cfg(feature = "frameworks")]
 #[proc_macro_error]
 #[proc_macro_attribute]
 pub fn main(attr: TokenStream, item: TokenStream) -> TokenStream {
     shuttle_main::r#impl(attr, item)
 }
 
+/// Setups up a `shuttle_app` to be generated. Allows
+/// for user to setup a `next` application.
```
oddgrd, line 29: Perhaps something like this here? And perhaps with links to axum router and hyper server?
```rust
/// Generates a wasm32-wasi module containing an Axum router with your endpoints, which is passed as a 
/// hyper::service::Service to a hyper::Server. 
```

oddgrd (commented):
Thanks @ethanholz, this looks great! I just have a couple of small comments I think we should address and this is good to go. Apologies for the late review!

PR #904, opened 2023-05-11
deployer(MockedBuilder): push to container registry

iulianbarbu:
## Description of change

* Added `MockedBuilder` support to push docker OCI tar images to a container registry.
* Update the `/project/:project_name` route to return a fake `deployment_id` as a response.

TODOs:
* ~~Add a few more unit tests to the `oci` module and make sure all of them pass.~~
* ~~Fix the documentation for the `oci` module.~~
* ~~Update the `oci` module to be `async` based (I think we can cover this in a next PR). @chesedo @oddgrd what do you think?~~
* ~~add an integration test for the auth code path (spawn a docker registry container with auth on circleci, push an image through the builder there, pull it with docker and use it to run a command on it)~~

Note: The `oci` module is based on a stripped version of the `ocipkg` crate.

Locally, by following the `deployer/README` and `deployer/src/builder/README`.


deployer/src/builder/oci/distribution/auth.rs
```diff
@@ -0,0 +1,141 @@
...
+            }
+            Err(ureq::Error::Status(..)) => Err(Error::AuthorizationFailed(token_url.to_string())),
+            Err(ureq::Error::Transport(e)) => Err(Error::Network(e.to_string())),
+        }
+    }
+
+    pub fn append(&mut self, other: Self) -> Result<()> {
+        for (key, value) in other.auths.into_iter() {
+            self.auths.insert(key, value);
+        }
```
chesedo, line 61: You might be able to use [`HashMap::extend()`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.extend) here.

deployer/src/builder/oci/distribution/auth.rs
```diff
@@ -0,0 +1,141 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AuthChallenge {
+    pub url: String,
+    pub service: String,
+    pub scope: String,
+}
+
+impl AuthChallenge {
+    pub fn from_header(header: &str) -> Result<Self> {
+        let err = || Error::UnSupportedAuthHeader(header.to_string());
```
chesedo, line 90: Nice :star_struck:

deployer/src/builder/oci/distribution/client.rs
```diff
@@ -0,0 +1,165 @@
...
+            agent: ureq::Agent::new(),
+            url,
+            name,
+            auth,
+            token: None,
+        })
+    }
+
+    fn call(&mut self, req: ureq::Request) -> Result<ureq::Response> {
+        if req.url().contains("localhost") {
```
chesedo, line 33: Maybe we should comment what a request to `localhost` means?

deployer/src/builder/oci/distribution/client.rs
```diff
@@ -0,0 +1,165 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    //
+    // Following tests need registry server. See test/fixture.sh for setting.
+    // These tests are ignored by default.
+    //
```
chesedo, line 147:
```suggestion
    // Following tests need a registry server. See test/fixture.sh for setting one up.
    // These tests are ignored by default.
    //
```

deployer/src/builder/oci/distribution/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+
+/// Push image to registry
+pub fn push_image(path: &Path) -> Result<()> {
+    if !path.is_file() {
+        return Err(Error::NotAFile(path.to_owned()));
+    }
+    let mut f = fs::File::open(path)?;
+    let mut ar = super::image::Archive::new(&mut f);
+    for (image_name, manifest) in ar.get_manifests()? {
+        info!("Push image: {}", image_name);
```
chesedo, line 25:
```suggestion
        info!(%image_name, "pushing image");
```

deployer/src/builder/oci/distribution/client.rs
```diff
@@ -0,0 +1,165 @@
...
+            name,
+            auth,
+            token: None,
+        })
+    }
+
+    fn call(&mut self, req: ureq::Request) -> Result<ureq::Response> {
+        if req.url().contains("localhost") {
+            return req
+                .call()
```
chesedo, line 35: Can this be async?

deployer/src/builder/oci/error.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+impl From<OciSpecError> for Error {
+    fn from(e: OciSpecError) -> Self {
+        match e {
+            OciSpecError::SerDe(e) => Error::InvalidJson(e),
+            OciSpecError::Io(e) => Error::UnknownIo(e),
+            OciSpecError::Builder(_) => unreachable!(),
```
chesedo, line 61: We should document why it is unreachable, if possible

chesedo (changes requested):
Seems about done. Thanks @iulianbarbu

chesedo:
> Note: The `oci` module is based on a stripped version of the `ocipkg` crate.

Why not use the `ocipkg` crate then?

iulianbarbu:
> > Note: The `oci` module is based on a stripped version of the `ocipkg` crate.
> 
> Why not use the `ocipkg` crate then?

Has more functionality than we'll need, like pulling images and loading them. Also, the interface is not convenient because when pushing an image you can do it only based on a path: https://github.com/search?q=repo%3Atermoshtt%2Focipkg%20push_image&type=code. 

As you said, we will most probably have the image in memory, so we would've needed to fork the crate, modify it and consume it from there. At the same time, given we would've done this, it made more sense to me to strip it down and get only what's necessary for our use case (modifying directly in our codebase if needed).

deployer/src/builder/oci/error.rs
iulianbarbu: Hmm, this is a leftover from `ocipkg` crate. I added details about it. Thanks! 👍

iulianbarbu (commented):
Thanks P! Addressed the majority of the comments. Let me know if something's missing or you have more feedback.

iulianbarbu (commented):
Thanks P! Addressed the majority of the comments. Let me know if something's missing or you have more feedback.

oddgrd (commented):
Thanks Iulian, I need to get more familiar with the OCI and OCI distribution specs, but this looks good to me! 

I saw your reasons for using a stripped version of the ocipkg crate and it makes sense. However, did you check out the newer [`oci-distribution`](https://github.com/krustlet/oci-distribution) crate? It seems to be actively developed by companies in a similar space. It may also have an API [more suited](https://docs.rs/oci-distribution/latest/oci_distribution/struct.Client.html#method.push) to our needs, what do you think? Although it aims to implement the entire spec, so it conflicts with your argument regarding abundant functionality, although I'm not sure that's a very important consideration for this service.

Oh and by the way, if we keep the stripped `ocipkg` implementation, should we have have an attribution in the `oci` module readme to indicate it's copied from the `ocipkg` crate?

iulianbarbu:
> Thanks Iulian, I need to get more familiar with the OCI and OCI distribution specs, but this looks good to me!

Thanks O! 🙏 I am not extremely familiar with OCI too so any help coming in is welcomed.

> I saw your reasons for using a stripped version of the ocipkg crate and it makes sense. However, did you check out the newer [`oci-distribution`](https://github.com/krustlet/oci-distribution) crate? It seems to be actively developed by companies in a similar space. It may also have an API [more suited](https://docs.rs/oci-distribution/latest/oci_distribution/struct.Client.html#method.push) to our needs, what do you think? Although it aims to implement the entire spec, so it conflicts with your argument regarding abundant functionality, although I'm not sure that's a very important consideration for this service.

`oci-distribution` crate seems to have it all but not just yet. I think this might be suitable down the line after it gets more features. The API is not ideal for our use case though. 

1) We're working with an OCI image as a tar archive. The `oci_distribution::push_image` image API takes a vector of `ImageLayer`s and sends them one by one. However, it computes first the blobs [digests](https://docs.rs/oci-distribution/latest/src/oci_distribution/client.rs.html#317). This is not needed in our case because we already have the digests and blobs to send, per the image below. This was for an image generated with docker as a tar archive OCI image. We would use nix build for prod, but it should look similar to this: http://lethalman.blogspot.com/2016/04/cheap-docker-images-with-nix_15.html.

Screenshot 2023-05-15 at 08 58 27 

2) For some reason pushing tar images to ECR should return an image location URL, per: https://docs.docker.com/registry/spec/api/#put-manifest, but it doesn't. As such, using `oci_distribution::push_image` with ECR will not be yet a plug-and-play because this will fail: https://docs.rs/oci-distribution/latest/src/oci_distribution/client.rs.html#347.

I think this is why we can not use `oci-distribution` just yet but in the near future, we might be able to.
 
> Oh and by the way, if we keep the stripped `ocipkg` implementation, should we have have an attribution in the `oci` module readme to indicate it's copied from the `ocipkg` crate?

😍 This is a great idea, yes! Added a mention to it.

oddgrd (approved):
Ah, I see, thanks for looking into it and for the explanation Iulian!

deployer/src/builder/error.rs
```diff
@@ -0,0 +1,25 @@
...
+    fn from(e: oci::error::Error) -> Self {
+        Error::Oci(e)
+    }
+}
+
+impl From<io::Error> for Error {
+    fn from(e: io::Error) -> Self {
+        Error::Io(e)
+    }
+}
```
chesedo, line 25: We should be able to use `#[from]` for these right?

deployer/src/builder/mod.rs
```diff
@@ -19,17 +19,38 @@ impl MockedBuilder {
...
+        debug!("Successfuly built and pushed the image to the container registry.");
+
+        Ok(uuid::Uuid::new_v4())
+    }
+
+    /// Push a built image to an container registry.
+    pub async fn push_image(&self, image: Vec<u8>) {
+        self::oci::distribution::push_image(image)
+            .await
+            .expect("to not fail");
```
chesedo, line 54: If this code won't change, maybe we want proper error handling?

deployer/src/builder/oci/distribution/auth.rs
```diff
@@ -0,0 +1,146 @@
...
+        })
+    }
+}
+
+#[derive(Deserialize)]
+struct Token {
+    token: String,
+}
+
+mod tests {}
```
chesedo, line 146:
```suggestion
```

deployer/src/builder/oci/distribution/client.rs
```diff
@@ -0,0 +1,216 @@
...
+        Ok(Client {
+            agent: reqwest::Client::new(),
+            url,
+            name,
+            auth,
+            token: None,
+        })
+    }
+
+    #[async_recursion::async_recursion]
```
chesedo, line 33: I've never seen this. What does it do? :star_struck:

deployer/src/builder/oci/distribution/client.rs
```diff
@@ -0,0 +1,216 @@
...
+        // 	"errors": [
+        // 	    {
+        //             "code": <error code>,
+        //             "message": "<error message>",
+        //             "detail": ...
+        //         },
+        //         ...
+        //     ]
+        // }
+        // ```
```
chesedo, line 72: Nice comment :sparkles:

deployer/src/builder/oci/distribution/client.rs
chesedo: Remember this one

deployer/src/builder/oci/image/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+        }
+    }
+
+    pub fn entries(&mut self) -> Result<tokio_tar::Entries<&'a [u8]>> {
+        // Calling entries must be preceded by a reseted archive, which is denoted
+        // by a non empty Option. Consumers of the archive, meaning scopes where
+        // `.entries()` is called will take out the value from the Option leaving
+        // it empty for the next `.entries()` calls.
+        if self.archive.is_some() {
+            let mut inner = self.archive.take().expect("to get the inner archive");
```
chesedo, line 122: Should we have proper error handling here too?

chesedo (approved):
nit: looking good. Thanks @iulianbarbu!

deployer/src/builder/error.rs
iulianbarbu: Yep, thanks! Done.

deployer/src/builder/mod.rs
iulianbarbu: Done.

deployer/src/builder/oci/distribution/auth.rs
iulianbarbu: Done.

deployer/src/builder/oci/distribution/client.rs
iulianbarbu: Hah, basically it's changing the return type of the function to a [`BoxedFuture`](https://docs.rs/futures/0.3.19/futures/future/type.BoxFuture.html).  A compiler error usually guides you to this when you write async recursive functions. An example can look like this:
```
 error[E0733]: recursion in an `async fn` requires boxing
--> src/main.rs:1:26
  |
1 | async fn fib(n : u32) -> u32 {
  |                          ^^^ recursive `async fn`
  |
  = note: a recursive `async fn` must be rewritten to return a boxed `dyn Future`.
  = note: consider using the `async_recursion` crate: https://crates.io/crates/async_recursion
```

deployer/src/builder/oci/distribution/client.rs
iulianbarbu: Thanks! Done.

deployer/src/builder/oci/image/mod.rs
iulianbarbu: Updated how this gets done. Our calls to `.entries()` must be preceded by the `self.archive` reinitialization, so I simply moved the reinit to the `.entries()` body, and now the `.expect(...)` call is justified. I don't think we'll need to handle an error anymore given how we construct this.

iulianbarbu:
Seems like the `workspace` test is failing on `shuttle-deployer-alpha` clippy. I think it's not worth fixing. I ran clippy on `deployer` and `resource-recorder` (the components we've worked on) and it didn't find issues with them, so I think we're fine merging this.

Issue #372, opened 2022-10-05
Does Shuttle support crontab?


piloer:

chesedo:
Not officially, but it should be possible to add a custom cron service. Docs for creating a custom service should be coming out soon

jonaro00:
Closed in favor of #909

PR #905, opened 2023-05-11
Add shuttle_crontab service

morlinbrot:
Closes shuttle-hq/shuttle#372

Following our discussions in the Q&A sessions and elsewhere, we implemented a crontab service with "full functionality" as shuttle_crontab service.

This is the "call api on a schedule service" as discussed in the original, much smaller example service in [this PR](https://github.com/shuttle-hq/shuttle-examples/pull/42), which is now obsolete.

Once this is merged, we can add a simple example demonstrating its usage, a [draft PR](https://github.com/shuttle-hq/shuttle-examples/pull/47) is already set up for reference but it needs another commit removing the local path from its `Cargo.toml`.

Tested by running locally and interacting with it via the example service.


jonaro00 (commented):
This is a good implementation for adding scheduled GET requests dynamically. And good documentation :)

However, if I wanted to use a service like this, I would probably want to be able to build more custom logic per endpoint (secrets, auth, parsing, calls to other shuttle services) and sacrificing the dynamic addition of schedules. What I have in mind is that you would create functions/closures/higher abstractions (structs/traits) and add them to the crontab service together with their schedule.

Overall, this way of using crontab feels more like an example implementation (shuttle-examples) than an official `shuttle-crontab` service, since the capability and customization of this is quite limited.

The use-case I describe above is quite easy to make without a crate to abstract it, so I'm not sure what a proper `shuttle-crontab` would look like. I'm willing to discuss ideas though.

morlinbrot:
Thanks for the feedback.

> However, if I wanted to use a service like this, I would probably want to be able to build more custom logic per endpoint (secrets, auth, parsing, calls to other shuttle services)

That sounds super interesting and I'd be glad to look into this although I wouldn't be exactly sure how to implement it right now.

What we did here was the outcome of the previous discussion we had where it was decided that going for this sort of "call URL on a schedule service" would be a good addition.

We're happy to discuss any further evolution of this if the team thinks this would be the right direction to go in. If so, I think it would be great to have some kind of design session on this (on Discord) where we would hammer out where exactly we would like to go with this.

services/shuttle-crontab/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+    async fn run_jobs(&mut self) {
+        if let Ok(tab) = self.persist.load::<Crontab>("crontab") {
+            debug!("Found {} jobs", tab.jobs.len());
+            for raw in tab.jobs {
+                debug!("Starting job: {:?}", raw);
+                let job = CronJob::from_raw(&raw);
+
+                tokio::spawn(async move {
+                    job.run().await;
+                });
```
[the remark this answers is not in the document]
morlinbrot: This doesn't seem to work any other way, since we need the `move` to not run into lifetime issues (We do want to completely move this to another thread after all).

services/shuttle-crontab/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+#[derive(Debug)]
+struct CronJob {
+    schedule: Schedule,
+    url: String,
+}
+
+impl CronJob {
+    fn from_raw(raw: &RawJob) -> Self {
```
iulianbarbu, line 49: I would use the idiomatic implementation of `From<RawJob>`  for `CronJob`.

services/shuttle-crontab/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+        while let Some(next_run) = self.schedule.upcoming(Utc).next() {
+            let next_run_in = next_run
+                .signed_duration_since(chrono::offset::Utc::now())
+                .to_std()
+                .unwrap();
+            sleep(next_run_in).await;
+
+            info!("Calling {}", self.url);
+
+            let req = reqwest::get(self.url.clone()).await.unwrap();
```
iulianbarbu, line 68: Comments about customizing this Shuttle service were given and I agree with them to some extent. The main point I would add is that this resource feels more like a `shuttle-http-request-scheduler`. A long name, I know, but if we stick with the `shuttle-crontab` name this would be misleading. The service is specialized. In addition, I would add the ability to schedule a request on the service instead of giving it just a `url` and a `schedule`.

As for a `shuttle-crontab` or `shuttle-crond`, as @jonaro00 said, I would make it more generic, as in making it able to schedule any job. I would be happy though to have a `shuttle-http-request-scheduler` in shuttle. I am curious about anyone's thoughts on this.

services/shuttle-crontab/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+Create a new `CrontabService` by providing a `shuttle_persist::PersistInstance`
+and an `axum::Router`:
+
+```
+// main.rs
+
+#[shuttle_runtime::main]
+async fn crontab(#[Persist] persist: PersistInstance) -> ShuttleCrontab {
+    let router = Router::new().route("/", get(my_own_route));
```
iulianbarbu, line 15: nit: `my_own_route` is great, but a closure is better `|| async { println!("triggered by the crontab service"); }` makes more sense. Maybe this example can be transferred as a crate public doc.

iulianbarbu (commented):
Part 1, will review the rest soon.

services/shuttle-crontab/src/lib.rs
morlinbrot: Obviously. I think I had something else here and refactored and then didn't realise...

services/shuttle-crontab/README.md
morlinbrot: The closure makes sense, although I used a different string to make it obvious that this is a route that is going to be mounted next to the cron service's.

services/shuttle-crontab/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+    sync::oneshot,
+    time::sleep,
+};
+
+mod error;
+use error::CrontabServiceError;
+
+mod router;
+
+pub type ShuttleCrontab = Result<CrontabService, shuttle_runtime::Error>;
```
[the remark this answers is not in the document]
morlinbrot: Done.

services/shuttle-crontab/src/lib.rs
jonaro00: Agreed. In its current form, I wouldn't say it is a relevant crate to publish, it would however be a great example in `shuttle-examples:custom-service/crontab/request-scheduler` (and why not add to the docs, hint: https://github.com/shuttle-hq/shuttle-docs/issues/86).
And by the way, now that I look, it surprises me how none of the Custom Services from the docs are not present in the examples repo (!) I was already planning on adding custom services there later, such as crontab, but others can take the opportunity if they want.

urbaniak: maybe the request part should be just example how to use the cron service, cause doing a proper "cron webhook" service (with all the options like basic auth, token auth, custom headers, custom methods, retry policy with backoff) it will be a much larger job (for sure not a small tag on product board) and not sure there's a high demand for a feature like that?

morlinbrot: Ok, from my point of view, I would propose the following:

1. Rename this to `request-scheduler` and merge it (here or in shuttle-examples), closing the original issue.
2. Open a new issue on adding a `shuttle-crond`, the design and scope of which to be discussed separately.
3. Discuss both points above on an in-person call on Discord.

The last point may be the most important one as it is becoming a bit disheartening for @urbaniak and I to keep up with what feels like imprecise and changing requirements. I think a design session with everybody who has a vision for this to define clear requirements would help a lot.

I will ping you guys on Discord to see if we can make that happen later today!

iulianbarbu: Thanks for taking the lead on clarifying this @morlinbrot and @urbaniak! These conversations are needed, especially because we don't have shuttle charter just yet, but this might change soon.

I concur with the points made by @jonaro00 and @urbaniak on having the `request-scheduler` as an example instead of a shuttle service for the time being. What concerns me the most is the lack of demand for such a service (hence the lack of clear requirements). 

What's reassuring though is that IMO we have a precedent that started with `shuttle-poise`, an opinionated Discord bot framework. We can chat more in our Discord in-person call today.

iulianbarbu:
Was great settling on this guys. Feel free to close and move this PR over to the `shuttle-examples`, as we discussed. We'll take it from there. Closing this.

PR #906, opened 2023-05-11
[scaling] feat: create a resource service

chesedo:
## Description of change
Makes resource management a dedicated service for our new upcoming scaler

A whole bunch of new tests


resource-recorder/migrations/0000_init.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS resources (
+    project_id TEXT,                                -- Identifier of the project this resource belongs to.
+    service_id TEXT,                                -- Identifier of the service this resource belongs to.
+    type TEXT,                                      -- Type of resource this is.
+    data TEXT,                                      -- Data about this resource.
+    config TEXT,                                    -- The config to create the object for this resource.
+    is_active boolean,                              -- The config to create the object for this resource.
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP, -- Time this resource was created.
+    PRIMARY KEY (project_id, service_id, type)
```
oddgrd, line 9: I recognize this as the second NSS proposal of your URN rfc, out of curiosity, does that mean we will only use this or will we also use the one with names?

resource-recorder/src/dal.rs
```diff
@@ -0,0 +1,413 @@
...
+        sqlx::query("UPDATE resources SET is_active = false WHERE service_id = ?")
+            .bind(service_id.to_string())
+            .execute(&mut transaction)
+            .await?;
+
+        // Making mutliple DB "connections" is fine since the sqlite is on the same machine
+        for resource in resources {
+            if let Some(r_project_id) = resource.project_id {
+                if r_project_id != project_id {
+                    warn!("adding a resource that belongs to another project");
```
oddgrd, line 103: Should this also fail and return an error?

resource-recorder/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+pub use dal::Sqlite;
+use tracing::error;
+use ulid::DecodeError;
+
+/// A wrapper to capture any error possible with this service
+#[derive(Error, Debug)]
+pub enum Error {
+    UlidDecode(#[from] DecodeError),
+    Dal(#[from] sqlx::Error),
+    String(String),
```
oddgrd, line 27: Should it be like this?

```suggestion
    #[error("could not parse resource type: {0}")
    String(String),
```

resource-recorder/tests/itegration_tests.rs
oddgrd, line 1: There is a typo in this filename: `itegration_tests.rs` :smile:

resource-recorder/tests/itegration_tests.rs
```diff
@@ -0,0 +1,335 @@
...
+    select! {
+        _ = server_future => panic!("server finished first"),
+        _ = test_future => {},
+    }
+}
+
+/// Layer to set JwtScopes on a request
+#[derive(Clone)]
+pub struct JwtScopesLayer {
+    /// Thes scopes to set
```
oddgrd, line 285:
```suggestion
    /// The scopes to set
```

oddgrd (approved):
This looks really good Pieter, thank you! :heart_eyes: I left a few comments, but mostly questions and typos.

iulianbarbu (commented):
It looks great P. Learned a lot from it. I have one main concern, but it might be a non-issue (`is_active` related comments).

resource-recorder/migrations/0000_init.sql
chesedo: At this level we will use the id one. But I am hoping we will be able to allow users to use the names one, then gateway + deployer should have enough information to convert the names one to this id one.

resource-recorder/src/lib.rs
chesedo: I'm implementing a custom `fmt::Display` to be able to hide sensitive `sqlx::Error` messages from users.

resource-recorder/migrations/0000_init.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS resources (
+    project_id TEXT,                                -- Identifier of the project this resource belongs to.
+    service_id TEXT,                                -- Identifier of the service this resource belongs to.
+    type TEXT,                                      -- Type of resource this is.
+    data TEXT,                                      -- Data about this resource.
+    config TEXT,                                    -- The config to create the object for this resource.
+    is_active boolean,                              -- The config to create the object for this resource.
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP, -- Time this resource was created.
```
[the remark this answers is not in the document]
chesedo: Oh, yes :heart_eyes:

resource-recorder/src/dal.rs
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    async fn delete_resource(&self, resource: &Resource) -> Result<(), sqlx::Error> {
+        sqlx::query("DELETE FROM resources WHERE project_id = ? AND service_id = ? AND type = ?")
+            .bind(resource.project_id.map(|u| u.to_string()))
+            .bind(resource.service_id.map(|u| u.to_string()))
+            .bind(resource.r#type)
+            .execute(&self.pool)
+            .await
+            .map(|_| ())
```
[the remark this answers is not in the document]
chesedo: I'm not sure yet. So far I'm thinking one of two routes:
1. Something else (deployer) call this service and provisioner. And it will check if a resource is inactive before trying to delete it.
2. This service will be the only service that calls provisioner. Then we should check in `lib.rs` if the resource is inactive first.

I'm not sure which route we will take yet. We'll see when we stick all the services together. So I'm leaving it for now.

Issue #903, opened 2023-05-11
[Feature]: Show output of failed tests


piewhat:
### Details

Current if a `cargo shuttle deploy` fails because a test failed, then only the following is returned in the logs:
```
2023-05-05T14:20:54.611619206Z  INFO error: test failed, to rerun pass `--bin hello-world`
```

### Requirements

It would be better if the entire test output, with the tests that failed, were in the log’s output. Not really sure what is missing to make this happen.

### Bonus

Not a requirement for this ticket, but it will be helpful if the solution did not use the `cargo` crate dependency (as it currently does) but rather made a process call to the `cargo` cli. This is because there are other open tickets to remove the `cargo` crate dependency.

### Duplicate declaration

- [X] I have searched the issues and this feature has not been requested before.

piewhat:
Clawed Crusaders

piewhat:
Would this be the desired output?
```
2023-05-10T22:12:52.801210684Z  INFO shuttle_deployer::deployment::queue: Running tests before starting up
2023-05-10T22:12:57.606957936Z  INFO
2023-05-10T22:12:57.617894877Z  INFO running 3 tests
2023-05-10T22:12:57.628967288Z  INFO test claims::tests::missing_bearer ... ok
2023-05-10T22:12:57.639810150Z  INFO test claims::tests::fail ... FAILED
2023-05-10T22:12:57.650867852Z  INFO test claims::tests::to_token_and_back ... ok
2023-05-10T22:12:57.661965310Z  INFO
2023-05-10T22:12:57.673049933Z  INFO failures:
2023-05-10T22:12:57.684073701Z  INFO
2023-05-10T22:12:57.695188432Z  INFO ---- claims::tests::fail stdout ----
2023-05-10T22:12:57.706195597Z  INFO thread 'claims::tests::fail' panicked at 'assertion failed: `(left == right)`
2023-05-10T22:12:57.717223623Z  INFO   left: `false`,
2023-05-10T22:12:57.728102975Z  INFO  right: `true`', src/claims.rs:126:9
2023-05-10T22:12:57.738985121Z  INFO note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2023-05-10T22:12:57.750114651Z  INFO
2023-05-10T22:12:57.761444262Z  INFO
2023-05-10T22:12:57.772608037Z  INFO failures:
2023-05-10T22:12:57.783908953Z  INFO     claims::tests::fail
[...]

PR #813, opened 2023-04-24
Removed cargo dependency from deployer

gautamprikshit1:
## Description of change

Please write a summary of your changes and why you made them. 

Removed Cargo lib dependency from `deployer`

Be sure to reference any related issues by adding `closes issue #`.
Part 2 of #756 

## How Has This Been Tested (if applicable)?

Not tested yet.

PR #907, opened 2023-05-11
feat: show output of failed tests

piewhat:
* feat: remove cargo dependency

## Description of change
Replace the cargo crate with a process call to the cargo cli and send stdout to the message pipe.

closes issue #903 

Tested by modifying the rocket authentication test to fail.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -49,15 +49,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


oddgrd (approved):
LGTM, and thanks for removing cargo as well! :partying_face:

jonaro00:
So does this close #813 now?

Issue #895, opened 2023-05-09
[Feature]: DynamoDB Resource


mikegin:
### Describe the feature

Extending available auto provisioned resources to include non SQL (AWS RDS) type. DynamoDB would be a good candidate as is popular and still based in AWS.

### Duplicate declaration

- [X] I have searched the issues and this feature has not been requested before.

paulotten:
We probably want to be able to run locally. See https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html

paulotten:
@oddgrd I assume you are aware of https://github.com/shuttle-hq/shuttle/pull/910/files#diff-9aa440c73eedfd63ea9f7221d4be2d948eb7af1bffaadd9eec69a379ab1c9cadR659?

That policy restricts users access to only tables that match the assigned prefix.

oddgrd:
Oh, sorry, that was a bit unclear. Our main concern was that we had to give permissions to the provisioner to create users and policies. So we want to find a way to support resources like dynamodb and s3 (where we can't just return a public URL like we do for RDS) without needing those permissions in the provisioner.

jonaro00:
Closing in favour of #1568

PR #910, opened 2023-05-12
feat: add dynamodb resource

mikegin:
**[DynamoDB auto connection](https://github.com/shuttle-hq/shuttle/issues/895)**

Auto connection to the first Key Value store on Shuttle 🥳🚀

Big shoutout to @paulotten for taking the lead on this task


paulotten:
Issue #895

provisioner/src/lib.rs
```diff
@@ -233,6 +254,193 @@ impl MyProvisioner {
...
+                {
+                    "Sid": "ListAndDescribe",
+                    "Effect": "Allow",
+                    "Action": [
+                        "dynamodb:List*",
+                        "dynamodb:DescribeReservedCapacity*",
+                        "dynamodb:DescribeLimits",
+                        "dynamodb:DescribeTimeToLive"
+                    ],
+                    "Resource": "*"
```
paulotten, line 280: Remove all of these ^

provisioner/src/lib.rs
```diff
@@ -233,6 +254,193 @@ impl MyProvisioner {
         }
     }
 
+    async fn get_prefix(&self, project_name: &str) -> String {
+        //TODO: add userid or something else unique here
+        format!("shuttle-dynamodb-{}-", project_name)
```
paulotten, line 259: Need to be 100% sure that users can't overlap these prefixes. (Project id or uid probably.)

provisioner/src/lib.rs
```diff
@@ -236,6 +261,269 @@ impl MyProvisioner {
...
+
+        self.create_iam_identity(&prefix).await?;
+
+        self.attach_user_policy(&prefix).await?;
+
+        let (aws_access_key_id, aws_secret_access_key)= self.get_iam_identity_keys(&prefix).await?;
+
+        let aws_default_region = self.dynamodb_client.conf().region().ok_or_else(|| Error::GetRegion("empty region".to_string()))?.to_string();
+
+        //TODO:
```
paulotten, line 500: TODO

provisioner/src/lib.rs
```diff
@@ -236,6 +261,269 @@ impl MyProvisioner {
...
+    fn get_access_key_file_name(&self, prefix: &str) -> String {
+        format!("{}.txt", prefix)
+    }
+
+    async fn delete_saved_access_key(&self, prefix: &str) -> Result<(), std::io::Error> {
+        std::fs::remove_file(self.get_access_key_file_name(prefix))?;
+        Ok(())
+    }
+
+    async fn save_access_key(&self, prefix: &str, access_key_id: &str, secret_access_key: &str) -> Result<(), std::io::Error> {
```
paulotten, line 409: Question for Shuttle devs: where and how would you like us to store these access keys?

Currently we're using .txt files. Presumably that's not the desired end solution though.

paulotten:
Matching PR for the new example is here: https://github.com/shuttle-hq/shuttle-examples/pull/55

paulotten:
```
#!/bin/bash -eo pipefail
git submodule update --init
Submodule 'examples' (git@github.com:shuttle-hq/examples.git) registered for path 'examples'
Cloning into '/home/circleci/project/examples'...
fatal: remote error: upload-pack: not our ref 9885c59b698cedef4b740bf701c967a2565c8904
fatal: Fetched in submodule path 'examples', but it did not contain 9885c59b698cedef4b740bf701c967a2565c8904. Direct fetching of that commit failed.

Exited with code exit status 128
CircleCI received exit code 128
```

These failed CI tests appear to be because we updated the examples git submodule. The PR for that is here: https://github.com/shuttle-hq/shuttle-examples/pull/55

It's a bit of a "chicken or the egg" problem about which to merge first.

cargo-shuttle/src/provisioner_server.rs
```diff
@@ -412,3 +619,107 @@ fn db_type_to_config(db_type: Type) -> EngineConfig {
...
+            .unwrap();
+
+        std::env::set_var("AWS_ACCESS_KEY_ID", info.aws_access_key_id);
+        std::env::set_var("AWS_SECRET_ACCESS_KEY", info.aws_secret_access_key);
+        std::env::set_var("AWS_REGION", info.aws_default_region);
+
+        let aws_config = aws_config::from_env()
+            .endpoint_url(info.endpoint.unwrap())
+            .load()
+            .await;
```
chesedo, line 693: I suspect the environment variables might not be needed. Using `aws_config::ConfigLoader::default().enpoint_url(...)` might avoid confusion here.

cargo-shuttle/src/provisioner_server.rs
```diff
@@ -412,3 +619,107 @@ fn db_type_to_config(db_type: Type) -> EngineConfig {
...
+            .delete_dynamodb_tables_by_prefix_in_container(&project_name)
+            .await
+            .unwrap();
+
+        // select from table (should fail now that tables have been deleted)
+        let result = select_from_table(&dynamodb_client, &table_name).await;
+
+        assert!(result.is_err());
+
+        println!("{result:?}");
```
chesedo, line 723:
```suggestion
        assert!(result.is_err(), "expected result to be an error, but found: {result:?}");
```

runtime/tests/integration/helpers.rs
```diff
@@ -105,4 +106,18 @@ impl Provisioner for DummyProvisioner {
...
+        _request: Request<DynamoDbRequest>,
+    ) -> Result<Response<DynamoDbResponse>, Status> {
+        panic!("did not expect any runtime test to use dynamodb");
+    }
+
+    async fn delete_dynamo_db(
+        &self,
+        _request: Request<DynamoDbRequest>,
+    ) -> Result<Response<DynamoDbDeletionResponse>, Status> {
+        panic!("no run tests should request delete a dynamodb");
```
chesedo, line 121:
```suggestion
        panic!("no run tests should delete a dynamodb");
```

provisioner/src/lib.rs
```diff
@@ -236,6 +258,294 @@ impl MyProvisioner {
...
+
+        // 43 characters long (4 characters correspond to 3 bytes of data)
+        // sha256 is 32 bytes. 32 / 3 * 4 ~ 43
+        // we care about this because various AWS identifiers have specific length constraints
+        Base64UrlUnpadded::encode_string(&hash)
+    }
+
+    async fn get_dynamodb_policy_name(&self, prefix: &str) -> String {
+        format!("{}-dynamo-policy", prefix)
+    }
```
chesedo, line 276: These methods don't use `&self`, maybe they should be outside of the struct `impl`?

provisioner/src/lib.rs
```diff
@@ -236,6 +258,294 @@ impl MyProvisioner {
...
+            .await
+            .map_err(Error::DeleteIAMPolicy)?;
+
+        Ok(())
+    }
+
+    async fn get_iam_identity_user_name(&self, prefix: &str) -> String {
+        // max characters is 64
+        format!("{}-dynamo-user", prefix)
+    }
```
chesedo, line 366: This could also be outside the struct `impl`

provisioner/src/lib.rs
```diff
@@ -236,6 +258,294 @@ impl MyProvisioner {
...
+        access_key_id: &str,
+        secret_access_key: &str,
+    ) -> Result<(), std::io::Error> {
+        use std::io::prelude::*;
+        let mut file = File::create(self.get_access_key_file_name(prefix))?;
+        let contents = format!("{}\n{}", access_key_id, secret_access_key);
+        file.write_all(contents.as_bytes())?;
+
+        Ok(())
+    }
```
chesedo, line 422: These could also be outside the struct `impl`.

provisioner/src/lib.rs
```diff
@@ -236,6 +258,294 @@ impl MyProvisioner {
...
+
+    pub async fn request_dynamodb(&self, project_name: &str) -> Result<DynamoDbResponse, Error> {
+        let prefix = self.get_prefix(project_name).await;
+
+        self.create_dynamodb_policy(&prefix).await?;
+
+        self.create_iam_identity(&prefix).await?;
+
+        self.attach_user_policy(&prefix).await?;
```
chesedo, line 514: I'm wondering, since all these methods just need the prefix and the IAM client, maybe all this logic can move to a separate struct. This struct would have a small lifetime and would be constructed here?

I'm mostly trying to find a way to make this current struct's `impl` simpler and move the new common logic to one place.

provisioner/src/lib.rs
```diff
@@ -522,3 +860,153 @@ fn engine_to_port(engine: aws_rds::Engine) -> String {
...
+    #[tokio::test]
+    async fn test_create_and_delete_dynamodb_policy() {
+        let provisioner = make_test_provisioner().await;
+
+        let prefix = provisioner.get_prefix("my_cool_project").await;
+
+        provisioner.create_dynamodb_policy(&prefix).await.unwrap();
+
+        provisioner.delete_dynamodb_policy(&prefix).await.unwrap();
+    }
```
chesedo, line 928: Nice!!! Will they also run successfully when a machine / CI does not have AWS credentials set? In which case the `ignore` attribute just has to be added.

chesedo (changes requested):
Great one @mikegin and @paulotten!! I left a few comments and suggestions.

Mind reminding me, the saving of the key is only needed the second time an existing DynamoDB table is being requested by a user right?

paulotten:
> Mind reminding me, the saving of the key is only needed the second time an existing DynamoDB table is being requested by a user right?

@chesedo The key is tied to the IAM identity. Both are created the first time the dynamoDB resource is requested by the `#[shuttle_dynamodb::DynamoDB]` attribute. They are then reused each further time.

cargo-shuttle/src/provisioner_server.rs
```diff
@@ -325,6 +512,26 @@ struct EngineConfig {
...
+}
+
+fn dynamodb_config() -> DynamoDbConfig {
+    DynamoDbConfig {
+        container_name: "shuttle_dynamodb".to_string(),
+        image: "amazon/dynamodb-local:latest".to_string(),
+        port: "8000/tcp".to_string(),
+        aws_access_key_id: "DUMMY_ID_EXAMPLE".to_string(),
+        aws_secret_access_key: "DUMMY_EXAMPLE_KEY".to_string(),
+        aws_default_region: "DUMMY_EXAMPLE_REGION".to_string(),
```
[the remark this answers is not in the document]
paulotten: The AWS config library requires these values to be set, even though they aren't used when running against a local dynamoDB docker container endpoint.

cargo-shuttle/src/provisioner_server.rs
paulotten: > I suspect the environment variables might not be needed.

You'll fail validation by `aws_config` without them.

The closest AWS documentation I can quickly find on this is

> The YAML scripts require that you specify an AWS access key and an AWS secret key, but they are not required to be valid AWS keys for you to access DynamoDB local.

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.DownloadingAndRunning.html

paulotten:
@mikegin did you have a list of all of the AWS permissions the (cloud) provisioner needs in order to create/attach/delete the required IAM identities and roles?

provisioner/src/lib.rs
paulotten: Yep, this test and the following ones don't work without AWS credentials. I'll `ignore` them.

paulotten: Done

cargo-shuttle/src/provisioner_server.rs
jonaro00: @paulotten I did this like yesterday, found out you don't need them. Had to add a bunch of deps tho. https://github.com/jonaro00/wallace-minion/commit/257f38e4aef873374a4a6acabce8cfcd3c55f09f#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR45

paulotten:
Of the failed CI jobs, this is the only one that looks like it's from us

> cargo-shuttle

```
thread 'provisioner_server::tests::test_create_and_delete_dynamodb' panicked at 'called `Result::unwrap()` on an `Err` value: DispatchFailure(DispatchFailure { source: ConnectorError { kind: Other(Some(TransientError)), source: hyper::Error(IncompleteMessage), connection: Unknown } })', cargo-shuttle/src/provisioner_server.rs:655:14
test provisioner_server::tests::test_create_and_delete_dynamodb ... FAILED
test config::tests::setting_name_overrides_name_in_config ... ok
test tests::make_archive_ignore_target_folder ... ok
test tests::make_archive_respect_ignore ... ok
test args::tests::project_name ... ok
test tests::make_archive_include_secrets ... ok
test args::tests::project_name_in_workspace ... ok
test tests::load_project_returns_proper_working_directory_in_project_args ... ok
test config::tests::get_local_config_finds_name_from_workspace_dir ... ok

failures:

failures:
    provisioner_server::tests::test_create_and_delete_dynamodb
```

Running it locally it looks like the AWS local testing docker container needs some extra after it is initially created before it's ready to handle requests. Time to add a sleep somewhere...

oddgrd:
We use the [`wait_for_ready`](https://github.com/shuttle-hq/shuttle/blob/main/cargo-shuttle/src/provisioner_server.rs#L168) function for the other ones, we should be able to use that for dynamo db as well, right? See the `get_db_connection_string` function.

oddgrd:
> > We use the [wait_for_ready](https://github.com/shuttle-hq/shuttle/blob/main/cargo-shuttle/src/provisioner_server.rs#L168) function for the other ones, we should be able to use that for dynamo db as well, right?
> 
> Best I seem to be able to do is use curl to check if DynamoDB is listening on it's port yet. Fortunately this seems to be good enough.

Nice, I think that should be fine!

resources/dynamodb/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "shuttle-dynamodb"
+version = "0.17.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+async-trait = "0.1.56"
+serde = { version = "1.0.148", features = ["derive"] }
+shuttle-service = { path = "../../service", version = "0.17.0", default-features = false }
```
oddgrd, line 11:
```suggestion
shuttle-service = { path = "../../service", version = "0.18.0", default-features = false }
```

provisioner/src/lib.rs
```diff
@@ -457,6 +555,301 @@ impl Provisioner for MyProvisioner {
...
+
+        let request = request.into_inner();
+
+        let reply = self.delete_dynamodb(&request.project_name).await?;
+
+        Ok(Response::new(reply))
+    }
+}
+
+struct DynamoDBHandler {
```
Shubham8287, line 588: I am thinking if it will be a good idea to separate Identity, policy creation and management part from this Handler, so that can be independently re-used for other future AWS resouces.
For one such instance - while working on s3 resource PR, [s3-resource](https://github.com/shuttle-hq/shuttle/pull/995/files#diff-9aa440c73eedfd63ea9f7221d4be2d948eb7af1bffaadd9eec69a379ab1c9cadR630) I am doing similar stuff of creating identity, creating policy and attaching.

provisioner/src/lib.rs
```diff
@@ -457,6 +555,301 @@ impl Provisioner for MyProvisioner {
...
+
+    async fn save_access_key(
+        &self,
+        access_key_id: &str,
+        secret_access_key: &str,
+    ) -> Result<(), std::io::Error> {
+        use std::io::prelude::*;
+        let mut file = File::create(self.get_access_key_file_name())?;
+        let contents = format!("{}\n{}", access_key_id, secret_access_key);
+        file.write_all(contents.as_bytes())?;
```
Shubham8287, line 836: @mikegin , @chesedo This is to enhance my knowledge but not comment on a PR. Sorry for spamming here.
As I am not well aware of shuttle deployment architecture, I am having a little hard time understanding how writing data to a file in a container will provide persistence. What if the provisioner container gets redeployed due to some reason? or are we okay to lose this data and the file is just a good to have cache?

iulianbarbu: The provisioner will have its own docker volume, which will persist across invocations. This wasn't the case before but will be introduced through this PR.

iulianbarbu:
@mikegin @paulotten FYI, I synced this branch with `main` and fixed the tests.

service/src/lib.rs
```diff
@@ -6,12 +6,67 @@ use async_trait::async_trait;
...
-    DatabaseReadyInfo, DbInput, DbOutput, SecretStore,
+    DatabaseReadyInfo, DbInput, DbOutput, DynamoDBInput, DynamoDbReadyInfo, SecretStore,
 };
 
 pub mod error;
 pub use error::{CustomError, Error};
 
+#[cfg(feature = "codegen")]
+extern crate shuttle_codegen;
+#[cfg(feature = "codegen")]
```
chesedo, line 17: Is this update correct? Won't it undo another PR?

cargo-shuttle/Cargo.toml
```diff
@@ -54,6 +54,8 @@ url = "2.3.1"
 uuid = { workspace = true, features = ["v4"] }
 webbrowser = "0.8.2"
 semver = "1.0.17"
+aws-sdk-dynamodb = { workspace = true }
+aws-config = { workspace = true }
```
chesedo, line 58: We should confirm these dependencies are needed in the CLI. I think only the tests use them :man_shrugging:

chesedo (commented):
@iulianbarbu I left two comments. This seems about ready, thanks!

iulianbarbu (commented):
Thanks @chesedo ! Resolved the comments.

oddgrd:
Hey @mikegin and @paulotten, we're going to close this as blocked until we can figure out how we will securily provision this resource. Thanks for all your hard work on this, we're sorry we couldn't get this over the line, it looked really great, and we'll likely return to your implementation in the future when we decide on the design. This PR also made it very clear which problem we need to solve next, so thanks for that too!

Issue #844, opened 2023-05-02
[Feature]: Refactor database resources to no longer cycle passwords (ENG-551)


ankitpatel96:
## Description

Historically we never stored the credentials for DBs and this meant we had to cycle the credentials for every new deploy.  But since we now stored the credentials, we no longer need to cycle them.  Cycling the RDS credentials also takes 1+ minutes.

## Requirements:

- refactor the functions for provisioning a database user in provisioner/lib.rs so that if a user already exists the password is no longer cycled
- Update deployer to always send down the details of persisted resources

### Suggestion or Example of how the feature would be used

@Shariffintech and I will be working on this

### Duplicate declaration

- [X] I have searched the issues and this feature has not been requested before.

Shariffintech:
@oddgrd could you also add me as a participant on this ticket?

oddgrd:
Hey @Shariffintech and @ankitpatel96, are you still interested in this issue? If not I'll just open it up again :slightly_smiling_face:

oddgrd:
Hey! So this issue has been resolved, we will now get the resource information from the resource-recorder if it already exists, meaning the call to provisioner that lead to the cycling of passwords no longer occurs. However, we might need to create a new command that allows users to cycle their database password, in case their password is leaked. We're currently working on resource deletion, but we'll discuss this internally and open a new ticket as soon as we have the capacity to review it.

Shariffintech:
🫡 Thank you for the update @oddgrd! I'm glad to hear that the problem has been addressed. Looking forward to contributing to the new command for resource deletion whenever it's the right time, happy to assist.

PR #912, opened 2023-05-12
[shared_mongo][Feature]: Refactor database resources to no longer cycle passwords (ENG-551) #844

Shariffintech:
## Description of change

This PR is a reference to #844.

Historically we never stored the credentials for DBs and this meant we had to cycle the credentials for every new deploy. But since we now stored the credentials, we no longer need to cycle them. Cycling the RDS credentials also takes 1+ minutes.

Requirements:
refactor the functions for provisioning a database user in provisioner/lib.rs so that if a user already exists the password is no longer cycled
Update deployer to always send down the details of persisted resources

1. I first refactored the logic of recycling a users password by removing it => Check if the user already exists before creating a new one. If the user exists, return the existing credentials without changing the password. 
2. Replacing the logic handle, if a new user is not an existing user the create the user otherwise bubble an error.
3. To handle the AlreadyExists error case I modified the error enum and included a variant.
4. In the shared_mongodb function I returned the AlreadyExists error variant for the of when the user already exists.
5. Refactored the test for function shared_mongodb to remove the password cycling  checks => dereference the new_user_stored_key variable to prevent a double borrow when comparing the stored key for a new user to the stored key for an existing user

Be sure to reference any related issues by adding `closes issue #`.

This refactor has been tested through running 

`cargo test --package shuttle=provisioner --all-features --test '*' -- --nocapture`

and here was the key output:
Screenshot 2023-05-12 at 4 57 54 PM

All test passing related to shared_mong_db role in provisioner/src/lib.rs , tests/provisioner

### DIFF
--- a/runtime/src/alpha/mod.rs
+++ b/runtime/src/alpha/mod.rs
@@ -84,7 +84,10 @@ pub async fn start(loader: impl Loader<ProvisionerFactory> + Send + 'static) {
         server_builder.add_service(svc)
     };
 
-    router.serve(addr).await.unwrap();
+    match router.serve(addr).await {
+        Ok(_) => {}
+        Err(e) => panic!("Error while serving address {addr}: {e}"),
+    };
 }
 
 pub struct Alpha<L, S> {


iulianbarbu:
Hey @Shariffintech ! Just a remainder for you on the message @oddgrd left above. It's hard for us to spot the changes you've made in relation to #844 .

Shariffintech:
Hey @iulianbarbu @oddgrd thanks for catching this and apologies on the delay. This was a mistake, I'm looking into my commits and where I went wrong.

Issue #913, opened 2023-05-13
[Feature]: Support custom features (`--features`) for `run` and `deploy` commands


orhun:
### Describe the feature

In my project, I'm using a separate cargo feature for enabling the shuttle builds:

```toml
[features]
default = []
shuttle = [
  "shuttle-actix-web",
  "shuttle-runtime",
  "shuttle-static-folder",
  "tokio",
]
```

However I can't use this feature for `cargo shuttle` thus any shuttle command that I run fails.

### Suggestion or Example of how the feature would be used

Add an argument for `cargo shuttle` subcommands for enabling custom features. For example:

```sh
$ cargo shuttle run --features shuttle
```

```
$ cargo shuttle deploy --features shuttle
```

I'm happy to work on a PR if it is applicable.

### Duplicate declaration

- [X] I have searched the issues and this feature has not been requested before.

orhun:
I added `--features` argument to `run` in #914 

I'm not sure how `deploy` subcommand works so I need some guidance there 🐻

jonaro00:
Custom feature flags were added in https://github.com/shuttle-hq/shuttle/releases/tag/v0.46.0

orhun:
hell yea

Issue #751, opened 2023-03-25
[Feature]: Support for choosing binary target


akhildevelops:
### Describe the feature

It would be helpful to have an option like `cargo shuttle deploy --bin <bin_name>` similar to running a specific binary target from cargo: `cargo run --bin <bin_name>`.

This will help to write multiple binaries for specific environment. Ex: A specific binary for shuttle environment and another binary that has it's own server.

Ref: https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries

### Duplicate declaration

- [X] I have searched the issues and this feature has not been requested before.

jonaro00:
I like it.
This could also be a field in `Shuttle.toml` like
```toml
name = "project"
bin = "shuttle"
```
Where `Cargo.toml` is something like
```toml
[[bin]]
name = "shuttle"

[[bin]]
name = "standalone"
```

Shubham8287:
@oddgrd Is it open to take? I will love to contribute to shuttle :)

jonaro00:
Sadly, it seems like this is the best we can do at the moment. https://github.com/shuttle-hq/shuttle-examples/tree/main/other/standalone-binary

PR #914, opened 2023-05-13
 feat: add `--features` argument to `cargo shuttle run`

orhun:
## Description of change

This PR adds `--features` argument to `run` subcommand for activating cargo features.

See #913 for more detail.

I ran it locally, worked fine.

### DIFF
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -229,6 +229,9 @@ pub struct RunArgs {
     /// Use 0.0.0.0 instead of localhost (for usage with local external devices)
     #[arg(long)]
     pub external: bool,
+    /// Set the list of features to activate
+    #[arg(long)]
+    pub features: Vec<String>,
     /// Use release mode for building the project
     #[arg(long, short = 'r')]
     pub release: bool,
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -793,7 +793,14 @@ impl Shuttle {
         );
 
         // Compile all the alpha or shuttle-next services in the workspace.
-        build_workspace(working_directory, run_args.release, tx, false).await
+        build_workspace(
+            working_directory,
+            run_args.features.clone(),
+            run_args.release,
+            tx,
+            false,
+        )
+        .await
     }
 
     async fn setup_local_provisioner(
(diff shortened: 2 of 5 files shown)


jonaro00:
It's a good feature request.
It would be suitable to make it work for deployments as well. Not sure if this complicates things.
It's worth considering how this affects #756, but it seems completely fine.

orhun:
> It would be suitable to make it work for deployments as well. Not sure if this complicates things.

I was going to implement it for deployments as well but couldn't figure out how `deploy` command works. I guess the build happens somewhere else since it only sends a request along with a compressed archive, right? I would be happy to go ahead and implement it if I can get a better understanding of the internals.

> It's worth considering how this affects #756, but it seems completely fine.

Oh, I think it is easily convertable to execute the `cargo` command instead, so I agree.

jonaro00:
It would probably work to add them as a URL parameter, like `no_test` is now.

But I remembered there is another challenge:
In the PR now, all shuttle crates in the workspace gets the same feature flags set.
Specifying flags to individual crates becomes a challenge.
Related:
- similar request as the --features: https://github.com/shuttle-hq/shuttle/issues/751
- unresolved questions about workspace: https://github.com/shuttle-hq/shuttle/pull/767#discussion_r1151750730

iamwacko:
#756 has been merged now, so you will have to make some changes.

oddgrd:
Hey @orhun, thanks for getting started on this! Apologies for the delayed review, but we're in the middle of a major refactor of how we build crates and deploy users projects. I think it makes sense to hold off on this until that is all merged. The PR for the builder should be up within a few days!

orhun:
Refactored the code to use the new cargo build mechanism. Let me know if it looks okay!

I might provide another PR for the deploy subcommand.

jonaro00:
There was a discussion about this the other day, regarding design of multi-service deploys.
Sadly, it looks like shuttle services will require the default binary and features to be the shuttle version of the crate.
It does not only apply to `--features`, so supporting customization of this sort becomes a catch up game with cargo's `build` and `run`

orhun:
I see, in that case, what exactly is needed?

jonaro00:
Something like this https://github.com/shuttle-hq/shuttle-examples/pull/61/files

orhun:
Unfortunately I'm still not following. I guess I will wait a bit until things stabilize 😅

jonaro00:
Yeah, with the major deployment rewrite coming up there are a lot of question marks from me as well.

oddgrd:
I'll close this for now @orhun, we can pick it up again when we're implementing the builder, but it will likely take a different shape.

PR #915, opened 2023-05-14
suggest next logical command

### LINKED ISSUE #898: [Feature]: suggest the next logical command (if relevant) on success of `cargo shuttle` subcommands
### Describe the feature

Some commands like `cargo shuttle project start` would almost always be followed by `cargo shuttle deploy`. The final output of `cargo shuttle project start` might as well therefore tell the user that they can run `cargo shuttle deploy` next to deploy their project.

_No response_

### Duplicate declaration

SyedFasiuddin:
closes #898

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -236,6 +236,8 @@ impl Shuttle {
 
             self.load_project(&mut project_args)?;
             self.project_create(&self.client()?, IDLE_MINUTES).await?;
+        } else {
+            println!("Run `cargo shuttle project start` to create project environment on Shuttle");
         }
 
         Ok(())
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -608,6 +608,14 @@ impl Shuttle {
             addr
         );
 
+        println!("
+    To host the application on Shuttle run
+       `cargo shuttle project start`
+    if project environment is not created previously and then run
+       `cargo shuttle deploy`
+    Your application will be hosted at https://{}.shuttleapp.rs
+", service_name);
+
         let start_request = StartRequest {
             ip: addr.to_string(),
         };
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -1020,6 +1020,7 @@ impl Shuttle {
             client,
         )
         .await?;
+        println!("Run `cargo shuttle deploy` to deploy your Shuttle service.");
 
         Ok(())
     }
@@ -1076,6 +1077,7 @@ impl Shuttle {
             client,
         )
         .await?;
+        println!("Run `cargo shuttle project start` to recreate project environment on Shuttle.");
 
         Ok(())
     }
@@ -1103,7 +1105,6 @@ impl Shuttle {
         }
         progress_bar.finish_and_clear();
         println!("{project}");
-        println!("Run `cargo shuttle deploy` to deploy your Shuttle service.");
         Ok(())
     }
 
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -791,6 +791,11 @@ impl Shuttle {
             }
         }
 
+        println!(
+            "Run `cargo shuttle project start` to create a project environment on Shuttle.\n\
+             Run `cargo shuttle deploy` to deploy your Shuttle service."
+        );
+
         // If prior signal received is set to true we must stop all the existing runtimes and
         // exit the `local_run`.
         if signal_received {
@@ -802,10 +807,6 @@ impl Shuttle {
                         trace!(status = ?err, "stopping the runtime errored out");
                     });
             }
-            println!(
-                "Run `cargo shuttle project start` to create a project environment on Shuttle.\n\
-                 Run `cargo shuttle deploy` to deploy your Shuttle service."
-            );
             return Ok(());
         }
 
@@ -853,13 +854,6 @@ impl Shuttle {
                     true
                 }
             };
-        }
-
-        if signal_received {
-            println!(
-                "Run `cargo shuttle project start` to create a project environment on Shuttle.\n\
-                 Run `cargo shuttle deploy` to deploy your Shuttle service."
-            );
         }
 
         Ok(())
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -243,8 +243,7 @@ impl Shuttle {
             self.project_create(&self.client()?, IDLE_MINUTES).await?;
         } else {
             println!(
-                "Run `cargo shuttle project start` to create a project environment on Shuttle.\n\
-                 Run `cargo shuttle deploy` to deploy your Shuttle service."
+                "Run `cargo shuttle project start` to create a project environment on Shuttle."
             );
         }
 
@@ -803,6 +802,10 @@ impl Shuttle {
                         trace!(status = ?err, "stopping the runtime errored out");
                     });
             }
+            println!(
+                "Run `cargo shuttle project start` to create a project environment on Shuttle.\n\
+                 Run `cargo shuttle deploy` to deploy your Shuttle service."
+            );
             return Ok(());
         }
 
@@ -852,6 +855,13 @@ impl Shuttle {
             };
         }
 
+        if signal_received {
+            println!(
+                "Run `cargo shuttle project start` to create a project environment on Shuttle.\n\
+                 Run `cargo shuttle deploy` to deploy your Shuttle service."
+            );
+        }
+
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -606,6 +608,14 @@ impl Shuttle {
             addr
         );
 
+        println!("
+    To host the application on Shuttle run
+       `cargo shuttle project start`
+    if project environment is not created previously and then run
+       `cargo shuttle deploy`
+    Your application will be hosted at https://{}.shuttleapp.rs
+", service_name);
```
oddgrd, line 617: I'm not sure we need this message in local run, but it might be useful :thinking: Either way we should not show the URL the project will be hosted at here, because they may use a name that is already taken, but that won't fail in local runs, it will only fail when they run `cargo shuttle project start`.

cargo-shuttle/src/lib.rs
```diff
@@ -236,6 +236,8 @@ impl Shuttle {
 
             self.load_project(&mut project_args)?;
             self.project_create(&self.client()?, IDLE_MINUTES).await?;
+        } else {
+            println!("Run `cargo shuttle project start` to create project environment on Shuttle");
```
oddgrd, line 240: Nice! Perhaps we should also have a message about running `cargo shuttle deploy` after this else block?

oddgrd (commented):
Nice, thanks for picking this up @SyedFasiuddin! I think we can also add a message to the deployment `stop` command, something like "Run `cargo shuttle deploy` to re-deploy your service". Perhaps we can also add one to `deployment list` about how you can use the deployment id in the `cargo shuttle logs <id>` command to get the logs for a given deployment.

cargo-shuttle/src/lib.rs
SyedFasiuddin: The users might do the local run multiple time, over and over again; do you think this message will get annoying to see every time? If then then we should remove this but the user already see a lot of output from cargo, we could do however print this to stderr, but there are no `eprintln` in the entire file just `println`.

SyedFasiuddin:
When `cargo shuttle stop` is ran the service stays deployed but is inactive i.e. stopped, is this right?

How am I now supposed to start the service back? Do I need to run `cargo shuttle deploy` again?
There is no `cargo shuttle start` to start a stopped service, perhaps we could have a `cargo shuttle start <id>` to do the same.

cargo-shuttle/src/lib.rs
SyedFasiuddin: Not after the else block but inside the else block, as `project_create` will call `wait_with_spinner` which already has the `cargo shuttle deploy` line printed.

SyedFasiuddin:
The git history looks bad sorry. Can you cherrypick commits on github? 😄

cargo-shuttle/src/lib.rs
```diff
@@ -623,6 +630,11 @@ impl Shuttle {
             addr
         );
 
+        println!(
+            "Run `cargo shuttle project start` to create a project environment on Shuttle.\n\
+                  Run `cargo shuttle deploy` to deploy your Shuttle service."
+        );
```
iulianbarbu, line 636: I wouldn't add this here or for anything related to the local run. The local run is blocking until the program is stopped (through signals), so it doesn't make too much sense in my opinion to print the next commands at the end of the command after it's stopped by a signal.

iulianbarbu:
> When `cargo shuttle stop` is ran the service stays deployed but is inactive i.e. stopped, is this right?

Correct.

> How am I now supposed to start the service back? Do I need to run `cargo shuttle deploy` again? There is no `cargo shuttle start` to start a stopped service, perhaps we could have a `cargo shuttle start <id>` to do the same.

A `cargo shuttle project start` is what you're looking for. This will restart the environment for the project deployment.

SyedFasiuddin:
As @oddgrd mentioned here
> I think we can also add a message to the deployment `stop` command, something like "Run `cargo shuttle deploy` to re-deploy your service". 

we could use `cargo shuttle deploy` and you are suggesting `cargo shuttle project start` ??

iulianbarbu:
> As @oddgrd mentioned here
> 
> > I think we can also add a message to the deployment `stop` command, something like "Run `cargo shuttle deploy` to re-deploy your service".
> 
> we could use `cargo shuttle deploy` and you are suggesting `cargo shuttle project start` ??

Hmm, seems a bit confusing I agree. Probably there is some misunderstanding, but you can try it yourself as well. Testing this is doable by following the `CONTRIBUTING.md`. Doing a sequence of `project start`, `deploy`, `project stop`, `deploy` results on my local machine with:
```text
Error: 503 Service Unavailable
message: project not ready
```

But if I am doing a `project start` between the last `stop` and `deploy`, it's working as expected. This is an issue that I think was documented somewhere. I can't find though where.

SyedFasiuddin:
`project start` command will call `project_create` method where I can suggest to run `deploy`, thats fine.
`project stop` command will call `project_delete` method where I can suggest to run `project start`, thats fine.

But `project restart` command will call `project_delete` and then `project_create` and in this case we only want to suggest to `deploy` and not `project start`

Or should we just leave it?

SyedFasiuddin:
`project restart` will look something like this then:
```
$ cargo shuttle project restart
project '...' is destroyed
Run `cargo shuttle project start` to recreate project environment on Shuttle.
project '...' is ready
Run `cargo shuttle deploy` to deploy your Shuttle service.
```

Both those suggestions coming from `project_delete` and `project_create` respectively.

iulianbarbu:
> `project restart` will look something like this then:
> 
> ```
> $ cargo shuttle project restart
> project '...' is destroyed
> Run `cargo shuttle project start` to recreate project environment on Shuttle.
> project '...' is ready
> Run `cargo shuttle deploy` to deploy your Shuttle service.
> ```
> 
> Both those suggestions coming from `project_delete` and `project_create` respectively.

Love it :D

cargo-shuttle/src/lib.rs
iulianbarbu: There is also the issue of possibly printing this multiple times because locally we'll spin a runtime for each workspace service. This log might be handy for people, so maybe we can add it in the `local_run` function, but before waiting for runtimes to finish. This is somehow conflicting with what I said previously, but I see the point here of having the next logical command be `project start`/`deploy`.

cargo-shuttle/src/lib.rs
iulianbarbu: Maybe we can mark this as resolved. We've clarified how the `println` should be placed in relation to `wait_with_spinner`.

cargo-shuttle/src/lib.rs
```diff
@@ -1036,6 +1046,7 @@ impl Shuttle {
         }
         progress_bar.finish_and_clear();
         println!("{project}");
+        println!("Run `cargo shuttle deploy` to deploy the project");
```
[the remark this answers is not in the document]
iulianbarbu: This seems to be done as well. Resolving it.

cargo-shuttle/src/lib.rs
SyedFasiuddin: `local_run` both of unix and windows call `spin_local_runtime` and I have added those two suggestion in this method rather than adding it to both `local_run` methods. Anything wrong with this?

SyedFasiuddin: oops sorry `spin_local_runtime` gets called multiple times if we have multiple services, I didn't see this because of my limited testing on hello world shuttle service.

SyedFasiuddin:
If every thing is resolved then still I don't think we should merge this PR as adding a few println statements should not require 9 commits, it should have been a single commit. I you agree with me then I'll create another PR.

cargo-shuttle/src/lib.rs
```diff
@@ -796,6 +802,10 @@ impl Shuttle {
                         trace!(status = ?err, "stopping the runtime errored out");
                     });
             }
+            println!(
+                "Run `cargo shuttle project start` to create a project environment on Shuttle.\n\
+                 Run `cargo shuttle deploy` to deploy your Shuttle service."
+            );
```
iulianbarbu, line 808: I would add this `println!` right above the `if signal_received`. We want to display it when doing a `local_run`, but not only if a signal was sent.

cargo-shuttle/src/lib.rs
```diff
@@ -845,6 +855,13 @@ impl Shuttle {
             };
         }
 
+        if signal_received {
+            println!(
+                "Run `cargo shuttle project start` to create a project environment on Shuttle.\n\
+                 Run `cargo shuttle deploy` to deploy your Shuttle service."
+            );
+        }
```
iulianbarbu, line 863: This shouldn't be needed, based on the previous comment.

iulianbarbu (commented):
Thanks, we're almost there. 👍

iulianbarbu:
> If every thing is resolved then still I don't think we should merge this PR as adding a few println statements should not require 9 commits, it should have been a single commit. If you agree with me then I'll create another PR.

No worries, we can keep all these commits for now. Usually, the commits of a shuttle PR are squashed and merged by using GitHub, after the PR was approved. So merging into `main` will consist of a single commit at the end.

Screenshot 2023-05-17 at 10 08 39

SyedFasiuddin:
> So merging into `main` will consist of a single commit at the end.

Didn't knew about this. Thanks.

iulianbarbu (approved):
Thanks! LGTM.

PR #919, opened 2023-05-15
fix: deployment state shown as running on startup crash

oddgrd:
## Description of change

We were using a last_state var to track the last state received from the logs. When we received `Running` we broke the loop. We then check if there is a deployment running, but if it immediately crashed it won't be and we go to the `not entered the running state` branch. Here we print information based on the `last_state`, which would still be `Running`. 

I fixed this by rather fetching data about the deployment we started in the event that it isn't running after the logs finish. 

Tested by local deployer runs with a successful deploy of axum hello world, and a crashed deploy of rocket postgres.

Crashed output:
```sh
Deployment has not entered the running state

State: Crashed - Deployment crashed after startup.

Run the following for more details

cargo shuttle logs 8cad9582-7775-4fb3-aea3-dd4815840848
```

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -967,10 +967,6 @@ impl Shuttle {
             println!("{}", "Deployment has not entered the running state".red());
             println!();
 
-            let deployment = client
-                .get_deployment_details(self.ctx.project_name(), &deployment.id)
-                .await?;
-
             match deployment.state {
                 shuttle_common::deployment::State::Stopped => {
                     println!("State: Stopped - Deployment was running, but has been stopped by the user.")


cargo-shuttle/src/lib.rs
```diff
@@ -954,35 +947,49 @@ impl Shuttle {
...
+            Ok(CommandOutcome::Ok)
         } else {
             println!("{}", "Deployment has not entered the running state".red());
             println!();
-            match last_state {
-                Some(shuttle_common::deployment::State::Stopped) => {
+
+            let deployment = client
+                .get_deployment_details(self.ctx.project_name(), &deployment.id)
+                .await?;
```
iulianbarbu, line 972: Why should we query for it again?

iulianbarbu (commented):
LGTM with one quick question.

cargo-shuttle/src/lib.rs
oddgrd: Woops, that's a leftover from testing, thanks!

iulianbarbu (approved):

PR #922, opened 2023-05-15
feat(service): get rid of `cargo` dependency

### LINKED ISSUE #756: [Enhancement]: Get rid of `cargo` dependency

iamwacko:
## Description of change

`is_next`, `is_alpha`, `ensure_bin`, `ensure_cdylib`, and `clean_crate` no longer require `cargo`.

partially closes #756, particularly service part 1.

I ran `cargo test`, `cargo fmt`, and  `cargo clippy`.


iamwacko:
I have time, so I can do service part 2 as well.

iulianbarbu:
> I have time, so I can do service part 2 as well.

Sounds good @iamwacko ! Let us know if you need help. We will review this whenever it's ready.

service/src/builder.rs
```diff
@@ -291,22 +199,110 @@ fn ensure_binary(package: &Package) -> anyhow::Result<()> {
...
+    project_path: PathBuf,
+) -> anyhow::Result<Vec<BuiltService>> {
+    let jobs = std::thread::available_parallelism()?.get();
+    let manifest_path = project_path.join("Cargo.toml");
+
+    let mut cargo = std::process::Command::new("cargo");
+
+    cargo
+        .arg("build")
+        .arg("-j")
```
jonaro00, line 226: There was a limit of 4 jobs when running in release mode previously. Now it would be more suitable to have that limit only apply during deployment, and not when running release mode locally.

service/src/builder.rs
```diff
@@ -80,209 +75,122 @@ fn extract_shuttle_toml_name(path: PathBuf) -> anyhow::Result<String> {
...
-            targets: Vec::new(),
-            requested_profile,
-            profile_specified: true,
-            doc: false,
-        };
-
-        clean(&ws, &opts).unwrap();
+        stdout_write.write_all(&output.clone().stdout).unwrap();
+        stderr_write.write_all(&output.stderr).unwrap();
+        status_write.write_all(status.as_bytes()).unwrap();
```
jonaro00, line 155: A tokio oneshot could probably be used here.

service/src/builder.rs
```diff
@@ -80,209 +75,122 @@ fn extract_shuttle_toml_name(path: PathBuf) -> anyhow::Result<String> {
...
+        .manifest_path(&manifest_path)
+        .exec()?;
+    trace!("Cargo metadata parsed");
 
     let mut alpha_packages = Vec::new();
     let mut next_packages = Vec::new();
 
-    for member in ws.members() {
+    for member in metadata.workspace_packages() {
+        println!("{}", member.name);
```
jonaro00, line 97: a debug print?

service/src/builder.rs
```diff
@@ -291,22 +199,110 @@ fn ensure_binary(package: &Package) -> anyhow::Result<()> {
...
+    let mut cargo = std::process::Command::new("cargo");
+
+    cargo
+        .arg("build")
+        .arg("-j")
+        .arg(jobs.to_string())
+        .arg("--manifest-path")
+        .arg(manifest_path);
+
+    for package in packages.clone() {
```
jonaro00, line 231: You can probably iterate over `&packages` here, and the last loop probably doesn't need a clone.

jonaro00 (commented):
It's hard to accurately review Rust code without rust-analyzer at hand, but here are some thoughts I had. Don't worry if I'm wrong about something.

service/src/builder.rs
```diff
@@ -80,209 +75,122 @@ fn extract_shuttle_toml_name(path: PathBuf) -> anyhow::Result<String> {
...
-                    if let Err(error) = tx.send(message) {
-                        error!("failed to send cargo message on channel: {error}");
-                    }
-                }
-                Err(error) => {
-                    error!("failed to parse cargo message: {error}");
-                }
-            }
-        }
-    });
```
iulianbarbu, line 104: We should have something similar to this when getting output from `cargo` because otherwise we'll not be able to provide compilation data when doing local run [1] and [2] the compilation logs are also important for the deployer, because they're persisted and retrieved by user when accessing deployment logs or when they subscribe to the deployment logs websocket.

[1] https://github.com/shuttle-hq/shuttle/blob/d0b2f32cdfeba02b7fb145c11d8df57d0ad9a3ba/cargo-shuttle/src/lib.rs#L716
[2]

iulianbarbu (changes requested):

iulianbarbu:
@iamwacko  let me know if you have questions on the changes I proposed and we can set up a call to get through them. The PR is on the right path and we'll just need to polish a bit the code async-wise, while keeping it on par with the old functionality.

We have a codebase walkthrough session next Monday (if that helps) and a mentoring session next Wednesday, but feel free to hit me up with qs in private on Discord if you feel stuck.

service/src/builder.rs
```diff
@@ -80,209 +75,122 @@ fn extract_shuttle_toml_name(path: PathBuf) -> anyhow::Result<String> {
...
-
-    let config = get_config(write)?;
     let manifest_path = project_path.join("Cargo.toml");
-    let ws = Workspace::new(&manifest_path, &config)?;
-    check_no_panic(&ws)?;
```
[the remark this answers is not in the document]
iulianbarbu: Actually, maybe not. @chesedo , can you expand a bit on the comment here: https://github.com/shuttle-hq/shuttle/issues/756#issuecomment-1488168471?

service/src/builder.rs
```diff
@@ -80,209 +75,122 @@ fn extract_shuttle_toml_name(path: PathBuf) -> anyhow::Result<String> {
...
+    if release_mode {
+        profile = "release";
+    }
+
+    // It is easier just to use several pipes
+    let (mut stderr_read, mut stderr_write) = pipe::pipe();
+    let (mut stdout_read, mut stdout_write) = pipe::pipe();
+    let (mut status_read, mut status_write) = pipe::pipe();
 
     tokio::task::spawn_blocking(move || {
```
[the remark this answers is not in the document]
jonaro00: I think it might be blocking due to being used in the deployment handler. Say, if 10 people run this against production at the same time, it shouldn't block other async tasks.

service/src/builder.rs
iamwacko: oops

service/src/builder.rs
iamwacko: I don't see any way to determine if it is running locally or a deployment. Maybe we should just get rid of the `-j`?

service/src/builder.rs
```diff
@@ -291,22 +199,110 @@ fn ensure_binary(package: &Package) -> anyhow::Result<()> {
...
+fn compiler(
+    packages: Vec<&Package>,
+    release_mode: bool,
+    wasm: bool,
+    project_path: PathBuf,
+) -> anyhow::Result<Vec<BuiltService>> {
+    let jobs = std::thread::available_parallelism()?.get();
+    let manifest_path = project_path.join("Cargo.toml");
+
+    let mut cargo = std::process::Command::new("cargo");
```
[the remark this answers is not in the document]
iamwacko: Is there any point to making the compiling async if `cargo` is going to be using all of the CPU's anyway?

service/src/builder.rs
iamwacko: It looks like getting the same behavior as the deleted code block is a pain, but it looks like I could use [duct](https://crates.io/crates/duct/0.13.6) to do that.

service/src/builder.rs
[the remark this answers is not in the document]
iamwacko: @chesedo sorry to ping again, but this is holding this PR up.

service/src/builder.rs
iulianbarbu: It's more like when built locally you can assume it is done with a `dev` profile, but when in production, the profile should be `release`.

service/src/builder.rs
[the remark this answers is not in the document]
iulianbarbu: Hmm, at this moment the deployer is per project, so I think it would be more like 1 person made 10 requests against it. Even when this is the case, we usually queue these requests in the `deployer/deployment/queue.rs` logic, so I don't see why we would still need the `spawn_blocking` based on this.

In any case the current `spawn_blocking` is currently preceded by this comment:
```
// This needs to be on a separate thread, else deployer will block (reason currently unknown :D)
```

I think we still don't know why this needs to be blocking, but given we're working with a spawned `cargo` process now we should be able to start the process and then wait on its output asynchronously.

service/src/builder.rs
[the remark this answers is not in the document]
iulianbarbu: I think so. The compilation is going to use most of the CPU, but while its task can be blocked on IO (e.g. reading files, downloading crates) it can yield the execution to a task that is not blocked and make sure we extract the most out of the asynchronous runtime. This becomes more important when this part of the `service/src/builder` will be owned by a service which will handle the building for more than one project.

service/src/builder.rs
iulianbarbu: I think we don't need to have the exact same logic. I think we can try to redirect the `cargo` process stderr/stdout to the parent process stderr/stdout for local run while for deployer, we should make sure we extract both the stdout and stderr (separated or concatenated) into the deployer persistence.

service/src/builder.rs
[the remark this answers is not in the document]
iulianbarbu: Thanks @chesedo ! Resolving this.

Cargo.toml
```diff
@@ -78,7 +77,7 @@ strum = { version = "0.24.1", features = ["derive"] }
 tempfile = "3.4.0"
 thiserror = "1.0.37"
 tar = "0.4.38"
-tokio = { version = "1.22.0" }
+tokio = { version = "1.22.0", features = ["process"] }
```
jonaro00, line 80: This feature flag can be moved to service/Cargo.toml

service/src/builder.rs
```diff
@@ -291,22 +182,107 @@ fn ensure_binary(package: &Package) -> anyhow::Result<()> {
...
+    project_path: PathBuf,
+) -> anyhow::Result<Vec<BuiltService>> {
+    let manifest_path = project_path.join("Cargo.toml");
+
+    let mut cargo = tokio::process::Command::new("cargo");
+
+    cargo.arg("build").arg("--manifest-path").arg(manifest_path);
+
+    #[cfg(not(debug_assertions))]
+    cargo.arg("-j").arg(4.to_string());
```
jonaro00, line 209: This should not be added based on if the Shuttle crate is compiled in dev/release mode.
Ideally, the limit should only be imposed in deployment, so I suggest adding a bool `in_deployment` to this function and callers up the chain. The call from cargo-shuttle would use false and the one in deployer true (test cases could use false I guess).

jonaro00 (changes requested):
I found some more small things.

I read some more about tokio's Command.
It sounds like it is the proper replacement for spawn_blocking + std Command, so that should be fine.

service/src/builder.rs
oddgrd: Indeed, I think we did this before adding support for local runs with release profile. For context, the reason we set this for deploys is because cargo will think that the deployer has full to all of our VMs cores, even though it is limited by docker. Additionally using a lot of threads for compiling will increase peak memory usage for builds.

oddgrd (commented):
This looks pretty close @iamwacko! I left a few small comments, but I think the main thing we need to fix is sending the build logs to the `build_workspace` caller, which Iulian pointed out. Let us know if you are stuck on this and need any help!

service/src/builder.rs
iamwacko: I've made a bit of progress. I'm using `os_pipe` instead of `duct`, to minimize the number of needed dependencies. I'm basically just doing [the second example](https://github.com/oconnor663/os_pipe.rs#examples).

service/src/builder.rs
```diff
@@ -291,22 +199,139 @@ fn ensure_binary(package: &Package) -> anyhow::Result<()> {
...
+                paused = true;
             }
+            read = reader.read(&mut buf).unwrap();
+            if let Err(error) = tx.send(Message::TextLine(
+                String::from_utf8(buf.clone().to_vec()).unwrap(),
+            )) {
+                error!("failed to send cargo message on channel: {error}");
+            };
+        }
+    });
```
oddgrd, line 275: Can we rather read lines here? This should work, let me know if you tried this already and you found some issues with it!
```suggestion
    tokio::task::spawn_blocking(move || {
        let reader = std::io::BufReader::new(reader);
        for line in reader.lines() {
            if let Ok(line) = line {
                if let Err(error) = tx.send(Message::TextLine(line)) {
                    error!("failed to send cargo message on channel: {error}");
                };
            } else {
                // TODO: handle error
                println!("err reading line");
            }
        }
    });
```

oddgrd (commented):
Nice progress! I took a look at your latest changes and saw what may be an easier way to read the output, let me know if you think it could work!

service/src/builder.rs
iamwacko: Yeah we could, my weird solution is the result of doing this at midnight.

service/Cargo.toml
```diff
@@ -13,14 +13,14 @@ doctest = false
 [dependencies]
 anyhow = { workspace = true }
 async-trait = { workspace = true }
-cargo = { workspace = true, optional = true }
 cargo_metadata = { workspace = true, optional = true }
 crossbeam-channel = { workspace = true, optional = true }
+os_pipe = "1.1.4"
 pipe = { workspace = true, optional = true }
```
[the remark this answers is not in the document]
iamwacko: No, thanks for pointing that out

iulianbarbu (approved):
Thanks @iamwacko ! This is great 🥳 .

iulianbarbu:
I think there is a regression on `shuttle-next`. I am not able to run the `next/hello-world` because of:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: status: Unknown, message: "failed to read input file", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Mon, 05 Jun 2023 20:43:08 GMT", "content-length": "0"} }', /Users/iulianbarbu/repos/shuttle/cargo-shuttle/src/lib.rs:779:51
```

iamwacko:
I see the problem. - gets turned into _ by cargo, but I didn't account for that so it tries to execute hello-world.wasm instead of hello_world.wasm

iamwacko:
Now the `cargo-shuttle` test is failing with `thread 'init::non_interactive_basic_init' panicked at 'Could not query the latest version of shuttle-runtime', cargo-shuttle/src/init.rs:994:33`. It works locally, so I have no idea what has gone wrong.

iamwacko:
Looks like the only test failures I have are #821

iulianbarbu (approved):
Thanks @iamwacko for fixing the `shuttle-next` issue. Yes, it seems we're getting the error corresponding #821 in your tests. The tests pass locally for me as well.

PR #924, opened 2023-05-17
feat(shuttle-axum) Make AxumService generic to be able to use state

morlinbrot:
## Description of change
Made `AxumService` generic to be able to use it with `axum::State` and updated the `From` impl for it. This won't have any impact on existing usages as it defaults to the unit type, e.g.:
```rust
pub struct AxumService<S = ()>(pub axum::Router<S>);
```

I also created an [`axum/with-state` example](https://github.com/morlinbrot/examples/tree/feat/axum-with-state) to test this, let me know if you want this in the examples repo.

Built all the axum examples with it.

### DIFF
--- a/services/shuttle-axum/src/lib.rs
+++ b/services/shuttle-axum/src/lib.rs
@@ -18,7 +18,7 @@ use shuttle_runtime::{CustomError, Error};
 use std::net::SocketAddr;
 
 /// A wrapper type for [axum::Router] so we can implement [shuttle_runtime::Service] for it.
-pub struct AxumService(pub axum::Router);
+pub struct AxumService<S = ()>(pub axum::Router<S>);
 
 #[shuttle_runtime::async_trait]
 impl shuttle_runtime::Service for AxumService {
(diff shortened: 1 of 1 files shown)


ivancernja:
@joshua-mo-143 Heads-up on this, might require some changes to the articles, in a good way! :)

iulianbarbu:
This PR looks good to me. I have few notes:
1) Shuttle docs for axum go through some examples and I am wondering if an example with axum state is a good candidate for the page: https://docs.shuttle.rs/examples/axum.
2) It would definitely help bringing over the `axum-state` example to `shuttle-examples` after merging this.

So I would say that it would be great if we can get across associated PRs across `shuttle-docs` and `shuttle-examples` that reference this PR. I would not block this PR for the docs or example, but I think it would be an invisible feature if not documented and shown in action.

@morlinbrot what are your thoughts?

joshua-mo-143:
Hi, thanks for mentioning me @ivancernja. First of all, good work @morlinbrot.

Please correct if wrong but it looks that there's no immediate changes required to docs as the previous usage for state within Axum is basically the same as it is in the created example.

iulianbarbu:
> Please correct if wrong but it looks that there's no immediate changes required to existing docs beyond adding the example as the previous usage for state within Axum is basically the same as it is in the created example.

No immediate changes are required to the existing docs, only adding the new example as a new way to provide axum state. I think that before the alternative was to implement an extension (and still is), but seems like the state is the recommended way of doing it: https://docs.rs/axum/latest/axum/index.html#middleware.

morlinbrot:
I agree, @iulianbarbu, no change to the docs is required since it's strictly adding new functionality and the old way of doing things will keep working and this can be merged independent from my example or an update to the docs.

I will put up my example as a PR for the `shuttle-examples` repo, I'd be happy to add to or update the documentation, @joshua-mo-143, let me know if I should do that.

Issue #930, opened 2023-05-21
[Feature]: Log paths of all files included in deployment archive


jonaro00:
### Describe the feature

Many questions are posted regarding which files are uploaded to Shuttle.
The answer is quite simple (This should be listed in the docs btw), but it is hard to verify exactly which files are uploaded. The code for handling this is in an obscure place and not very straight forward.
Therefore, either cargo-shuttle or the logs from deployment should log all files that were included in the archive, i.e. either when making the archive or unpacking it.

Perhaps this will spam the logs a bit, so perhaps a `--verbose` flag could be introduced for deployments?

Alternatively, it could be a development build logging feature.

### Suggestion or Example of how the feature would be used

Packing file .gitignore
Packing file Cargo.toml
...
Packing file src/main.rs
...

### Duplicate declaration

- [X] I have searched the issues and this feature has not been requested before.

imor:
Another idea might be to write the archive to disk locally for debugging. Users can then inspect the archive to confirm which files are included in it.

hseeberger:
I already considered running the deployer locally and add logging. So I like this idea. `trace` level makes sense for me for this use case.

I also like the alternative. Write the archive to disk and log (`info` level) the path.

jonaro00:
I like the idea with a tracing log in cargo-shuttle. It show up for the end user if they ran deploy with RUST_LOG, right? In that case, perhaps an info level log would be good.

imor:
Yes these logs can be enabled by setting the RUST_LOG env variable. E.g. to see only cargo_shuttle info logs, set RUST_LOG=cargo_shuttle=info. Added PR #931 for it.

PR #931, opened 2023-05-21
fix: log files packed in archive

imor:
## Description of change
This PR adds info level log for each file added to the archive. Users will be able to see these logs if they set the appropriate level in the RUST_LOG env var. E.g. `RUST_LOG=cargo_shuttle=info`.

closes issue #930 .

Ran the `cargo shuttle deploy` command locally.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -49,7 +49,7 @@ use shuttle_service::builder::{build_workspace, BuiltService};
 use std::fmt::Write;
 use strum::IntoEnumIterator;
 use tar::Builder;
-use tracing::{error, info, trace, warn};
+use tracing::{debug, error, trace, warn};
 use uuid::Uuid;
 
 use crate::args::{DeploymentCommand, ProjectCommand, ProjectStartArgs, ResourceCommand};
@@ -1175,7 +1175,7 @@ impl Shuttle {
 
         // Append all the entries to the archive.
         for (k, v) in entries {
-            info!("Packing {k:?}");
+            debug!("Packing {k:?}");
             tar.append_path_with_name(k, v)?;
         }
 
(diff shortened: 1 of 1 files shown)


hseeberger:
Should we better use debug instead of info?

imor:
@hseeberger I agree. I think it should be either debug or trace. Currently most logging in cargo-shuttle is at the trace level. We were having a discussion about it in #930 @jonaro00 what do you think?

jonaro00:
Debug is fine

iulianbarbu (commented):
Thanks @imor!

I am wondering how this should be considered: relevant only from a debugging perspective or relevant also from an usage perspective? 

The idea is simple, if users are wondering about what files are included in the archive that gets deployed, maybe we should make this optional (hidden behind a flag) but still usable without a `RUST_LOG` env variable. I know the demarcation from a user and developer is not that clear, but I would frame it as a contract rather than a developer feature, because the contract is more explicit/looked for by users. 

Thinking about something of this sort (extracted from `tar` command):
```
     -v, --verbose
             Produce verbose output.  In create and extract modes, tar will list each file name as it is read from or
             written to the archive.  In list mode, tar will produce output similar to that of ls(1).  An additional -v
             option will also provide ls-like details in create and extract mode.
```

I am fine with the existing approach too, but would like to see your thoughts.

jonaro00:
@iulianbarbu It can definitely be the case that enough users want this as a feature, but thus far it has only been relevant for debugging "did this file get deployed" issues.
It is a nice transparency tool for the docs FAQ, aka "Run `RUST_LOG="..." cargo shuttle deploy` to see exactly which files are uploaded".

I say we include it as is. The future will tell if it needs to upgrade to a more "official" feature.

imor:
I agree with @jonaro00. For now this PR can be merged. A `--verbose` flag can be added later if required.

oddgrd (approved):
I like Iulian's suggestion too, but I agree with you that this is a good intermediate solution either way. Thanks!

PR #942, opened 2023-05-23
Update Cargo.lock

orhun:
## Description of change

I came across this error while building the Arch Linux package:

```
error: the lock file /build/cargo-shuttle/src/cargo-shuttle/Cargo.lock needs to be updated but --locked was passed to prevent this
If you want to try to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.
```

This is due to the usage of `--locked` flag and it is necessary for reproducible builds as stated [in the wiki](https://wiki.archlinux.org/title/Rust_package_guidelines).

This PR is simply the output of `cargo generate-lockfile` for updating `Cargo.lock`.

Tested locally with `--locked` flag.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -49,10 +49,19 @@ dependencies = [
 ]
 
 [[package]]
-name = "ambient-authority"
-version = "0.0.1"
+name = "aho-corasick"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


orhun:
Arch Linux package is updated with this patch applied: https://gitlab.archlinux.org/archlinux/packaging/packages/cargo-shuttle/-/commit/80eec76ed7627091cf2310e8ca7f2c6bd076426a

oddgrd (approved):
Nice, thanks @orhun!

PR #946, opened 2023-05-26
refactor: sanitize all path on the user's proxy

chesedo:
## Description of change
If users don't properly handle input paths then they can open themselves up to path traversal vulnerabilities. We can protect all our users by taking care of this at our proxy layer.

Using the local dev environment to ensure path traversals are removed from all proxy requests.


jonaro00:
Cool crate 👍

oddgrd (approved):
LGTM, and I love the crate as well! :partying_face:

PR #948, opened 2023-05-26
docs: Update links and commands

jonaro00:
## Description of change

Various docs fixes.

Moved hint prints for local run to after the run is done.


oddgrd (approved):
Great, thanks @jonaro00!

PR #950, opened 2023-05-28
Add helpful error if port cannot be used

beyarkay:
## Description of change

A lot of the discussion happened on the discord server, under "help", and the issue named "Trying to setup postgres locally, getting "failed to connect to provisioner"".

Basically there was an `unwrap` which was emitting a not-too-helpful error message, and my root cause was just something was using the port that shuttle wanted to use.

I'm not sure if you have a preferred error-emitting system, so if you  tell me how you like your errors to look I'm more than happy to edit this to keep everything consistent.

Thanks for the great tool!

### DIFF
--- a/runtime/src/alpha/mod.rs
+++ b/runtime/src/alpha/mod.rs
@@ -84,7 +84,10 @@ pub async fn start(loader: impl Loader<ProvisionerFactory> + Send + 'static) {
         server_builder.add_service(svc)
     };
 
-    router.serve(addr).await.unwrap();
+    match router.serve(addr).await {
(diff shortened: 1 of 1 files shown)


oddgrd (approved):
Thanks! It seems we are not correctly shutting down the runtime server in the event of certain crashes. We'll need to look into that, but this should be helpful either way!

beyarkay:
For future travellers, here's a shell one-liner to find the running process and kill it. Use at your pleasure/peril:

```
lsof -nP -iTCP:7999 | tail -1 | cut -d' ' -f2 | xargs kill
```

PR #957, opened 2023-05-31
Add a SQLite resource

### LINKED ISSUE #874: [Feature]: SQLite support
### Describe the feature

My Rust app has as a database a Sqlite file. Without support for that I'd have to rearchitect and migrate my app to another database which for small applications really only adds operational complexity.

I'm not sure how you could support this, either:
- Offer SQLite as a supported database type
- Offer read/write file storage so the application could put its database file there and interact with it like that

_No response_

### Duplicate declaration

morlinbrot:
Closes #874 

Adds a `shuttle-qlite` resource using `sqlx` under the hood. We're exposing almost all of sqlx's [`SqliteConnectOptions`](https://docs.rs/sqlx/latest/sqlx/sqlite/struct.SqliteConnectOptions.html) to the user which can be configured with the `shuttle_sqlite::SQLiteConnOpts` builder like so:
```rust
#[shuttle_runtime::main]
async fn axum(
    #[SQLite(opts = SQLiteConnOpts::new().filename("custom.sqlite"))] pool: sqlx::SqlitePool,
) -> shuttle_axum::ShuttleAxum { /* ... */ }
```

This gives the user a highly customisable SQLite instance being run in-process by the shuttle runtime. Options we omitted are those we regarded as unneeded or too complex to implement with only marginal gains for the user.

## Status
- ~Currently, **_the in-memory option is not working_** and I'm struggling to find a solution, any help would be appreciated.~ EDIT: fixed.
- A useful feature would be a way to export the created database (or import an existing one), some design work is needed for this.

- There's an example [over here](https://github.com/morlinbrot/shuttle-examples/tree/sqlite-example/axum/sqlite) that we have been testing with.
- It would be great to add some tests but I only see actual use in some kind of integration or e2e tests and I'm not sure there is a way to do this currently. I'd be happy to be corrected on this.


resources/sqlite/src/conn_opts.rs
```diff
@@ -0,0 +1,263 @@
...
+            serialized: false,
+            command_channel_size: 50,
+            row_channel_size: 50,
+        }
+    }
+
+    /// Use this to set the database to in memory mode.
+    pub fn in_memory(mut self, on: bool) -> Self {
+        self.in_memory = on;
+        self
```
jonaro00, line 116: This being public does not make sense if a user were to turn this off but still have the default :memory: file. I think it can use in memory by default, and only calls to filename() will flip this flag and set the new filename.

jonaro00 (changes requested):
Looks promising! I wonder what implications this has for future development of libsql/Turso integration.

Is the plan to have the file be persisted if written to file, but not if in memory?

resources/sqlite/src/conn_opts.rs
morlinbrot: We don't want to give the user access to the connection string directly because for security reasons (user could construct arbitrary paths). That's why we're constructing the conn string internally and need to expose this is some other way.

resources/sqlite/src/conn_opts.rs
```diff
@@ -0,0 +1,270 @@
...
+
+        pragmas.insert("locking_mode".into(), None);
+
+        pragmas.insert("journal_mode".into(), None);
+
+        pragmas.insert("foreign_keys".into(), Some("ON".into()));
+
+        pragmas.insert("synchronous".into(), None);
+
+        pragmas.insert("auto_vacuum".into(), None);
```
oddgrd, line 92: Do we need to insert all of these here, or can we just do it if/when they are set?

oddgrd (commented):
Hey Linus, apologies for the delayed review! This is looking great, but I left some questions and comments in the review. 

Additionally I am wondering if we should be exposing this much config, since it can make it more complex to use and maintain (since we need to keep it synced with sqlx). For example we only expose the `local_uri` config for our other databases.  Perhaps just exposing a few core options is enough for now, and we can add more if we see a demand? I think the most important are journal mode and synchronous setting, as well as maybe the filename which can also be used to make the database in-memory. This is all very open to discussion though.

I also saw that we made you wait so long that [sqlx 0.7](https://github.com/launchbadge/sqlx/blob/main/CHANGELOG.md#070---2023-06-30) is out :sweat_smile: , but I think we could go ahead with 0.6 for this PR, and then upgrade it later (we are on 0.6 in other resources as well).

morlinbrot:
Hi Oddbjørn,

Thanks for the feedback, here are some replies:

> Additionally I am wondering if we should be exposing this much config, since it can make it more complex to use and maintain

This is what we were wondering in the beginning. At some point it was decided that since I had to "mirror" the config anyway, it wouldn't be much more effort to expose it completely for maximum usability/flexibility. I do see the point of higher maintenance burden though but that would only occur if bumping a sqlite version.

>  For example we only expose the local_uri config for our other databases

We decided against this (following a discussion on Discord IIRC) for security reasons, because this would expose the possibility to the user to pass arbitrary file names and with it arbitrary file paths on the host (e.g. you can pass absolute file paths via the uri). The way it's implemented now, the actual construction of the db uri is entirely internal and controlled by us.

> did you get the in memory to work?

I did, will update the description.

> And the e2e testing did not work out for you, right

No, it didn't, unfortunately. Also in a discussion on Discord we decided to go ahead with this without those e2e tests for now, they could be added later (by someone who doesn't have the "doesn't work on only my machine" problem that I had 😃)

I hope this clears up the main points you made, for the remaining review points, I'll make sure to update the PR with the requested changes or answer the questions there directly.

resources/sqlite/src/conn_opts.rs
morlinbrot: Hm, so following the principle of mirroring the original config, I put it here to make sure that we're not missing any interactions that we don't know about, e.g. making sure that all the original defaults are set. Take the `foreign_keys` pragma for example:

When the `foreign_keys` method is being used, it sets `ON` or `OFF` but the `ON` variant is being set as the default, what happens if it's not set I don't really know. True, it is a bit verbose but this way we have a catch-all way of making sure that the config is initialised correctly without knowing of all the internal interdependencies that there may be.

resources/sqlite/src/conn_opts.rs
```diff
@@ -0,0 +1,270 @@
...
+            read_only,
+            create_if_missing,
+            statement_cache_capacity,
+            busy_timeout,
+            immutable,
+            vfs,
+            serialized,
+            command_channel_size,
+            row_channel_size,
+            ..
```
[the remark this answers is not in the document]
morlinbrot: You are right. Now I remember that at some piont I realised that I don't have an easy way to pass pragmas to the final config we're creating in `try_from`. I should have removed the whole pragma business I guess.

This is actually a good argument for exposing only a limited set of options. For now, we could simply expose all the options that are being set directly and remove the pragma logic completely. This would actually not change what is being exposed right now as I already put a disclaimer in the docs that pragmas are not exposed.

resources/sqlite/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    ) -> Result<Self::Output, shuttle_service::Error> {
+        // We construct an absolute path using `storage_path` to prevent user access to other parts of the file system.
+        let storage_path = factory.get_storage_path()?;
+        self.opts.storage_path = storage_path;
+        Ok(self.opts)
+    }
+
+    /// Build this resource from its config output
+    async fn build(build_data: &Self::Output) -> Result<sqlx::SqlitePool, shuttle_service::Error> {
+        let opts = SqliteConnectOptions::try_from(build_data).unwrap();
```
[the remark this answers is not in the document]
morlinbrot: Not sure where my comment went that I had here at some point. This should actually never fail if our `try_from` method is implemented correctly, which is guaranteed by tests. I'll add both a comment and a `?`.

resources/sqlite/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+use sqlx::sqlite::SqliteConnectOptions;
+/// The [`sqlx::SqlitePool`](https://docs.rs/sqlx/latest/sqlx/type.SqlitePool.html) that is being returned to the user.
+///
+pub use sqlx::SqlitePool;
+
+/// Builder struct used to configure the database, e.g. `SQLite(opts = SQLiteConnOpts::new())`.
+#[derive(Serialize)]
+pub struct SQLite {
+    opts: SQLiteConnOpts,
+}
```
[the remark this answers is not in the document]
morlinbrot: Right. I liked this because it didn't collide with the original structs like `sqlx::Sqlite`. But I agree, naming conventions should be followed, will rename to something like `ShuttleSqlite`.

resources/sqlite/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+        // We construct an absolute path using `storage_path` to prevent user access to other parts of the file system.
+        let storage_path = factory.get_storage_path()?;
+        self.opts.storage_path = storage_path;
+        Ok(self.opts)
+    }
+
+    /// Build this resource from its config output
+    async fn build(build_data: &Self::Output) -> Result<sqlx::SqlitePool, shuttle_service::Error> {
+        let opts = SqliteConnectOptions::try_from(build_data).unwrap();
+        // debug!("{:#?}", opts);
```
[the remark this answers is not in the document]
morlinbrot: Removed.

resources/sqlite/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+//! async fn axum(
+//!     #[shuttle_sqlite::SQLite(opts = SQLiteConnOpts::new().filename("custom.sqlite"))] pool: shuttle_sqlite::SqlitePool,
+//! ) -> shuttle_axum::ShuttleAxum { /* ... */ }
+//! ```
+//! Note that Shuttle does currently not support the `collation`, `thread_name`, `log_settings`, `pragma`, `extension`,
+//! `shared_cache` options.
+use async_trait::async_trait;
+use serde::Serialize;
+use shuttle_service::{Factory, ResourceBuilder, Type};
+// use tracing::debug;
```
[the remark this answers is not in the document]
morlinbrot: Removed.

oddgrd (commented):
Hey again Linus, thanks for the detailed response! I added a few more lines of feedback, and also some naming suggestions, everything is open to discussion of course, this goes without saying but naming is hard :joy:

Oh, and getting this branch updated and conflicts resolved would be great when you get a chance!

resources/sqlite/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+# Configuration
+The database can be configured using [`SQLiteConnOpts`](https://docs.rs/shuttle-sqlite/latest/shuttle_sqlite/struct.SQLiteConnOpts.html) 
+which mirrors sqlx's [`SqliteConnectOptions`](https://docs.rs/sqlx/latest/sqlx/sqlite/struct.SqliteConnectOptions.html) 
+for the options it exposes.
+
+```rust
+#[shuttle_runtime::main]
+async fn axum(
+    #[shuttle_sqlite::SQLite(opts = SQLiteConnOpts::new().filename("custom.sqlite"))] pool: shuttle_sqlite::SqlitePool,
```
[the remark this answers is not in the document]
morlinbrot: Good catch, updated.

resources/sqlite/src/conn_opts.rs
```diff
@@ -0,0 +1,215 @@
...
+/// for the options it exposes, see their docs for reference.
+///
+/// Construction of the full connection string is handled internally for security reasons and defaults to creating a
+/// file-based database named `default_db.sqlite` with `create_if_missing == true`. Use the `filename` and/or
+/// `in_memory` methods to configure the type of database created.
+///
+/// Note that Shuttle does currently not support the `collation`, `thread_name`, `log_settings`, `pragma`, `extension`,
+/// `shared_cache` options. The following options are internally controlled by pragmas and hence also not exposed: `foreign_keys`, `locking_mode`, `auto_vacuum`, `page_size`.
+#[derive(Debug, Deserialize, Serialize)]
+pub struct ShuttleSqliteConnOpts {
```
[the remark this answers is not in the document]
morlinbrot: I had updated the names already but didn't push yet because I hadn't rebased onto the newest main yet. I opted for the longer but explicit `ShuttleSqlite` and `ShuttleSqliteConnOpts`, hope that's fine!

resources/sqlite/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+pub use conn_opts::*;
+
+use sqlx::sqlite::SqliteConnectOptions;
+/// The [`sqlx::SqlitePool`](https://docs.rs/sqlx/latest/sqlx/type.SqlitePool.html) that is being returned to the user.
+///
+pub use sqlx::SqlitePool;
+
+/// Builder struct used to configure the database, e.g. `SQLite(opts = SQLiteConnOpts::new())`.
+#[derive(Serialize)]
+pub struct ShuttleSqlite {
```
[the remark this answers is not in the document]
morlinbrot: See above.

resources/sqlite/src/conn_opts.rs
```diff
@@ -0,0 +1,215 @@
...
+            .command_buffer_size(*command_channel_size)
+            .row_buffer_size(*row_channel_size);
+
+        if let Some(journal_mode) = journal_mode {
+            opts = opts.pragma("journal_mode", journal_mode.as_str());
+        }
+
+        if let Some(synchronous) = synchronous {
+            opts = opts.pragma("synchronous", synchronous.as_str());
+        }
```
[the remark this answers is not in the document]
morlinbrot: Same as the original problem with `SqliteConnectOptions`, they're not serializable so we need to have our own wrappers. But I updated the code to not use `pragma` but set them directly and implemented `From` for them, which actually shortened their impl blocks a bit.

morlinbrot:
Ok, I just pushed another update. I had already done some changes (ex. renaming the types) before you commented but hadn't pushed because I was blocked by not having the newest rustc and hence couldn't rebase. That is done now and I hope the names are fine, I can of course change them again if desired.

Thanks for the other small nitpicks which I missed, I corrected them, too.

oddgrd:
I forgot to mention this, but it would be great to have an example of using this resource (with e.g. axum) in the examples repo, then we can use that for testing as well. Other than that I am going on holiday (again :smile:), but Pieter or Iulian will do a final review of this when they get a chance. Thanks for all your work on this, it's looking good!

morlinbrot:
Luckily, that already exists, I just hadn't put up a PR yet: shuttle-hq/shuttle-examples/pull/72

resources/sqlite/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+        Self {
+            opts: ShuttleSqliteConnOpts::default(),
+        }
+    }
+
+    /// Get the internal config state of the builder
+    ///
+    /// If the exact same config was returned by a previous deployement that used this resource, then [Self::output()]
+    /// will not be called to get the builder output again. Rather the output state of the previous deployment
+    /// will be passed to [Self::build()].
```
chesedo, line 92: You don't have to repeat comments on a trait method since the comments from the trait will always be used. Ie. these will never be shown to the user and will have to be kept in-sync with the trait comments

resources/sqlite/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+            .await
+            .map_err(|e| shuttle_service::Error::Database(e.to_string()))?;
+
+        Ok(pool)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use assert_json_diff::assert_json_eq;
```
chesedo, line 127: [`pretty_assertions`](https://docs.rs/pretty_assertions) will reduce the need to convert it to json first

chesedo (commented):
A few comments to mostly make the user experience better. :smile: 

You will want to add the new type inside `resource_recorder` after merging in `main`

resources/sqlite/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+use serde::Serialize;
+use shuttle_service::{Factory, ResourceBuilder, Type};
+
+mod conn_opts;
+pub use conn_opts::*;
+
+use sqlx::sqlite::SqliteConnectOptions;
+/// The [`sqlx::SqlitePool`](https://docs.rs/sqlx/latest/sqlx/type.SqlitePool.html) that is being returned to the user.
+///
+pub use sqlx::SqlitePool;
```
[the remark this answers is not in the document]
morlinbrot: Leftover from some earlier iteration I think, I removed it and updated the example, too.

resources/sqlite/src/conn_opts.rs
```diff
@@ -0,0 +1,214 @@
...
+        if let Some(synchronous) = synchronous {
+            opts = opts.synchronous(synchronous.into());
+        }
+
+        if let Some(vfs) = vfs {
+            opts = opts.vfs(vfs.clone());
+        }
+
+        // `shared_cache` must be enabled to use a conn pool (instead of single conns) with an in-memory sqlite db.
+        if *in_memory {
```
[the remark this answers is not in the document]
morlinbrot: It's destructured from `opts` which is of type `&ShuttleSqliteConnOpts` so a `&bool` which the compiler complained about. Not sure if there's a more elegant way to do this?

resources/sqlite/README.md
```diff
@@ -0,0 +1,30 @@
...
+# Usage
+Add `shuttle-sqlite` to the dependencies for your service and use the resource by adding the `shuttle_sqlite::ShuttleSqlite` 
+attribute to main.
+
+It returns a [`sqlx::SqlitePool`](https://docs.rs/sqlx/latest/sqlx/type.SqlitePool.html).
+
+```rust
+#[shuttle_runtime::main]
+async fn axum(
+    #[shuttle_sqlite::ShuttleSqlite] pool: shuttle_sqlite::SqlitePool,
```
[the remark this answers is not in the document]
morlinbrot: Yup, updated.

resources/sqlite/src/lib.rs
morlinbrot: Didn't know that. I originally just put it there for myself to make sense of the trait.

resources/sqlite/src/lib.rs
morlinbrot: So *that's* the name of the crate I couldn't rememberˆˆ

resources/sqlite/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+    fn config(&self) -> &Self::Config {
+        &self
+    }
+
+    async fn output(
+        mut self,
+        factory: &mut dyn Factory,
+    ) -> Result<Self::Output, shuttle_service::Error> {
+        // We construct an absolute path using `storage_path` to prevent user access to other parts of the file system.
+        let storage_path = factory.get_storage_path()?;
```
jonaro00, line 97: I have some major refactorings of the deployer storage system in sight with #1050. Although that is far from complete, it would probably be simpler if this was delayed until after #1050 (and the Beta arch in general) to not complicate that rewrite even more. WDYT @chesedo @oddgrd ?

morlinbrot: If I may jump in here, I would love to have this merged beforehand as that plan sounds like it would make some heavy refactorings necessary for which I don't really have much time right now. This has been gotten out of sync a few times and required some updates which already took a lot longer than I would have preferred.

jonaro00: I know it sucks to have a huge PR waiting around 😞. I'll see if we can escalate the priority of this one.

resources/sqlite/src/conn_opts.rs
```diff
@@ -0,0 +1,214 @@
...
+            statement_cache_capacity,
+            busy_timeout,
+            immutable,
+            vfs,
+            serialized,
+            command_channel_size,
+            row_channel_size,
+        } = opts;
+
+        let db_path = storage_path.join(&filename);
```
jonaro00, line 175: Only needed if in_memory is false. I'm also wondering: which folder is storage_path pointing to currently (locally vs when deployed)?

resources/sqlite/src/conn_opts.rs
```diff
@@ -0,0 +1,214 @@
...
+            .command_buffer_size(*command_channel_size)
+            .row_buffer_size(*row_channel_size);
+
+        if let Some(journal_mode) = journal_mode {
+            opts = opts.journal_mode(journal_mode.into());
+        }
+
+        if let Some(synchronous) = synchronous {
+            opts = opts.synchronous(synchronous.into());
+        }
```
jonaro00, line 201: Why are these enums wrapped in custom types? I don't see where the benefit is.

resources/sqlite/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+version = "0.22.0"
+edition = "2021"
+
+[dependencies]
+async-trait = "0.1.68"
+indexmap = { version = "1.9.3", features = ["serde"] }
+regex = "1.8.3"
+serde = "1.0.163"
+shuttle-service = { version = "0.22.0", path = "../../service" }
+sqlx = { version = "0.6.3", features = ["sqlite", "runtime-tokio-rustls"] }
```
jonaro00, line 12: nit: please bump to sqlx 0.7 here and in the example.

resources/sqlite/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+        let ours = ShuttleSqliteConnOpts::new().filename(&filename);
+        let opts_from = SqliteConnectOptions::try_from(&ours).unwrap();
+        let str_from = format!("{:?}", opts_from);
+
+        assert_eq!(str_sqlx, str_from);
+    }
+
+    #[test]
+    fn try_from_in_memory() {
+        // `shared_cache` must be enabled to use a conn pool (instead of single conns) with an in-memory sqlite db.
```
jonaro00, line 138: Is this something worth noting in the docs of this crate?

jonaro00 (changes requested):
Some general thoughts:

Might want to include a note that a native sqlite3 lib is required for running locally (just link to relevant docs?).
I assume this is already installed in deployer container.

resources/sqlite/src/conn_opts.rs
jonaro00: Is path injection possible with the current impl? i.e. `filename = "../../../some/path"`
Btw, is that what you mean by arbitrary paths, or is there other behaviors to avoid?

jonaro00:
Hi @morlinbrot.
We had a discussion internally about this feature today.
In short, we are a bit unsure about the need for this resource, considering we have decent support for Turso and we are a bit unsure if we product-wise are aiming to provide more "local" resources like this, or external ones like Turso and other cloud DBs.
In either case, we think merging this before the upcoming architectural changes will introduce a lot of migration pain, so we will have to wait and see where we end up.
Sorry for leaving you to wait longer.

jonaro00:
We are reconsidering how we can support SQLite after future reworks of our storage offerings. Closing this in favour of #1566

PR #864, opened 2023-05-04
feat: add on_new_span impl to runtime Logger

vroussea:
Add spans impl to runtime Logger that allows all alltributes to be passed down to shuttle.

Added new span test
Before:
Screenshot from 2023-05-04 17-18-08

After:
Screenshot from 2023-05-04 17-19-35

PR #958, opened 2023-05-31
fix: Ignore span logs below WARN

### LINKED ISSUE #955: [Bug]: Serenity bot crashes due to excessive logs
### What happened?

Serenity projects emit way too many log lines since #864, and can lead a bot to crash (not 100% confirmed).

The excessive logs are due to functions annotated with `#[instrument]` (spans) are set to INFO by default.

I think we can compromise by having the minimum level increased in the Shuttle log consumer for spans,
but it would be awesome if this was separately configurable with #842 (set separate span and event levels, per crate)

### Version

v0.17.0

Linux, In deployment

x86_64, In deployment

### Relevant log output

```sh
2023-05-31T02:03:10.586481275Z  INFO {shard="Shard { shard_info: ShardInfo { id: 0, total: 1 }, .. }"} serenity::gateway::bridge::shard_runner: [span] recv
2023-05-31T02:03:10.586582375Z  INFO serenity::gateway::shard: [span] do_heartbeat
2023-05-31T02:03:10.586616675Z  INFO {shard="Shard { shard_info: ShardInfo { id: 0, total: 1 }, .. }"} serenity::gateway::bridge::shard_runner: [span] recv_event
```

### Duplicate declaration

jonaro00:
## Description of change

Closes #955

Tested on my serenity bot (0.12) and hello world example (0.11) - no more spam, back to normal amount of logs as before
Tide hello world gives the same amount of logs before and after

### DIFF
--- a/runtime/src/logger.rs
+++ b/runtime/src/logger.rs
@@ -112,18 +112,17 @@ mod tests {
 
         let _guard = tracing_subscriber::registry().with(logger).set_default();
 
-        let span = tracing::debug_span!("this is a span");
+        let span = tracing::warn_span!("this is a warn span");
         span.in_scope(|| {
             tracing::debug!("this is");
             tracing::info!("hi");
+        });
+        let span = tracing::info_span!("this is an info span");
+        span.in_scope(|| {
             tracing::warn!("from");
             tracing::error!("logger");
         });
 
-        assert_eq!(
-            r.blocking_recv().map(to_tuple),
-            Some(("[span] this is a span".to_string(), LogLevel::Debug as i32))
-        );
         assert_eq!(
             r.blocking_recv().map(to_tuple),
             Some(("this is".to_string(), LogLevel::Debug as i32))
@@ -131,6 +130,10 @@ mod tests {
         assert_eq!(
             r.blocking_recv().map(to_tuple),
             Some(("hi".to_string(), LogLevel::Info as i32))
+        );
+        assert_eq!(
+            r.blocking_recv().map(to_tuple),
+            Some(("[span] this is an info span".to_string(), LogLevel::Debug as i32))
         );
         assert_eq!(
             r.blocking_recv().map(to_tuple),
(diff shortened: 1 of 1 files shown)


jonaro00:
@oddgrd I don't see any difference in tide logs between 0.15 -> 0.17 -> this PR. What was the issue that #864 solved?

oddgrd:
Ah, cool, this looks promising! The issue with tide logs was that it was missing request metadata when running on shuttle. But if this doesn't cause a regression in that regard, I think we're good.

jonaro00:
The span logs need to be configurable by the user since they can spam a lot. I'd say we raise the bar to WARN like this until a better solution arises.

oddgrd:
Yeah, that's true. I think we can do this as an intermediate until we make it configurable, and then when we do we can document the required steps to get the expected log output in tide in the tide-integration docs. And again, another point in favor of this, I don't think the missing log output in tide is affecting a lot of users (if any).

jonaro00:
I moved the return statement up a bit, skipping the JsonVisitor. I'm not sure what it does though. Should it be before or after?

runtime/src/logger.rs
```diff
@@ -35,8 +35,16 @@ where
...
             let metadata = attrs.metadata();
-            let mut visitor = JsonVisitor::default();
 
+            let level = LogLevel::from(metadata.level()) as i32;
+
+            // Ignore span logs from the default level for #[instrument] (INFO) and below.
+            // TODO: make this configurable
+            if level < LogLevel::Warn as i32 {
+                return;
+            }
```
chesedo, line 45: I think we only need the `as i32` conversion when it goes into a `LogItem`. You should also not need to convert it to a `LogLevel` yet - aka delay the conversion :smile:

chesedo (commented):
Seems good. Thanks @jonaro00. Just left a note about converting less.

jonaro00:
Goofy tracing. They invert the order so it is not the same as the underlying enum int.
The shuttle LogLevel uses the integer.
```rs
println!("{}", Level::INFO < Level::WARN); // false
println!("{}", LogLevel::Info < LogLevel::Warn); // true
```
Should work now.

oddgrd (approved):
Thanks so much for getting to the bottom of this issue and finding a solution!

Issue #928, opened 2023-05-21
[Bug]: .cargo/config.toml is not deployed


hseeberger:
### What happened?

While working on an example app, I found out that `.cargo/config.toml` is not deployed. Yet this is needed to successfully compile, because the app is using uuid v7 which requires setting specific rustflags and that by default happens in `.cargo/config.toml`.

More details can be found at https://discord.com/channels/803236282088161321/1109452835214590123.

To reproduce the issue, use tag shuttle-issue-928: https://github.com/hseeberger/ticklist/releases/tag/shuttle-issue-928.

### Version

0.16.0

### Which operating systems are you seeing the problem on?

macOS

### Which CPU architectures are you seeing the problem on?

ARM64

### Duplicate declaration

- [X] I have searched the issues and there are none like this.

hseeberger:
According to @jonaro00, "hidden" files are not deployed. I suggest to modify the relevant logic to include "hidden" default files/directories like `.cargo`.

hseeberger:
Looking at `cargo_shuttle::make_archive`, I think I disagree with @jonaro00: `hidden(false)` is called on the `WalkBuilder`, i.e. hidden files are **not** excluded by default.

As `.cargo/config.toml` is not in any `.gitignore` or `.ignore` files, I have no clue why the deploy fails.

jonaro00:
Yeah, you're right. I misremembered.
What toolchain are you using locally?
This comment says that the rustcflags should be in a [target] field, and the string should also be split into two. https://www.reddit.com/r/rust/comments/yindaf/comment/iujnbgo/
Perhaps it is an issue with mismatching environments locally/in deployment.

jonaro00:
This sparked an idea: https://github.com/shuttle-hq/shuttle/issues/930

imor:
I tried locally and in my case at least the `.cargo/config.toml` file was included in the archive. So it looks like it may not be due to the file missing from the archive. @hseeberger Do you have the project you were deploying available on Github? I can try deploying it as well.

imor:
Sorry @hseeberger missed the link in the description. I also ran into the same problem. The project builds locally for me but when deployed it fails because of missing `Uuid::now_v7()`. It has something to do with how shuttle builds the project.

jonaro00:
Are you both using Rust 1.69? Deployment uses 1.68. I didn't see anything related to this but it is after all a source of errors, so just double checking.

imor:
[...]

PR #960, opened 2023-05-31
feat: pre-installed build environment in deployer

### LINKED ISSUE #807: [Feature]: pre-installed environment on `deploy` build machines
### Describe the feature

This is related to #703.

It is not uncommon that a project may depend on crates that wrap code in other languages.
To have a hassle free experience at least the build tools for the most common cases should be installed on the machine that handles the `shuttle deploy` build step. 

I would assume this to be at least what a GH runner comes pre-installed with.

See [here](https://github.com/actions/runner-images#software-and-image-support) for what this entails on the resp. target platform/OS.

I would suggest starting with a C/C++ compiler and I would have some projects to test this.

_No response_

### Duplicate declaration

jonaro00:
## Description of change

Closes #807.

Added some common (C) build tools when installing deployer.
This allows crates such as `audiopus` to build.

If there are more tools needed, they can be suggested or added in subsequent PRs.

Deployed a crate with `audiopus v0.2.0` locally. https://crates.io/crates/audiopus/

### DIFF
--- a/deployer/prepare.sh
+++ b/deployer/prepare.sh
@@ -68,5 +68,6 @@ done
 # cargo fetch
 
 # Install common build tools for external crates
+# The image should already have these: https://github.com/docker-library/buildpack-deps/blob/65d69325ad741cea6dee20781c1faaab2e003d87/debian/buster/Dockerfile
 apt update
-apt install -y llvm-dev libclang-dev gcc clang cmake make automake autoconf libtool
+apt install -y llvm-dev libclang-dev clang cmake
--- a/deployer/prepare.sh
+++ b/deployer/prepare.sh
@@ -66,3 +66,7 @@ done
 # TODO: restore when we know how to prefetch from our mirror
 # cd /usr/src/shuttle/service
 # cargo fetch
+
+# Install common build tools for external crates
+apt update
+apt install -y llvm-dev libclang-dev gcc clang cmake make automake autoconf libtool


oddgrd (commented):
Thanks! For the new builder we intend to allow easy configuration of build dependencies, but this will be good to have in the meantime. I do believe we have some of the ones you added covered already though, since we use a rust image with the [buster](https://github.com/docker-library/buildpack-deps/blob/65d69325ad741cea6dee20781c1faaab2e003d87/debian/buster/Dockerfile) buildpack-deps.

jonaro00:
Is build time env vars planned aswell?

oddgrd:
You mean like --cfg flags to rustc (https://github.com/shuttle-hq/shuttle/issues/928), or more than that? @chesedo is working on the new builder as we speak, so it would be great to hear what functionality you'd like to see in this regard!

jonaro00:
Sooner or later there will be dependencies that need env vars at build time for arbitrary purposes. Secrets come in too late for that. While .cargo/config.toml can fix flags to rustc, some users will want to set env vars.
https://discord.com/channels/803236282088161321/1107140940759568384/1107511887157022820

(p.s this PR should be ready)

oddgrd (approved):
Alright, thanks! We'll keep it in mind for the builder.

jonaro00:
You can also consider the possibility of running an arbitrary script (install third party deps) before compilation starts.

oddgrd:
I think since we're using nix for the builder adding third-party deps should be quite straightforward with nix derivations.

PR #968, opened 2023-06-02
fix: include the base domain shuttleapp.rs in the wildcard certificate

### LINKED ISSUE #833: [Bug]: wildcard certificate for *.shuttleapp.rs does not cover base domain shuttleapp.rs
### What happened?

https://shuttleapp.rs/ on Chrome currently yields a certificate error. The certificate appears to be for "*.shuttleapp.rs" which does not include "shuttleapp.rs".

The certificate appears to be from "Let's Encrypt". Its been a while but I think I remember being able to include the base domain when creating a wildcard certificate with "Let's Encrypt".

### Version

n/a

### Which operating systems are you seeing the problem on?

Windows

_No response_

### Relevant log output

_No response_

### Duplicate declaration

paulotten:
closes #833

## Description of change

TODO

TODO

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1117,7 +1117,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-shuttle"
-version = "0.16.0"
+version = "0.17.0"
 dependencies = [
  "anyhow",
  "assert_cmd",
@@ -5163,7 +5163,7 @@ dependencies = [
 
 [[package]]
 name = "shuttle-admin"
-version = "0.16.0"
+version = "0.17.0"
 dependencies = [
  "anyhow",
  "clap",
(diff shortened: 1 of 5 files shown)


jonaro00:
Hi @paulotten, would you be interested in continuing on this? It has been left stale due to the base name not being very important, but I can see how having a nice landing page or redirect to the main website has value.

paulotten:
Hi @jonaro00, my dance card is full until the new year. Feel free to take over on this if you wish.

From my recollection I thought these current changes should be sufficient to get the wildcard certificate working also for the base domain. I never managed to properly test that however. @iulianbarbu had some feedback on how to do this on issue #833. I forget exactly where I ran into problems.

Once the certificate is working for the base domain I imagine adding a redirect would be trivial.

jonaro00:
We took this suggestion and have implemented it (a redirect to the main website) on an upcoming platform update, so this won't be needed. Thanks for the suggestion and work towards solving it.

PR #970, opened 2023-06-03
fix/release(prod): unstable AWS creds clashed with prod

iulianbarbu:
## Description of change

Our dev AWS creds were set for the `.circleci` machine when deploying Shuttle to prod because the environment variables names are the standard AWS environment variables names for credentials and are picked up by `aws-ecr/ecr-login` step, which conflicts with how we set the ecr-login in our `.circleci/config.yaml` file. At the same time, the PROD variables' names used in the `.circleci/config.yaml` had a typo 🙈  (which might have been the main reason why the ecr-login failed, besides the standard env variable names issue). To fix everything this PR changed the dev environment variables' names to non-standard names, so as to not be picked up by the `aws-ecr/ecr-login` step and double-checked the env vars used in the `.circleci/config.yaml` to be the ones from the circleci project settings.

Tested the AWS creds on this PR, checking if we can list the prod/dev public ECR repos. Tested the creation of a postgres project on `unstable` as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   docker-buildx: sensu/docker-buildx@1.1.1
-  aws-ecr: circleci/aws-ecr@8.1.2
+  aws-ecr: circleci/aws-ecr@8.2.1
   win: circleci/windows@5.0
 
 executors:
@@ -316,6 +316,9 @@ jobs:
           aws-access-key-id: << parameters.aws-access-key-id >>
           aws-secret-access-key: << parameters.aws-secret-access-key >>
           public-registry: << parameters.public-registry >>
+          # `registry-id`` field is required, although we don't need it (used for private registry).
+          # We give it a non-empty env variable name to bypass the `ecr-login` empty check.
+          registry-id: TAG
       - run:
           name: Make and push images
           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -362,15 +365,25 @@ jobs:
         name: Deploy images
         command: |
           DOCKER_HOST=ssh://ec2-user@master.<< parameters.ssh-host >> USE_TLS=enable PROD=<< parameters.production >> DD_API_KEY=$DD_API_KEY \
-          POSTGRES_PASSWORD=<< parameters.postgres-password >> \
-          MONGO_INITDB_ROOT_PASSWORD=<< parameters.mongodb-password >> \
+          POSTGRES_PASSWORD=${<< parameters.postgres-password >>} \
+          MONGO_INITDB_ROOT_PASSWORD=${<< parameters.mongodb-password >>} \
```
oddgrd, line 369: Cool! Did you check on unstable that the provisioner doesn't error with the changes to how we handle these vars?

oddgrd (approved):
LGTM, just one question. Thanks for sorting this out Iulian!

.circleci/config.yml
iulianbarbu: All good in terms of the logs it seems! Resolving this.

PR #982, opened 2023-06-07
fix: logs --latest flipped order

jonaro00:
#862 flipped the order of deployment list, making the --latest flag get the oldest deployment.


oddgrd (approved):
Great, thanks!

Issue #991, opened 2023-06-10
[Improvement]: Suggest user to run `cargo shuttle run` instead of `cargo run`


jonaro00:
### Describe the improvement

Currently, if a user accidentally runs `cargo run` instead of `cargo shuttle run`, they get an unhelpful error.
```
thread 'main' panicked at 'could not parse arguments: MissingRequiredArgument { arg: "--port" }', /home/jonaro00/.cargo/registry/src/github.com-1ecc6299db9ec823/shuttle-runtime-0.18.0/src/alpha/mod.rs:53:30
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
This is due to the internal project binary not starting with the correct args from cargo-shuttle.

Make it instead print a suggestion to use the correct command.
- Modify the error handling of `Args::parse()` in `runtime/src/alpha/mod.rs`. The error type returned contains variants for many reasons of parsing failure, but in this case we can expect cargo-shuttle to always be correct, so printing the suggestion regardless of what error type happened should be sufficient.

### Duplicate declaration

- [X] I have searched the issues and this improvement has not been requested before.

akansjain:
Hey @jonaro00 , I would like to work on this issue.

jonaro00:
Hi @akansjain. This was already fixed in #993 and we forgot to close this issue.

PR #993, opened 2023-06-10
provide better error message

lecoqjacob:
Solves #991 

image


jonaro00 (approved):

jonaro00:
Next projects are libraries, so they can't be `cargo run`'ed

oddgrd (approved):
Thanks!

Issue #668, opened 2023-03-02
Add support for S3 objects


brokad:
Currently, for ad-hoc persistent storage, we only support local volumes. We'd like to add support for S3 object storage. The user-facing API should follow the same entrypoint as other [resources](https://docs.shuttle.rs/resources/shuttle-shared-db). IOW, there should be a new plugin crate allowing the proc-macro to support adding S3 clients from the [aws-sdk-s3](https://docs.rs/aws-sdk-s3/0.24.0/aws_sdk_s3/) crate to the arguments of a `shuttle_runtime::main` function.

I can think of a rough plan to deliver this (which could be separate but complete PRs):
- Add the required methods to [provisioner](https://github.com/shuttle-hq/shuttle/tree/main/provisioner) to create/delete buckets. (started in #685, which is no longer in progress, but we should be able to use some of the work done there.)
- Create a plugin crate (as mentioned above) in the same fashion as other [resources](https://github.com/shuttle-hq/shuttle/tree/main/resources), this will also require updating the [Factory](https://github.com/shuttle-hq/shuttle/blob/main/service/src/lib.rs#L78) trait and it's implementations to make it able to request S3 storage.

After this, we also have to update the local provisioner in [cargo-shuttle](https://github.com/shuttle-hq/shuttle/blob/main/cargo-shuttle/src/provisioner_server.rs), which requires thinking about how we want this to work in a local run. Should we use the users ambient AWS credentials and provision the resource on their own account? Or should we use an [S3 mock server](https://hub.docker.com/r/adobe/s3mock/)?

Our terraform repository is private (it will be made public in the future), but editing our terraform is not required for this PR. This PR should however indicate which roles are needed to create the S3 buckets.

dphov:
@ivancernja @brokad I'm interested to collaborate with you on this issue

dphov:
@oddgrd I'm sorry, but I will step down from this task, too many constraints for me at this moment.

Shubham8287:
Hi, I am actively looking to contribute to Shuttle. I also read some code base already, I would love to pick this up and I will start by sharing my approach in couple of days, please assign this to me.

oddgrd:
No worries @dphov, thanks for all your effort! I hope to see you again in the future.

@Shubham8287 cool, sounds good! Feel free to reach out in our discord in the #contributors channel if you have questions.

ibrahim-akrab:
Hey @Shubham8287 , are you still working on this?

[...]

PR #917, opened 2023-05-15
feat: added support for s3 bucket in provisioner (#668)

Shubham8287:
* feat: update provisioner.proto to add methods for provisioning storage

* feat: implemented provision_storage to create a bucket and return user credentials to access the bucket

This is a provisioner side change for  task /attempt #668, Steps for provisioning storage look like -

- Create a S3 bucket by appending an uuid to project_name to make it globally unique.
- Create a User to interact with the bucket
- Create a Policy and attach that to User
- Return access keys for the user, so that resource builder can access the bucket.

PR #995, opened 2023-06-11
S3 plugin

Shubham8287:
feat: update provisioner.proto to add methods for provisioning storage

feat: implemented provision_storage to create a bucket and return user credentials to access the bucket

This is a provisioner side change for task /attempt https://github.com/shuttle-hq/shuttle/issues/668, Steps for provisioning storage look like -

Create a S3 bucket by appending an uuid to project_name to make it globally unique.
Create a User to interact with the bucket
Create a Policy and attach that to User
Return access keys for the user, so that resource builder can access the bucket.


Shubham8287:
draft implementation of plugin for #668

jonaro00:
I suppose this is paired with #917

Shubham8287:
yes but not ready to review yet :)

iulianbarbu:
Hey @Shubham8287 ! This PR slipped from our list of priorities lately. We'll be a bit low on bandwidth in the next weeks because of a bigger release and an offsite happening shortly, but we'll try to catch up on this in the meantime. 

We thought that this PR will have a common portion shared with #910, regarding the AWS credentials management, so would be great to have alignment between the two. I haven't personally looked that much into how this is done on both PRs, and it would be great if you can share any thoughts you formed on this subject. Also, please check out #910 approach, would be great if you can spot something there that is conflicting with this PR needs, if any.

Shubham8287:
I am  holding this PR for #910 as it shares a common portion and I think that user policy part has some security concerns as well. Would resume it once we get other one merged or some better approach. Let me if this is fine to you folks.

iulianbarbu:
Thanks @Shubham8287! Indeed, at this time we have some security concerns that required postponing the DynamoDB/S3 PRs. We need to go back to the drawing board and see what architecture/design changes would need to be done to accommodate integrating these resources with Shuttle. BTW, thanks a lot for staying close to S3/DynamoDB development track. We'll try to give more updates about the way forward once we'll plan this internally.

iulianbarbu:
Closing this. Context provided here: https://github.com/shuttle-hq/shuttle/issues/668#issuecomment-1731190978.

PR #996, opened 2023-06-12
feat(resources): add support for turso client w/o provisioning

### LINKED ISSUE #986: [Feature]: Add support for Turso
### Describe the feature

[Turso](https://turso.tech/) is an edge SQLite database. We would like for Shuttle to be able to provision and connect to it, exposing a [`libsql-client::Client`](https://github.com/libsql/libsql-client-rs).

Currently Turso doesn't offer a way through their SDK to create a database. The first would then be to simply add a way to initiate a client through a macro annotation. We can then later on add a way to instantiate one.

```rust
#[shuttle_runtime::main]
async fn main(
    #[shuttle_turso(addr = "...")] client: libsql_client::Client,
) -> shuttle_axum::ShuttleAxum { 
    ...
}
```

### Duplicate declaration

Kazy:
## Description of change

This adds support for Turso but without provisioning.

Close #986.

Locally yes, both with a local and a remote database.

## TODO

- Clarify the few `XXX` in the code.

### DIFF
--- a/resources/turso/Cargo.toml
+++ b/resources/turso/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "shuttle-turso"
-version = "0.19.0"
+version = "0.20.0"
 edition = "2021"
 license = "Apache-2.0"
 description = "Plugin to obtain a client connected to a Turso database"
--- a/resources/turso/Cargo.toml
+++ b/resources/turso/Cargo.toml
@@ -12,3 +12,8 @@ libsql-client = { version = "0.27" }
 serde = { version = "1.0.148", features = ["derive"] }
 shuttle-service = { path = "../../service", version = "0.18.0", default-features = false }
 url = { version = "2.3.1", features = ["serde"] }
+
+
+[dev-dependencies]
+tempfile = "3.3.0"
+tokio = "1.28.2"
--- a/resources/turso/Cargo.toml
+++ b/resources/turso/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "shuttle-turso"
+version = "0.18.0"
+edition = "2021"
+license = "Apache-2.0"
+description = "Plugin to obtain a client connected to a Turso database"
+keywords = ["shuttle-service", "turso"]
+
+[dependencies]
+async-trait = "0.1.56"
+libsql-client = { version = "0.27" }
+serde = { version = "1.0.148", features = ["derive"] }
+shuttle-service = { path = "../../service", version = "0.18.0", default-features = false }
+url = { version = "2.3.1", features = ["serde"] }
--- a/resources/turso/Cargo.toml
+++ b/resources/turso/Cargo.toml
@@ -8,7 +8,7 @@ keywords = ["shuttle-service", "turso"]
 
 [dependencies]
 async-trait = "0.1.56"
-libsql-client = { version = "0.27" }
+libsql-client = { version = "=0.30.1" }
 serde = { version = "1.0.148", features = ["derive"] }
 shuttle-service = { path = "../../service", version = "0.19.0", default-features = false }
 url = { version = "2.3.1", features = ["serde"] }
--- a/resources/turso/Cargo.toml
+++ b/resources/turso/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "shuttle-turso"
-version = "0.18.0"
+version = "0.19.0"
 edition = "2021"
 license = "Apache-2.0"
 description = "Plugin to obtain a client connected to a Turso database"
@@ -10,7 +10,7 @@ keywords = ["shuttle-service", "turso"]
 async-trait = "0.1.56"
 libsql-client = { version = "0.27" }
 serde = { version = "1.0.148", features = ["derive"] }
-shuttle-service = { path = "../../service", version = "0.18.0", default-features = false }
+shuttle-service = { path = "../../service", version = "0.19.0", default-features = false }
 url = { version = "2.3.1", features = ["serde"] }
 
 
--- a/resources/turso/Cargo.toml
+++ b/resources/turso/Cargo.toml
@@ -10,7 +10,7 @@ keywords = ["shuttle-service", "turso"]
 async-trait = "0.1.56"
 libsql-client = { version = "=0.30.1" }
 serde = { version = "1.0.148", features = ["derive"] }
-shuttle-service = { path = "../../service", version = "0.19.0", default-features = false }
+shuttle-service = { path = "../../service", version = "0.20.0", default-features = false }
 url = { version = "2.3.1", features = ["serde"] }
 
 
--- a/resources/turso/README.md
+++ b/resources/turso/README.md
@@ -9,19 +9,23 @@ This plugin allows services to connect to a Turso database. Turso is an edge-hos
 Add `shuttle-turso` to the dependencies for your service.
 This resource will be provided by adding the `shuttle_turso::Turso` attribute to your Shuttle `main` decorated function.
 
-It returns a `libsql_client::Client`. When running locally it will instantiate a local SQLite database of the name of your service instead of connecting to your edge database.
+It returns a `libsql_client::Client`. When running in production, the token used to connect to your database will be read from `TURSO_DB_TOKEN` in your `Secrets.toml` file. This can be overridden with the `token_secret` parameter. When running locally it will instantiate a local SQLite database of the name of your service instead of connecting to your edge database.
+
+If you want to connect to a remote database when running locally, you can specify the `local_addr` parameter. In that case, the token will be read from your `Secrets.dev.toml` file.
 
 ### Example
 
 ```rust
+use libsql_client::client::Client;
+
 #[shuttle_runtime::main]
-async fn app(#[shuttle_turso::Turso(addr="advanced-lightspeed.turso.io", auth_token=Some("token")] client: Client) -> __ { }
+async fn app(#[shuttle_turso::Turso(addr="libsql://advanced-lightspeed.turso.io")] client: Client) -> __ { }
 ```
 
 ### Parameters
 
-| Parameter  | Type        | Default | Description                                                                                                                   |
-| ---------- | ----------- | ------- | ----------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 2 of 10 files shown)


resources/turso/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    async fn output(
+        self,
+        factory: &mut dyn Factory,
+    ) -> Result<Self::Output, shuttle_service::Error> {
+        match factory.get_environment() {
+            shuttle_service::Environment::Production => {
+                if self.addr.is_empty() {
+                    // XXX: should we raise an error even not running in a production
+                    // environment ?
```
AlphaKeks, line 67: Not sure I understand this. If we are in this match arm, we are definitely running in production, are we not?

resources/turso/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+                        // instead of just giving the responsability to the user.
+                        conn_url: Url::parse(&format!("libsql://{}", self.addr))
+                            .map_err(Error::UrlParseError)?,
+                        token: Some(self.token),
+                    })
+                }
+            }
+            shuttle_service::Environment::Local => {
+                // Default to a local db of the name of the service.
+                // XXX: should we check that the file doesn't exist before using that name ?
```
AlphaKeks, line 82: Does `.get_build_path()` already covert that case? It might lead to confusing error messages if the file doesn't exist or can't be created.

resources/turso/src/lib.rs
Kazy: The question was that if we run in a local environment, should we raise an error ? The idea behind it is to be able to warn even when running locally that you're going to have an issue when deploying.

The other integrations don't seem to do that though, and maybe a warning message is enough.

resources/turso/src/lib.rs
Kazy: `.get_build_path()` covers the case of the folder not exiting which would lead to the errors you mention. The case I was talking about is when inside the folder returned by `.get_build_path()`, what should happen if database file already exists ? This is obviously okay if that's the database that has been created previously, but it might collude with a pre-existing file. Should we instead force the user to define the name of the database to use locally instead of choosing one for them ? Or do we just don't care ?

AlphaKeks: If the file already exists and is not a valid database, the code will end up throwing errors somewhere else. I think we should just make sure we print useful error messages in those cases rather than putting that burden on the user.

resources/turso/src/lib.rs
AlphaKeks: Well, right now, if the `addr` is `None`, we fall back to a local database file, which is the intended way of running locally anyway, right?

resources/turso/src/lib.rs
Kazy: When you try to read or do anything with the database, the error is `File opened that is not a database file`, which seems fine to me :)

resources/turso/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+    ) -> Result<Self::Output, shuttle_service::Error> {
+        match factory.get_environment() {
+            shuttle_service::Environment::Production => {
+                if self.addr.is_empty() {
+                    Err(ShuttleError::Custom(CustomError::msg("missing addr")))
+                } else {
+                    Ok(TursoOutput {
+                        // XXX: should we allow for other kind of connection string ? Not having libsql at
+                        // the start of the url even though the Turso CLI is printing it might be confusing
+                        // instead of just giving the responsability to the user.
```
iulianbarbu, line 71: Can we check if the link starts with `libsql://` and if not, try connecting to the url but with the added `libsql://` prefix? In this way, we'll cover both situations I think, unless I haven't understood what you meant.

iulianbarbu (commented):
Thanks @Kazy ! This looks on the right path. I would need to take a closer look at the tests and do a manual one also. Consider this part 1. :D 

We'll be a bit low on bandwidth reviewing this during the next two weeks because we're planning for a bigger release and an off-site soon after. We'll try our best to provide meaningful feedback though in the meanwhile, so please shoot if you have other qs.

resources/turso/src/lib.rs
Kazy: I meant to allow the possibility to connect using an other kind of connection that the `libsql-client` library support, in particular the `reqwest` one, but I don't think there is a reason to support that since Turso's documentation only mention the `libsql` one.

resources/turso/README.md
```diff
@@ -0,0 +1,27 @@
...
+Add `shuttle-turso` to the dependencies for your service.
+This resource will be provided by adding the `shuttle_turso::Turso` attribute to your Shuttle `main` decorated function.
+
+It returns a `libsql_client::Client`. When running locally it will instantiate a local SQLite database of the name of your service instead of connecting to your edge database.
+
+### Example
+
+```rust
+#[shuttle_runtime::main]
+async fn app(#[shuttle_turso::Turso(addr="advanced-lightspeed.turso.io", auth_token=Some("token")] client: Client) -> __ { }
```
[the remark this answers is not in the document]
Kazy: It was, and I've updated the code to read from the secrets instead.

resources/turso/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+edition = "2021"
+license = "Apache-2.0"
+description = "Plugin to obtain a client connected to a Turso database"
+keywords = ["shuttle-service", "turso"]
+
+[dependencies]
+async-trait = "0.1.56"
+libsql-client = { version = "0.27" }
+serde = { version = "1.0.148", features = ["derive"] }
+shuttle-service = { path = "../../service", version = "0.18.0", default-features = false }
```
oddgrd, line 13: Just some delayed-review version bumps :smile:
```suggestion
version = "0.19.0"
edition = "2021"
license = "Apache-2.0"
description = "Plugin to obtain a client connected to a Turso database"
keywords = ["shuttle-service", "turso"]

[dependencies]
async-trait = "0.1.56"
libsql-client = { version = "0.27" }
serde = { version = "1.0.148", features = ["derive"] }
shuttle-service = { path = "../../service", version = "0.19.0", default-features = false }
```

resources/turso/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+If you want to connect to a remote database when running locally, you can specify the `local_addr` parameter. In that case, the token will be read from your `Secrets.dev.toml` file.
+
+### Example
+
+```rust
+use libsql_client::client::Client;
+
+#[shuttle_runtime::main]
+async fn app(#[shuttle_turso::Turso(addr="libsql://advanced-lightspeed.turso.io")] client: Client) -> __ { }
```
oddgrd, line 22: We can use secrets string interpolation here, see [this section](https://docs.shuttle.rs/resources/shuttle-aws-rds#parameters) in the rds docs for an example. I think we should also add the token param to the example and interpolate it from secrets, and maybe also make that one required.

I'm also thinking we should use one of the frameworks for the example so it's clear that using turso resource doesn't change the return type, what do you think?
```suggestion
async fn app(#[shuttle_turso::Turso(addr="libsql://advanced-lightspeed.turso.io")] client: Client) -> ShuttleAxum { ... }
```

resources/turso/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+                conn_url: Url::parse(addr).map_err(Error::UrlParseError)?,
+                token: Some(token.to_string()),
+            }),
+            None => Err(ShuttleError::Custom(CustomError::msg(format!(
+                "could't find secret {}",
+                self.token_secret
+            )))),
+        }
+    }
+}
```
oddgrd, line 76: I'm thinking maybe we can replace this with the secrets string interpolation?

resources/turso/README.md
```diff
@@ -0,0 +1,31 @@
+# Shuttle Turso
+
+This plugin allows services to connect to a Turso database. Turso is an edge-hosted distributed database based on libSQL.
+
+## Usage
+
+**IMPORTANT**: Currently Shuttle isn't able to provision a database for you (yet). This means you will have to create an account on their [website](https://turso.tech/) and follow the few steps required to create a database and create a token to access it.
+
+Add `shuttle-turso` to the dependencies for your service.
+This resource will be provided by adding the `shuttle_turso::Turso` attribute to your Shuttle `main` decorated function.
```
oddgrd, line 10: Excellent docs, thank you! Perhaps we can suggest to `cargo add shuttle-turso`?

oddgrd (commented):
Thank you @Kazy, this is really great! Apologies for the delayed review, but I'd say this is very nearly finished, I just have some suggestions for some small changes we could make. Thanks for adding tests and docs too :heart:

resources/turso/README.md
oddgrd: Perhaps we don't need to use secrets interpolation for the address, since it doesn't contain secrets like a sql connection string, rather just the token

oddgrd: Yes, Pieter implemented it a while ago but we haven't really documented it very clearly. :smile: It's pretty cool!

Kazy:
Sorry for the forced push. I missed that you updated the fork and ended up with some unwanted commits, and I preferred to have a clean state :)

oddgrd (approved):
This looks great, thank you @Kazy! I left a couple of comments but they are just nits :smile:

resources/turso/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+edition = "2021"
+license = "Apache-2.0"
+description = "Plugin to obtain a client connected to a Turso database"
+keywords = ["shuttle-service", "turso"]
+
+[dependencies]
+async-trait = "0.1.56"
+libsql-client = { version = "=0.30.1" }
+serde = { version = "1.0.148", features = ["derive"] }
+shuttle-service = { path = "../../service", version = "0.19.0", default-features = false }
```
oddgrd, line 13:
```suggestion
shuttle-service = { path = "../../service", version = "0.20.0", default-features = false }
```

resources/turso/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "shuttle-turso"
+version = "0.19.0"
```
oddgrd, line 3:
```suggestion
version = "0.20.0"
```

PR #998, opened 2023-06-12
fix: re-add `--version` flag

### LINKED ISSUE #992: [Bug]: Version check of existing shuttle-next binary panics
### What happened?

Running a next project installs shuttle-next if not present.
If it exists, the --version flag is used to check it.
Since https://github.com/shuttle-hq/shuttle/pull/822/files#diff-651941ab3b205fa0a1385341412700b34be3a850f36cf002fe60810f4ba89e2aL4 , the version flag is no longer there.
Therefore, running `cargo shuttle run` with shuttle-next installed panics, so it only works the first time.

I think the automatic install (in prod build) is strange.
I'd prefer getting prompted to `cargo install` so that I actually know what's happening. The user already installed cargo-shuttle and does not need to be spoonfed IMO.

The version check is still required though, so it could be solved by checking for "--version" explicitly before `NextArgs::parse()`.

### Version

v0.18.0

Linux

x86_64

### Relevant log output

```sh
Installing shuttle-next runtime. This can take a while...
thread 'main' panicked at 'shuttle-runtime version should be in the `name version` format', /home/jonaro00/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-shuttle-0.18.0/src/lib.rs:1283:14
```

### Duplicate declaration

paulotten:
## Description of change

Re-adds the `--version` flag.

Closes #992.

Ran locally.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2069,21 +2069,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
-name = "foreign-types"
-version = "0.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f6f339eb8adc052cd2ca78910fda869aefa38d22d5cb648e6485e4d3fc06f3b1"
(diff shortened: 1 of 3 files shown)


jonaro00 (approved):
👌

PR #1005, opened 2023-06-14
feat: session-based auth for auth service

oddgrd:
## Description of change

This implements session-based auth for the auth service, which aims to be equivalent of what we had with the axum API.

TODO:

- [x] - sign cookies
- [ ] - utilities for setting/getting cookies from tonic::Request
- [x] - expiration in session storage

Restored the old session tests.


chesedo (approved):
LGTM, thanks O!

PR #1006, opened 2023-06-14
feat(secrets): Implement into_iter for SecretStore

jonaro00:
## Description of change

Title~
Came up in a discussion about secrets https://discord.com/channels/803236282088161321/1107140940759568384/1118625700824485928

Added test

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -195,11 +195,18 @@ impl SecretStore {
     }
 }
 
+impl IntoIterator for SecretStore {
+    type Item = (String, String);
+    type IntoIter = <BTreeMap<String, String> as IntoIterator>::IntoIter;
+    fn into_iter(self) -> Self::IntoIter {
+        self.secrets.into_iter()
+    }
+}
+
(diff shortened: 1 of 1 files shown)


dteare:
For context for others who don't want to wade through the entire Discord thread: basically I'm wanting to use Rocket's outstanding [figment](https://api.rocket.rs/master/rocket/struct.Rocket.html#method.figment) feature that allows you to extract config values directly into structs without futzing around. For example, `rocket.figment().extract::<OpenAiConfig>()` gives me all I need to configure OpenAI without working with each field individually. 

Rocket does this automatically based on environment variables. [Shuttle Secrets](https://docs.shuttle.rs/resources/shuttle-secrets) (set via `Secrets.toml`) do not get injected into applications as environment variables, and so figment and others were sad. 

The idea was to work around this limitation by iterating over all the secrets within SecretStore and call `std::env::set_var` for each one. SecretStore doesn't provide a convenient way to do this, hence Shuttle Hero @jonaro00 jumped into action with this MR.

PR #1012, opened 2023-06-18
fix(runtime): Remove 2s startup sleep

jonaro00:
## Description of change

- Got rid of 2 second sleep by polling instead. It was at first needed due to the runtime not having time to open a socket on that port. I now use a 5ms polling time for 500ms. My test took <3ms to connect (with 1ms poll time), so I set the poll time to 5ms.
- Code and docs cleanup

EDIT: Moved .git exclusion to #1036 to not clutter this PR too much.

Local run.
Let's see if CI likes it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,11 +56,13 @@ You can now start a local deployment of Shuttle and the required containers with
 USE_PANAMAX=disable make up
 ```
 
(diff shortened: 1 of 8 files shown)


chesedo (approved):
CI seems happy. LGTM, thanks @jonaro00

PR #1013, opened 2023-06-19
chore: move codegen::main from service to runtime

jonaro00:
## Description of change

- Clean up the service crate by moving codegen::main to runtime (its only user). This removes the codegen feature from service, and the main function is no longer behind a feature due to being in runtime.
- Clean up Cargo files, docs, and reorder imports/exports.

CI pass


chesedo (approved):
Nice catch! Thanks for the cleanup @jonaro00

PR #1017, opened 2023-06-19
chore: add resources and services to workspace

jonaro00:
## Description of change

- add resources and services to workspace, and bump many minimum versions (semver compatible)
- cargo update
- added rustls feature flags to aws-rds


jonaro00:
`aws-rds` and `shared-db` were left excluded from the workspace, as the `--all-features` tests would break on them due to sqlx feature flags.
I adjusted CI accordingly.

.circleci/config.yml
```diff
@@ -175,24 +178,28 @@ jobs:
...
+    parameters:
+      crate:
+        description: "Crate in workspace to test"
+        type: string
+    executor: docker-rust
+    steps:
+      - checkout
+      - restore-cargo-cache
+      - apply-patches
+      - run: cargo test --package << parameters.crate >> --all-features -- --nocapture
```
iulianbarbu, line 202: How's this different than check-standalone?

.circleci/config.yml
```diff
@@ -596,35 +601,36 @@ workflows:
...
       - check-standalone:
         # shuttle-shared-db has mutually exclusive features
         # so we run checks for each feature separately
           name: "resources/shared-db: << matrix.features >>"
           matrix:
             alias: check-standalone-shared-db
             parameters:
-              path: [resources/shared-db]
+              path:
+                # resources/aws-rds has no tests, but also has mutually exclusive features
```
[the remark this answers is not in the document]
jonaro00: aws-rds crate would need a similar matrix, but that crate has no tests at all, so I left it out.

.circleci/config.yml
jonaro00: check-standalone does a full fmt+clippy+test, but when the crates are in the workspace fmt+clippy is already done, so test-member only tests a workspace member, unlike platform-test which also does integration tests with examples etc. I can clarify the names a bit.

resources/shared-db/Cargo.toml
```diff
@@ -7,11 +7,11 @@ description = "Plugin for managing shared databases on shuttle"
...
 [dependencies]
-async-trait = "0.1.56"
-mongodb = { version = "2.3.0", optional = true }
-serde = { version = "1.0.148", features = ["derive"] }
-shuttle-service = { path = "../../service", version = "0.19.0", default-features = false }
-sqlx = { version = "0.6.2", optional = true }
+async-trait = "0.1.68"
+mongodb = { version = "2.5.0", optional = true }
+serde = { version = "1.0.164", features = ["derive"] }
+shuttle-service = { path = "../../service", default-features = false }
```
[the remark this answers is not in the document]
jonaro00: You are probably right.

services/shuttle-rocket/Cargo.toml
```diff
@@ -1,16 +1,17 @@
...
+repository.workspace = true
 description = "Service implementation to run a rocket webserver on shuttle"
 keywords = ["shuttle-service", "rocket"]
 
-[workspace]
-
 [dependencies]
-rocket = { version = "0.5.0-rc.2" }
-shuttle-runtime = { path = "../../runtime", version = "0.19.0" }
+rocket = "0.5.0-rc.3"
```
jonaro00, line 11: Is this bump fine btw?

jonaro00:
Turso is also excluded, a comment in the root toml explains why.

jonaro00:
SQLX 0.7 is released, which should solve the turso thing, but that is a bigger job for a different PR.

services/shuttle-rocket/Cargo.toml
iulianbarbu: Should be.

Cargo.toml
```diff
@@ -28,10 +36,13 @@ repository = "https://github.com/shuttle-hq/shuttle"
...
 [workspace.dependencies]
-shuttle-codegen = { path = "codegen", version = "0.20.0" }
-shuttle-common = { path = "common", version = "0.20.0" }
-shuttle-proto = { path = "proto", version = "0.20.0" }
-shuttle-service = { path = "service", version = "0.20.0" }
+# libraries
+shuttle-codegen = { path = "codegen" }
+shuttle-common = { path = "common" }
+shuttle-proto = { path = "proto" }
+shuttle-service = { path = "service" }
```
jonaro00, line 43: @oddgrd Do you think removing the versions here will break something in the release process? I'm down to just "if it ain't broke, don't fix it" and add them back.

oddgrd: Yes, I think we would need the versions to be able to publish the crates that use these. You should be able to test this with [`cargo publish --dry-run`](https://doc.rust-lang.org/cargo/commands/cargo-publish.html#publish-options) in one of the crates.

jonaro00:
When I tried to `cargo update` on this branch now, I got this cyclic dependency
```
error: cyclic package dependency: package `ahash v0.8.3` depends on itself. Cycle:
package `ahash v0.8.3`
    ... which satisfies dependency `ahash = "^0.8.0"` of package `hashbrown v0.14.0`
    ... which satisfies dependency `hashbrown = "^0.14"` of package `indexmap v2.0.0`
    ... which satisfies dependency `indexmap = "^2"` of package `serde_json v1.0.104`
    ... which satisfies dependency `serde_json = "^1.0"` of package `wasm-bindgen v0.2.87`
    ... which satisfies dependency `wasm-bindgen = "^0.2.87"` of package `js-sys v0.3.64`
    ... which satisfies dependency `js-sys = "^0.3"` of package `getrandom v0.2.10`
    ... which satisfies dependency `getrandom = "^0.2.0"` of package `const-random-macro v0.1.15`
    ... which satisfies dependency `const-random-macro = "^0.1.15"` of package `const-random v0.1.15`
    ... which satisfies dependency `const-random = "^0.1.12"` of package `ahash v0.8.3`
    ... which satisfies dependency `ahash = "^0.8"` of package `actix-http v3.3.1`
```
...which might not be our issue, but these kinds of things increase in probability if we jam more and more packages into the workspace 😕. I think we should be careful about the decision to merge this.

oddgrd:
Yes, we have stumbled upon this [hidden treasure](https://github.com/tkaitchuck/aHash/issues/95) in the rust ecosystem before. :sweat_smile: I agree that we should be careful merging this, since the advantage of doing so is not that significant either (correct me if I'm wrong). The PR has also grown quite large, perhaps we should rather split it up, merge the easy wins, and then re-consider how we will handle services and resources in the workspace in the future?

jonaro00:
The only easy wins I see are small CI improvements and the extended feature flags on aws-rds. Can be done but not very significant.

jonaro00:
#1167

Issue #1004, opened 2023-06-14
[Feature]: Add qdrant resource


paulotten:
### Describe the feature

Requested in [Building Semantic Search in Rust with Qdrant & Shuttle Workshop](https://www.youtube.com/watch?v=YLWSeiDh2o0)

It looks like there is a docker container image: https://github.com/qdrant/qdrant#client-server

### Duplicate declaration

- [X] I have searched the issues and this feature has not been requested before.

PR #1025, opened 2023-06-20
feat: Qdrant resource

paulotten:
## Description of change

Adds a Qdrant resource.

For local runs (`cargo shuttle run`) it starts Qdrant in a local docker container.

For production deployment (`cargo shuttle deploy`) the user is required to provide valid Qdrant Cloud credentials. See https://qdrant.tech/documentation/cloud/cloud-quick-start/

`cargo shuttle run` and `cargo shuttle deploy` both ran locally against new example.


jonaro00:
Closes #1004

Srylax:
What is blocking this Feature?

jonaro00:
@Srylax We have let this one slide for a while, but now there are some people asking for it. There is nothing blocking it, so I'm thinking it can get done soon. Due to its age it will need some re-writing.

jonaro00:
Merged main, will look at simplifying our resource logic before merging this, as it will make this much easier.

jonaro00:
I removed the `Qdrant` resource type and made this a `Custom` instead (Turso and other plugins should go this route as well). This is due to this plugin being just a client, and Shuttle not touching or managing it. The gRPC call to provisioner was still needed in order to allow the docker integration on local runs, which raises some design concers for this type of library: this type of library wants to provision docker locally without affecting the Factory trait that is also used in Shuttle's backend. We should look into generalising the local provisioner behaviour so that a plugin like this can exist without changing the Shuttle codebase.

provisioner/src/lib.rs
```diff
@@ -581,6 +582,16 @@ impl Provisioner for MyProvisioner {
...
 
+    #[tracing::instrument(skip(self))]
+    async fn provision_qdrant(
+        &self,
+        _request: Request<QdrantRequest>,
+    ) -> Result<Response<QdrantResponse>, Status> {
+        Err(Status::invalid_argument(
+            "Provisioning Qdrant on Shuttle is not supported",
+        ))
+    }
```
oddgrd, line 593: The fact that this should only be used locally and not in the provisioner makes me think that this should rather be an extension trait, that's only used where it's needed.

resources/qdrant/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+                None => {
+                    let url = factory.get_qdrant_connection(md.project_name).await?.url;
+                    Ok(QdrantClientConfigWrap { url, api_key: None })
+                }
+            },
+        }
+    }
+
+    async fn build(client_config: &Self::Output) -> Result<QdrantClient, shuttle_service::Error> {
+        // TODO: Handle error better?
```
oddgrd, line 93: Does this still need to be addressed?

proto/provisioner.proto
```diff
@@ -43,5 +44,14 @@ message DatabaseResponse {
...
 
 message DatabaseDeletionResponse {}
+
+message QdrantRequest {
+  string project_name = 1;
+}
+
+message QdrantResponse {
+  string url = 1;
+}
```
oddgrd, line 54: Should we name this in a less qdrant specific way? We may add similar resources in the future that could also use these messages.

cargo-shuttle/src/provisioner_server.rs
```diff
@@ -289,6 +346,17 @@ impl Provisioner for LocalProvisioner {
         panic!("local runner should not try to delete databases");
     }
 
+    async fn provision_qdrant(
+        &self,
+        request: Request<QdrantRequest>,
+    ) -> Result<Response<QdrantResponse>, Status> {
+        let QdrantRequest { project_name } = request.into_inner();
+
+        let res = self.get_qdrant_connection_info(&project_name).await?;
```
oddgrd, line 355: Could we refactor these functions so they could be used for resources that will be started similarly to qdrant in the future?

oddgrd (commented):
Thanks guys! I left some comments and questions about possible refactors, but this is looking good!

proto/provisioner.proto
jonaro00: Yeah I'm inclined to go down this path. Would have to look at how similarly other integrations want to use the local provisioner server for docker, and see if there is an easy pattern to follow. It might be that each type of plugin is too unique for a generalized solution.

cargo-shuttle/src/provisioner_server.rs
jonaro00: Same as other comment

provisioner/src/lib.rs
jonaro00: Yep, the other changes suggested might prompt more differences between local and deployed provisioner.

jonaro00:
Refactored to an endpoint that allows arbitrary containers to start with 1 port, should be extendable for a fair number of use cases.
Also took a more direct approach and exposed a proto type at the service trait level (which I like - The alternative is a wrapper type with bidirectional `into`).

provisioner/src/lib.rs
jonaro00: Extension trait was not possible due to resource builder taking `&mut dyn Factory`, no big deal

oddgrd (approved):
Thanks!

proto/provisioner.proto
```diff
@@ -3,6 +3,7 @@ package provisioner;
 
 service Provisioner {
   rpc ProvisionDatabase(DatabaseRequest) returns (DatabaseResponse);
+  rpc ProvisionArbitraryContainer(ContainerRequest) returns (ContainerResponse);
```
chesedo, line 6: I'm not sure we want to add this method to provisioner :thinking: I get that we want something generic, but this seems to limit the generic options to containers.

Maybe we should just have a `Provision` method with a project name, type (string), details (json), and env? And just have it return json?

chesedo (commented):
Looks good. Would just like to make one element a bit more future proof.

proto/provisioner.proto
jonaro00: I don't think this endpoint needs to be future proof. If we are adding a more generic way of doing this in the future, it would very likely be a breaking change anyways and a new endpoint would suit that. Since this is only used in local runs, the breaking change would not affect deployers.

Not in favour of returning JSON inside protobufs yet since we have limited knowledge of how future use cases look like. I'm in favour of solving this for more generic cases when we have generic cases to solve this for and the current impl is not enough.

chesedo: I did not think of this being scoped to local runs. Fair enough! :+1:

chesedo (approved):

PR #1030, opened 2023-06-21
[scaling] refactor: remove project commands from cli

chesedo:
## Description of change
Each project will no longer have its own container, so we no longer have to start / stop them all :partying_face: 

Using `cargo check` :see_no_evil:


cargo-shuttle/src/client.rs
```diff
@@ -92,48 +92,12 @@ impl Client {
...
-        let path = format!("/projects/{}/clean", project.as_str(),);
-
-        self.post(path, Option::<String>::None)
-            .await
-            .context("failed to get clean output")?
-            .to_json()
-            .await
-    }
-
-    pub async fn get_project(&self, project: &ProjectName) -> Result<project::Response> {
```
oddgrd, line 119: I did not remove this, but I am not sure how much use it would have over getting a deployment/service status.

oddgrd (approved):
LGTM! One comment on get_project, but I'm not sure we need it, we can always add it back when we decide.

PR #1031, opened 2023-06-21
fix: increase body size limit for deploy

oddgrd:
## Description of change

Since we no longer just stream the entire body when receiving an archive to deploy, we are hitting the default body size limit in axum of 2MB for several projects. This PR increases the limit to 50MB.

Thanks to @jonaro00 for debugging and testing this!

Tested with a 5MB project, which worked after the layer was applied.

### DIFF
--- a/deployer/src/handlers/mod.rs
+++ b/deployer/src/handlers/mod.rs
@@ -114,7 +114,7 @@ impl RouterBuilder {
                     .post(
                         create_service
                             // Set the body size limit for this endpoint to 50MB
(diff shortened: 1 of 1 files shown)


jonaro00 (approved):

iulianbarbu (approved):
LGTM! Thanks @oddgrd and @jonaro00 ! I guess a good improvement in the future is to check the size of the serialized data (also, I am not sure if it gets compressed on the network) and prompt the users eagerly.

PR #1032, opened 2023-06-21
Fix(common): format logs in correct local timezone

### LINKED ISSUE #834: [Bug]: `shuttle deploy` incorrectly reports times as "Zulu"
### What happened?

```
> cargo shuttle deploy --allow-dirty

2023-05-01T12:21:31.543482956Z  INFO Entering queued state
[...]
```

I'm in the EDT timezone. So that should be `2023-05-01T16:21:31.543482956Z` or just `2023-05-01T12:21:31.543482956`.

A `Z` on the end of a timestamp indicates "Zulu" time which is the military name for UTC.

### Version

0.15.0

### Which operating systems are you seeing the problem on?

Windows

_No response_

### Relevant log output

```sh
> cargo shuttle deploy --allow-dirty

2023-05-01T12:21:31.543482956Z  INFO Entering queued state
2023-05-01T12:21:31.543503663Z TRACE shuttle_deployer::deployment::queue: getting a build slot
2023-05-01T12:21:31.548494787Z TRACE {uri="http://gateway:8001/stats/load"} shuttle_deployer::deployment::gateway_client: calling gateway
2023-05-01T12:21:31.550793860Z DEBUG hyper::client::connect::http: connecting to 10.99.82.119:8001
2023-05-01T12:21:31.553419377Z DEBUG hyper::client::connect::http: connected to 10.99.82.119:8001
2023-05-01T12:21:31.555555361Z DEBUG hyper::client::pool: pooling idle connection for ("http", gateway:8001)
[...]
```

### Duplicate declaration

### LINKED ISSUE #843: [Feature]: Improve errors for panics in the main function

timonv:
## Description of change
Logs were formatted in UTC regardless of local timezone, resulting in incorrect times. We now format the datetime with rfc3339 which includes the correct timezone.

Testing is a bit hacky with the mutex, as per the comment, Time uses libc to get the timezone. By forcing the env var to the expected timezone we get the result we want. Originally had more than one test, hence the mutex. Decided to leave it in, as any additional test using timezones would yield crazy racing errors. Happy to remove it if needed.

Closes #834 

Added unit tests.
Having some issues running the full test suite / local env. Should be alright but timezones.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -1,3 +1,7 @@
+#[cfg(test)]
+#[cfg(macro_use)]
+extern crate lazy_static;
+
 #[cfg(feature = "backend")]
 pub mod backends;
 #[cfg(feature = "claims")]
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -1,7 +1,3 @@
-#[cfg(test)]
-#[cfg(macro_use)]
-extern crate lazy_static;
-
 #[cfg(feature = "backend")]
 pub mod backends;
 #[cfg(feature = "claims")]
--- a/common/src/log.rs
+++ b/common/src/log.rs
@@ -78,7 +78,7 @@ impl std::fmt::Display for Item {
         write!(
             f,
             "{} {} {}",
-            datetime.to_rfc3339(),
+            datetime.to_rfc3339().dim(),
             self.level.get_colored(),
             message
         )
--- a/common/src/log.rs
+++ b/common/src/log.rs
@@ -78,7 +78,7 @@ impl std::fmt::Display for Item {
         write!(
             f,
             "{} {} {}",
-            datetime.format("%Y-%m-%dT%H:%M:%S.%fZ").to_string().dim(),
+            datetime.to_rfc3339(),
             self.level.get_colored(),
             message
         )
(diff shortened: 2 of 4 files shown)


common/src/lib.rs
```diff
@@ -1,3 +1,7 @@
+#[cfg(test)]
+#[cfg(macro_use)]
+extern crate lazy_static;
+
```
chesedo, line 4: This might not be needed since the test is including (`use`ing) it too. It's also a bit unexpected

common/src/log.rs
```diff
@@ -121,3 +121,54 @@ impl From<&tracing::Level> for Level {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use lazy_static::lazy_static;
+
+    lazy_static! {
+        static ref TZ_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
```
chesedo, line 131: Can we rather use `once_cell`?

chesedo: Well, maybe removing it will be better since there is only one test now :smile:

chesedo (commented):
Nice, thanks @timonv! :partying_face: 

I left some cleanup related comments

timonv:
Thanks for the feedback, added a clarifying comment, cleaned up and readded dim()

timonv:
> Having some issues running the full test suite / local env. Should be alright but timezones.

image

oddgrd (approved):
This looks great, thanks Timon! Thanks for the test, the screenshot and the thorough PR description as well :heart_eyes:

Issue #562, opened 2023-01-02
Add resource to get Shuttle service info


sd2k:
It's not ideal, but there may be some cases where the app will need to know some information about the environment it's running on, most notably the public URL. I'm currently using a [custom resource](https://github.com/sd2k/recipes/blob/eeb0e512906742baf215dffb5a31587de90b8f74/crates/recipe-shuttle/src/service_info.rs) to get the service name and creating a URL out of that, but it might be nice to provide a first-class Resource for the same thing. Happy to send a (more generalized) PR if you think it'd be useful!

oddgrd:
Hey @sd2k, apologies for the delayed reply. Is this something you're still interested in working on? We've had another request for this now, and your proposed solution seems interesting!

sd2k:
Hey @oddgrd, yeah I can fire off a PR. I originally wrote it against Shuttle 0.8.0 or something and there's been some big changes to custom resources since then so I'm not 100% confident in it, but that's what reviews are for, right? 😂

oddgrd:
Great, happy to hear that! 🥳 Yeah, it's changed quite a bit since then, so feel free to open a PR or a draft PR with any questions you may have. It may be useful to look at PRs for other resources that have been added more recently, like the Turso resource.

PR #1036, opened 2023-06-21
fix: Don't deploy .git folder to save space

jonaro00:
## Description of change

Moved this to a separate PR instead of together with unrelated stuff in #1012 

Original comment:

Hardcode to not upload the .git folder at deploy (the metadata is gathered before upload). This can shrink the upload size by a bit. If a project uses the local repo at build time for inserting, say, commit ID or something, this will break it. So perhaps add an option to include it? Or make a bunch of metadata available as a resource, like https://github.com/shuttle-hq/shuttle/issues/562

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -1224,6 +1224,8 @@ impl Shuttle {
 
         // Make sure the target folder is excluded at all times
         let overrides = OverrideBuilder::new(working_directory)
+            .add("!.git/")
+            .context("add `!.git/` override")?
             .add("!target/")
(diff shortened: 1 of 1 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -1224,6 +1224,8 @@ impl Shuttle {
 
         // Make sure the target folder is excluded at all times
```
oddgrd, line 1225:
```suggestion
        // Make sure the target  and .git folders are excluded at all times
```

oddgrd (approved):
Thanks!

PR #1038, opened 2023-06-22
[scaling] deployer: added project events streaming

iulianbarbu:
## Description of change

* Added project events streaming for the deployer.
* Added project_id and last_updated to the services table.

Tested locally by:
* creating user services and looking at the event stream messages, seeing how the services go through all the states
* very important, once the service is in `Running` state, it should send the service socket addr.

### DIFF
--- a/common/src/backends/auth.rs
+++ b/common/src/backends/auth.rs
@@ -411,7 +411,7 @@ mod tests {
     fn to_token_and_back() {
         let mut claim = Claim::new(
             "ferries".to_string(),
-            vec![Scope::DeploymentRead, Scope::Project],
+            vec![Scope::DeploymentRead, Scope::ProjectRead],
         );
 
         let doc = signature::Ed25519KeyPair::generate_pkcs8(&rand::SystemRandom::new()).unwrap();
@@ -433,7 +433,7 @@ mod tests {
     async fn authorization_layer() {
         let claim = Claim::new(
             "ferries".to_string(),
-            vec![Scope::DeploymentRead, Scope::Project],
+            vec![Scope::DeploymentRead, Scope::ProjectRead],
         );
 
         let doc = signature::Ed25519KeyPair::generate_pkcs8(&rand::SystemRandom::new()).unwrap();
@@ -455,7 +455,7 @@ mod tests {
                             let public_key = public_key.clone();
                             async move { public_key.clone() }
                         }))
-                        .layer(ScopedLayer::new(vec![Scope::Project])),
+                        .layer(ScopedLayer::new(vec![Scope::ProjectRead])),
                 );
 
         //////////////////////////////////////////////////////////////////////////
@@ -531,7 +531,7 @@ mod tests {
     fn hack_symmetric_alg() {
         let claim = Claim::new(
             "hacker-hs256".to_string(),
-            vec![Scope::DeploymentRead, Scope::Project],
+            vec![Scope::DeploymentRead, Scope::ProjectRead],
         );
 
         let doc = signature::Ed25519KeyPair::generate_pkcs8(&rand::SystemRandom::new()).unwrap();
@@ -571,7 +571,7 @@ mod tests {
     fn hack_no_alg() {
         let claim = Claim::new(
             "hacker-no-alg".to_string(),
-            vec![Scope::DeploymentRead, Scope::Project],
+            vec![Scope::DeploymentRead, Scope::ProjectRead],
         );
 
         let doc = signature::Ed25519KeyPair::generate_pkcs8(&rand::SystemRandom::new()).unwrap();
@@ -603,7 +603,7 @@ mod tests {
     fn hack_no_sig() {
         let claim = Claim::new(
             "hacker-no-sig".to_string(),
-            vec![Scope::DeploymentRead, Scope::Project],
+            vec![Scope::DeploymentRead, Scope::ProjectRead],
         );
 
         let doc = signature::Ed25519KeyPair::generate_pkcs8(&rand::SystemRandom::new()).unwrap();
@@ -626,7 +626,7 @@ mod tests {
     fn hack_bad_iss() {
         let claim = Claim::new(
             "hacker-iss".to_string(),
-            vec![Scope::DeploymentRead, Scope::Project],
+            vec![Scope::DeploymentRead, Scope::ProjectRead],
         );
 
         let doc = signature::Ed25519KeyPair::generate_pkcs8(&rand::SystemRandom::new()).unwrap();
--- a/deployer/src/lib.rs
+++ b/deployer/src/lib.rs
@@ -233,6 +233,7 @@ impl<D: Dal + Send + Sync + 'static> DeployerService<D> {
                 name: runnable_deployment.service_name.clone(),
                 state_variant: creating.to_string(),
                 state: creating,
+                last_update: Utc::now(),
             };
             self.dal
                 .insert_service_if_absent(service)
--- a/deployer/src/lib.rs
+++ b/deployer/src/lib.rs
@@ -164,6 +164,7 @@ impl<D: Dal + Send + Sync + 'static> DeployerService<D> {
 
             self.instate_service(
                 runnable_deployment,
+                existing_deployment.project_id,
                 GitMetadata::default(),
                 image_name.clone(),
                 existing_deployment.idle_minutes,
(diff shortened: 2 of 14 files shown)


common/src/backends/auth.rs
```diff
@@ -411,7 +411,7 @@ mod tests {
     fn to_token_and_back() {
         let mut claim = Claim::new(
             "ferries".to_string(),
-            vec![Scope::DeploymentRead, Scope::Project],
+            vec![Scope::DeploymentRead, Scope::ProjectRead],
```
chesedo, line 414: Hmm, why were these tests updated?

deployer/src/lib.rs
```diff
@@ -91,6 +94,13 @@ pub struct DeployerService<D: Dal + Send + Sync + 'static> {
     deployment_state_machine_channel:
         tokio::sync::mpsc::Sender<Box<dyn Task<(), Output = (), Error = project::error::Error>>>,
     runtime_start_channel: Sender<RunnableDeployment>,
+
+    // These channel are intended to work with one subscriber. They can be reinitialized as well.
+    // They are wrapped in Mutex first because we want them with interior muttability (see unsubscribe method).
+    // They are wrapped in Arc because they get sent on all deployment tasks.
+    project_events_channel_tx: Arc<Mutex<Option<UnboundedSender<ProjectEvent>>>>,
+    project_events_channel_rx: Arc<Mutex<Option<UnboundedReceiver<ProjectEvent>>>>,
```
chesedo, line 102: Why do we only want one subscriber?

deployer/src/lib.rs
```diff
@@ -293,16 +312,18 @@ impl<D: Dal + Sync + 'static> Deployer for DeployerService<D> {
         request: tonic::Request<DeployRequest>,
     ) -> TonicResult<tonic::Response<DeployResponse>, tonic::Status> {
         // Authorize the request.
-        request.verify(Scope::DeploymentWrite)?;
+        // request.verify(Scope::DeploymentWrite)?;
 
-        let claim = request.extensions().get::<Claim>().cloned();
+        // let claim = request.extensions().get::<Claim>().cloned();
```
chesedo, line 317: Comments :smile:

deployer/src/lib.rs
```diff
@@ -314,7 +335,7 @@ impl<D: Dal + Sync + 'static> Deployer for DeployerService<D> {
             service_name,
             service_id,
             tracing_context: Default::default(),
-            claim,
+            claim: None,
```
chesedo, line 338: Also from comments

deployer/src/lib.rs
```diff
@@ -349,7 +371,7 @@ impl<D: Dal + Sync + 'static> Deployer for DeployerService<D> {
         request: tonic::Request<DestroyDeploymentRequest>,
     ) -> TonicResult<tonic::Response<DestroyDeploymentResponse>, tonic::Status> {
         // Authorize the request.
-        request.verify(Scope::DeploymentWrite)?;
+        // request.verify(Scope::DeploymentWrite)?;
```
chesedo, line 374: Comment

deployer/src/lib.rs
```diff
@@ -411,96 +437,78 @@ impl<D: Dal + Sync + 'static> Deployer for DeployerService<D> {
...
-                .map_err(|err| tonic::Status::internal(err.to_string()))?;
-            let service_name = container
-                .service_name()
-                .map_err(|err| tonic::Status::internal(err.to_string()))?;
-            let deployment =
-                self.dal
-                    .deployment(&deployment_id)
+        request: tonic::Request<SubscribeProjectsRequest>,
+    ) -> TonicResult<tonic::Response<Self::SubscribeProjectsStream>, tonic::Status> {
+        // request.verify(Scope::ProjectRead)?;
```
chesedo, line 448: This comes from gateway, so not sure which claim it should have :thinking:

deployer/src/lib.rs
```diff
@@ -411,96 +437,78 @@ impl<D: Dal + Sync + 'static> Deployer for DeployerService<D> {
...
-
-        Ok(Response::new(TargetIpResponse {
-            target_ip: new_target_ip.to_string(),
-        }))
+    #[instrument(skip_all)]
+    async fn unsubscribe_projects(
+        &self,
+        request: tonic::Request<UnsubscribeProjectsRequest>,
+    ) -> TonicResult<tonic::Response<UnsubscribeProjectsResponse>, tonic::Status> {
+        // request.verify(Scope::ProjectWrite)?;
```
chesedo, line 506: Same as above

deployer/src/project/docker.rs
```diff
@@ -141,26 +146,34 @@ pub struct ServiceDockerContext {
...
     runtime_manager: RuntimeManager,
+    events_tx: Arc<Mutex<Option<UnboundedSender<ProjectEvent>>>>,
 }
 
 impl ServiceDockerContext {
-    pub fn new(docker: Docker, runtime_manager: RuntimeManager) -> Self {
+    pub fn new(
+        docker: Docker,
+        runtime_manager: RuntimeManager,
+        events_tx: Arc<Mutex<Option<UnboundedSender<ProjectEvent>>>>,
```
chesedo, line 156: Is there anyway we can make the sender async so that it does not have to be behind a mutex?

chesedo (commented):
Looking good, left about 2 questions :smile:

common/src/backends/auth.rs
iulianbarbu: Oh, it's because I changed the name of the Scope from `Project` to `ProjectRead` (to be more informative in terms of the naming about what actions are happening within the system, as an effect of the request). However, I've changed the needed scopes of the `subscribe/unsubscribe` RPC to `Admin`.  

Btw, just realized, I am not sure if renaming the scope enum names would be breaking for our existing users. I have to check.

deployer/src/lib.rs
iulianbarbu: The idea is that only the gateway subscribes at this moment to the deployer project events. We would want eventually to support more than one subscriber, but if we want it we should keep track of each of them, and handle cleaning up connections if they get an EOF as well.

deployer/src/lib.rs
iulianbarbu: yep, thanks! Done.

deployer/src/lib.rs
iulianbarbu: Done.

deployer/src/lib.rs
iulianbarbu: Exactly. I thought about an `Admin` scope but I am not yet clear how it will be able to get the JWT for it.

deployer/src/lib.rs
iulianbarbu: True. Keeping this open because it will have the same solution if we find it. I'll resolve both after we decide what to do.

deployer/src/project/docker.rs
iulianbarbu: Not sure what you mean. The sender is already `async` (https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.UnboundedSender.html).

chesedo: Ahh :man_facepalming: 

But this does not have to be behind a `Mutex` right?

deployer/src/lib.rs
chesedo: Maybe we can put only the `Receiver` behind a mutex then?

chesedo: Nvm, you will need the mutex to reset the channel

deployer/src/project/docker.rs
chesedo: Nvm, we will need the mutex to still reset the channel

chesedo: Then maybe we should have the `lock()` on the mutex usage do an immediate `clone()` to cause the guard to live as short as possible?

iulianbarbu: I believe the guard already lives as short as possible. It should live as long as the operation involving the guarded object. In our case, we send an event after each service state transition, but also when creating the service (which doesn't correspond to a service state transition). Let me know if you thought about something else though.

deployer/src/lib.rs
iulianbarbu: Decided to avoid it for now. The deployer is not on the user network.

PR #1039, opened 2023-06-22
fix: Target directory from config, Windows .exe suffix

### LINKED ISSUE #990: [Bug]: target directory is hardcoded
### What happened?

Currently shuttle uses hardcoded paths for the output directory which is unfortunate for people who change this value in their global cargo settings (in my case that is on a different drive). I got a pretty confusing error message when running shuttle locally about the project binary not being found.

```
No such file or directory (os error 2)
```

After doing some digging I found that the target directory is hardcoded in a bunch of places across the code base. I suggest we replace all of those with the `target_directory` value provided by `cargo metadata`. To me it seems to be a consistent way of getting the _actual_ output directory.

If I go into my project and run `cargo metadata | jq '.target_directory'` it will give me `"/mnt/dev/cargo-global-target"`, which is correct according to how I have cargo configured.

### Version

v0.18.0

Linux

x86_64

### Relevant log output

```sh
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: spawning runtime process

Caused by:
    No such file or directory (os error 2)', /mnt/dev/GitHub/Kazy/shuttle.git/secrecy/cargo-shuttle/src/lib.rs:785:51
```

### Duplicate declaration

### LINKED ISSUE #1034: [Bug]: The system cannot find the file specified. (os error 2)
### What happened?

There are two bugs causing this on local runs:
1. target directory is hard-coded, see #990 
2. Misconfiguration causing the incorrect binary filename to be executed, see https://discord.com/channels/803236282088161321/1120727478332313760/1121093833250328576

This issue can be closed when (2) is completed. (1) is tracked in #990.

The suggested solution can be found at the discord link above.

### Version

v0.19.0

Windows

x86_64

### Relevant log output

```sh
Error: spawning runtime process

Caused by:
    The system cannot find the file specified. (os error 2)
```

### Duplicate declaration

jonaro00:
## Description of change

Closes #990, Closes #1034. Two birds with one stone, or as we say in Sweden: Two flies with one whack

Haven't double checked in a Windows env, but it should work, confirmed by https://discord.com/channels/803236282088161321/1120727478332313760/1121093833250328576

### DIFF
--- a/deployer/src/deployment/queue.rs
+++ b/deployer/src/deployment/queue.rs
@@ -209,10 +209,10 @@ impl Queued {
         let project_path = project_path.canonicalize()?;
 
         // Currently returns the first found shuttle service in a given workspace.
-        let runtime = build_deployment(&project_path, tx.clone()).await?;
+        let built_service = build_deployment(&project_path, tx.clone()).await?;
 
         // Get the Secrets.toml from the shuttle service in the workspace.
-        let secrets = get_secrets(&runtime.working_directory).await?;
+        let secrets = get_secrets(&built_service.working_directory).await?;
 
         // Set the secrets from the service, ignoring any Secrets.toml if it is in the root of the workspace.
         // TODO: refactor this when we support starting multiple services. Do we want to set secrets in the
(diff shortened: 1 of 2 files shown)


oddgrd (approved):
We say that in Norway too :joy: Thanks for this, looks great!

PR #1040, opened 2023-06-22
build(shell.nix): add openssl package to the build dependencies

timonv:
## Description of change
While building the client locally, it was complaining about missing openssl. Added it back and all is good.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -29,6 +29,7 @@ in
       fastmod
       pebble
       kondo
+      openssl
     ];
 
     PROTOC = "${protobuf}/bin/protoc";


oddgrd (commented):
Thanks Timon! We recently migrated to rustls across the board, but it may have been a bit optimistic, when we added cargo-generate for `cargo shuttle init` we got the openssl dependency back :sweat_smile: 

For the nix file, openssl should rather be in `nativeBuildInputs`, by the way

timonv:
> Thanks Timon! We recently migrated to rustls across the board, but it may have been a bit optimistic, when we added cargo-generate for `cargo shuttle init` we got the openssl dependency back 😅
> 
> For the nix file, openssl should rather be in `nativeBuildInputs`, by the way

Thanks! Did just that 🙏

oddgrd:
Nice! We used to only have it in `nativeBuildInputs`, do we need it in both? Sorry for the nitpicking :smile:

timonv:
No problem, oversight on my end, nice catch 💯

oddgrd (approved):

PR #1042, opened 2023-06-22
Feat/refactor gateway state

oddgrd:
## Description of change

This PR refactors the gateway state to fit the new architecture through the gateway migrations, changing the primary key of projects to the new ulid project_id, as well as dropping or replacing some columns. Doing this is a bit convoluted in sqlite since we can't  simply [`ALTER TABLE`](https://www.sqlite.org/lang_altertable.html) to change FK/PK. 

This PR also aims to implement the state changes in the gateway code.

TODO

### DIFF
--- a/gateway/src/api/latest.rs
+++ b/gateway/src/api/latest.rs
@@ -95,6 +95,7 @@ impl StatusResponse {
     }
 }
 
+#[instrument]
 #[utoipa::path(
     get,
     path = "/",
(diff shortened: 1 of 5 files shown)


gateway/src/api/latest.rs
```diff
@@ -95,6 +95,7 @@ impl StatusResponse {
     }
 }
 
+#[instrument]
```
chesedo, line 98: In theory these should not be needed on endpoints, but maybe the tracing layer has issues :smile:

chesedo (approved):
LGTM

gateway/src/api/latest.rs
oddgrd: Ah, I see. Shall we remove them from all the endpoints then? I will merge this now either way, though.

PR #1048, opened 2023-06-24
feat: implement `ResourceBuilder` for `diesel::postgres::Pool`

RobWalt:
## Description of change

This change also slightly restructures the `postgres` module by splitting it up into two modules, one for each `sqlx` and `diesel`.

The rough design as well as the use of the `deadpool_diesel` crate were inspired by the [`diesel-postgres` example in the `axum` crate](https://github.com/tokio-rs/axum/blob/main/examples/diesel-postgres/src/main.rs)

## Breaking changes

This commit also includes a breaking change. Since the existing struct named `Postgres` would be ambiguous in the face of two possible postgres database connection pool type, it was renamed to `PostgresSQLX` to make that clearer. I need you to make a decision and pick one of the two following:

1. We keep the change. Then I'm going to apply it to all code existing in the `shuttle` and `shuttle-examples` repo.
2. We discard the change. In this case the `sqlx::PgPool` related struct will keep its name `Postgres` and I need you to provide a possible name for the `diesel` version.

I'll also provide an example which implements a very basic example (axum-diesel-postgres). You'll find the link to the example [here](https://github.com/shuttle-hq/shuttle-examples/pull/66) (It currently uses the patched `shuttle` crate as a `path` dependencies which needs to be fixed but the code is there. I only tested the example locally on my own machine (NixOS 23.05)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "examples"]
 	path = examples
-	url = ../examples.git
+	url = https://github.com/shuttle-hq/shuttle-examples.git
--- a/resources/shared-db/Cargo.toml
+++ b/resources/shared-db/Cargo.toml
@@ -8,11 +8,12 @@ keywords = ["shuttle-service", "database"]
 
 [dependencies]
 async-trait = "0.1.56"
+deadpool-diesel = { version = "0.4.1", features = ["postgres"], optional = true }
 mongodb = { version = "2.3.0", optional = true }
 serde = { version = "1.0.148", features = ["derive"] }
 shuttle-service = { path = "../../service", version = "0.19.0" }
 sqlx = { version = "0.6.2", optional = true }
 
 [features]
-postgres = ["sqlx/postgres", "sqlx/runtime-tokio-native-tls"]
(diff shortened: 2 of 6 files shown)


RobWalt:
Note to self: revert 2e475f2 and rebase if [this](https://github.com/shuttle-hq/shuttle/pull/1049) is merged.

oddgrd:
Hey @RobWalt, thanks for this! Apologies for the delayed response, we were on a company retreat last week. We'll try to get a review in this week :slightly_smiling_face:

oddgrd:
Hey @RobWalt! Apologies for the delay again :smile: So for this integration, I'm wondering if you'd rather be willing to make this an external community crate? Someone from our community has created a similar crate for another flavor of diesel: https://github.com/aumetra/shuttle-diesel-async/tree/main

The reason I ask is that having this integrated directly into `shuttle_shared_db` will make it increasingly complex, and there are other ORMs that people may want supported in the future. That's not to say we will never have some internal integration for diesel, but this we way we can see what people end up using and how the integration works out first.

RobWalt:
Yeah makes sense, I'll close this PR then.

PR #1054, opened 2023-06-26
fix: windows local run + log clarifications

jonaro00:
## Description of change

Fixes more Windows issues discovered here https://discord.com/channels/803236282088161321/1122643649503694919

- Added more timeout to the runtime wait.
- The stopped_tx send calls are now ignored if they fail, because there is only a subscriber in deployment.

No further proof than Samarth's code working.

### DIFF
--- a/proto/src/lib.rs
+++ b/proto/src/lib.rs
@@ -265,13 +265,16 @@ pub mod runtime {
 
         // Wait for the spawned process to open the endpoint port.
         // Connecting instantly does not give it enough time.
-        let channel = tokio::time::timeout(Duration::from_millis(500), async move {
+        let channel = tokio::time::timeout(Duration::from_millis(7000), async move {
+            let mut ms = 5;
             loop {
                 if let Ok(channel) = conn.connect().await {
                     break channel;
                 }
                 trace!("waiting for runtime endpoint to open");
-                tokio::time::sleep(Duration::from_millis(5)).await;
+                // exponential backoff
+                tokio::time::sleep(Duration::from_millis(ms)).await;
+                ms *= 2;
             }
         })
         .await
--- a/runtime/src/alpha/mod.rs
+++ b/runtime/src/alpha/mod.rs
@@ -52,10 +52,12 @@ mod args;
 pub async fn start(loader: impl Loader<ProvisionerFactory> + Send + 'static) {
     let args = match Args::parse() {
         Ok(args) => args,
-        Err(_) => {
+        Err(e) => {
+            println!("{e}");
             let help_str = "[HINT]: Run shuttle with `cargo shuttle run`";
             let wrapper_str = "-".repeat(help_str.len());
-            return println!("{wrapper_str}\n{help_str}\n{wrapper_str}",);
+            println!("{wrapper_str}\n{help_str}\n{wrapper_str}");
+            return;
         }
     };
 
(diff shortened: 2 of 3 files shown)


runtime/src/alpha/mod.rs
```diff
@@ -333,7 +335,7 @@ where
                     match res {
                         Ok(_) => {
                             info!("service stopped all on its own");
-                            stopped_tx.send((StopReason::End, String::new())).unwrap();
+                            let _ = stopped_tx.send((StopReason::End, String::new()));
```
oddgrd, line 338: I'm wondering if we should log these errors in a `trace!` call? It might be useful in the future.

oddgrd (approved):
LGTM! Just a few suggestions for tracing calls

runtime/src/alpha/mod.rs
jonaro00: Sure. For now it is better to not unwrap them IMO.

runtime/src/alpha/mod.rs
```diff
@@ -52,10 +52,12 @@ mod args;
 pub async fn start(loader: impl Loader<ProvisionerFactory> + Send + 'static) {
     let args = match Args::parse() {
         Ok(args) => args,
-        Err(_) => {
+        Err(e) => {
+            println!("{e}");
```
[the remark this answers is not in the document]
jonaro00: A tracing call there is not visible without log env variable. The intention is to show what caused the arg parsing error. 

Now that I'm testing it, the Display impl for the Error type does not seem to work...

proto/src/lib.rs
```diff
@@ -265,13 +265,16 @@ pub mod runtime {
 
         // Wait for the spawned process to open the endpoint port.
         // Connecting instantly does not give it enough time.
-        let channel = tokio::time::timeout(Duration::from_millis(500), async move {
+        let channel = tokio::time::timeout(Duration::from_millis(7000), async move {
```
iulianbarbu, line 268: nit: Why do we need to change this given you've introduced the exponential backoff?

runtime/src/alpha/mod.rs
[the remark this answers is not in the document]
jonaro00: Nevermind, I had misconfigured my [patch]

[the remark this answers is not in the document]
jonaro00: But yeah, since only the developer will be interested in which error actually caused the arg parse to fail, it can be a tracing error.

runtime/src/alpha/mod.rs
iulianbarbu: If we unwrap without a trace we're losing the error, right? Tracing it will at least log the errors and make them searchable.

proto/src/lib.rs
jonaro00: The discord thread saw a case where the startup took -> 510 <- ms 😅. So I increased it a lot to be sure.

iulianbarbu: Got it. It is more than a 10x increase. My thinking is that the exponential backoff is more than enough to handle any variance. The 500 vs 510 vs 7000 would be handled very nicely by it. I wouldn't change the initial 500ms we're waiting for.

jonaro00: 500/7000 is the maximum time for the entire backoff to happen within. This means several attempts will be made if the connect() call fails instantly. It seemed to me like the windows connect call took much longer (perhaps all of those 510 ms?) which means we want a looser max time.
The backoff is mainly for preventing log spam in the case of failure.

iulianbarbu: Oh, I misread this. Yeh, keeping the backoff within the 7000 ms should be good. Resolving this.

iulianbarbu (approved):
LGTM! thanks

PR #1056, opened 2023-06-27
gateway: status check includes info about auth & provisioner

iulianbarbu:
## Description

The gateway status endpoint returns back an array with the statuses for gateway, auth and provisioner.

Manually tested it locally.


gateway/src/api/latest.rs
```diff
@@ -255,23 +259,53 @@ async fn route_project(
...
     } else {
-        (StatusCode::OK, StatusResponse::healthy())
+        StatusResponse::unhealthy()
     };
 
-    let body = serde_json::to_vec(&body).unwrap();
+    statuses.push(("shuttle-provisioner", provisioner_status));
+
+    // Compute auth status.
+    let auth_status = if let Ok(response) = reqwest::get(service.auth_uri().to_string()).await {
```
oddgrd, line 293: Nit: should we rather have a reqwest client in the router state, or in a lazy_static? So we don't create it for each request.

oddgrd (approved):
LGTM! Just one small comment about potential performance improvement

gateway/src/api/latest.rs
jonaro00: tip: OnceCell / OnceLock in Rust 1.70 😉

iulianbarbu: Thanks guys! Used a `OnceLock` because it might be shared between threads.

PR #989, opened 2023-06-09
fix: dockerfile and ci improvements

jonaro00:
## Description of change

- Cleanup and upgrades in Dockerfiles and CI.
- Minor fixes in builder code.

## How has this been tested? (if applicable)

Tested some deploys with a minimal service.

PR #1060, opened 2023-06-28
fix: bash bug in containerfile

oddgrd:
## Description of change

There was a bug where if the bash conditional evaluated to false, it would fail, I am not sure why. Changing it away from the shorthand resolved it.

Tested by running `make shuttle-provisioner` with and without PROD=true and both cases work.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -16,8 +16,9 @@ COPY . .
 RUN find ${SRC_CRATES} \( -name "*.proto" -or -name "*.rs" -or -name "*.toml" -or -name "Cargo.lock" -or -name "README.md" -or -name "*.sql" \) -type f -exec install -D \{\} /build/\{\} \;
 # This is used to carry over in the docker images any *.pem files from shuttle root directory,
 # to be used for TLS testing, as described here in the admin README.md.
-RUN [ "$CARGO_PROFILE" != "release" ] && \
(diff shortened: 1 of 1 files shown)


jonaro00:
That's strange, it was introduced a long time ago. https://github.com/shuttle-hq/shuttle/commit/68438742f0f2d43bc980a72acd2143794ce1e33f#diff-5fcdf9b4580789697d834d1456a22bcfaa236d668fc180cad4775afc36ed5914R23

Could my bump of the Dockerfile syntax version be the cause? It was only rc -> official.

Anyways, if the condition evaluates to false, the command exits with non zero, so that's the "reason" it fails. Not sure why it worked before tho :)

oddgrd:
Oops, I mean the build failed if it evaluated to true. I believe it always evaluated to false before your PR since it was missing the ARG CARGO_PROFILE, so your PR fixed it but that revealed a bash bug that I don't quite understand :sweat_smile:

jonaro00:
Then it could be that Dockerfile evaluates RUN lines with `sh` and not `bash`, which would explain why it fails at `[[`. 
The same bug was fixed here https://github.com/shuttle-hq/shuttle/pull/989/files#diff-682a893302ebb3f2ed26573ec16e2f46e6642492a65a6853405963b7d989300eR1

PR #1063, opened 2023-06-28
fix: remove md that is incompatible with docs

oddgrd:
## Description of change

This fixes the parsing error in [the docs](https://docs.shuttle.rs/community/contribute). 

Tested with `yarn mintlify dev` in the docs repo on latest mintlify.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,13 +56,11 @@ You can now start a local deployment of Shuttle and the required containers with
 USE_PANAMAX=disable make up
 ```
 
-<!-- markdownlint-disable no-blanks-blockquote -->
 > Note: `make up` does not start [panamax](https://github.com/panamax-rs/panamax) by default, if you do need to start panamax for local development, run this command with `make COMPOSE_PROFILES=panamax up`.
 
 > Note: `make up` can also be run with `SHUTTLE_DETACH=disable`, which means docker-compose will not be run with `--detach`. This is often desirable for local testing.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -400,4 +398,4 @@ git config --global core.autocrlf true
 
 After you run this command, you should be able to checkout projects that are maintained using CRLF (Windows) again.
 
-[^1]: <https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf>
+[^1]: https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf
```
oddgrd, line 401: @jonaro00 do you recall why you changed this? Removing it fixes the parsing error in the docs, but it might be needed?

jonaro00: Mainly due to markdown lint rules, but as you also have discovered, Mintlify's markdown parser does not even support comments...

jonaro00: In my other large docs PR, using Mintlify v3, I changed the Contributing sidebar item to be a link to CONTRIBUTING.md instead. Sacrificing the quality (graph etc) from GitHub's rendering to mirror it is a waste IMO. A contributor should not be afraid of starting out on a markdown file on GH.

oddgrd: Yes, that is a very fair point, there certainly are some footguns, I encountered a few myself testing this. Our understanding was that they would support mermaid soon, but they must not have had time to integrate it yet. :sweat_smile: 

I will merge this for now so we fix the parsing error, but I think just linking to github is definitely worth considering. I'll ask Ivan if they replied about mermaid as well.

Issue #1052, opened 2023-06-25
[Improvement]: Add list and remove operations to shuttle-persist


jonaro00:
### Describe the improvement

The API of shuttle-persist consists of `save` and `load` methods.

Adding `list` (iterate over keys), and `remove` should be easy to add to the current implementation ~~so that I can finish my TODO list app FINALLY AFTER SIX MONTHS (jk)~~.

### Duplicate declaration

- [X] I have searched the issues and this improvement has not been requested before.

sentinel1909:
I'll take this one, seems like a good starting point for me.  It will take me awhile though, and I might have to ask for some coaching along the way. :)

sentinel1909:
Have been turning this issue over and around in my mind for the last couple of days. Trying to force myself to think about solutions before I start slinging code.  For the list method, I'd like to be able to iterate over all the keys to get their values. You can't iterate over the fields of a struct, so some Googling this morning revealed that the struct could be converted into something iterable, like an array.  Once you have an array, then you can iterate and do things.  This feels like a brute force approach, and I'm missing something more elegant, am I on the right track though?

jonaro00:
Hey Jeff, here are some tips:

- (If the solution in my head works, which it might not 🤞) To solve this task, you need to add two public methods to the PersistInstace impl:
```rust
pub fn list(&self) -> Result<Vec<String>, PersistError> {
    ...
}

// "remove" or "delete", you choose the name
pub fn delete(&self) -> Result<(), PersistError> {
    ...
}

// bonus points: a remove_all/delete_all/clear that deletes everything
```
- Some things that will get you on the right track towards implementing them:
  - Where are keys & values saved? Where is the key string stored (and are there and bugs/limitations because of that?) and where is the value stored? (the data format is irrelevant for this task)
  - What type of operations are needed to achieve the tasks? (think which terminal commands they are similar to)
  - Which methods in the standard library can perform these operations? They return errors, so to handle that nicely, create two new variants of PersistError that represent and wrap that error type. Then, follow the same pattern as the save and load methods to convert the internal error to a PersistError.

Hope that helps! You can also ask in #build-together for more hints :D

sentinel1909:
Thanks so much! Very helpful breadcrumbs for sure :)

jonaro00:
[...]

PR #1066, opened 2023-07-02
[Improvement]: Add list, remove, clear, and size operations to shuttle-persist

sentinel1909:
## Description of change

This PR adds list and remove operations to shuttle-persist, closes issue #1052. 
Looking for feed back to see if I'm on the right track. I feel I'm not handling errors in quite the right way yet.  The only way to learn is to practice and ask for feedback.

Created two unit tests, test_list and test_remove, to test the new functionality. All tests passed.


jonaro00:
1. Use cargo fmt to get past that step in CI.
2. You can refactor the error handling by following the pattern in the other methods. The key is `map_err` and the `?` shorthand return.
3. You can probably squeeze in the remove operation in this same PR. This feature is not in a rush, and it is nice if the updates are released together.

sentinel1909:
> 1. Use cargo fmt to get past that step in CI.
> 2. You can refactor the error handling by following the pattern in the other methods. The key is `map_err` and the `?` shorthand return.
> 3. You can probably squeeze in the remove operation in this same PR. This feature is not in a rush, and it is nice if the updates are released together.

I will be going back to the drawing board, as the saying goes.  I'm having great difficulty with the ? operator and error conversion.

Will definitely work in the remove feature.  

Glad there's no rush here because this is a meaty bone for me to chew on and it's going to take more time.

sentinel1909:
One thing I forgot to ask, comments or no comments?  I noticed there aren't any elsewhere around where I'm working.

sentinel1909:
Tests are a failin' where it matters...having a "but it worked on my machine" moment. :)  Back to the head banging.

jonaro00:
It is starting to look really good now!

One thing that I would like to see it for the file name to be stripped of `.bin` when listing, so that (pseudo-code) `remove(list()[0])` will work.

Did you also want to add `clear()`? It should not be much more than one line of code.

sentinel1909:
> 

Ooooo...challenge accepted :)  The clear() method would nuke everything from orbit I assume?

sentinel1909:
> Ooooo...challenge accepted :) The clear() method would nuke all the keys from orbit I assume?

Looks like fs::remove_dir_all will do the trick and is indeed a one-liner.  fs::remove_dir won't cut it because it won't work if directory isn't empty.

sentinel1909:
> One thing that I would like to see it for the file name to be stripped of .bin when listing, so that (pseudo-code) remove(list()[0]) will work.

Ok, clear() was easy and is done with a test.  For the refinement to list, it should strip off the .bin from the file extension somewhere before going into the list vector that gets returned, correct?  This would allow you to pass in a list with the index as you say to another method like remove.  This will cause some headscratching but I'm gonna get it! :)

sentinel1909:
My first pass at the functionality noted earlier, remove() will now accept a Vec<String> (like what comes out of list(), get the index of the item in it, then use that to pass to get_storage_file() to get the file back and subsquently delete it.  The clear() functionality is also here.  Tests pass still.  I'm not sure the updated remove functionality is efficient and I feel like I might need to think about it further.

sentinel1909:
The remove method will accept a list_item parameter, which is a string.  This could be the output from list(), called as you outlined in the commend above, i.e. `remove(list()[0])`

jonaro00:
The current state of the remove function will just iterate over each item and then delete the last index. I still think remove should just delete one key. What I meant with `remove(list()[0])` is that remove() should be able to successfully delete any arbitrary key from the vector that list() returns (in my example, the 0'th). This was why I hinted about removing the `.bin` suffix.

sentinel1909:
> The current state of the remove function will just iterate over each item and then delete the last index. I still think remove should just delete one key. What I meant with `remove(list()[0])` is that remove() should be able to successfully delete any arbitrary key from the vector that list() returns (in my example, the 0'th). This was why I hinted about removing the `.bin` suffix.

Understood.

resources/persist/src/lib.rs
```diff
@@ -41,6 +47,41 @@ impl PersistInstance {
...
+
+        let entries = fs::read_dir(storage_folder).map_err(PersistError::ListFolder)?;
+        for entry in entries {
+            let dir = entry.map_err(PersistError::ListFolder)?;
+            list.push(
+                dir.path()
+                    .into_os_string()
+                    .into_string()
+                    .unwrap_or_else(|_| "path contains non-UTF-8 characters".to_string())
+                    .split(".bin")
```
jonaro00, line 64: Splitting here works, but can break: if a key `aaa.binbbb` is saved, that would make the file `aaa.binbbb.bin`, which after this split+collect becomes `aaabbb`. I would say trimming the file extension from the end is better. You can check out `std::path::Path::with_extension`.

resources/persist/src/lib.rs
```diff
@@ -41,6 +47,41 @@ impl PersistInstance {
...
+
+    /// clear method removes the storage folder from the PersistInstance
+    pub fn clear(&self) -> Result<(), PersistError> {
+        let folder_path = self.get_storage_folder();
+        fs::remove_dir_all(folder_path).map_err(PersistError::ClearFolder)?;
+        Ok(())
+    }
+
+    /// remove method deletes a key from the PersistInstance
+    pub fn remove(&self, key: String) -> Result<(), PersistError> {
```
jonaro00, line 79: Since save and load takes the key as a `&str`, it makes sense that this also does. This means that you will need `.as_str()` in the test case.

resources/persist/src/lib.rs
```diff
@@ -111,13 +152,53 @@ mod tests {
...
 
+    #[test]
+    fn test_list() {
+        let persist = PersistInstance {
+            service_name: ServiceName::from_str("test_list").unwrap(),
+        };
+
+        persist.save("test_list", "test_list").unwrap();
+
+        let result = vec!["shuttle_persist/test_list/test_list".to_string()];
```
jonaro00, line 163: Should the resulting list item really contain the full path like this?

resources/persist/src/lib.rs
```diff
@@ -111,13 +152,53 @@ mod tests {
...
+            service_name: ServiceName::from_str("test_clear").unwrap(),
+        };
+
+        persist.save("test_clear", "test_clear").unwrap();
+        persist.clear().unwrap();
+        let clear_result = persist.list().unwrap_err();
+        assert_eq!(
+            clear_result.to_string(),
+            "failed to list contents of folder: No such file or directory (os error 2)"
+        );
```
jonaro00, line 180: Maybe it is unintuitive that calling list() on an empty persist store returns an error. If you also call `create_dir_all` in list() like save() does, you can make sure that the folder always exists when asked for, and an empty Vec can be returned.

jonaro00 (commented):
You are making good progress! 🥳 Here are some suggestions.

resources/persist/src/lib.rs
sentinel1909: Thanks.  Waffled endlessly here.  Thanks for confirming that staying consistent with the rest is best.

resources/persist/src/lib.rs
sentinel1909: Struggling here.  Not seeing how std:::path::Path::with_extension helps. I'm getting better at interpreting the docs content, but I have a ways to go. Will resume the head banging.

jonaro00: My mistake. The `dir.path()` method gives you a PathBuf, and there the method is called `set_extension()`. The idea is to modify the path of the dir entry before converting it to a string, making it more "properly" handled.

sentinel1909: Have adjusted the list method to a good form now, I feel.

resources/persist/src/lib.rs
sentinel1909: Now using .as_str() in the assocated test case.

resources/persist/src/lib.rs
sentinel1909: Fixed, list returns the key name only.

resources/persist/src/lib.rs
sentinel1909: Now just removing the files instead of the entire folder, test adjusted accordingly.

sentinel1909:
I would like to embelish the test for the clear method to create multiple keys within the sample instance of persist, list the result (should contain multiple keys), then remove a random result.  This doesn't seem to work.  I only ever get one instance of persist with one key.  Creating multiples (with a for loop) just overwrites over top of the same one.  I will continue to think about this.

jonaro00:
Another thing: The '.bin' files from the tests should not be checked in. You can remove them in a commit, and then help future yourself and others by adding a gitingore file in the persist directory that ignores `*.bin`.

The .size() method should also be very simple to implement.

sentinel1909:
list(), remove(), clear() and size() methods all now added, along with assocated tests.  I've added an element of randomness to the tests, which will now create a randomized list of keys before testing the associated method.

jonaro00:
The implementation of size() surprised me, as the initial suggestion was

> Nice! I think it would be interesting to have a .size() method, right? It could just count the amount of files in the folder

I was thinking this would simply be a shorthand for checking the len() of list(), but that is quite trivial for the end user. Knowing the persist's storage size on disk might be useful if the future introduces storage limits, but I don't see it being useful for a user yet. We can keep it as is if the doc comment clearly states what it does, so it is not confused with the above.

---

The randomness added in the tests are strange to me. You generally want to avoid randomness in tests. While the loops you constructed will seem to always work, the randomness adds no "proof" that it works.
Now, the number of keys *N* handled in those tests are between 1 and 20, but ensuring it works for 1 and 2 should be enough to be sure that it works for larger *N*.
One thing that the tests are missing is *N*=0 (aka does the persist store work as expected when empty?).
You could add *N*=0 checks before and after some of the tests, for example
```rs
persist.remove("abc") // assert this is an Err (or just unwrap_err)
persist.save("abc", "xyz")
persist.remove("abc") // assert this is an Ok (or just unwrap)
persist.remove("abc") // assert this is an Err (or just unwrap_err)
```
The same pattern can repeat where relevant, for example, checking if list() gives an empty list or an error when empty.

sentinel1909:
Going back to the drawing board on size() as well as the tests. Sorry I missed the memo on size(), my mind just went to a literal place, not a simple counting of the strings in the list.

For the tests, I blew it and will re-think them.

sentinel1909:
I added a new() method, which looks like this: 

```Rust
pub fn new(name: &str) -> Self {
     PersistInstance { service_name: ServiceName::from_str(name).unwrap() }
```

The name gets passed in as a parameter when you call new() and then you get back your PersistInstance. I used this in the tests and it works nicely.  Two things:

1) Not sure if this is wanted, if I'm going off script, let me know and I'll change it back.
2) The unwrap() worries me and I'm not sure how to propgate the error nicely with the ? operator.

I've thrown all the tests out and redid them.  I haven't committed yet as I still need to address  the N=0 checks as noted above.

jonaro00:
I'd say we can wait with involving ServiceName since it looks like there are huge changes regarding those coming soon. (To propagate with ?, you would have made the `new()` function return `Result<Self, ...>`)

clear() could take the approach of just removing and re-creating the directory.

sentinel1909:
Looks like I failed a couple of the integration tests for cargo-shuttle. Looked at the details, not sure how to resolve them.

resources/persist/src/lib.rs
```diff
@@ -111,13 +158,95 @@ mod tests {
         assert_eq!(result, "test");
     }
 
+    #[test]
+    fn test_list_and_size() {
+        let persist = PersistInstance {
+            service_name: ServiceName::from_str("test1").unwrap(),
```
jonaro00, line 164: The tests are looking really good now! However, there is one case that is not yet covered, but it might simplify the code.

If list/load/remove etc are called before any key has been saved, they will fail due to the folder not existing yet. This can be avoided if we ensure the folder exists before any method is called.

My idea is to add a method `new(service_name: ServiceName) -> Result<Self, PersistError>` that creates the folder upfront. Creating the folder at this time also means it is not needed in `save`.
This would mean all constructions of `PersistInstance { ... }` change to `PersistInstance::new(...)` + `.unwrap()`/`?`. It also means that the error needs to `map_err` into a `shuttle_service::Error` (or other way of converting) in `ResourceBuilder::output`.

jonaro00 (commented):
Well done! We are close to being finished now :)

resources/persist/src/lib.rs
```diff
@@ -41,6 +47,47 @@ impl PersistInstance {
...
+
+        let entries = fs::read_dir(storage_folder).map_err(PersistError::ListFolder)?;
+        for entry in entries {
+            let key = entry.map_err(PersistError::ListFolder)?;
+            let key_name = key
+                .path()
+                .file_stem()
+                .unwrap_or_default()
+                .to_str()
+                .unwrap_or("file name contains non-UTF-8 characters")
```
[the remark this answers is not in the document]
sentinel1909: This one is causing a bit of head bashing, but will continue to bash. I'm not seeing how to propagate right in this moment, but will get there.

resources/persist/src/lib.rs
sentinel1909: So a new method after all then...got it.

resources/persist/src/lib.rs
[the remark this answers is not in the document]
jonaro00: You can take a shortcut and map the error to a `ListFolder` and propagate with `?` like the statement before it.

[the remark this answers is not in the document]
jonaro00: Aha, well then don't you already have a `Result<&str, PersistError>` after doing `.ok_or(PersistError::ListFolder)`? If so, you can do the `?` immediately, then `to_string()`.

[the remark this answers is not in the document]
sentinel1909: Hmmmmm...might just have to take my laptop with me to work today and continue to work on this at lunch :)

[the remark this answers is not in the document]
sentinel1909: Ok, I think I've got this worked out. I figured out how to use ok_or to convert the error related to the potential for invalid Unicode characters.  Now I just have to work out a test...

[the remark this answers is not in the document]
jonaro00: I would say that error does not need a test case, since it should realistically never happen. Since the keys (file names) are `String`s when they are saved, they should be valid utf-8 when reading the directory as well.

resources/persist/src/lib.rs
sentinel1909: I'm doing a push as a checkpoint to show where I'm at.  As a start, I've made a skeleton new method that instantiates a PersistInstance struct, given a ServiceName. I've substituted this in the tests and all works as before. I'm having difficulty now understanding how to get the folder creation incorporated.  If I put a &self in as a parameter to new, how do I then instantiate in the tests?

sentinel1909: > It also means that the error needs to `map_err` into a `shuttle_service::Error` (or other way of converting) in `ResourceBuilder::output`.

Having some difficulty now with this piece.  I understand what you're saying but am not sure how to make it happen. I will reflect through the day today.

sentinel1909: Alright, time to admit it.  After sititng staring at the code for a fair bit of time, I have no clue how to convert the errors into a shuttle_service::Error in ResourceBuilder::output.  I think I need a hint.

jonaro00: This works
```rs
// Get the PersistError in, let's say, a match statement, then
return Err(shuttle_service::Error::Custom(
    /*The PersistError*/.into(),
));
```

sentinel1909: Ok, I feel I've got the new method working. It will convert a PersistError into a shuttle_service::Error. I'm having difficulty with a test though. All I managed to do yesterday evening was test ServiceName, which panics if you pass it something invalid. This doesn't really test the new method erroring out because of an issue with create_dir_all.  Also, the cargo-shuttle circleci check is still failing, and I'm not clear why (I have tried to read the log output) or how to resolve it.

jonaro00: Yeah, the create_dir_all is hard to test failures of. Perhaps creating a folder in a root-owned directory, but that might not be reproducible on all machines. I would say that we can trust it without a test.

The CI fail on cargo-shuttle init is a sporadic error (a hard one to fix :/ ), and not related to your changes.

resources/persist/src/lib.rs
```diff
@@ -25,22 +33,77 @@ pub enum PersistError {
...
+    /// new method constructs a new PersistInstance along with its associated storage folder
+    pub fn new(service_name: ServiceName) -> Result<Self, shuttle_service::Error> {
+        let instance = Self { service_name };
+        let storage_folder = instance.get_storage_folder();
+        match fs::create_dir_all(storage_folder) {
+            Ok(_) => &instance,
+            Err(e) => {
+                return Err(shuttle_service::Error::Custom(
+                    PersistError::CreateFolder(e).into(),
+                ))
```
jonaro00, line 51: My thought with converting to a `shuttle_service::Error::Custom` was to have that happen in `ResourceBuilder::output()` (since the trait requires a `shuttle_service::Error` type in that result). The `new` function can simply return this `PersistError` with `map_err` and `?`, similar to all other errors in this struct. The construction of `PersistInstance` in `output` should then use `::new(...)` and convert the error if it happens there (with this match clause or map_err).
(The code looks correct, it is just missing the usage of `new()` in the builder 😄)

sentinel1909: Got it.  Will sort it out this weekend.

sentinel1909: I think I got it!  Except for the failing cargo-shuttle circleci :(

jonaro00:
This is looking pretty complete now 🥳! Thanks for the consistent work and tolerating my annoying feedback 😄. I'll let the team know this is ready for review.

sentinel1909:
> This is looking pretty complete now 🥳! Thanks for the consistent work and tolerating my annoying feedback 😄. I'll let the team know this is ready for review.

Thank you for your patience and guidance! Working this solidified a lot of things for me, it was extremely valuable.  The code base is way less scary now :)

sentinel1909:
Wondering if there is any further work to be done here? Are upcoming changes going to make this obsolete and not necessary? Happy to refine this work further if needed.

oddgrd:
Sorry for the delay, Jeff! This looks great, we just haven't been able to get to the final review yet. We'll get to it this week or early next week. :slightly_smiling_face:

orhun (approved):

resources/persist/src/lib.rs
```diff
@@ -25,22 +33,70 @@ pub enum PersistError {
...
+    }
+
+    pub fn save<T: Serialize>(&self, key: &str, struc: T) -> Result<(), PersistError> {
         let file_path = self.get_storage_file(key);
         let file = File::create(file_path).map_err(PersistError::Open)?;
         let mut writer = BufWriter::new(file);
         Ok(serialize_into(&mut writer, &struc).map_err(PersistError::Serialize))?
     }
 
+    /// list method returns a vector of strings containing all the keys associated with a PersistInstance
```
oddgrd, line 58: Nit: should we capitalize these doc comments like we do elsewhere in the codebase? I also don't think we need "list method" as the first part of the comment.
```suggestion
    /// Returns a vector of strings containing all the keys associated with a [PersistInstance]
```

oddgrd (approved):
Looks great to me, thanks for all your work on this @sentinel1909! :partying_face: And thanks for the mentoring @jonaro00! I left a very minor comment about the doc comments, we may want to change them a bit.

Oh, and it would be great to have the persist docs in the docs repo updated with these new features as well. If you have time, if not we can also do it. :slightly_smiling_face:

PR #1068, opened 2023-07-03
fix: rds deploy crash

### LINKED ISSUE #1029: [Bug]: Deployment fails when using shuttle-aws-rds
### What happened?

I was trying to deploy a service using `shuttle-aws-rds` and it was consistently erroring on each deployment. To my surprise, because I couldn't see anything in the logs, the service was deployed and the database was provisioned, so my guess is that the error in the deployment was due to some sort of timeout as the provisioning of the aws resource maybe takes longer than what the cli is expecting.

Is this a known issue? I can see how this may be confusing to some people. The only way to get the connection string to the database was to run `cargo shuttle resource list`.

See the logs below.

Once I see the logs, I can see this:

```txt
2023-06-21T10:28:56.238745673Z  INFO {success="true"} shuttle_deployer::deployment::run: loading response
2023-06-21T10:28:56.252101329Z  INFO Entering running state
2023-06-21T10:28:56.262932293Z DEBUG {service.ready=true} tower::buffer::worker: processing request
2023-06-21T10:28:56.263229824Z  INFO shuttle_deployer::deployment::run: starting service
2023-06-21T10:28:56.265522486Z DEBUG {service.ready=true} tower::buffer::worker: processing request
2023-06-21T10:28:56.268446787Z TRACE shuttle_runtime::alpha: alpha starting
2023-06-21T10:28:56.268456594Z TRACE {service_address="127.0.0.1:19957"} shuttle_runtime::alpha: starting
2023-06-21T10:28:56.268559336Z  INFO {response="StartResponse { success: true }"} shuttle_deployer::deployment::run: start client response: 
2023-06-21T10:28:56.285627368Z  INFO actix_server::builder: starting 4 workers
2023-06-21T10:28:56.285643226Z  INFO actix_server::server: Tokio runtime found; starting in existing Tokio runtime
```

### Version

v0,18,0

In deployment

In deployment

### Relevant log output

```sh
2023-06-21T10:27:28.319393940Z  INFO Entering built state

2023-06-21T10:27:28.319570485Z  INFO Entering loading state
2023-06-21T10:27:28.324570067Z TRACE shuttle_deployer::runtime_manager: making new client
2023-06-21T10:27:28.329432175Z DEBUG shuttle_deployer::runtime_manager: Starting alpha runtime at: /opt/shuttle/shuttle-executables/b0bc819d-ce4d-45f2-a04b-a41da09e5162
2023-06-21T10:27:30.332814984Z  INFO shuttle_proto::runtime: connecting runtime client
2023-06-21T10:27:30.332889398Z DEBUG hyper::client::connect::http: connecting to 127.0.0.1:20952
2023-06-21T10:27:30.335734551Z DEBUG hyper::client::connect::http: connected to 127.0.0.1:20952
2023-06-21T10:27:30.338143033Z DEBUG {service.ready=true} tower::buffer::worker: processing request
2023-06-21T10:27:30.340761682Z TRACE {old_id="ed13908e-7085-4553-954c-b406938a5fa2"} shuttle_deployer::deployment::run: stopping old deployment
2023-06-21T10:27:30.342512658Z TRACE {id="ed13908e-7085-4553-954c-b406938a5fa2"} shuttle_deployer::runtime_manager: sending stop signal for deployment
2023-06-21T10:27:30.345004912Z DEBUG {service.ready=true} tower::buffer::worker: processing request
2023-06-21T10:27:30.348988321Z TRACE {response="Response { metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\", \"date\": \"Wed, 21 Jun 2023 08:27:30 GMT\", \"grpc-status\": \"0\"} }, message: StopResponse { success: true }, extensions: Extensions }"} shuttle_deployer::runtime_manager: stop deployment response
2023-06-21T10:27:30.349623673Z  INFO shuttle_deployer::deployment::run: loading project from: /opt/shuttle/shuttle-executables/b0bc819d-ce4d-45f2-a04b-a41da09e5162
2023-06-21T10:27:30.351972206Z DEBUG shuttle_deployer::deployment::run: loading service
2023-06-21T10:27:30.360895772Z DEBUG {service.ready=true} tower::buffer::worker: processing request
Deployment has not entered the running state

Deployment encountered an unexpected error - Please create a ticket to report this.

Run the following for more details

cargo shuttle logs b0bc819d-ce4d-45f2-a04b-a41da09e5162
```

### Duplicate declaration

oddgrd:
## Description of change

We were incorrectly assuming the state to be running in the logs received from a deployment, but they could also be loading (and they will be loading for a while in the case of rds). This lead to the client reporting that the deployment crashed since it received the running state in the logs stream, but the deployment was still loading (an unexpected state since we received running in logs). 

Closes #1029 

TODO: 
- We're not getting the logs from runtime during loading, although I think we used to (I remember seeing the "deploying <db_type>, this could take a while..." output). :thinking:  This means we get "loading service" from the deployer run function, and then nothing until loading is complete.
- Should we use rwlock for the logger state? Does it matter?

Tested by deploying tide rds example to unstable, as well as some other examples (axum hello world, poem postgres).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ BUILDX_FLAGS=$(BUILDX_OP) $(PLATFORM_FLAGS) $(CACHE_FLAGS)
 # ensuring all user crates are compiled with the same rustc toolchain
 RUSTUP_TOOLCHAIN=1.70.0
 
-TAG?=$(shell git describe --tags)
+TAG?=$(shell git describe --tags --abbrev=0)
 BACKEND_TAG?=$(TAG)
 DEPLOYER_TAG?=$(TAG)
 PROVISIONER_TAG?=$(TAG)
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -1034,23 +1034,9 @@ impl Shuttle {
 
                             return Ok(CommandOutcome::DeploymentFailure);
                         }
-                        shuttle_common::deployment::State::Running => {
-                            println!("got Running in logs stream");
-                            let fields: serde_json::Value =
-                                serde_json::from_slice(log_item.fields.as_slice()).unwrap();
-                            println!("log item: {:?}", fields);
-                            break;
-                        }
-                        shuttle_common::deployment::State::Completed => {
-                            println!("got Completed in logs stream");
-                            break;
-                        }
-                        shuttle_common::deployment::State::Stopped => {
-                            println!("got Stopped in logs stream");
-                            break;
-                        }
-                        shuttle_common::deployment::State::Unknown => {
-                            println!("got Unknown in logs stream");
+                        // Break on remaining end states: Running, Stopped, Completed or Unknown.
+                        end_state => {
+                            debug!(state = %end_state, "received end state, breaking deployment stream");
                             break;
                         }
                     };
@@ -1107,7 +1093,7 @@ impl Shuttle {
                     );
                 }
                 state => {
-                    println!("{state}");
+                    debug!("deployment logs stream received state: {state} when it expected to receive state: running");
                     println!(
                     "Deployment entered an unexpected state - Please create a ticket to report this."
                 );
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -1031,10 +1031,20 @@ impl Shuttle {
 
                             return Ok(CommandOutcome::DeploymentFailure);
                         }
-                        shuttle_common::deployment::State::Running
-                        | shuttle_common::deployment::State::Completed
-                        | shuttle_common::deployment::State::Stopped
-                        | shuttle_common::deployment::State::Unknown => {
+                        shuttle_common::deployment::State::Running => {
+                            println!("got Running in logs stream");
+                            break;
+                        }
+                        shuttle_common::deployment::State::Completed => {
+                            println!("got Completed in logs stream");
+                            break;
+                        }
+                        shuttle_common::deployment::State::Stopped => {
+                            println!("got Stopped in logs stream");
+                            break;
+                        }
+                        shuttle_common::deployment::State::Unknown => {
+                            println!("got Unknown in logs stream");
                             break;
                         }
                     };
@@ -1090,9 +1100,12 @@ impl Shuttle {
                         "State: Crashed - Deployment crashed after startup.".red()
                     );
                 }
-                _ => println!(
-                    "Deployment encountered an unexpected error - Please create a ticket to report this."
-                ),
+                state => {
+                    println!("{state}");
+                    println!(
+                    "Deployment entered an unexpected state - Please create a ticket to report this."
+                );
+                }
             }
 
             println!();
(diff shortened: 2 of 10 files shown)


Makefile
```diff
@@ -24,7 +24,7 @@ BUILDX_FLAGS=$(BUILDX_OP) $(PLATFORM_FLAGS) $(CACHE_FLAGS)
 # ensuring all user crates are compiled with the same rustc toolchain
 RUSTUP_TOOLCHAIN=1.70.0
 
-TAG?=$(shell git describe --tags)
+TAG?=$(shell git describe --tags --abbrev=0)
```
oddgrd, line 27: I can remove this from this PR, but I don't see any problem with it, and it means less rebuilding from scratch since we don't get a new tag on every commit.

runtime/src/logger.rs
```diff
@@ -13,11 +16,15 @@ use tracing_subscriber::Layer;
 
 pub struct Logger {
     tx: UnboundedSender<LogItem>,
+    state: Arc<Mutex<State>>,
```
iulianbarbu, line 19: I thought the deployment state is determined based on the instrumentation happening in the run queue + the deployer tracing layer. Is this something that plays into how the deployment state is computed?

oddgrd: In the deployer it is, but we were incorrectly defaulting to `State::Running` for logs received from the runtime here, in try_from: https://github.com/shuttle-hq/shuttle/blob/main/deployer/src/runtime_manager.rs#L131. 

So with this change we rather default to `State::Loading` for runtime logs, and then change it to `State::Running` after the "start" event. There may well be a more graceful way to do this though, I'll try to think about it more!

cargo-shuttle/src/lib.rs
```diff
@@ -1093,9 +1092,12 @@ impl Shuttle {
                         "State: Crashed - Deployment crashed after startup.".red()
                     );
                 }
-                _ => println!(
-                    "Deployment encountered an unexpected error - Please create a ticket to report this."
-                ),
+                state => {
+                    debug!("deployment logs stream received state: {state} when it expected to receive state: running");
```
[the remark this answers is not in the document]
oddgrd: Good point!

runtime/src/logger.rs
iulianbarbu: I think this looks good though.

What I missed is that we're subscribing to the runtime logs in the `get_runtime_client`. All runtime logs emitted during the `load` and `run` phases are sent to persistence, but just before that, the deployment state is updated according to the log state field.

Also, the runtime logger gets instantiated in the `load` phase but is moved alongside the `loader` in a task that outlives the `load` handler scope, and even the `load` phase, being registered as a tracing subscriber. This means all the runtime logs (including the `start` phase ones) are sent on the channel created by the `SubscribeLogs` RPC, called in the `get_runtime_client`, so eventually the runtime's log state will be sent to the deployer, and the deployment state updated.

iulianbarbu (approved):
LGTM @oddgrd ! This is a great fix.

iulianbarbu:
Related to the TODOs:
1) The missing logs sound like a regression, but I don't see why runtime logs are not sent to the deployer. I would double-check this separately though (seems like this PR doesn't depend on the `load` phase logs at least).
2) I think the RW lock is the safest choice, but I also think we can use the `Mutex` as well for now. I don't see how the runtime can have contention read/write-wise on the logger, to make the use of Mutex inefficient. In case that might happen (by having multiple runtime tasks do >= INFO level logging) then the RW lock would help.

PR #1070, opened 2023-07-04
Explain 413 error when a user tries to create a project larger than the limit

### LINKED ISSUE #1035: [Improvement]: Explain why 413 Request Too Large happens
### Describe the improvement

0.19 introduces a 50MB limit on the deploy request (a compressed TAR archive of your files).

#1012 helps with this due to not uploading .git, but a user's project might still be too large.

The current error message is:
```
Error: 413 Payload Too Large
message: an unexpected error occurred. Please create a ticket to report this
```

We can
1. Print a friendly message when 413 happens (why it happened)
2. Check the body size of the outgoing request and deny the user if it will be too large. (refactor to use the same constant on both ends) (https://github.com/shuttle-hq/shuttle/pull/1031#issuecomment-1600757951)

(2) might make (1) obsolete, but I imagine there could be some rare case that slips through anyways...

@iulianbarbu Do you think we need (1), or just (2)?

### Duplicate declaration

ShouvikGhosh2048:
## Description of change

Adds a client check for project size. If too large, informs the user about the project size limit.
Closes #1035 

Tried it locally on a large project

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -13,6 +13,7 @@ use std::path::{Path, PathBuf};
 use std::process::exit;
 use std::str::FromStr;
 
+use shuttle_common::models::deployment::CREATE_SERVICE_BODY_LIMIT;
 use shuttle_common::{
     claims::{ClaimService, InjectPropagation},
     models::{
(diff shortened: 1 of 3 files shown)


iulianbarbu (approved):
LGTM! Thanks

PR #1071, opened 2023-07-04
Clarify persist.save semantics (not requiring move)

laundmo:
## Description of change
Previously, `PersistInstance.save` required that the value to save will be moved into the function.

Even though this doesn't change any functionality, as passing in a reference worked previously, this PR does clear up the semantics of only needing a reference for persist serialisation.

Use case: in my application, i have a rather big piece of data i want to save when its changed, but which is also read from a lot. I spent a long time struggling before finding out i can just pass a reference to .save anyway.

i've tested it in the main function of a fresh axum template, using this snippet:
```rust
    let mut b = String::from("hello");
    persist.save("test", &b).unwrap();
    b += " world";

    dbg!(b, persist.load::<String>("test"));
```

this is a breaking change, requiring users to add a & to their .save calls.

### DIFF
--- a/resources/persist/src/lib.rs
+++ b/resources/persist/src/lib.rs
@@ -31,14 +31,14 @@ pub struct PersistInstance {
 }
 
 impl PersistInstance {
-    pub fn save<T: Serialize>(&self, key: &str, struc: T) -> Result<(), PersistError> {
+    pub fn save<T: Serialize>(&self, key: &str, struc: &T) -> Result<(), PersistError> {
         let storage_folder = self.get_storage_folder();
         fs::create_dir_all(storage_folder).map_err(PersistError::CreateFolder)?;
 
         let file_path = self.get_storage_file(key);
         let file = File::create(file_path).map_err(PersistError::Open)?;
         let mut writer = BufWriter::new(file);
-        Ok(serialize_into(&mut writer, &struc).map_err(PersistError::Serialize))?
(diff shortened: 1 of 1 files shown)


laundmo:
i'm confused about the CI failure, locally that works just fine.

iulianbarbu:
Hey @laundmo ! The shuttle-persist test fails locally for me.  Can you try reproducing it again? I am `cd'ing`  in the `resources/persist` dir and running `cargo test`. You've done the same?

Btw, I think using `&T` might not work because of the `Sized` marker. We can relax the constraint but afterward, functionally speaking, I am not sure the new interface will work as expected. For example, running this test fails for me every time:
```rust
    #[test]
    fn test_save_and_load_1() {
        let x: u16 = 15;
        let persist = PersistInstance {
            service_name: ServiceName::from_str("test").unwrap(),
        };

        persist.save("test", &x).unwrap();
        let result: u16 = persist.load("test").unwrap();
        assert_eq!(result, 15);
    }
```

I am also wondering if the `T: Serialize + ?Sized` constraint actually would work fine with references. I don't see any implementation for them under serde: https://docs.rs/serde/1.0.171/serde/ser/index.html#implementations-of-serialize-provided-by-serde, and then implementing them might fall under the orphan rule.

jonaro00:
Closing due to inactivity. Thanks for the initiative, and feel free to re-open if you think this should be re-considered.

PR #1075, opened 2023-07-09
fix: broken gateway pagination and status check tests

oddgrd:
## Description of change

When testing h2 project support, I discovered some broken gateway tests. These have (understandably) slipped through since we don't run them in CI, which is the bigger problem here. :smile:  The status test was not updated to reflect the changes we made to the gateway status check for the shuttle status page, and the pagination tests in service were not deterministic.

Run `USE_PANAMAX=disable make shuttle-deployer`, then run the test command from the gateway readme: 
```
SHUTTLE_TESTS_RUNTIME_IMAGE=public.ecr.aws/shuttle-dev/deployer:latest SHUTTLE_TESTS_NETWORK=shuttle-dev_user-net cargo test --package shuttle-gateway --all-features -- --nocapture
```

### DIFF
--- a/gateway/src/api/latest.rs
+++ b/gateway/src/api/latest.rs
@@ -880,6 +880,7 @@ pub mod tests {
     use axum::http::Request;
     use futures::TryFutureExt;
     use hyper::StatusCode;
+    use serde_json::Value;
     use tokio::sync::mpsc::channel;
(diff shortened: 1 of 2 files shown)


iulianbarbu (approved):
LGTM! Thanks @oddgrd !

Issue #872, opened 2023-05-05
[Feature]: Handling regular signals sent to cargo-shuttle on Windows


pedromfedricci:
### Describe the feature

We must handle gracefully when the local deployment of user services is interrupted by sending a signal to the cargo-shuttle project, or any of the user’s services panic, or the cargo-shuttle process is gracefully killed. At this moment we cover some of these in cargo-shuttle for *unix systems (SIGINT, SIGTERM, and user service panic). We must do similar handling for regular signals sent to cargo-shuttle process on Windows.

Requirements:
* reproduce the issue first, so you can verify your solution works.
* add support for Windows common signals handling by killing the shuttle-runtimes gracefully upon the cargo-shuttle process receiving them, or user’s services panic at load time.
* use tokio::signal windows support: https://docs.rs/tokio/latest/tokio/signal/windows/index.html.
* Optional: think about what would be required to gracefully stop all the runtimes upon one of them receiving a signal or panics later during the runtime (for both windows and linux).

### Duplicate declaration

- [X] I have searched the issues and this feature has not been requested before.

pedromfedricci:
ping buddy @piewhat

Shubham8287:
I went through the SIG* handlers for UNIX, and able to reproduce the issue on windows. Stabbing into it.
/attempt #852

PR #1077, opened 2023-07-09
feat: Handling regular signals sent to cargo-shuttle on Windows

Shubham8287:
## Description of change
Gracefully handling windows signal to interrupt local run for #872.
/claim #872

Tested by running on windows, with Ctrl+C

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -926,33 +926,81 @@ impl Shuttle {
     async fn local_run(&self, run_args: RunArgs) -> Result<()> {
         let services = self.pre_local_run(&run_args).await?;
         let (provisioner_server, provisioner_port) = Shuttle::setup_local_provisioner().await?;
-
+        let mut ctlr_c_notif =
+            tokio::signal::windows::ctrl_c().expect("Can not get the CtrlC signal receptor");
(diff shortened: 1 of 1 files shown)


Shubham8287:
Right now Implemented handler for Ctrl+C only, it can be extended to other ones. Would love to get some input on which one to add from https://docs.rs/tokio/latest/tokio/signal/windows/index.html.
Also, Signal handler looks is mostly repeated for Unix and windows and between different handlers. Will try to have signal handling as wrapper or something.

iulianbarbu:
I think we need to handle all of them: https://learn.microsoft.com/en-us/windows/console/setconsolectrlhandler#remarks. A default handler for all these signals calls `ExitProcess`.

Shubham8287:
Added all the handlers, please review

iulianbarbu:
Seems like tests are failing because of `cargo fmt`. Can you fix it?

Shubham8287:
@iulianbarbu  passed all checks now.

iulianbarbu:
Thanks @Shubham8287 ! We just need to do some manual testing on a windows machine. We'll follow up after.

iulianbarbu (approved):
Testing worked well on my side @Shubham8287 ! thanks.

PR #1080, opened 2023-07-10
feat: support http2 projects

oddgrd:
## Description of change

Currently it's not possible to host H2 based applications behind the shuttle proxies for two main reasons:

- We check for the host header in the request in the gateway proxy, and use the name from this to find the correct project. However, an H2 request won't have a Host header, in H2 we get the host from the host subcomponent of the URI authority.

- We only start one proxy client in the gateway and deployer proxies, this cannot handle both http1 and h2 interchangeably (they can't share a connection pool anyway), so we need one client for each protocol.

I started implementing solutions for these two points in this PR, but I am going on holiday :beach_umbrella: so I can't finish it now (although it's very nearly there). I'm opening this as a draft in case anyone wants to take a look until I return in two weeks.

To test the full flow here we need to start gateway and deployer, so I tested it with the full environment (make images -> make up). You can recreate just gateway and deployer images when you make changes. 

I tested this with the full environment, running `curl --http2-prior-knowledge --header "Host: test-h2.unstable.shuttleapp.rs"  localhost:8000 -v` for testing http2 and `curl --header "Host: test-h2.unstable.shuttleapp.rs" localhost:8000 -v` for testing http1. I reviewed the logs and they were as expected, I also ran benchmarks that showed http1 having about 10% higher RPS but http2 used less than a third of the bandwidth.

I also ran the gateway tests, which all pass. See gateway readme for instructions on running those tests, they are not currently running in CI.

Benchmarks in release mode on this branch (I did not have time to benchmark the current main, but it should be easy and it will prove there is no performance regression):

```
examples/axum/hello-world (11307da) [$!] is 📦 v0.1.0 via 🦀 v1.70.0 took 10s
❯ rewrk  --http2 -c 40 -t 8 -d 10s -H "Host: test-h2.unstable.shuttleapp.rs"  -h http://localhost:8000
Beginning round 1...
Benchmarking 40 connections @ http://localhost:8000 for 10 second(s)
  Latencies:
    Avg      Stdev    Min      Max
    8.59ms   10.74ms  1.17ms   63.06ms
  Requests:
    Total:  46397  Req/Sec: 4643.50
  Transfer:
    Total: 1.78 MB Transfer Rate: 182.61 KB/Sec


examples/axum/hello-world (11307da) [$!] is 📦 v0.1.0 via 🦀 v1.70.0 took 10s
❯ rewrk  -c 40 -t 8 -d 10s -H "Host: test-h2.unstable.shuttleapp.rs"  -h http://localhost:8000
Beginning round 1...
Benchmarking 40 connections @ http://localhost:8000 for 10 second(s)
  Latencies:
    Avg      Stdev    Min      Max
    8.33ms   2.72ms   0.80ms   46.59ms
  Requests:
    Total:  47998  Req/Sec: 4800.23
  Transfer:
    Total: 6.87 MB Transfer Rate: 703.16 KB/Sec
```


deployer/src/proxy.rs
```diff
@@ -36,41 +40,27 @@ pub async fn handle(
...
             return Ok(Response::builder()
                 .status(StatusCode::BAD_REQUEST)
-                .body(Body::empty())
+                .body(Body::from("request has no host header or URI authority host subcomponent"))
                 .unwrap());
-        }
-    };
+        };
 
-    if host != fqdn {
```
oddgrd, line 55: So the reason I removed this check is that for http2 requests that are proxied via gateway, the URI will be the gateway address and there won't be a host header. Therefore the host will not match the fqdn of the deployer.

What I'm wondering is if there is a problem with removing this? Will anyone outside of the controller be able to reach this proxy? :thinking:  Additionally, we do still check that the `XShuttleProject` header is here.

deployer/src/proxy.rs
```diff
@@ -36,41 +40,27 @@ pub async fn handle(
...
-            .to_owned(),
-        None => {
-            trace!("proxy request has no host header");
+    // First try to get the Host header. This header is not allowed in H2, so if we
+    // can't find it we try to get the host subcomponent from the URI authority.
+    let Some(host) = req
+        .headers()
+        .typed_get::<Host>()
+        .map(|host| fqdn!(host.hostname()))
+        .or_else(|| req.uri().host().map(|host| fqdn!(host))) else {
```
oddgrd, line 49: This is a common way to get the host from http1 | http2, actix-web uses a similar strategy: https://docs.rs/actix-web/4.3.1/src/actix_web/guard/host.rs.html#58-63

deployer/src/proxy.rs
```diff
@@ -97,7 +87,19 @@ pub async fn handle(
...
     };
 
-    match reverse_proxy(remote_address.ip(), &proxy_address.to_string(), req).await {
+    let client = match req.version() {
+        Version::HTTP_10 | Version::HTTP_11 => &H1_PROXY_CLIENT,
+        Version::HTTP_2 => &H2_PROXY_CLIENT,
+        protocol => {
+            error!(protocol = ?protocol, "received request with unsupported protocol");
+            return Ok(Response::builder()
+                .status(StatusCode::INTERNAL_SERVER_ERROR)
```
christos-h, line 96: Should this not be Bad Request?

oddgrd: Oh, thanks! Actually I think it should be [505](https://httpwg.org/specs/rfc9110.html#status.505 with a message detailing which versions are supported.

oddgrd:
I will close this since it's very old, the proxy in the deployer has been removed, and the gateway proxy has been refactored. We do still want to support this feature, but we can do so in a new PR when we are able to prioritize iit.

PR #1082, opened 2023-07-11
Added the missing support for creating custom resources

iulianbarbu:
## Description of change

The `shuttle-common::resource::Type` misses the `Type::Custom` variant that gets to be used when provisioning custom resources according to the `service` crate docs.

Locally against https://github.com/shuttle-hq/shuttle-examples/pull/73 and on unstable.

### DIFF
--- a/common/src/models/resource.rs
+++ b/common/src/models/resource.rs
@@ -22,6 +22,7 @@ pub fn get_resources_table(resources: &Vec<Response>, service_name: &str) -> Str
                 Type::StaticFolder => "Static Folder",
                 Type::Persist => "Persist",
                 Type::Turso => "Turso",
+                Type::Custom => "Custom",
             };
 
             let elements = acc.entry(title).or_insert(Vec::new());
@@ -46,6 +47,10 @@ pub fn get_resources_table(resources: &Vec<Response>, service_name: &str) -> Str
 
         if let Some(persist) = resource_groups.get("Persist") {
             output.push(get_persist_table(persist, service_name));
+        };
+
+        if let Some(custom) = resource_groups.get("Custom") {
+            output.push(get_custom_resources_table(custom, service_name));
         };
 
         output.join("\n")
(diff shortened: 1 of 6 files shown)


common/src/models/resource.rs
```diff
@@ -159,3 +167,32 @@ fn get_persist_table(persist_instances: &[&Response], service_name: &str) -> Str
...
+        .load_preset(UTF8_FULL)
+        .apply_modifier(UTF8_ROUND_CORNERS)
+        .set_header(vec![Cell::new("Instances")
+            .set_alignment(CellAlignment::Center)
+            .add_attribute(Attribute::Bold)]);
+
+    for (idx, _) in custom_resource_instances.iter().enumerate() {
+        // TODO: add some information that would make the custom resources identifiable.
+        // This requires changing the backend resource list response to include a resource identifier
+        // that can be used to query for more info related to a resource.
```
chesedo, line 187: I think we can just print some JSON output for now

common/src/models/resource.rs
```diff
@@ -148,6 +153,9 @@ fn get_persist_table(persist_instances: &[&Response], service_name: &str) -> Str
             .add_attribute(Attribute::Bold)]);
 
     for _ in persist_instances {
+        // TODO: add some information that would make the persist resource identifiable.
+        // This requires changing the backend resource list response to include a resource identifier
+        // that can be used to query for more info related to a resource.
```
chesedo, line 158: I think we can leave the persistence as is for now... unless if there was a request to change :smile:

chesedo (commented):
Left a suggestion for what we can output.

We should also update the resource recorder that just got merged

common/src/models/resource.rs
iulianbarbu: I am not sure what that would be. We can print the `config` or the `data`. Some of them are subject to leaking secrets. I would keep this as is for now and we'll see how this can be improved later on. What do you think?

chesedo: Yip, sounds good to change it once we have a better idea of what to print

PR #1084, opened 2023-07-12
feat: resource recorder

chesedo:
## Description of change
Add the new resource recorder as an isolates service

Btw, this used the following command to only extract the resource recorder from the feature branch
```
git cherry-pick $(g rev-list --reverse upstream/feat/shuttle-runtime-scaling -- resource-recorder)
```

With added integration tests

### DIFF
--- a/resource-recorder/src/lib.rs
+++ b/resource-recorder/src/lib.rs
@@ -1,0 +1,192 @@
+use async_trait::async_trait;
+use dal::{Dal, DalError, Resource};
+use prost_types::TimestampError;
+use shuttle_common::{backends::auth::VerifyClaim, claims::Scope};
+use shuttle_proto::resource_recorder::{
+    self, resource_recorder_server::ResourceRecorder, ProjectResourcesRequest, RecordRequest,
+    ResourcesResponse, ResultResponse, ServiceResourcesRequest,
+};
+use thiserror::Error;
+use tonic::{Request, Response, Status};
+
+pub mod args;
+mod dal;
+mod r#type;
+
+pub use dal::Sqlite;
+use tracing::error;
+use ulid::DecodeError;
+
+/// A wrapper to capture any error possible with this service
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("could not decode id: {0}")]
+    UlidDecode(#[from] DecodeError),
+
+    #[error("failed to interact with database: {0}")]
+    Dal(#[from] DalError),
+
+    #[error("could not parse resource type: {0}")]
+    String(String),
+
+    #[error("could not parse timestamp: {0}")]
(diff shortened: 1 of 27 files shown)


resource-recorder/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+
+        Ok(resources.into_iter().map(Into::into).collect())
+    }
+
+    /// Delete a resource
+    async fn delete_resource(&self, resource: resource_recorder::Resource) -> Result<(), Error> {
+        self.dal.delete_resource(&resource.try_into()?).await?;
+
+        Ok(())
+    }
```
iulianbarbu, line 97: Yeh, wondering if this is enough. Should we cleanup our infra as well? I would cover this in another PR probably, but the endpoint shouldn't be exposed yet given we remove the resources from the persistence and we might be left with dangling resources in the infra, but maybe I am missing something? Also, we should factor in this endpoint in the custom resources definitions too (and infra that lives outside our management e.g. Turso).

resource-recorder/src/type/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+use sqlx::Database;
+
+pub mod database;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Type {
+    Database(database::Type),
+    Secrets,
+    StaticFolder,
+    Persist,
```
iulianbarbu, line 12: I am just wondering if we should have `Turso` here as well. Potentially the custom resource once we merge the #1082 . What do you think?

iulianbarbu (commented):
Just a few questions related to Turso and custom resources.

resource-recorder/src/lib.rs
chesedo: Yip, this should not be exposed yet. Whatever will call this (deployer) will have to also call provisioner to cleanup the infra.

resource-recorder/src/type/mod.rs
chesedo: Yip, I'll have to add `Turso`

resource-recorder/src/lib.rs
iulianbarbu: Do we still want to expose this RPC (even internally)? Maybe we can add a clarifying comment on the RPC method implementation, that this handler still lacks the infrastructure teardown. What do you think?

iulianbarbu (approved):

resource-recorder/src/lib.rs
chesedo: Nah, the infrastructure teardown won't go through this RPC anyway (something else will do the teardown and just record it here). And trying to add the comment here feels out of place.

PR #1088, opened 2023-07-15
feat(runtime): add alpha runtime version check

### LINKED ISSUE #1053: [Improvement]: Warn user if cargo-shuttle and shuttle-runtime are not compatible
### Describe the improvement

This issue covers "alpha" projects (not shuttle-next, which already has a version check).

1. Add an explicit version flag check to `runtime/src/alpha/mod.rs`, just like the one in `runtime/src/bin/shuttle-next.rs`.
2. In `cargo-shuttle/src/lib.rs` in `get_runtime_executable`, before the outer else-clause returns the executable path, run the executable with the version flag and check if it is semver compatible with cargo-shuttle. Code in the `check_version()` function can be re-used or refactored to achieve this.
3. If the versions are not compatible, print a visible warning that notifies the user of this difference and suggests what action is recommended (cargo install or upgrade dependencies), but keep running the program.

### Duplicate declaration

d4ckard:
Add a version flag check to the alpha runtime and check if its version is the same as the version of cargo-shuttle. Warn the user if that is not the case and give hints. The next runtime already has such a check.

Closes #1053

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -604,6 +604,18 @@ impl Shuttle {
                 runtime_path
             } else {
                 trace!(path = ?executable_path, "using alpha runtime");
+                if let Err(err) = check_version(&executable_path) {
+                    println!("Warning: {}.", err);
+                    println!(
+                        "[HINT]: cargo-shuttle is on version {VERSION}. \
+                        Check this project's Cargo.toml if shuttle-runtime matches this."
+                    );
+                    println!(
+                        "[HINT]: If cargo-shuttle was installed \
+                        using cargo, you can get the latest version \
+                        by running `cargo install cargo-shuttle`."
+                    );
+                }
                 executable_path.clone()
             }
         };
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -1397,11 +1397,11 @@ fn check_version(runtime_path: &Path) -> Result<()> {
 /// nonzero digit is the same."
 fn semvers_are_compatible(a: &semver::Version, b: &semver::Version) -> bool {
     if a.major != 0 || b.major != 0 {
-        return a.major == b.major;
+        a.major == b.major
     } else if a.minor != 0 || b.minor != 0 {
-        return a.minor == b.minor;
+        a.minor == b.minor
     } else {
-        return a.patch == b.patch;
+        a.patch == b.patch
     }
 }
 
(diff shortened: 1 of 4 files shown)


jonaro00:
I'll continue the conversation here.

First, the version printing can use the next binary's `print_version` (move to lib.rs).

> What exactly qualifies two versions as being compatible? Currently check_version simply tests for equality between the two versions.

I see that this comparison is done after converting the versions back to strings, but I would hope that the `semver::Version` structs has methods for checking compatible versions. If not, it should be good enough to check that the major and minor numbers are equal.

d4ckard:
> I would hope that the semver::Version structs has methods for checking compatible versions.

I  seems like it doesn't. Looking at [the source](https://docs.rs/semver/latest/src/semver/lib.rs.html#161) it seems that comparison between versions is derived and enforced only by the struct's members. Since the version number fields are plain `u64`s we can be sure that they don't integrate any special logic.

jonaro00:
I found cargo's compatibility check. https://github.com/rust-lang/cargo/blob/master/src/cargo/core/resolver/context.rs#L62

Since it does not seem to be part of the SemVer standard, we can probably implement a simplified check based on cargo's logic "Versions are considered compatible if their left-most non-zero major/minor/patch component is the same"

cargo-shuttle/src/lib.rs
```diff
@@ -1567,4 +1623,23 @@ version = "0.1.0"
...
     }
+
+    #[test]
+    fn semver_compatibility_check_works() {
+        let semver_tests = &[
+            ("1.0.0", "1.0.0", true),
+            ("1.8.0", "1.0.0", true),
+            ("0.1.0", "0.2.1", false),
+            ("0.9.0", "0.2.0", false),
+            ("0.9.0", "0.2.0", false),
```
jonaro00, line 1634: small duplication

jonaro00 (approved):
Looks good

d4ckard:
All checks pass. I think that's all we've discusses so for it for this issue :partying_face:.

iulianbarbu (approved):
LGTM! Thanks.

cargo-shuttle/src/lib.rs
```diff
@@ -604,6 +604,29 @@ impl Shuttle {
...
+                                "[HINT]: You should update cargo-shuttle. \
+                                If cargo-shuttle was installed using cargo, \
+                                you can get the latest version by running \
+                                `cargo install cargo-shuttle`."
+                            );
+                        } else {
+                            println!(
+                                "[HINT]: You should update shuttle-runtime. \
+                                Change its version to {} in this project's \
+                                Cargo.toml to use a compatible version.",
```
oddgrd, line 624: Nit: it's not necessarily incompatible, so we could rephrase it to this (or something like this).
```suggestion
                                "[HINT]: A newer version of shuttle-runtime is available. \
                                Change its version to {} in this project's Cargo.toml to update it.",
```

oddgrd (commented):
Hey @d4ckard, this is great, thanks! Just a minor nit about the warning, since they are not necessarily incompatible, but the warning to update is still useful.

d4ckard:
I agree. It's also more concise this way.

PR #1091, opened 2023-07-17
feat: add a project_id to gateway

chesedo:
## Description of change
This adds a ULID for project IDs to gateway. This will be needed for the new resource recorder.

It also replaces the need for a `created_at` field since the ULID has a timestamp in it.

Most of gateway still uses the project name (and not this new id) everywhere. I think this is fine for now as the deployer update will take care to switch over to the IDs.

By running all the tests locally


iulianbarbu (approved):
LGTM!

PR #1092, opened 2023-07-17
resource-recorder: update the docker compose file

iulianbarbu:
## Description of change

Added a docker-compose service for resource-recorder and `Makefile` tag env variable.

Started Shuttle locally and it worked. Tested on unstable too.

### DIFF
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -325,7 +325,7 @@ impl InitTemplateArg {
             Tide => "tide/hello-world",
             Tower => "tower/hello-world",
             Warp => "warp/hello-world",
-            None => "custom/none",
+            None => "custom-service/none",
         };
 
         TemplateLocation {
@@ -408,7 +408,7 @@ mod tests {
             init_args.git_template().unwrap(),
             Some(TemplateLocation {
                 auto_path: EXAMPLES_REPO.into(),
-                subfolder: Some("custom/none".into())
+                subfolder: Some("custom-service/none".into())
             })
         );
 
--- a/resource-recorder/prepare.sh
+++ b/resource-recorder/prepare.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+###############################################################################
+# This file is used by our common Containerfile incase the container for this #
+# service might need some extra preparation steps for its final image         #
+###############################################################################
+
+# Stuff that depends on local source files
+if [ "$1" = "--after-src" ]; then
+    exit 0
+fi
+
+# Patch crates to be on same versions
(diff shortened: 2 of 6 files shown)


resource-recorder/prepare.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Patch crates to be on same versions
+mkdir -p $CARGO_HOME
+touch $CARGO_HOME/config.toml
+if [[ $PROD != "true" ]]; then
+    echo '
+    [patch.crates-io]
+    shuttle-common = { path = "/usr/src/shuttle/common" }
+    shuttle-proto = { path = "/usr/src/shuttle/proto" }' > $CARGO_HOME/config.toml
+fi
```
chesedo, line 21: Why does resource recorder require patches? It does not compile anything like deployer does

iulianbarbu: True that. The workspace deps for common and proto are path based, so it's all good. Will change this.

.circleci/config.yml
```diff
@@ -660,19 +660,19 @@ workflows:
               only: production
       - approve-push-unstable:
           type: approval
-          filters:
-            branches:
-              only: main
+          # filters:
+          #   branches:
+          #     only: main
```
chesedo, line 665: Just have to remove comments :smile:

chesedo (commented):
Just have to undo comments

.circleci/config.yml
iulianbarbu: Done.

cargo-shuttle/src/args.rs
```diff
@@ -325,7 +325,7 @@ impl InitTemplateArg {
             Tide => "tide/hello-world",
             Tower => "tower/hello-world",
             Warp => "warp/hello-world",
-            None => "custom/none",
+            None => "custom-service/none",
```
chesedo, line 328: Not sure how the change relates to this PR?

chesedo (commented):
PR just has an unexpected change

cargo-shuttle/src/args.rs
iulianbarbu: It doesn't. It's a fix required for the tests to pass and a regression that came with custom-resources. I modified `shuttle-examples`, by changing the `custom` directory name to `custom-service`. The PR is merged on `shuttle-examples` main and will cause failure to initing a `none` example with `cargo-shuttle`.

chesedo: Ah, I see. This should ideally have been a separate PR

chesedo (approved):

PR #1095, opened 2023-07-18
Change service id from UUIDv4 to ULID

iulianbarbu:
## Description of change

Changed service_id from `UUID` to `ULID`.
Added deployer db migration as well, based on https://github.com/shuttle-hq/sqlite-ulid/pull/1.
This is based on: https://github.com/shuttle-hq/shuttle/pull/1092.

Deployed it on unstable and tested it.


common/src/models/deployment.rs
```diff
@@ -7,18 +7,33 @@ use comfy_table::{
...
 #[derive(Deserialize, Serialize)]
 #[cfg_attr(feature = "openapi", derive(ToSchema))]
 #[cfg_attr(feature = "openapi", schema(as = shuttle_common::models::deployment::Response))]
 pub struct Response {
     #[cfg_attr(feature = "openapi", schema(value_type = KnownFormat::Uuid))]
     pub id: Uuid,
-    #[cfg_attr(feature = "openapi", schema(value_type = KnownFormat::Uuid))]
-    pub service_id: Uuid,
+    #[cfg_attr(feature = "openapi", schema(schema_with = ulid_type))]
+    pub service_id: Ulid,
```
chesedo, line 36: I'm wondering if we can make this backwards compatible somehow?

deployer/migrations/0004_change_service_id_to_ulid.sql
```diff
@@ -0,0 +1,107 @@
...
+CREATE TABLE IF NOT EXISTS resources_copy (
+    service_id TEXT,   -- Identifier of the service this resource belongs to.
+    type TEXT,         -- Type of resource this is.
+    data TEXT,         -- Data about this resource.
+    config TEXT,       -- Resource configuration.
+    PRIMARY KEY (service_id, type),
+);
+INSERT INTO resources_copy (service_id, type, data, config)
+SELECT
+  uuid_to_ulid(resources.service_id),
```
chesedo, line 52: Is this deterministic?

chesedo: Also, shouldn't the ulid0 plugin be loaded in `persistence/mod.rs` for this to work?

chesedo (commented):
Just left a few comments

common/src/models/deployment.rs
iulianbarbu: Great catch. Yeh, I think newer `cargo-shuttle` clients will have trouble fetching `deployment` info based on `Uuid`, given the newer `common::deployment::Response` expects the deployer to hand over an `Ulid`. We can make this field plain, relying on the textual representation of `Uuid` and `Ulid`. This means old deployers can still serve `Uuid` (serialized they come as text and `cargo-shuttle` can use them like that), while the new deployer can serve the new `Ulid` (again, having `cargo-shuttle` display it in plain, as a regular string). Have to try it out though, but what do you think?

deployer/migrations/0004_change_service_id_to_ulid.sql
iulianbarbu: > Is this deterministic?

I think so: https://github.com/dylanhart/ulid-rs/blob/master/src/uuid.rs#L8.

>  Also, shouldn't the ulid0 plugin be loaded in persistence/mod.rs for this to work?

Thanks, good catch. Fixed it.

iulianbarbu: We haven't used `uuid_to_ulid` because the uuid misses the time information, which makes the associated ulid unpredictable in terms of time information. Used `uuid_with_datetime` instead with '2023-06-01` as time info for the services ulids that were created before releasing this PR.

common/src/models/deployment.rs
chesedo: Having them plain all the way sounds good :+1:

chesedo (approved):
LGTM :partying_face:

iulianbarbu:
Closing this in favour of #1101.

PR #1098, opened 2023-07-20
fix(codegen): allow `main` function name, prevent clippy warning

jonaro00:
## Description of change

Prefix the main macro's function name to allow the main function to be called `main`.
QUESTION: Should it just be *replaced* with something like `__shuttle_main` instead of being prefixed?

The drop of the `vars` hashmap prevents a clippy warning in user's projects if `clippy::nursery` rules are enabled (https://discord.com/channels/803236282088161321/1076126506323030017/1131270095075823687). This saves the tiniest bit of memory 🤓. Maybe it should just have been clippy-allowed...

cargo test codegen, and verified that a user project with `main()` compiles.


codegen/src/shuttle_main/mod.rs
```diff
@@ -85,15 +85,11 @@ impl Parse for BuilderOption {
...
-        let fn_ident = item_fn.sig.ident.clone();
-
-        if fn_ident.to_string().as_str() == "main" {
-            emit_error!(
-                fn_ident,
-                "shuttle_runtime::main functions cannot be named `main`"
-            );
-            return None;
-        }
+        // rename function to allow any name, such as 'main'
```
oddgrd, line 88: Nit: we use capitalized comments elsewhere, so it would be good to be consistent.

oddgrd (commented):
On the main macro change, nice! I should have thought about it when I implemented the duplicate main test :sweat_smile: 

On the lint, I'm a bit torn on whether we should fix it, since it is an optimization that is very unlikely to ever be significant. We shouldn't strive to follow clippy nursery either, since it does have bad lints ([example](https://github.com/rust-lang/rust-clippy/issues/6947)). This is not blocking though, we can merge as is if others think we should.

chesedo (approved):
LGTM

PR #1099, opened 2023-07-20
deployer: made the resource manager call the resource recorder

iulianbarbu:
## Description of change

Made the deployer resource manager call into the resource recorder.
This is based on #1095 .

## How has this been tested? (if applicable)
TODO: test on unstable

PR #1101, opened 2023-07-21
deployer: sync resource-recorder with persistence

iulianbarbu:
## Description of change

Made the deployer push existing resources persisted locally to the resource-recorder.
Based on #1099 .

Pending unstable testing, on older projects.


deployer/src/persistence/mod.rs
```diff
@@ -71,19 +94,136 @@ impl Persistence {
...
+                // This uses the ulid0.so file in the crate root, with the
+                // LD_LIBRARY_PATH env set in build.rs.
+                .extension("ulid0"),
+        )
+        .await
+        .unwrap();
         Self::from_pool(pool).await
     }
 
+    async fn configure(
```
chesedo, line 123: How is this method different from `from_pool`?

gateway/src/service.rs
```diff
@@ -263,6 +266,7 @@ impl GatewayService {
         let headers = req.headers_mut();
         headers.typed_insert(XShuttleAccountName(account_name.to_string()));
         headers.typed_insert(XShuttleAdminSecret(control_key));
+        headers.typed_insert(XShuttleProjectId(project_id));
```
chesedo, line 269: I'm not sure why we do this? I see there is a new arg on `deployer` which I expect does the same?

chesedo (commented):
Nice, thanks @iulianbarbu !

gateway/src/service.rs
iulianbarbu: Indeed. That was done previously to avoid sending the `project_id` as an arg to the deployer, and it remained like this although we went on the arg path. It is trivial now to use the `project_id` stored on the `Persistence` instance, attached as an extension, and avoid setting this header. Good catch! I will remove it since we don't need it anymore.

deployer/src/deployment/deploy_layer.rs
```diff
@@ -558,23 +563,39 @@ mod tests {
...
+        async fn get_resources(
+            &mut self,
+            _service_id: &ulid::Ulid,
+            _claim: Claim,
+        ) -> Result<ResourcesResponse, Self::Err> {
+            Ok(ResourcesResponse {
+                success: true,
+                message: "dummy impl".to_string(),
+                resources: Vec::new(),
+            })
```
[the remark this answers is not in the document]
iulianbarbu: They are not used but we need them because they're implement for the `StubResourceManager`, which is used as a mock to instantiate a `DeploymentManager`, which is used in multiple places in the tests.

deployer/migrations/0004_change_service_id_to_ulid.sql
```diff
@@ -0,0 +1,169 @@
...
+-- the rest of the tables are still having an FK on service_id.
+CREATE TABLE IF NOT EXISTS services_copy (
+    id TEXT PRIMARY KEY,  -- Ulid identifier of the service.
+    uuid TEXT,            -- Old Uuid identifier of the service.
+    name TEXT UNIQUE      -- Name of the service.
+);
+
+INSERT INTO services_copy (id, uuid, name)
+SELECT
+  upper(ulid_with_datetime(strftime('%Y-%m-%d %H:%M:%f', datetime('2023-06-01')))) as id,
```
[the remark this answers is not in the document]
iulianbarbu: Needed because `ulid_with_datetime` creates ULIDs in lowercase format, while `ulid` crate and the implementation for `Ulid::to_string/from_string` convert them to/from uppercase format. We have to match the case between the two to be able to keep the services state consistent right after the migration.

deployer/src/persistence/mod.rs
iulianbarbu: Not much, it just accepts more arguments needed now by the `Persistence` struct. I deduped the code here... we could probably have a new struct created that wraps the `logs` senders too... but I don't find it extremely valuable atm. LMK what you think.

common/src/claims.rs
```diff
@@ -135,6 +135,7 @@ impl ScopeBuilder {
             Scope::DeploymentPush, // To start an idle deploy
             Scope::Resources,      // To get past resources for an idle deploy
             Scope::Service,        // To get the running deploy for a service
+            Scope::ResourcesWrite,
```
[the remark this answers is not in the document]
iulianbarbu: Done.

deployer/src/deployment/deploy_layer.rs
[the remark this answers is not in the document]
chesedo: Ahh, the return type changed :man_facepalming:

.circleci/config.yml
```diff
@@ -662,24 +666,25 @@ workflows:
               only: production
       - approve-push-unstable:
           type: approval
-          filters:
-            branches:
-              only: main
+          # filters:
+          #   branches:
+          #     only: main
```
chesedo, line 671: Remember to remove these

chesedo (approved):
LGTM :partying_face:

common/src/claims.rs
```diff
@@ -135,6 +135,13 @@ impl ScopeBuilder {
             Scope::DeploymentPush, // To start an idle deploy
             Scope::Resources,      // To get past resources for an idle deploy
             Scope::Service,        // To get the running deploy for a service
+            // To add the locally persisted resources from the older deployers
+            // to the resource-recorder the first time the old deployer is updated
+            // to the new deployer (based on resource-recorder). After this moment
+            // all requests are routed to the resource-recorder instead of local
+            // persistence. We also use it when new project resources are provisioned
+            // and must be saved to the resource-recorder (for new deployments).
```
oddgrd, line 143: Hmm, won't new deployments have a token/claim from the user? I may be out of the loop here, though.

iulianbarbu: Hmm, you're actually right. 🤦 This is for the deployer permissions. Removed the clarifying note. Thanks :D

oddgrd (approved):
LGTM too, thanks for this as well as all the testing Iulian! :partying_face:

PR #1111, opened 2023-07-27
feat: add deployer tier to auth

chesedo:
## Description of change
Makes it possible to add a new machine user that will have deployer rights

By creating new tests and running the updated contributing flow

### DIFF
--- a/auth/src/args.rs
+++ b/auth/src/args.rs
@@ -15,7 +15,8 @@ pub struct Args {
 #[derive(Subcommand, Debug)]
 pub enum Commands {
     Start(StartArgs),
-    Init(InitArgs),
+    InitAdmin(InitArgs),
+    InitDeployer(InitArgs),
 }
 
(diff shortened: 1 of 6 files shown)


auth/src/args.rs
```diff
@@ -15,7 +15,8 @@ pub struct Args {
 #[derive(Subcommand, Debug)]
 pub enum Commands {
     Start(StartArgs),
-    Init(InitArgs),
+    InitAdmin(InitArgs),
+    InitDeployer(InitArgs),
```
iulianbarbu, line 19: nit: I thought about `InitMachineUser` that can have deploy rights.

iulianbarbu (approved):
LGTM 🥳

auth/src/args.rs
chesedo: Hmm, interesting. I will prefer the keep the mapping to the tier type (`Deployer`) closer though

PR #1114, opened 2023-07-28
feat: expand gateways args to receive the key for the machine user

chesedo:
## Description of change
Expands gateways args with the key that will allow it to start up old deploys.

The work of converting the key to a JWT and starting the last idle deploys forms part of another ticket.

By trying to start gateway locally (with the Makefile) to confirm every works correctly


iulianbarbu (approved):
LGTM!

PR #1115, opened 2023-07-28
feat: add a route to deployer to start a past deployment

chesedo:
## Description of change
Adds a `PUT` endpoint to allow the starting of older deploys.

By updating the tests

### DIFF
--- a/deployer/src/handlers/mod.rs
+++ b/deployer/src/handlers/mod.rs
@@ -1,6 +1,6 @@
 mod error;
 
-use crate::deployment::{DeploymentManager, Queued};
+use crate::deployment::{Built, DeploymentManager, Queued};
 use crate::persistence::{Deployment, Log, Persistence, ResourceManager, SecretGetter, State};
 use async_trait::async_trait;
 use axum::extract::{
@@ -132,7 +132,8 @@ impl RouterBuilder {
             .route(
                 "/projects/:project_name/deployments/:deployment_id",
                 get(get_deployment.layer(ScopedLayer::new(vec![Scope::Deployment])))
-                    .delete(delete_deployment.layer(ScopedLayer::new(vec![Scope::DeploymentPush]))),
+                    .delete(delete_deployment.layer(ScopedLayer::new(vec![Scope::DeploymentPush])))
+                    .put(start_deployment.layer(ScopedLayer::new(vec![Scope::DeploymentPush]))),
             )
             .route(
                 "/projects/:project_name/ws/deployments/:deployment_id/logs",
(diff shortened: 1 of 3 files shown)


deployer/src/handlers/mod.rs
```diff
@@ -493,6 +494,43 @@ pub async fn delete_deployment(
...
+        ("deployment_id" = String, Path, description = "The deployment id in uuid format.")
+    )
+)]
+pub async fn start_deployment(
+    Extension(persistence): Extension<Persistence>,
+    Extension(deployment_manager): Extension<DeploymentManager>,
+    Extension(claim): Extension<Claim>,
+    Path((project_name, deployment_id)): Path<(String, Uuid)>,
+) -> Result<()> {
+    if let Some(deployment) = persistence.get_runnable_deployment(&deployment_id).await? {
```
iulianbarbu, line 517: Given we have a deployment per service, I think we could have assumed that we can start the latest deployment as well, but only if it's `Stopped` or `Running`. I wonder if you thought about sending the `deployment_id` from the gateway because of the possibility of having a deployment per service, in a workspace-based Shuttle project.

deployer/src/persistence/mod.rs
```diff
@@ -296,6 +296,24 @@ impl Persistence {
...
+        sqlx::query_as(
+            r#"SELECT d.id, service_id, s.name AS service_name, d.is_next
+                FROM deployments AS d
+                JOIN services AS s ON s.id = d.service_id
+                WHERE state IN (?, ?, ?)
+                AND d.id = ?"#,
+        )
+        .bind(State::Running)
+        .bind(State::Stopped)
+        .bind(State::Completed)
```
iulianbarbu, line 310: nit: the `Completed` deployments might be stateful and be prevented from running again (like a `FnOnce`?), but they can also be stateless. Not sure if it is worth including them here, but can't see them as a big problem (they would also be restarted only if they are idled for some reason, but that would be conflicting with the fact that they are completed, just because that means they should exit by themselves and the project state should be `ProjectStopped` probably?).

chesedo: Hmm, interesting. :thinking:

deployer/src/handlers/mod.rs
chesedo: Yes, but then we won't be able to use this for a `cargo shuttle start` command

iulianbarbu: We want the idle logic to be removed from the deployer and moved to the gateway, so the gateway will decide what deployment to start, leaving just the deployer as a layer of managing runtimes and storing state/exposing state.

iulianbarbu (approved):

deployer/src/persistence/mod.rs
chesedo: I think not covering `Complete` means users won't be able to start stateless projects ever.

But covering `Complete` means users can use some idempotency to get around it. So this seems to cover more options in the future :smile:

PR #1116, opened 2023-07-30
feat: add idle timeout warning after a deploy

christos-h:
## Description of change
Lot's of users get confused when their project goes to sleep after exceeding the idle timeout limit without any activity.

This should hopefully help them.

Not yet, hence the draft.


iulianbarbu (commented):
Thanks @christos-h ! I am also thinking about displaying the `idle_minutes` info in response to `project_create` which gets called for `cargo shuttle project start`, which accounts for the `idle_minutes` behind the scenes even if not set.

As for testing, I think we would need to add an e2e test where we make some kind of CLI screenshot test against the `idle_minutes`. I don't think we have the infra for this yet, so it would be something we'll need to add in this PR. Also `e2e` tests run only on the `production` branch, but I think we can comment the `.cricleci` filter in order to run them as part of the CI for this PR too. 

However, it might be a bit of a stretch as for the scope of this PR to include such level of testing. Some manual tests and screenshots with the output could be of help too, but let us know what you think.

jonaro00 (approved):

orhun (approved):

oddgrd (commented):
Thanks @christos-h! I'm thinking we should also add the warning to "project_create", since that's when it's set. Arguably we should only have it there, but if users miss that then the reminder when deploying might be useful. Edit: I see Iulian already suggested this :smile:

I should be able to implement these changes and resolve the conflicts this week if you don't have time, by the way!

iulianbarbu (approved):

PR #1118, opened 2023-07-31
Chore/bump sqlx

oddgrd:
## Description of change

Bump sqlx internally to 0.7.1.

Tested in CI and cargo checked locally.

### DIFF
--- a/provisioner/src/lib.rs
+++ b/provisioner/src/lib.rs
@@ -1,3 +1,4 @@
+use std::ops::Deref;
 use std::time::Duration;
 
 pub use args::Args;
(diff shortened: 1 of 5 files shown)


provisioner/src/lib.rs
```diff
@@ -173,7 +174,13 @@ impl MyProvisioner {
 
             // Make sure database can't see other databases or other users
             // For #557
-            let options = self.pool.connect_options().clone().database(&database_name);
+            let options = self
+                .pool
+                .connect_options()
+                .deref()
+                .clone()
+                .database(&database_name);
```
oddgrd, line 182: This is how they do it in sqlx's tests: https://github.com/launchbadge/sqlx/blob/main/sqlx-postgres/src/testing/mod.rs#L176-L180

iulianbarbu (approved):
LGTM! I couldn't find yet why we need the `rt-multi-thread` explicitly mentioned now, but I think we already used it as you mentioned in our deployer, so we should be good.

PR #1126, opened 2023-08-02
fix: add volume to mongo container

oddgrd:
## Description of change

We had two reports of MongoDB databases resetting recently. The reason is that the shared MongoDB container was remade, and it was never assigned a volume, so the data was not persisted. 

Deployed the container with the new volume without losing the current data in our staging environment.


chesedo (approved):
LGTM

PR #1127, opened 2023-08-03
fix: install shuttle-next runtime in deployers

### LINKED ISSUE #1096: [Bug]: cargo shuttle deploy (shuttle_deployer::runtime_manager) does not detect the correct path for the cargo executable
### What happened?

I created a hello world shuttle-next app

Running: 
cargo shuttle deploy

compiles and tests ok... then:

```
..
2023-07-19T13:09:11.831655974-05:00  INFO Entering loading state
2023-07-19T13:09:11.836847151-05:00 TRACE shuttle_deployer::runtime_manager: making new client
2023-07-19T13:09:11.841483797-05:00 DEBUG shuttle_deployer::runtime_manager: Returning path to shuttle-next runtime
2023-07-19T13:09:11.844150484-05:00 TRACE shuttle_proto::runtime: Spawning runtime process "/usr/local/cargo/bin/shuttle-next" ["--port", "17236"]

Deployment crashed
```

On my system:

```
% ls /usr/local/cargo/
ls: /usr/local/cargo/: No such file or directory

% which cargo
/Users/----/.cargo/bin/cargo

% which shuttle-next
/Users/----/.cargo/bin/shuttle-next 
```

### Version

v0.21.0

macOS

ARM64

### Relevant log output

```sh
See above
```

### Duplicate declaration

oddgrd:
## Description of change

~~This ensures the shuttle-next runtime binary is installed in the deployer, which shuttle-next services expect. This means building the deployer image will always need to install shuttle-next, which is not ideal.~~ 

The shuttle-next binary was failing to install for production builds since the prepare.sh had already ran once and set the registry to panamax, so when the prepare.sh for shuttle-next install was ran it was trying to fetch the crates from panamax (which is not available when we build the images, it is for deployers to use at runtime). 

In the future, we should consider releasing the shuttle-next binary like we do for the cargo-shuttle CLI, so it can simply be downloaded from GitHub.

Closes #1096 

Deployed new image to unstable and it's working now.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -57,7 +57,8 @@ RUN /prepare.sh "${prepare_args}"
 
 COPY --from=cache /build /usr/src/shuttle/
 
-# Any prepare steps that depend on the COPY from src cache
+# Any prepare steps that depend on the COPY from src cache.
+# This step currently installs shuttle-next and adds the panamax mirror config.
 RUN /prepare.sh --after-src "${prepare_args}"
 
 COPY --from=builder /build/target/${CARGO_PROFILE}/shuttle-${folder} /usr/local/bin/service
--- a/Containerfile
+++ b/Containerfile
@@ -58,7 +58,7 @@ RUN /prepare.sh "${prepare_args}"
 COPY --from=cache /build /usr/src/shuttle/
 
 # Any prepare steps that depend on the COPY from src cache.
-# This step currently installs shuttle-next and adds the panamax mirror config.
+# In the deployer shuttle-next is installed and the panamax mirror config is added in this step.
 RUN /prepare.sh --after-src "${prepare_args}"
 
 COPY --from=builder /build/target/${CARGO_PROFILE}/shuttle-${folder} /usr/local/bin/service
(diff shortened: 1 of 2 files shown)


oddgrd:
Hmm, I misread the prepare.sh script, it should have been installed without my changes too. I'll need to dig further here to find out why it's missing in production.

jonaro00 (approved):
Good catch

Containerfile
```diff
@@ -57,7 +57,8 @@ RUN /prepare.sh "${prepare_args}"
 
 COPY --from=cache /build /usr/src/shuttle/
 
-# Any prepare steps that depend on the COPY from src cache
+# Any prepare steps that depend on the COPY from src cache.
+# This step currently installs shuttle-next and adds the panamax mirror config.
```
jonaro00, line 61: nit: This is done for every Container that follows this structure, so the comment might be too specific? But tbh deployer is kind of the only user of prepare scripts.

jonaro00 (approved):
LGTM

Containerfile
oddgrd: Oh, good point! I'll refactor it to say what this step does in the deployer.

chesedo (approved):
LGTM

PR #1135, opened 2023-08-07
ci: always build images in release profile

oddgrd:
## Description of change

We shouldn't build for staging and production with different images, this could lead to us missing bugs that don't appear in the debug profile. Additionally, this PR changes the docker build output to plain for CI.

Ran with and without PROD=true, and also ran build and deploy to unstable in CI.


iulianbarbu (approved):
LGTM

PR #1136, opened 2023-08-08
ci: upgrade mac bin build machine

oddgrd:
## Description of change

Upgrading to the m1 mac for the mac binary builds should improve speed, see: https://circleci.com/blog/m1-mac-resource-class/

The mac machine we were using was also about to be deprecated in october.

Tested in CI. Reduced total time of job from ~11 min to ~4 min, and I verified the binary still works on arm and intel macs.


iulianbarbu (approved):
LGTM! Please remove the build binaries filter comments before merging.

PR #1139, opened 2023-08-08
feat: merge logger service from feat/shuttle-runtime-scaling

oddgrd:
## Description of change

Merge the commits related to the new `logger` service to main using the following command (as well as manual fix commits):
`git cherry-pick $(g rev-list --reverse upstream/feat/shuttle-runtime-scaling -- logger)`

Ran the integration tests for logger locally and added them to CI.


runtime/src/next/mod.rs
```diff
@@ -288,8 +255,8 @@ impl Router {
         tokio::task::spawn_blocking(move || {
             let mut iter = logs_stream.bytes().filter_map(Result::ok);
 
-            while let Some(log) = Log::from_bytes(&mut iter) {
-                logs_tx.blocking_send(Ok(log.into())).expect("to send log");
+            while let Some(_log) = Log::from_bytes(&mut iter) {
+                // TODO: send to the log server here
```
iulianbarbu, line 259: We can follow up here to send logs in a separate PR for a `shuttle-next` runtime, but we should open a ticket to track this.

runtime/tests/integration/helpers.rs
```diff
@@ -20,11 +23,13 @@ use tonic::{
     transport::{Channel, Server},
     Request, Response, Status,
 };
+use ulid::Ulid;
 
 pub struct TestRuntime {
     pub runtime_client: RuntimeClient<ClaimService<InjectPropagation<Channel>>>,
     pub bin_path: String,
     pub service_name: String,
+    pub deployment_id: Ulid,
```
iulianbarbu, line 32: nit: Not sure if this is intentional, but we currently use `UUIDv4` for deployment ids. We've migrated only the service ids to `ulid`s, for the time being (we'll probably do this for deployments with the advent of the new deployer). For testing purposes I don't think it matters what we use since the runtime is agnostic (it accepts a `String`), but it just misleads/adds confusion if it doesn't correspond with the upstream data type.

oddgrd: Ah, yes, good catch! This is probably since the new deployer used ulid for deployment ID, I will switch it back to uuid.

runtime/src/next/mod.rs
oddgrd: Indeed, will do!

common/src/lib.rs
```diff
@@ -231,6 +231,7 @@ mod tests {
         }
     }
 
+    #[cfg(feature = "persist")]
```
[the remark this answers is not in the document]
oddgrd: I think we can safely remove it, yes

Makefile
```diff
@@ -1,4 +1,4 @@
-SRC_CRATES=deployer common codegen cargo-shuttle proto provisioner service
+SRC_CRATES=auth cargo-shuttle codegen common deployer gateway logger proto provisioner resource-recorder service
```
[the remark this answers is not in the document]
chesedo: Yip, I think that was the idea with them

cargo-shuttle/src/logger_server.rs
```diff
@@ -0,0 +1,95 @@
...
+
+        let logs = request.resource_logs.into_iter().flat_map(|logs| {
+            logs.scope_logs
+                .into_iter()
+                .flat_map(|scope| scope.log_records.into_iter().flat_map(try_from_log_record))
+        });
+
+        for log in logs {
+            println!("{log}");
+        }
```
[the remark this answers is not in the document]
chesedo: Yip, this should print the same logs for local runs

common/src/lib.rs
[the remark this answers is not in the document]
chesedo: This was to make clippy happy when not using `--all-features`

docker-compose.yml
```diff
@@ -7,6 +7,7 @@ volumes:
   panamax-io-index-vol:
   resource-recorder-vol:
   mongodb-vol:
+  logger-vol:
```
chesedo, line 10: We should remember to remove this when switching to Postgres

docker-compose.yml
```diff
@@ -75,6 +76,7 @@ services:
       # This image needs to run highly privileged in order to
       # orchestrate user runtimes safely
       - ${DOCKER_SOCK}:/var/run/docker.sock
+      - gateway-vol:/var/lib
```
chesedo, line 79: It might be worth double checking this change

logger/migrations/0000_init.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS logs (
+    deployment_id TEXT,        -- The deployment that this log line pertains to.
+    shuttle_service_name TEXT, -- The shuttle service which created this log.
```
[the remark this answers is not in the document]
chesedo: Oh, this is the name of the service inside shuttle `deployer`, `builder`, `resource_recorder`, etc, that generated this log. This will allow us to get access to the resource logs and other things

runtime/tests/integration/helpers.rs
chesedo: Or switch it to `String` to make the future change easier

oddgrd: This was the case in a couple of other places too, addressed in [de8c2c1](https://github.com/shuttle-hq/shuttle/pull/1139/commits/de8c2c16f1d149ad7679cf0f9a8bbe67b42b7b8b)

runtime/src/next/mod.rs
oddgrd: ENG-899 opened for this.

docker-compose.yml
oddgrd: Oh, I think I just missed this, this was the case for several of the volumes, but I opted to leave them as they were instead. I will do the same for this.

oddgrd: Removed in favor of the old one.

docker-compose.yml
oddgrd: I will add a note about this to the logger postgres ticket.

logger/migrations/0000_init.sql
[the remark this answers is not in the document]
iulianbarbu: Hmm.. I think we're still exporting those logs to honeycomb. Are we planning on exporting them to the logger service?

docker-compose.yml
```diff
@@ -99,6 +101,33 @@ services:
       timeout: 15s
       retries: 15
       #start_period: 30s
+  logger:
```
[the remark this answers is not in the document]
oddgrd: Done, thanks!

runtime/tests/integration/helpers.rs
oddgrd: I switched it to String in the logger in [53ab231](https://github.com/shuttle-hq/shuttle/pull/1139/commits/53ab2312540bc350ab4a70023733a269fcbeb25d)

logger/migrations/0000_init.sql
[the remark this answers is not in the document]
iulianbarbu: Clarified this internally, the answer is yes. We want to send logs relating to a service that are relevant for a specific user deployment too.

docker-compose.yml
```diff
@@ -49,6 +49,7 @@ services:
   gateway:
     image: "${CONTAINER_REGISTRY}/gateway:${GATEWAY_TAG}"
     depends_on:
+      - provisioner
```
chesedo, line 52: We probably don't want this :smile:

common/src/backends/tracing.rs
```diff
@@ -34,6 +34,7 @@ use crate::tracing::{
 };
 
 const OTLP_ADDRESS: &str = "http://otel-collector:4317";
+const LOGGER_URI: &str = "http://127.0.0.1:8080";
```
chesedo, line 37: Maybe this should be `http://logger:8000`?

chesedo (commented):
We might want to update the defaults to work in production

docker-compose.yml
oddgrd: We had it on main so changing it seemed unrelated to this PR. Perhaps it's because the gateway will start deployers, and the deployers will need a provisioner connection?

common/src/backends/tracing.rs
oddgrd: Yes, that makes sense.

logger/src/dal.rs
```diff
@@ -0,0 +1,292 @@
...
+            .events
+            .into_iter()
+            .flat_map(|event| {
+                let message = event.name;
+
+                let mut fields = from_any_value_kv_to_serde_json_map(event.attributes);
+                fields.insert("message".to_string(), message.into());
+
+                // Every event should have a level field, but we store it in the level column,
+                // so we can remove it from the event fields.
```
chesedo, line 245: Mind explaining this more please :smile:

logger/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+
+[dependencies]
+async-broadcast = "0.5.1"
+async-trait = { workspace = true }
+chrono = { workspace = true }
+clap = { workspace = true }
+opentelemetry-proto = { workspace = true, features = ["gen-tonic", "logs", "traces"] }
```
chesedo, line 13:
```suggestion
opentelemetry-proto = { workspace = true, features = ["gen-tonic", "traces"] }
```
Since we are no longer capturing logs

logger/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+
+        // println!("trace service received request: \n{:#?}", request);
+        let logs: Vec<_> = request
+            .resource_spans
+            .into_iter()
+            .flat_map(Log::try_from_scope_span)
+            .flatten()
+            .collect();
+
+        // println!("flattened logs: {:#?}", logs);
```
chesedo, line 58: We 
```suggestion
        let logs: Vec<_> = request
            .resource_spans
            .into_iter()
            .flat_map(Log::try_from_scope_span)
            .flatten()
            .collect();

```

logger/tests/integration_tests.rs
```diff
@@ -0,0 +1,395 @@
...
+use shuttle_proto::logger::{
+    logger_client::LoggerClient, logger_server::LoggerServer, LogItem, LogLevel, LogsRequest,
+};
+use tokio::time::timeout;
+use tonic::{transport::Server, Request};
+use tracing::{debug, error, info, instrument, trace, warn};
+use tracing_subscriber::prelude::*;
+
+// TODO: find out why these tests affect one-another. If running them together setting the timeouts
+// low will cause them to fail spuriously. If running single tests they always pass.
```
chesedo, line 23: Is this still the case?

logger/tests/integration_tests.rs
```diff
@@ -0,0 +1,395 @@
...
+            fields: json!({"message": "bar"}),
+        },
+    );
+}
+
+/// Helper function to setup a tracing subscriber and run an instrumented fn to produce logs.
+/// The runtime flag determines if we add the deployment_id as an attribute in the tracer config,
+/// which the shuttle-runtime will do, this way all traces from runtime are recorded. If we set
+/// it to false, only spans instrumented with the deployment_id field will be captured.
+fn generate_logs(port: u16, deployment_id: String, generator: fn(String), runtime: bool) {
```
chesedo, line 296: Is there a way to refactor this into two functions so that you don't have to send down the `runtime` argument?

runtime/Cargo.toml
```diff
@@ -16,6 +16,9 @@ doctest = false
 anyhow = { workspace = true }
 async-trait = { workspace = true }
 chrono = { workspace = true }
+opentelemetry = { workspace = true }
+opentelemetry-proto = { workspace = true, features = ["gen-tonic", "logs"] }
```
chesedo, line 20:
```suggestion
opentelemetry-proto = { workspace = true, features = ["gen-tonic", "traces"] }
```
I might be wrong here :thinking:

runtime/src/logger.rs
chesedo, line 1: :partying_face:

runtime/Cargo.toml
chesedo: ... maybe most of these can be removed instead since we are using the tracing otel layer directly

chesedo (commented):
Nice, thanks for all the hard work on this O!

codegen/src/shuttle_main/mod.rs
```diff
@@ -270,24 +270,55 @@ impl ToTokens for Loader {
...
                 use shuttle_runtime::Context;
                 use shuttle_runtime::tracing_subscriber::prelude::*;
+                use shuttle_runtime::opentelemetry_otlp::WithExportConfig;
                 #extra_imports
 
-                let filter_layer =
-                    shuttle_runtime::tracing_subscriber::EnvFilter::try_from_default_env()
-                        .or_else(|_| shuttle_runtime::tracing_subscriber::EnvFilter::try_new("INFO"))
-                        .unwrap();
+                shuttle_runtime::opentelemetry::global::set_text_map_propagator(shuttle_runtime::opentelemetry::sdk::propagation::TraceContextPropagator::new());
```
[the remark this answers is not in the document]
oddgrd: Ah, I'm not either, I think I just pasted everything from setup_tracing when I first started this. It's not in the logger integration tests either. I will remove it. :+1:

common/src/backends/tracing.rs
```diff
@@ -137,3 +154,435 @@ where
         ExtractPropagationFuture { response_future }
     }
 }
+
+/// Record a log for a deployment task
+pub trait DeploymentLogRecorder {
+    fn record_log(&self, details: &DeploymentDetails, visitor: JsonVisitor, metadata: &Metadata);
+}
```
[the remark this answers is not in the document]
oddgrd: True, I was thinking of cleaning the old stuff out in a follow-up, but I can also do it in this PR

logger/tests/integration_tests.rs
oddgrd: It is, yes.

logger/tests/integration_tests.rs
oddgrd: Probably! :smile: This is a bit messy, I'll look into it.

runtime/Cargo.toml
oddgrd: We can try it out :+1:

oddgrd: We need the types from these deps for the shuttle_main codegen, so I don't think we can remove them. :thinking:

oddgrd: But we can make opentelemetry_proto a dev dependency, it's only used in the tests.

common/src/tracing.rs
```diff
@@ -1,6 +1,12 @@
 use serde_json::json;
 use tracing::field::Visit;
 
+pub const MESSAGE_KEY: &str = "message";
+pub const FILEPATH_KEY: &str = "code.filepath";
+pub const LINENO_KEY: &str = "code.lineno";
+pub const NAMESPACE_KEY: &str = "code.namespace";
+pub const TARGET_KEY: &str = "target";
```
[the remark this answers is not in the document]
oddgrd: We do still use them in the the logger integration test, as well as the message key in the logger span filter. But we might not need them as consts in common anymore. :thinking:

common/src/log.rs
```diff
@@ -122,6 +122,41 @@ impl From<&tracing::Level> for Level {
     }
 }
 
+#[cfg(feature = "backend")]
+impl From<opentelemetry_proto::tonic::logs::v1::SeverityNumber> for Level {
```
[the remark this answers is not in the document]
oddgrd: It isn't, the severity number was used for the LogRecords.

common/src/backends/tracing.rs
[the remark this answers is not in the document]
oddgrd: Removed now.

logger/src/dal.rs
oddgrd: Clarified the comment now.

logger/tests/integration_tests.rs
oddgrd: I refactored this into one function for runtime and one for services, they are both almost the same so they just call the old function with the runtime arg. Perhaps you had a better idea?

logger/tests/integration_tests.rs
chesedo: Okay, I added it to the todo list :+1:

logger/tests/integration_tests.rs
chesedo: About, I was thinking this function could take in a resources vec which those two functions can set rather than the `runtime` arg.

oddgrd: Ah! I changed it to that now, and also added comments to the runtime/service functions.

chesedo (approved):
LGTM :partying_face: :confetti_ball:

PR #1141, opened 2023-08-08
fix: project date

GugaGongadze:
## Description of change
This PR works with https://github.com/shuttle-hq/console-ui/pull/44
Appends project ID to the response payload so that its creation date can be deduced.

Yes, tested in combination with Console UI

### DIFF
--- a/gateway/src/api/latest.rs
+++ b/gateway/src/api/latest.rs
@@ -113,12 +113,11 @@ async fn get_project(
     State(RouterState { service, .. }): State<RouterState>,
     ScopedUser { scope, .. }: ScopedUser,
 ) -> Result<AxumJson<project::Response>, Error> {
-    let id = service.find_project_id(&scope).await?;
-    let uppercase_id = id.map(|s| s.to_uppercase());
+    let id = service.find_project_id(&scope).await?.to_uppercase();
 
     let state = service.find_project(&scope).await?.into();
     let response = project::Response {
-        id: uppercase_id,
+        id,
         name: scope.to_string(),
         state,
     };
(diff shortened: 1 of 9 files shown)


iulianbarbu (commented):
Just a few improvements, otherwise looks good  👍

iulianbarbu (commented):
LGTM! We'll only need to use `get_project` instead of `find_project_id` to squeeze two db calls into one.

gateway/src/api/latest.rs
```diff
@@ -145,8 +146,9 @@ async fn get_projects_list(
         .iter_user_projects_detailed(&name, limit * page, limit)
         .await?
         .map(|project| project::Response {
-            name: project.0.to_string(),
-            state: project.1.into(),
+            id: project.0.to_string().to_uppercase(),
+            name: project.1.to_string(),
+            state: project.2.into(),
```
iulianbarbu, line 151: @GugaGongadze can we actually create structs that wrap around the info that gets sent around as a tuple? I am finding it difficult to read it based on tuple indexing. The structs can be defined in `service.rs` I think, but LMK what you think. This isn't necessarily blocking this PR either.

I see some tests are failing too. Can you look into them?

oddgrd (approved):
Looks great, thank you! :partying_face:

gateway/src/api/latest.rs
iulianbarbu: nit: I am fine with leaving this as is for this PR, but something similar to `FindProjectPayload` would be needed here as well.

Resolving this.

iulianbarbu (approved):
LGTM! 👍

PR #1145, opened 2023-08-10
Create the local setup for replacing shuttle-logger sqlite with postgres

iulianbarbu:
## Description of change

Added a dedicated `logger-postgres` service that would be started only by `make up` and stopped by `make down`. This service is not supposed to run in staging or production.

`make up` and `make down`, and additionally, by starting only the `logger` with: `LOGGER_POSTGRES_HOST=logger-postgres LOGGER_POSTGRES_PASSWORD=postgres docker-compose -f docker-compose.rendered.yml up logger`.


docker-compose.dev.yml
```diff
@@ -38,3 +42,32 @@ services:
...
+        do
+          >&2 echo "PG is not available yet - sleeping"
+          sleep 1
+        done
+
+        >&2 echo "DBs are available - starting shuttle-logger"
+
+        exec /usr/local/bin/service "$${@:0}"
+  logger-postgres:
+    image: "${CONTAINER_REGISTRY}/logger-postgres:${LOGGER_POSTGRES_TAG}"
```
chesedo, line 63: Why can't we just use the stock postgres image?

extras/logger-postgres/Containerfile
```diff
@@ -0,0 +1,15 @@
+ARG LOGGER_POSTGRES_TAG=
+
+FROM docker.io/postgres:${LOGGER_POSTGRES_TAG}
+
+RUN apt update && apt install -y curl python3 python3-aiohttp
+
+COPY watch /usr/sbin/watch
```
chesedo, line 7: I don't think we need the watches and all for local development

logger/src/args.rs
```diff
@@ -9,9 +9,9 @@ pub struct Args {
     #[arg(long, default_value = "127.0.0.1:8000")]
     pub address: SocketAddr,
 
-    /// Where to store resources state
+    /// Remote database connection url
     #[arg(long, default_value = "./")]
-    pub state: PathBuf,
+    pub db_connection_uri: Uri,
```
chesedo, line 14: Maybe we should remove the default value?

logger/tests/helpers.rs
```diff
@@ -0,0 +1,116 @@
...
+        args.extend(env.iter().flat_map(|e| ["-e", e]));
+
+        args.push(image);
+
+        Command::new("docker").args(args).spawn().unwrap();
+
+        Self::wait_ready(Duration::from_secs(120), &is_ready_cmd);
+
+        // TODO: find out why this is needed for tests to pass reliably.
+        sleep(Duration::from_millis(300));
```
chesedo, line 57: The postgres image enters the ready state and then reboots. So your first ready check would have passed. This sleep essentially gets it past this reboot.

You should be able to have a small delay after the first `wait_ready` and then have another `wait_ready` to have it work all the time.

chesedo (commented):
Looks good. Left a few comments for optimizations.

logger/tests/helpers.rs
oddgrd: Oh, I see, it was like that in the provisioner, but the comment said it was because of MongoDB. I will revert it here then. Thanks!

logger/src/args.rs
oddgrd: Agreed.

extras/logger-postgres/Containerfile
oddgrd: Cool, I'll remove it.

docker-compose.dev.yml
oddgrd: Not sure, I'll try to figure out why, and if I can't I'll remove it.

oddgrd: I think we can remove the logger-postgres from the makefile as well, right? And just have it in the docker-compose.dev.yml

extras/logger-postgres/Containerfile
oddgrd: I think we can remove the logger-postgres from the makefile as well, right? And just have it in the docker-compose.dev.yml

chesedo: Yes, we can remove it from the Makefile too :smile:

chesedo (approved):
LGTM :partying_face:

PR #1148, opened 2023-08-15
chore(container): use pre-installed cargo-chef image

orhun:
## Description of change

This PR updates the base image of Containerfile to `lukemathwalker/cargo-chef` to avoid installing `cargo-chef` via `cargo install`.

\-

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -3,8 +3,7 @@
 
 # Base image for builds and cache
 ARG RUSTUP_TOOLCHAIN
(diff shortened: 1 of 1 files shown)


jonaro00:
One downside here is that this image might get released a bit later than the Rust image on new versions of Rust, but since we are not yet pursuing to support the latest version(s) all the time, it should be fine.

oddgrd (approved):
Thanks @orhun, LGTM!

PR #1150, opened 2023-08-15
fix: secrets not updating

iulianbarbu:
## Description of change

Fixed the bug related to secrets not being updated after the first deployment.

By starting the deployer locally and querying for updated secrets on a second deployment.


oddgrd (approved):
Makes sense, thanks!

PR #1151, opened 2023-08-16
fix: static folder is not updated

iulianbarbu:
## Description of change

We want to copy the contents of the build storage for the static file to the runtime storage. That can happen during the `ResourceBuilder::build` call.

Locally, deploying an axum static-files project, and checking that at a second deployment the assets were updated.

### DIFF
--- a/resources/static-folder/src/lib.rs
+++ b/resources/static-folder/src/lib.rs
@@ -99,7 +99,7 @@ impl<'a> ResourceBuilder<PathBuf> for StaticFolder<'a> {
 
     async fn build(build_data: &Self::Output) -> Result<PathBuf, shuttle_service::Error> {
         let input_dir = &build_data.input;
-        let output_dir = build_data.output.join(&build_data.assets.clone());
+        let output_dir = build_data.output.join(&build_data.assets);
 
         if &output_dir == input_dir {
             return Ok(output_dir);
--- a/resources/static-folder/src/lib.rs
+++ b/resources/static-folder/src/lib.rs
@@ -34,7 +34,7 @@ impl<'a> ResourceBuilder<PathBuf> for StaticFolder<'a> {
 
     type Config = &'a str;
 
-    type Output = PathBuf;
+    type Output = (PathBuf, PathBuf, PathBuf);
 
     fn new() -> Self {
         Self { folder: "static" }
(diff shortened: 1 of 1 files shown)


resources/static-folder/src/lib.rs
```diff
@@ -80,13 +90,25 @@ impl<'a> ResourceBuilder<PathBuf> for StaticFolder<'a> {
...
+        Ok(Paths {
+            input: input_dir,
+            output: output_dir,
+            assets: folder.to_path_buf(),
+        })
+    }
+
+    async fn build(build_data: &Self::Output) -> Result<PathBuf, shuttle_service::Error> {
+        let input_dir = &build_data.input;
+        let output_dir = build_data.output.join(&build_data.assets.clone());
```
orhun, line 102:
```suggestion
        let output_dir = build_data.output.join(&build_data.assets);
```

chesedo (approved):
LGTM

resources/static-folder/src/lib.rs
iulianbarbu: Done, thanks!

resources/static-folder/src/lib.rs
```diff
@@ -20,6 +20,16 @@ pub enum Error {
     Copy(fs_extra::error::Error),
 }
 
+#[derive(Serialize, Deserialize)]
+pub struct Paths {
+    // Build storage, where service files including assets are downloaded into
+    input: PathBuf,
+    // The service storage, used at runtime
+    output: PathBuf,
```
robertohuertasm, line 28: It would be nice to make the fields public. This would make it easy to extend or use the static folder. I was relying on the output being accessible to make [this library](https://github.com/robertohuertasm/shuttle-env-vars/blob/master/src/lib.rs#L98) work. Now it's not possible to construct the `Paths` struct.

iulianbarbu: Ah, I see. I wasn't aware about this use-case. Does this mean 0.24.0 is breaking your app, since it relies on the library?

robertohuertasm: It works, so no problem for now. Anyway, I still think it would be nice to change the api of this library to make it easier extending it.

iulianbarbu: Yes, we'll probably want to expose just the output dir for now.

robertohuertasm: The problem of only making one field public is that the struct cannot be created outside the original crate. I'm not suggesting to drastically change the API if it's not needed, but I'm wondering what would be the problem of exposing everything, given that it's not possible to access to that resource from the public API anyway.

iulianbarbu: That's a good point, I am not sure if exposing everything will cause problems but we'll have to double-check.

PR #1164, opened 2023-08-20
refactor: Containerfile+Makefile improvement: build crates together, then distribute the binaries

jonaro00:

`USE_PANAMAX=disable make images` with debug mode now takes less than 10min from scratch (used to take >20min). Can still be improved a lot by improvements to cargo-chef.


Makefile
```diff
@@ -112,13 +118,15 @@ DOCKER_COMPOSE_ENV=\
...
+.PHONY: images clean src up down deploy shuttle-% shuttle-images postgres docker-compose.rendered.yml test bump-% deploy-examples publish publish-% --validate-version
 
 clean:
 	rm .shuttle-*
 	rm docker-compose.rendered.yml
 
-images: shuttle-provisioner shuttle-deployer shuttle-gateway shuttle-auth shuttle-resource-recorder postgres panamax otel
+images: shuttle-images postgres panamax otel
+
+shuttle-images: shuttle-auth shuttle-deployer shuttle-gateway shuttle-provisioner shuttle-resource-recorder
```
chesedo, line 129: I like this separation :star_struck:

auth/prepare.sh
```diff
@@ -1,13 +1,8 @@
-#!/usr/bin/env sh
```
chesedo, line 1: You'll want this back as `/bin/bash` does not work on all systems like my nix does not have bash at `/bin`. `/usr/bin/env` is more resilient.... although I don't really see anyone running these outside Docker :man_shrugging:

Containerfile
```diff
@@ -3,57 +3,70 @@
...
 ARG RUSTUP_TOOLCHAIN
 FROM docker.io/library/rust:${RUSTUP_TOOLCHAIN}-buster as shuttle-crate
+ARG CARGO_PROFILE
 ARG folder
+ARG crate
 ARG prepare_args
 # used as env variable in prepare script
 ARG PROD
-ARG RUSTUP_TOOLCHAIN
-ENV RUSTUP_TOOLCHAIN=${RUSTUP_TOOLCHAIN}
```
oddgrd, line 47: Did you move this or remove it? This is actually a fix for a strange bug we used to have, where users projects with dependencies that had toolchain overrides cause the build to fail with a version mismatch. It's been a while though, and I'm not sure we still have this issue now that we don't use the cargo library. We used to be able to reliably reproduce it by deploying a serenity project with the cache feature enabled (which pulled in dashmap).

chesedo (approved):
Looks good for me. LGTM

auth/prepare.sh
jonaro00: Yeah, they should only run in docker.... :D I know env is good but took a shortcut here 😶

Containerfile
```diff
@@ -3,57 +3,92 @@
...
+FROM shuttle-auth AS shuttle-auth-dev
+
+FROM shuttle-crate-base AS shuttle-deployer
+    ARG CARGO_PROFILE
+    COPY --from=builder /build/target/${CARGO_PROFILE}/shuttle-next /usr/local/cargo/bin/
+    COPY deployer/prepare.sh /prepare.sh
+    RUN /prepare.sh "${prepare_args}"
+FROM shuttle-deployer AS shuttle-deployer-dev
+    # Source code needed for compiling with [patch.crates-io]
+    COPY --from=planner /build /usr/src/shuttle/
```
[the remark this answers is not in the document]
jonaro00: Thought it looked better :)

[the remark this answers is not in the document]
orhun: My formatter removes the indentation :/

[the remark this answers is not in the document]
jonaro00: Alright, I'll flatten it then

Containerfile
```diff
@@ -50,23 +50,45 @@ RUN cargo build \
...
 ARG RUSTUP_TOOLCHAIN
-FROM docker.io/library/rust:${RUSTUP_TOOLCHAIN}-buster as shuttle-crate
+FROM docker.io/library/rust:${RUSTUP_TOOLCHAIN}-buster as shuttle-crate-base
 ARG CARGO_PROFILE
 ARG folder
 ARG crate
 ARG prepare_args
-# used as env variable in prepare script
+# Fixes some dependencies compiled with incompatible versions of rustc
+ENV RUSTUP_TOOLCHAIN=${RUSTUP_TOOLCHAIN}
```
oddgrd, line 61: We need this to be set in the deployer containers, and I don't think this will do that (at least it doesn't look like it on staging currently).

iulianbarbu (approved):
lgtm!

Containerfile
jonaro00: Ah, I missed that the ARG before FROM is cleared after the FROM.

orhun (approved):

PR #1166, opened 2023-08-21
logger: store span names 

iulianbarbu:
## Description of change

Previously we dropped on any span name information, but we'll need them to be shown in `cargo-shuttle`.

Through integration tests. Need to follow up with more local/unstable testing soon.

### DIFF
--- a/logger/src/dal.rs
+++ b/logger/src/dal.rs
@@ -220,9 +220,10 @@ impl Log {
         deployment_id: Option<String>,
     ) -> Option<Vec<Self>> {
         // If we didn't find the id in the resource span, check the inner spans.
-        let deployment_id = deployment_id.or(get_attribute(span.attributes, "deployment_id"))?;
+        let deployment_id =
+            deployment_id.or(get_attribute(span.attributes.clone(), "deployment_id"))?;
 
-        let logs = span
+        let mut logs: Vec<Self> = span
             .events
             .into_iter()
             .flat_map(|event| {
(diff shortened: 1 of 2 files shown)


logger/src/dal.rs
```diff
@@ -254,6 +256,30 @@ impl Log {
...
+                    (span.start_time_unix_nano / 1_000_000_000)
+                        .try_into()
+                        .unwrap_or_default(),
+                    (span.start_time_unix_nano % 1_000_000_000) as u32,
+                )
+                .unwrap_or_default(),
+                Utc,
+            ),
+            // Span level doesn't exist so this info is not relevant.
+            level: LogLevel::Info,
```
chesedo, line 279: Wait, doesn't `tracing` allow setting the span level?

chesedo (commented):
Just one comment about the level. Else seems good :partying_face:

logger/src/dal.rs
iulianbarbu: Sadly, I couldn't find anything. Also this: https://github.com/open-telemetry/opentelemetry-specification/issues/3205

oddgrd: Oh, interesting... A `tracing` span does have a verbosity level (it's [required](https://docs.rs/tracing/latest/tracing/struct.Metadata.html)), but from your link Iulian it seems it's not included in the exported trace since it's not a part of the otel spec. Perhaps we can work around it somehow by adding the span levels to the span attributes? :thinking: (perhaps this should also be explored in a follow-up PR)

iulianbarbu: Yeh, I will try this tomorrow. 👌

chesedo (approved):
LGTM

orhun (approved):

PR #1167, opened 2023-08-21
ci: refactor and improve speed and caching. add rustls flags to aws-rds

jonaro00:
## Description of change

Replaces #1017. 

- CI improvements. Optimized caches and core count to both speed up and reduce costs.
- added rustls feature flags to aws-rds
- cargo update

Lots of testing done: https://app.circleci.com/pipelines/github/shuttle-hq/shuttle?branch=pull%2F1167

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,9 +11,9 @@ executors:
     docker:
       - image: cimg/rust:1.70.0
     resource_class: small
-  image-ubuntu:
+  machine-ubuntu:
     machine:
-      image: ubuntu-2204:2022.04.1
+      image: ubuntu-2204:2023.07.2
       docker_layer_caching: true
     resource_class: xlarge
 
@@ -89,8 +89,11 @@ commands:
           command: |
             cat\<< EOF > ~/.cargo/config.toml
             [patch.crates-io]
-            shuttle-service = { path = "$PWD/service" }
+            shuttle-codegen = { path = "$PWD/codegen" }
+            shuttle-common = { path = "$PWD/common" }
+            shuttle-proto = { path = "$PWD/proto" }
             shuttle-runtime = { path = "$PWD/runtime" }
+            shuttle-service = { path = "$PWD/service" }
 
             shuttle-aws-rds = { path = "$PWD/resources/aws-rds" }
             shuttle-persist = { path = "$PWD/resources/persist" }
@@ -121,19 +124,16 @@ commands:
             curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.70.0 --target add wasm32-wasi
             sudo apt update && sudo apt install -y libssl1.1
   install-protoc:
-    parameters:
-      arch:
-        description: "The architecture of protoc to install"
-        type: string
-        default: "linux-x86_64"
     steps:
       - run:
           name: Install protoc
+          environment:
+            ARCH: "linux-x86_64"
+            VERSION: "22.2"
           command: |
-            curl -OL https://github.com/protocolbuffers/protobuf/releases/download/v21.9/protoc-21.9-<< parameters.arch >>.zip &&\
-              sudo unzip -o protoc-21.9-<< parameters.arch >>.zip -d /usr/local bin/protoc &&\
-              sudo unzip -o protoc-21.9-<< parameters.arch >>.zip -d /usr/local 'include/*' &&\
-              rm -f protoc-21.9-<< parameters.arch >>.zip
+            curl -OL "https://github.com/protocolbuffers/protobuf/releases/download/v$VERSION/protoc-$VERSION-$ARCH.zip" && \
+                unzip -o "protoc-$VERSION-$ARCH.zip" bin/protoc "include/*" -d /usr/local && \
+                rm -f "protoc-$VERSION-$ARCH.zip"
   make-artifact:
     parameters:
       target:
@@ -170,6 +170,24 @@ commands:
           paths:
             - << parameters.target >>/*
             - << parameters.target >>.env
+  do-integration:
+    steps:
+      - install-rust
+      - checkout
+      - run: git submodule sync
+      - run: git submodule update --init
+      - restore-cargo-cache
+      - apply-patches
+      - run:
+          name: Run unit tests
+          command: cargo test --package << parameters.crate >> --all-features --lib -- --nocapture
+      - run:
+          name: Run integration tests
+          # Only run integration tests if there are any
+          command: |
+            set +o pipefail
+            (cargo test --package << parameters.crate >> --all-features --test '*' -- --list 2>&1 | grep -q "no test target matches pattern") && echo "nothing to test" || cargo test --package << parameters.crate >> --all-features --test '*' -- --nocapture
+      - save-cargo-cache
 
 jobs:
   workspace:
@@ -182,7 +200,7 @@ jobs:
       # the Cargo.lock.
       - run:
           command: |
-            [[ -z $(cat Cargo.lock | grep "\[\[patch.unused\]\]") ]]
+            [ -z $(grep "\[\[patch.unused\]\]" Cargo.lock) ]
       - run: cargo fmt --all --check
       # TODO: this is incompatible with workspace inheritance, uncomment when
       # https://github.com/DevinR528/cargo-sort/pull/29 is merged
@@ -197,7 +215,7 @@ jobs:
                        -A clippy::let-unit-value \
                        -A clippy::format-push-string
       - save-cargo-cache
-  check-standalone:
+  test-standalone:
     parameters:
       path:
         description: "Path to crate external from workspace"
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -22,41 +22,46 @@ executors:
...
   restore-cargo-cache:
     steps:
-      # Restore cargo cache before installing anything with the cargo command (ie cargo install ...)
       - restore_cache:
-          name: Restore cargo cache
-          keys:
-            - cargo-{{ checksum "Cargo.lock" }}
+          name: Restore cargo registry cache
+          key: cargo-{{ checksum "Cargo.lock" }}
+          # Don't use fallback key to prevent registry cache from growing indefinitely.
```
iulianbarbu, line 28: What does this mean?

.circleci/config.yml
```diff
@@ -22,41 +22,46 @@ executors:
...
+                > ~/.cargo/bin/sccache \
+              && chmod +x ~/.cargo/bin/sccache
+            echo 'export RUSTC_WRAPPER=~/.cargo/bin/sccache' >> $BASH_ENV
+            echo 'export SCCACHE_CACHE_SIZE=500M' >> $BASH_ENV
       - restore_cache:
           name: Restore sccache cache
-          key: sccache-cache-{{ .Environment.CIRCLE_JOB }}
+          keys: # This can use fallback due to size limit.
+            - sccache-cache-{{ .Environment.CIRCLE_JOB }}-{{ .Branch }}-
+            - sccache-cache-{{ .Environment.CIRCLE_JOB }}-
```
iulianbarbu, line 46: Why are the `-` needed at the end?

.circleci/config.yml
```diff
@@ -22,41 +22,46 @@ executors:
...
+              || curl -L https://github.com/mozilla/sccache/releases/download/$SCCACHE_VERSION/sccache-$SCCACHE_VERSION-x86_64-unknown-linux-musl.tar.gz \
+                | tar -xOz sccache-$SCCACHE_VERSION-x86_64-unknown-linux-musl/sccache \
+                > ~/.cargo/bin/sccache \
+              && chmod +x ~/.cargo/bin/sccache
+            echo 'export RUSTC_WRAPPER=~/.cargo/bin/sccache' >> $BASH_ENV
+            echo 'export SCCACHE_CACHE_SIZE=500M' >> $BASH_ENV
       - restore_cache:
           name: Restore sccache cache
-          key: sccache-cache-{{ .Environment.CIRCLE_JOB }}
+          keys: # This can use fallback due to size limit.
```
iulianbarbu, line 44: Can you explain how this works?

.circleci/config.yml
```diff
@@ -180,24 +184,21 @@ jobs:
...
+      # - run: cargo binstall cargo-sort
       # - run: cargo sort --check --workspace
       - run: |
           cargo clippy --tests \
                        --all-targets \
                        --all-features \
                        --no-deps -- \
-                       --D warnings \
-                       -A clippy::let-unit-value \
-                       -A clippy::format-push-string
```
iulianbarbu, line 198: Why have we removed these?

jonaro00: For parity with the expectations of the dev who runs plain `cargo clippy`. If we want custom rules, it should be done with config files or a task runner like cargo xtask like we mentioned at some point.

.circleci/config.yml
```diff
@@ -118,22 +124,20 @@ commands:
       - run:
           name: Install Rust
           command: |
-            curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.70.0 --target add wasm32-wasi
-            sudo apt update && sudo apt install -y libssl1.1
+            which cargo || curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.70.0
+            # sudo apt update && sudo apt install -y libssl1.1
```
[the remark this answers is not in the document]
jonaro00: It failed for some reason and I tried without it to see why it's needed. Do you know what it's for?

resources/aws-rds/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![doc = include_str!("../README.md")]
+#![allow(unused_imports)]
```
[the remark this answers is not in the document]
jonaro00: The way the crate uses macro and features causes unused imports to be present. Not sure why CI didn't warn about it yet.

.circleci/config.yml
jonaro00: Clarifying why we don't use fallback key there. It can theoretically cause a cache to grow more and more over time since there is no function to prune cargo's registry.

.circleci/config.yml
jonaro00: It will look for a cache with longest matching prefix as it is set up now. They are tagged with Job name and Branch name to carry over to the correct job in the next ci run.

.circleci/config.yml
jonaro00: Doesn't matter if they are there or not. But I think it shows more clearly that they are trying to match a prefix.

resources/aws-rds/src/lib.rs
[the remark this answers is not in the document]
iulianbarbu: Maybe we can address this if we're here. We can import the types by their full path: e.g. `serde::Serialize` instead of `Serialize`, and the same for the rest.

.circleci/config.yml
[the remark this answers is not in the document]
iulianbarbu: I suspect this has to do with cargo installation, but not sure. If `cargo` installs fine without `libssl1.1` then we can remove it.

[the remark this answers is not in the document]
jonaro00: Well, we are running rustup on the line before that so I see no issue.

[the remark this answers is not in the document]
jonaro00: Or you mean with `cargo install`? We should refrain from doing that in CI. Only download binaries.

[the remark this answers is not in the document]
iulianbarbu: Oh, no, I meant remove the `libssl1.1` install if `cargo` installation on the line above works fine without it.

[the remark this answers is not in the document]
jonaro00: Why wouldn't it? 😂 👍

iulianbarbu (approved):
lgtm!

.circleci/config.yml
```diff
@@ -180,24 +184,21 @@ jobs:
...
-          command: |
-            [[ -z $(cat Cargo.lock | grep "\[\[patch.unused\]\]") ]]
+      - run: '[ -z $(grep "\[\[patch.unused\]\]" Cargo.lock) ]'
       - run: cargo fmt --all --check
       # TODO: this is incompatible with workspace inheritance, uncomment when
       # https://github.com/DevinR528/cargo-sort/pull/29 is merged
-      # - run: cargo install cargo-sort
+      # - run: # install binstall with script
+      # - run: cargo binstall cargo-sort
       # - run: cargo sort --check --workspace
```
oddgrd, line 193: Should we just remove these?

.circleci/config.yml
```diff
@@ -635,40 +680,69 @@ workflows:
...
+      #       parameters:
+      #         path:
+      #           - resources/aws-rds
+      #         features:
+      #           - "-F postgres"
+      #           - "-F postgres-rustls"
+      #           - "-F mysql"
+      #           - "-F mysql-rustls"
+      #           - "-F mariadb"
+      #           - "-F mariadb-rustls"
```
oddgrd, line 710: Should we remove this then?

.circleci/config.yml
```diff
@@ -635,40 +680,69 @@ workflows:
...
+                - cargo-shuttle
+                - shuttle-provisioner
       - e2e-test:
           requires:
-            - platform-test
-            - check-standalone
-            - check-standalone-shared-db
+            - test-standalone
+            - test-standalone-shared-db
+            # - test-standalone-aws-rds
```
oddgrd, line 743:
```suggestion
```

resources/aws-rds/README.md
```diff
@@ -16,6 +16,8 @@ Add `shuttle-aws-rds` to the dependencies for your service. Every engine is behi
 | MySql    | mysql        | shuttle_aws_rds::MySql    |
 | MariaDB  | mariadb      | shuttle_aws_rds::MariaDB  |
 
+Note: You can also add `-rustls` after the feature flag, `postgres-rustls` etc
```
oddgrd, line 19: Perhaps we should expand on this a bit and mention that it will use sqlx with native-tls by default, you can add rustls if you'd rather use that.

oddgrd (approved):
Thanks! Just left some suggestions for clean up, nothing major.

.circleci/config.yml
jonaro00: I'm intending to have that comment as a "how it would be done if it was needed".
Or is tests in that crate irrelevant and this comment too bloated?

oddgrd: I think looking at the test-standalone for shared db makes it easy to understand how to test rds as well if we add tests in the future, so I don't think we need this.

PR #1168, opened 2023-08-22
test(logger): run integration tests serially

orhun:
Utilize `serial_test` crate to run the logger tests serially. Running them in parallel or with `--test-threads 1` causes spurious errors.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5287,6 +5287,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "serial_test"
+version = "2.0.0"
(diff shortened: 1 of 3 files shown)


orhun:
Yeah, it is still not fixed somehow. I will mark this PR as draft.

oddgrd:
> Yeah, it is still not fixed somehow. I will mark this PR as draft.

Since this one seems like quite a conundrum, we might leave it as it was for now as well, and focus on something else. I am very curious about why this is failing, though. :smile:

We had some thoughts on potential changes to the logger that would also affect these tests, we can talk about that in the daily tomorrow.

PR #1172, opened 2023-08-23
logger: adjust logger to receive logs blobs

iulianbarbu:
## Description of change

Enabled storing data for logs blobs.

Locally, through integration testing.


codegen/src/shuttle_main/mod.rs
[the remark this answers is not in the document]
oddgrd: Oh, and the tests here need to be updated.

logger/tests/integration_tests.rs
```diff
@@ -323,64 +269,47 @@ async fn generate_and_stream_logs() {
...
 /// instrumented with the deployment_id field, this way we can choose which spans we want
 /// to associate with a deployment and record in the logger.
-fn generate_service_logs(port: u16, deployment_id: String, generator: fn(String)) {
-    generate_logs(
-        port,
-        deployment_id,
-        generator,
-        vec![KeyValue::new("service.name", "test")],
-    );
+fn generate_service_logs(
```
oddgrd, line 280: It looks like this and generate_runtime_logs are duplicated now

logger/src/lib.rs
```diff
@@ -95,6 +56,46 @@ impl<D> Logger for Service<D>
 where
     D: Dal + Send + Sync + 'static,
 {
+    async fn store_logs(
+        &self,
+        request: Request<StoreLogsRequest>,
+    ) -> Result<Response<StoreLogsResponse>, Status> {
+        request.verify(Scope::DeploymentPush)?;
```
chesedo, line 63: I don't think we have to check the claim - nor the runtime or our own services will have one.

oddgrd (commented):
I love all these red lines!

logger/tests/integration_tests.rs
```diff
@@ -323,64 +269,47 @@ async fn generate_and_stream_logs() {
...
-        .set_default();
+    // Set up tracing subscriber connected to the logger server
+    let layer = tracing_subscriber::fmt::layer()
+        .with_writer(Mutex::new(LoggerLayer::new(
+            deployment_id.clone(),
+            SHUTTLE_SERVICE.into(),
+            client,
+        )))
+        .with_ansi(false);
+    let _guard = tracing_subscriber::registry().with(layer).set_default();
```
chesedo, line 312: I'm not sure setting up tracing for these tests still makes sense. We should now be okay will calling the `StoreLogs` method directly and can have fever tests

chesedo (commented):
Have a few comments

logger/src/lib.rs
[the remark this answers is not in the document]
iulianbarbu: Removed the claim.

iulianbarbu: Removed the claim.

logger/src/lib.rs
```diff
@@ -95,6 +56,46 @@ impl<D> Logger for Service<D>
...
+
+        let request = request.into_inner();
+        let logs = request.logs;
+        if !logs.is_empty() {
+            _ = self
+                .logs_tx
+                .broadcast(
+                    logs.into_iter()
+                        .map(|li| {
```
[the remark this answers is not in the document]
iulianbarbu: It is more for `log`, but I refactored this to use a `from_stored` method.

logger/src/lib.rs
```diff
@@ -95,6 +56,46 @@ impl<D> Logger for Service<D>
...
+                                tx_timestamp: DateTime::from_utc(
+                                    NaiveDateTime::from_timestamp_opt(
+                                        timestamp.seconds,
+                                        timestamp.nanos.try_into().unwrap_or_default(),
+                                    )
+                                    .unwrap_or_default(),
+                                    Utc,
+                                ),
+                                data: li.data,
+                            }
```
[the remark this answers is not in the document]
iulianbarbu: Done. I wouldn't implement `From` for this because we would have to do it in `proto` and expose the `shuttle_logger::dal::Log` there. I implemented though `Log::from_stored` method which is doing the same thing.

logger/tests/integration_tests.rs
iulianbarbu: Removed the tracing and tested just the RPCs.

logger/tests/integration_tests.rs
iulianbarbu: Yep, fixed this by removing the tracing layer, and testing just the RPCs.

logger/tests/integration_tests.rs
```diff
@@ -210,8 +166,8 @@ async fn generate_and_get_service_logs() {
     );
 
     // Generate some logs with a fn not instrumented with deployment_id.
-    generate_service_logs(port, DEPLOYMENT_ID.into(), deploy);
-
+    generate_service_logs(client.clone(), DEPLOYMENT_ID.into(), deploy);
+    tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
```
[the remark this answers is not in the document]
iulianbarbu: Shouldn't be the case anymore with the tests change.

runtime/Cargo.toml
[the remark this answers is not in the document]
chesedo: I think we can remove the `backend` feature from `shuttle-common` too

logger/tests/integration_tests.rs
```diff
@@ -1,465 +1,188 @@
...
 };
 use tokio::time::timeout;
 use tonic::{transport::Server, Request};
-use tracing::{debug, error, info, instrument, trace, warn};
-use tracing_subscriber::prelude::*;
+
+const SHUTTLE_SERVICE: &str = "test";
 
 // TODO: find out why these tests affect one-another. If running them together setting the timeouts
 // low will cause them to fail spuriously. If running single tests they always pass.
```
chesedo, line 22: Do they still affect each other?

logger/tests/integration_tests.rs
```diff
@@ -1,465 +1,188 @@
...
+    let sqlite = Sqlite::new_in_memory().await;
+    let sqlite_clone = sqlite.clone();
+    let server = tokio::task::spawn(async move {
         Server::builder()
-            .layer(JwtScopesLayer::new(vec![Scope::Logs]))
-            .add_service(TraceServiceServer::new(ShuttleLogsOtlp::new(
-                sqlite.get_sender(),
+            .layer(JwtScopesLayer::new(vec![
+                Scope::Logs,
+                Scope::DeploymentPush,
```
chesedo, line 36:
```suggestion
```

chesedo (commented):
Looking good. Left small comments

codegen/src/lib.rs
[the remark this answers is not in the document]
iulianbarbu: Lint issues (saying they are unused).

runtime/Cargo.toml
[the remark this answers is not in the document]
iulianbarbu: Done.

logger/tests/integration_tests.rs
orhun: That was my question. Will check now :D

iulianbarbu: Removed it in advance.

logger/tests/integration_tests.rs
iulianbarbu: Done.

logger/tests/integration_tests.rs
orhun: Nope, they seem to work fine now.

codegen/src/lib.rs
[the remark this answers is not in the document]
iulianbarbu: To be honest, they shouldn't pop up as a lint issue, possibly I just had an intermediate lint problem while working on the code. I will revert this.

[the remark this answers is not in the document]
iulianbarbu: Nope, still have the issue when running just `cargo test` in `codegen`. 🙈

logger/tests/integration_tests.rs
```diff
@@ -1,465 +1,180 @@
...
+    }
 }
 
 #[tokio::test]
-async fn generate_and_stream_logs() {
+async fn get_stream_logs() {
     let port = pick_unused_port().unwrap();
     let addr = SocketAddr::new(Ipv4Addr::LOCALHOST.into(), port);
-    const DEPLOYMENT_ID: &str = "stream-logs-deployment-id";
+    const DEPLOYMENT_ID: &str = "runtime-fetch-logs-deployment-id";
```
orhun, line 102: I would say let's make this a non-constant since having a constant in a scope does not make sense.

We can simply:

```rs
let deployment_id = "runtime-fetch-logs-deployment-id";
```

logger/tests/integration_tests.rs
```diff
@@ -1,465 +1,180 @@
...
-                sqlite.get_sender(),
+            .add_service(LoggerServer::new(Service::new(
+                sqlite_clone.get_sender(),
+                sqlite_clone,
             )))
-            .add_service(LoggerServer::new(Service::new(sqlite.get_sender(), sqlite)))
             .serve(addr)
             .await
             .unwrap()
     });
```
orhun, line 40: We can extract this to a function called `spawn_server` and avoid having code repetitions for tests.

logger/src/lib.rs
```diff
@@ -95,6 +56,25 @@ impl<D> Logger for Service<D>
...
+        let request = request.into_inner();
+        let logs = request.logs;
+        if !logs.is_empty() {
+            _ = self
+                .logs_tx
+                .broadcast(logs.into_iter().map(Log::from_stored).collect())
+                .await
+                .map_err(|err| {
+                    println!("failed to send log to storage: {}", err);
+                });
```
orhun, line 72: The error is emitted here, is that something we want? Maybe we can handle this gracefully and return it.

iulianbarbu: Done.

logger/tests/integration_tests.rs
iulianbarbu: Done.

orhun (approved):

codegen/src/shuttle_main/mod.rs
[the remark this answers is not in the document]
jonaro00: Do we want to add back a macro annotation for enabling a default tracing subscriber like the one we've been using thus far?

logger/migrations/0000_init.sql
```diff
@@ -1,7 +1,6 @@
 CREATE TABLE IF NOT EXISTS logs (
     deployment_id TEXT,        -- The deployment that this log line pertains to.
     shuttle_service_name TEXT, -- The shuttle service which created this log.
-    timestamp INTEGER,         -- Unix epoch timestamp.
-    level INTEGER,             -- The log level
-    fields TEXT                -- Log fields object.
+    tx_timestamp INTEGER,      -- Unix epoch timestamp.
```
jonaro00, line 4: This is millisecond accuracy right? Feels weird to go from nanosecond timestamps to millis. Micros could be of interest. (what if logs get out of order due to this?)

proto/logger.proto
```diff
@@ -4,32 +4,41 @@ package logger;
...
-enum LogLevel {
-  Trace = 0;
-  Debug = 1;
-  Info = 2;
-  Warn = 3;
-  Error = 4;
+message FetchedLogItem {
+  string service_name = 1;
+  google.protobuf.Timestamp tx_timestamp = 2;
+  bytes data = 3;
```
jonaro00, line 43: How bout this structure? (feel free to change the names)
```rs
message LogItem {
  string deployment_id = 1;
  LogLine log_line = 2;
}

message LogLine {
  string internal_origin = 1;
  google.protobuf.Timestamp timestamp = 2;
  string line = 3;
}

```

logger/migrations/0000_init.sql
iulianbarbu: We'll have though to see how the proto timestamp gets serialized to see if it can fit `INTEGER` type here, which can be at maximum 8 bytes, but this is also signed. This is a good catch!

proto/logger.proto
iulianbarbu: Done 👍

iulianbarbu: Upsy, I have some duplicates. Will fix that.

PR #1173, opened 2023-08-24
feat(deployer): send runtime logs to the logger service

orhun:
Simply capture logs and send them to the logger service.

yolo.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -8,7 +8,7 @@ mod provisioner_server;
 use std::collections::{BTreeMap, HashMap};
 use std::ffi::OsString;
 use std::fs::{read_to_string, File};
-use std::io::{stdout, Write};
+use std::io::stdout;
 use std::net::{Ipv4Addr, SocketAddr};
 use std::path::{Path, PathBuf};
 use std::process::exit;
@@ -665,8 +665,7 @@ impl Shuttle {
         let mut reader = BufReader::new(child_stdout).lines();
         tokio::spawn(async move {
             while let Some(line) = reader.next_line().await.unwrap() {
-                let mut stdout = stdout().lock();
-                writeln!(stdout, "{}", line).unwrap();
+                println!("{}", line);
             }
         });
 
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -641,6 +641,7 @@ impl Shuttle {
         };
 
         // Child process and gRPC client for sending requests to it
+        // TODO: Is child stdout should be handled?
         let (mut runtime, mut runtime_client) = runtime::start(
             is_wasm,
             StorageManagerType::WorkingDir(working_directory.to_path_buf()),
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -8,7 +8,7 @@ mod provisioner_server;
 use std::collections::{BTreeMap, HashMap};
 use std::ffi::OsString;
 use std::fs::{read_to_string, File};
-use std::io::stdout;
+use std::io::{stdout, Write};
 use std::net::{Ipv4Addr, SocketAddr};
 use std::path::{Path, PathBuf};
 use std::process::exit;
@@ -46,9 +46,10 @@ use ignore::overrides::OverrideBuilder;
 use ignore::WalkBuilder;
 use indicatif::ProgressBar;
 use indoc::printdoc;
-use std::fmt::Write;
+use std::fmt::Write as FmtWrite;
 use strum::IntoEnumIterator;
 use tar::Builder;
+use tokio::io::{AsyncBufReadExt, BufReader};
 use tokio::process::Child;
 use tokio::task::JoinHandle;
 use tonic::transport::Channel;
@@ -641,7 +642,6 @@ impl Shuttle {
         };
 
         // Child process and gRPC client for sending requests to it
-        // TODO: Is child stdout should be handled?
         let (mut runtime, mut runtime_client) = runtime::start(
             is_wasm,
             StorageManagerType::WorkingDir(working_directory.to_path_buf()),
@@ -657,6 +657,18 @@ impl Shuttle {
             logger_server.abort();
             err
         })?;
+
+        let child_stdout = runtime
+            .stdout
+            .take()
+            .context("child process did not have a handle to stdout")?;
+        let mut reader = BufReader::new(child_stdout).lines();
+        tokio::spawn(async move {
+            while let Some(line) = reader.next_line().await.unwrap() {
+                let mut stdout = stdout().lock();
+                writeln!(stdout, "{}", line).unwrap();
+            }
+        });
 
         let service_name = service.service_name()?;
         let load_request = tonic::Request::new(LoadRequest {
--- a/deployer/src/deployment/deploy_layer.rs
+++ b/deployer/src/deployment/deploy_layer.rs
@@ -460,10 +460,7 @@ mod tests {
             .await
             .expect("failed to connect to logger");
 
-        let channel = ServiceBuilder::new()
-            .layer(ClaimLayer)
-            .layer(InjectPropagationLayer)
-            .service(channel);
+        let channel = ServiceBuilder::new().service(channel);
 
         let logger_client = LoggerClient::new(channel);
 
--- a/deployer/src/deployment/deploy_layer.rs
+++ b/deployer/src/deployment/deploy_layer.rs
@@ -286,8 +286,9 @@ mod tests {
     use ctor::ctor;
     use flate2::{write::GzEncoder, Compression};
     use portpicker::pick_unused_port;
-    use shuttle_common::claims::Claim;
+    use shuttle_common::claims::{Claim, ClaimLayer, InjectPropagationLayer};
     use shuttle_proto::{
+        logger::logger_client::LoggerClient,
         provisioner::{
             provisioner_server::{Provisioner, ProvisionerServer},
             DatabaseDeletionResponse, DatabaseRequest, DatabaseResponse, Ping, Pong,
@@ -296,7 +297,8 @@ mod tests {
     };
     use tempfile::Builder;
     use tokio::{select, time::sleep};
-    use tonic::transport::Server;
+    use tonic::transport::{Endpoint, Server};
+    use tower::ServiceBuilder;
     use tracing_subscriber::{fmt, prelude::*, EnvFilter};
     use ulid::Ulid;
     use uuid::Uuid;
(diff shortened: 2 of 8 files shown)


deployer/src/runtime_manager.rs
```diff
@@ -15,7 +19,7 @@ type Runtimes = Arc<
         HashMap<
             Uuid,
             (
-                process::Child,
+                Arc<Mutex<process::Child>>,
```
iulianbarbu, line 22: Do we need the `Arc<Mutex<..` here? I think `Child` is `Send` and `Sync` already [1], [2].

[1] https://docs.rs/tokio/latest/tokio/process/struct.Child.html#impl-Send-for-Child
[2] https://docs.rs/tokio/latest/tokio/process/struct.Child.html#impl-Sync-for-Child

deployer/src/runtime_manager.rs
```diff
@@ -115,19 +131,56 @@ impl RuntimeManager {
...
+        // Ensure the child process is spawned in the runtime so it can
+        // make progress on its own while we await for any output.
+        tokio::spawn(async move {
+            process_cloned
+                .lock()
+                .await
+                .wait()
+                .await
+                .expect("child process encountered an error");
+        });
```
iulianbarbu, line 157: Just wondering, is it worth waiting in a task that isn't awaited?

deployer/src/runtime_manager.rs
```diff
@@ -115,19 +131,56 @@ impl RuntimeManager {
...
+
+        let mut reader = BufReader::new(stdout).lines();
+        let mut logger_client = self.logger_client.clone();
+        tokio::spawn(async move {
+            while let Some(line) = reader.next_line().await.unwrap() {
+                // TODO: `store_logs` accepts a Vec but logs are sent one by one. Is this feasible?
```
iulianbarbu, line 163: I think we can probably start by adding some default batching logic here. It should account lets say for batching 100 logs one time, but it should also account for some time window. E.g. whenever deciding if we want to store the logs, we would verify the batch size against the default value or check whether the window has passed.

cargo-shuttle/src/lib.rs
```diff
@@ -641,6 +641,7 @@ impl Shuttle {
         };
 
         // Child process and gRPC client for sending requests to it
+        // TODO: Is child stdout should be handled?
```
chesedo, line 644: This should just print to stdout

deployer/src/runtime_manager.rs
chesedo, line 163: Correct, we'll make another PR to bundle them somehow

chesedo (commented):
For the CLI, we just have to print the logs to stdout too :smile:

deployer/src/runtime_manager.rs
```diff
@@ -115,19 +131,56 @@ impl RuntimeManager {
...
+                .lock()
+                .await
+                .wait()
+                .await
+                .expect("child process encountered an error");
+        });
+
+        let mut reader = BufReader::new(stdout).lines();
+        let mut logger_client = self.logger_client.clone();
+        tokio::spawn(async move {
```
iulianbarbu, line 161: I am wondering if it makes sense to handle this logging related task in the outer scope. Probably it means propagating this up to `task` from `run.rs`.

deployer/src/runtime_manager.rs
```diff
@@ -115,19 +131,56 @@ impl RuntimeManager {
...
+            while let Some(line) = reader.next_line().await.unwrap() {
+                // TODO: `store_logs` accepts a Vec but logs are sent one by one. Is this feasible?
+                let logs = vec![LogItem {
+                    deployment_id: id.to_string(),
+                    log_line: Some(LogLine {
+                        service_name: service_name.to_string(),
+                        tx_timestamp: Some(Timestamp::from(SystemTime::UNIX_EPOCH)),
```
iulianbarbu, line 168: We'll need to use here something like `Utc::now()`.

deployer/src/runtime_manager.rs
orhun: Yeah, but that's not the reason why I used `Mutex`.

See: https://github.com/orhun/shuttle/blob/1370efaaaf68f774094c180842d69422f0202355/deployer/src/runtime_manager.rs#L134-L140

Here, `process` is being moved to `self.runtimes` and we need to use it mutably afterwards in another closure (`tokio::spawn`) so there isn't a way to do that without an async `Mutex` and atomic reference counter since it \*does not\* implement `Clone`.

deployer/src/runtime_manager.rs
orhun: From the docs:

> The provided future will start running in the background immediately when spawn is called, even if you don’t await the returned `JoinHandle`.

deployer/src/runtime_manager.rs
orhun: f647dab028b08aa76898aec9bb136ea6116003cb

deployer/src/runtime_manager.rs
chesedo: `run.rs` might not have access to the logger client though

iulianbarbu: I meant just propagating the task handle, no need for the client in that scope. In some cases logging sending might break so detecting clearly when that happens should help.

deployer/src/runtime_manager.rs
iulianbarbu: Would be great though to add something like this for nanos: https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.timestamp_subsec_nanos.

cargo-shuttle/src/lib.rs
orhun: I'm getting the following error for `cargo shuttle run`:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: runtime client endpoint did not open in time
```

Still investigating, do you have any ideas?

deployer/src/runtime_manager.rs
iulianbarbu: Agreed. Resolving this.

cargo-shuttle/src/lib.rs
orhun: Ah, I think it is because I'm not using `git` dependencies for my test project.

deployer/src/runtime_manager.rs
iulianbarbu: Is this task where the process is moved the one which calls `.wait()` on it. I don't find it necessarily useful, but lmk if you disagree.

cargo-shuttle/src/lib.rs
orhun: da38f862bcce4520f213673e51d2d451c5201f78

deployer/src/runtime_manager.rs
orhun: Yeah, I tested in da38f862bcce4520f213673e51d2d451c5201f78 and it worked fine without that.

Removed in 002d0c01a408f49bfd49bbe4a82f042bdb6b5847

deployer/src/runtime_manager.rs
orhun: Yup, 002d0c01a408f49bfd49bbe4a82f042bdb6b5847

cargo-shuttle/src/lib.rs
```diff
@@ -658,6 +658,18 @@ impl Shuttle {
...
 
+        let child_stdout = runtime
+            .stdout
+            .take()
+            .context("child process did not have a handle to stdout")?;
+        let mut reader = BufReader::new(child_stdout).lines();
+        tokio::spawn(async move {
+            while let Some(line) = reader.next_line().await.unwrap() {
+                let mut stdout = stdout().lock();
+                writeln!(stdout, "{}", line).unwrap();
```
iulianbarbu, line 669: Why not use `println!`?

deployer/src/runtime_manager.rs
orhun: f2d42cbda13a34558f238b8a6f97c769b51af2d0

cargo-shuttle/src/lib.rs
orhun: From `println` docs:

> The println! macro will lock the standard output on each call. If you call println! within a hot loop, this behavior may be the bottleneck of the loop. To avoid this, lock stdout with [io::stdout().lock()](https://doc.rust-lang.org/std/io/struct.Stdout.html)

https://doc.rust-lang.org/std/macro.println.html

Also, I believe using `stdout` makes the implementation more explicit.

iulianbarbu: Ah, interesting. However, wouldn't the locking happening at each call of `println!` be the same as locking on `stdout inside the loop every time before the `writeln!`? I think the idea from the doc is to `lock` outside the loop and use the guard to write continuously to the stdout - meaning lock once for the entirety of the loop instead of locking on each iteration.

iulianbarbu: >  I think the idea from the doc is to lock outside the loop and use the guard to write continuously to the stdout - meaning lock once for the entirety of the loop instead of locking on each iteration.

However, thinking better about this, we might need to write to the stdout of the cargo-shuttle process from different tasks too. I think this would block the other tasks.

jonaro00: The speed of this loop is not super important either.

iulianbarbu: @orhun what are your thoughts on this? Do you still find the `stdout` locking approach useful? We might be able to close this, I think this is the last unresolved comment.

orhun: I'm okay with either way. If you're fine with `stdout` approach, feel free to give it a final review and we can proceed!

iulianbarbu: Not sure what you mean by the final review (I did review the `stdout` approach), but I would remove the `stdout` and replace it with `println!` just because using it like you did they should be the same from my understanding. Correct me if I am wrong.

orhun: 14d7dfe27ae54d0d92a01d971bad8e884ff06dc5

deployer/src/deployment/deploy_layer.rs
```diff
@@ -449,10 +454,24 @@ mod tests {
...
 
+        let channel = Endpoint::try_from(logger_uri.to_string())
+            .unwrap()
+            .connect()
+            .await
+            .expect("failed to connect to logger");
+
+        let channel = ServiceBuilder::new()
+            .layer(ClaimLayer)
+            .layer(InjectPropagationLayer)
```
chesedo, line 465: Hmm, we should be fine without these layers in the tests.

deployer/src/deployment/run.rs
```diff
@@ -526,10 +538,24 @@ mod tests {
...
 
+        let channel = Endpoint::try_from(logger_uri.to_string())
+            .unwrap()
+            .connect()
+            .await
+            .expect("failed to connect to logger");
+
+        let channel = ServiceBuilder::new()
+            .layer(ClaimLayer)
+            .layer(InjectPropagationLayer)
```
chesedo, line 549: Here too

chesedo (commented):
Just simplification notes

deployer/src/deployment/deploy_layer.rs
orhun: e2edeac07a50b5166c7f2fb1c8332a65020bc2d4

deployer/src/deployment/run.rs
orhun: e2edeac07a50b5166c7f2fb1c8332a65020bc2d4

deployer/src/deployment/deploy_layer.rs
oddgrd: The runtime manager expects a client with those layers, so we'll need to keep them (or refactor the runtime manager to create the client instead). I think we discussed removing the claim layer too, right @chesedo?

chesedo: Actually the server wants those layers. Since we will mock / stub the server, we can control the layers needed in each test

chesedo (approved):
LGTM

PR #1176, opened 2023-08-28
chore: Rust 1.72.0

jonaro00:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
 members = [
   "admin",
   "auth",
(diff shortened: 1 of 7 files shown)


orhun (approved):
🦀

Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
```
oddgrd, line 2: Is this required? It should be [implied](https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html#details) with the 2021 edition. Or did you see failures without it?

PR #1185, opened 2023-08-29
refactor: remove tracing from runtime

chesedo:
## Description of change
Removes the tracing dependency from runtime since we can use stdout instead.

Using `cargo shuttle run` on a postgres project


oddgrd (approved):
LGTM! Perhaps we should do a PR for the workspace CI failures too?

orhun (approved):

PR #1188, opened 2023-08-29
feat: logs batching

chesedo:
## Description of change
Makes sure logs are sent in batches to avoid traffic overhead.

With new tests

### DIFF
--- a/proto/src/lib.rs
+++ b/proto/src/lib.rs
@@ -232,5 +232,114 @@ pub mod resource_recorder {
 }
 
 pub mod logger {
+    use std::sync::mpsc;
+
+    use tonic::async_trait;
+
     include!("generated/logger.rs");
+
+    /// Adapter to some client which expects to receive a vector of items
+    #[async_trait]
+    pub trait VecReceiver: Send {
+        type Item;
+
+        async fn receive(&mut self, items: Vec<Self::Item>);
+    }
+
+    /// Wrapper to batch together items before forwarding them to some vector receiver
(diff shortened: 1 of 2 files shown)


proto/src/lib.rs
```diff
@@ -232,5 +232,202 @@ pub mod resource_recorder {
...
+            match self
+                .store_logs(Request::new(StoreLogsRequest { logs: items }))
+                .await
+            {
+                Ok(_) => {}
+                Err(error) => error!(
+                    error = &error as &dyn std::error::Error,
+                    "failed to send batch logs to logger"
+                ),
+            }
```
orhun, line 279: We can use `if let Err(error) = ...` if we are not going to handle the `Ok(_)` case here.

proto/src/lib.rs
```diff
@@ -232,5 +232,202 @@ pub mod resource_recorder {
...
+        pub fn wrap(inner: I) -> Self {
+            Self::new(inner, 2048, Duration::from_secs(5))
+        }
+
+        /// Send a single item into this batcher
+        pub fn send(&self, item: I::Item) {
+            match self.tx.send(item) {
+                Ok(_) => {}
+                Err(_) => unreachable!("the receiver will never drop"),
+            }
```
orhun, line 316: Can be simplified to only handle the `Err(_)` case via `if let`.

chesedo: Nice, addressed :smile:

oddgrd (approved):

PR #1190, opened 2023-08-30
refactor(runtime): replace trace with println

orhun:
Fixes the following build failure:

```
error[E0432]: unresolved import `tracing`
  --> runtime/src/next/mod.rs:26:5
   |
26 | use tracing::{error, trace, warn};
   |     ^^^^^^^ help: a similar path exists: `shuttle_common::tracing`

```

### DIFF
--- a/runtime/src/next/mod.rs
+++ b/runtime/src/next/mod.rs
@@ -23,7 +23,6 @@ use shuttle_proto::runtime::{
 use tokio::sync::{broadcast, mpsc, oneshot};
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::Status;
-use tracing::{error, trace, warn};
 use wasi_common::file::FileCaps;
 use wasmtime::{Engine, Linker, Module, Store};
 use wasmtime_wasi::sync::net::UnixStream as WasiUnixStream;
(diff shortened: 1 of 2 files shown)


runtime/Cargo.toml
```diff
@@ -21,6 +21,7 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 strfmt = { workspace = true }
 thiserror = { workspace = true }
+tracing = { workspace = true }
```
chesedo, line 24: Since no tracing is setup in the runtimes, we should replace those log lines with `println`s instead

chesedo (commented):
We should use `println` instead

runtime/Cargo.toml
orhun: Done: 8647dd3a352747ce042cc75b0e9ac9a44e879364

chesedo (approved):
LGTM

runtime/src/next/mod.rs
```diff
@@ -69,7 +68,7 @@ impl Runtime for AxumWasm {
         request: tonic::Request<LoadRequest>,
     ) -> Result<tonic::Response<LoadResponse>, Status> {
         let wasm_path = request.into_inner().path;
-        trace!(wasm_path, "loading shuttle-next project");
+        println!("loading shuttle-next project");
```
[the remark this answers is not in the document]
orhun: 4a9eae582a0b4f4190f2e0448e97b524df929593

PR #1192, opened 2023-08-31
feat(gateway): special error if own project is already running

### LINKED ISSUE #1155: [Improvement]: Correct behavior when running `project start`
### Describe the improvement

Currently, when your own project is running, you get
```
$ cargo shuttle project start
Error: 400 Bad Request
message: a project with the same name already exists
```
which is correct, but it would be better if it behaved differently when you are the owner of that project.

### Suggested Solution:

Output a different error message if you are the owner, such as `project is already running`.

### Duplicate declaration

d4ckard:
## Description of change

The command `cargo shuttle project start` fails if a project with the same name is already present. The error message that's printed now tells the caller that their project is already running if they are the project owner.

Closes #1155

One assertion has been added to the relevant test for each new case. They make sure that the new error variant `ErrorKind::ProjectAlreadyRunning` is returned if either the owner or an admin tries to recreate a running project.

### DIFF
--- a/common/src/models/error.rs
+++ b/common/src/models/error.rs
@@ -98,9 +98,10 @@ impl From<ErrorKind> for ApiError {
                 StatusCode::BAD_REQUEST,
                 "a project with the same name already exists",
             ),
-            ErrorKind::ProjectAlreadyRunning => {
-                (StatusCode::BAD_REQUEST, "project is already running")
-            }
+            ErrorKind::ProjectAlreadyRunning => (
+                StatusCode::BAD_REQUEST,
+                "it looks like your project is already running. You can find out more with `cargo shuttle project status`",
+            ),
             ErrorKind::InvalidCustomDomain => (StatusCode::BAD_REQUEST, "invalid custom domain"),
             ErrorKind::CustomDomainNotFound => (StatusCode::NOT_FOUND, "custom domain not found"),
             ErrorKind::CustomDomainAlreadyExists => {
--- a/common/src/models/error.rs
+++ b/common/src/models/error.rs
@@ -42,6 +42,7 @@ pub enum ErrorKind {
     ProjectNotFound,
     InvalidProjectName,
     ProjectAlreadyExists,
+    ProjectAlreadyRunning,
     ProjectNotReady,
     ProjectUnavailable,
     CustomDomainNotFound,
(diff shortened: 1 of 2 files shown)


jonaro00:
My main concern is: What if the project is not in `running` state? Is it not weird to say "already running" then? I.e. it can be errored, starting, ...
Maybe this can be done by embedding the State (which is not `Destroyed`) in the new ErrorKind enum variant, and have that state's string representation in the error message. A bit of playing with `&str` vs `String` might have to happen.
What do you think?

d4ckard:
If we can limit the number of cases, I would prefer not to add any strings to `ErrorKind`. This will likely increase it's complexity by an unnecessary amount. To begin with, I have to check which additional cases must be considered. The new error message is kind of a shot in the dark. The worst case solution would be to rephrase it as a question if we can't be sure that the project is already running. It could also give a hint on how check the project's state or similar.

jonaro00 (approved):
Alrighty, this will do fine as a first solution.

orhun (approved):

oddgrd (approved):
LGTM, thanks for adding tests as well! :star_struck: Regarding it being a few potential states, we could add "try running cargo shuttle project status" or something similar to the error message. We can merge this either way, though, let me know what you think!

oddgrd:
Cool, sounds good!

d4ckard:
How about "it looks like your project is already running. You can find out more with `cargo shuttle project status` "? This adds the hint to check the project status. The "it looks like" at the start signals that something different might be going on too. However, we could omit it for the sake of brevity, too.

d4ckard:
I gathered that `cargo shuttle project status` does nothing except printing the name and the state returned from gateway (based on cargo-shuttle's `Shuttle::project_status` and the `Response` and `State` in `common/src/models/projects.rs`).

It seems to me that in `GatewayService::create_project` (where the new  error is created), all of this information is available too, since the database query returns the `Project` that's running at the moment.

If I am correct, we could make `GatewayService::create_project` generate an error message that basically already contains all of the information that a user would get by running `cargo shuttle project status`. I would be happy to do so if that's desired.

jonaro00:
Yeah, that was roughly what I was thinking at first. A great task for a follow-up PR! 😃

PR #1193, opened 2023-08-31
feat(logger): support pagination

orhun:
↕️

### DIFF
--- a/deployer/src/handlers/mod.rs
+++ b/deployer/src/handlers/mod.rs
@@ -551,16 +551,20 @@ pub async fn start_deployment(
     ),
     params(
         ("project_name" = String, Path, description = "Name of the project that owns the deployment."),
-        ("deployment_id" = String, Path, description = "The deployment id in uuid format.")
+        ("deployment_id" = String, Path, description = "The deployment id in uuid format."),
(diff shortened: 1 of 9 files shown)


deployer/src/handlers/mod.rs
```diff
@@ -609,6 +613,7 @@ async fn logs_websocket_handler(
 ) {
     let mut logs_request: tonic::Request<LogsRequest> = tonic::Request::new(LogsRequest {
         deployment_id: deployment_id.to_string(),
+        ..Default::default()
```
oddgrd, line 616: We should consider allowing users to set a limit here too, but if it's tricky we can do it later. For example, in docker you can do `docker logs <id> --follow --tail 1000`. Although, tail is different from limit, perhaps we should consider supporting that rather than limit and page.

oddgrd (commented):
Thanks Orhun, this looks great! :heart_eyes:  I think it's very nearly there, but I left some questions we might want to discuss. Also, it would be great with an integration tests for the paginated logging.

orhun:
I added integration tests in cd0e4311604465fe99be78ba8984b28d7b1706b9

PR #1194, opened 2023-09-02
feat(gateway): inform project owner about running state

d4ckard:
## Description of change

`cargo shuttle project start` fails if the project already exists. If the existing project is owned by the caller of the command, new error messages are now  displayed. They automatically provide the project owner with the current status of their project, and they indicate that her or his project is running. Depending on the current state, different hints are also provided.

Extends #1192

The assertions added in #1192 have been updated to match the changes.

### DIFF
--- a/common/src/models/error.rs
+++ b/common/src/models/error.rs
@@ -30,7 +30,7 @@ impl Display for ApiError {
 
 impl std::error::Error for ApiError {}
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, strum::Display)]
+#[derive(Debug, Clone, PartialEq, Eq, strum::Display)]
 pub enum ErrorKind {
     KeyMissing,
     BadHost,
@@ -41,8 +41,12 @@ pub enum ErrorKind {
     UserAlreadyExists,
     ProjectNotFound,
     InvalidProjectName,
-    ProjectAlreadyExists,
-    ProjectAlreadyRunning,
+    ProjectAlreadyExists {
+        // A message describing a running state of the project.
+        // Used if the project already exists but is owned
+        // by the caller, which means they can modify the project.
+        owner_state_msg: Option<String>,
+    },
     ProjectNotReady,
     ProjectUnavailable,
     CustomDomainNotFound,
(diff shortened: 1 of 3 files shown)


jonaro00 (approved):

orhun (approved):

gateway/src/service.rs
```diff
@@ -480,10 +481,35 @@ impl GatewayService {
...
+                    | State::Stopping
+                    | State::Rebooting
+                    | State::Destroying => {
+                        format!("project '{project_name}' is already {state}. You can check the status again using `cargo shuttle project status`.")
+                    }
+                    // Use different message than the default for `State::Ready`.
+                    State::Ready => {
+                        format!("project '{project_name}' is already running")
+                    }
+                    State::Started | State::Stopped | State::Destroyed => {
```
oddgrd, line 503: If a project is in the `Stopped` state it is idled, so perhaps we should have a separate error for this saying that the project is idled, and perhaps link to the idle projects docs?

common/src/models/error.rs
```diff
@@ -41,8 +41,12 @@ pub enum ErrorKind {
     UserAlreadyExists,
     ProjectNotFound,
     InvalidProjectName,
-    ProjectAlreadyExists,
-    ProjectAlreadyRunning,
+    ProjectAlreadyExists {
+        // A message describing a running state of the project.
+        // Used if the project already exists but is owned
+        // by the caller, which means they can modify the project.
+        owner_state_msg: Option<String>,
```
oddgrd, line 48: Hmm, should we change this to `state_msg` or just `message`? And we should be able to make this a doc comment: `///`.

oddgrd (commented):
Thanks @d4ckard! I just have one small comment about the `Stopped` state, and a naming nit :smile:

common/src/models/error.rs
d4ckard: Yes, sounds like a good choice. I felt like this name was suboptimal anyways. I tried to make it clear that this is a message which is directly aimed at the project owner but not for any other case. Alternatively the entire `Option` could be replaced by an additional variant in the `enum` too.

gateway/src/service.rs
d4ckard: Ok, sounds good. I was unsure about the purpose of the `Stopped` state, so I chose to keep the message pretty vague. Thanks for explaining its purpose.

oddgrd: No worries, the naming of these states is not the best, we haven't really updated the names as we changed what they represent. This is all going away in a refactor we are working on, project management will be abstracted from the user and rolled into `deploy`, but it will be some time before that is released.

common/src/models/error.rs
oddgrd: Ah, maybe we could just have `ProjectAlreadyExists` and have a message field on that? :thinking:

jonaro00 (approved):

oddgrd (approved):
LGTM, thank you @d4ckard!

PR #1195, opened 2023-09-02
fix(persist): don't use lifetime in error

jonaro00:
## Description of change

Small simplification of persist API that also f i x e s [an error in examples ](https://github.com/shuttle-hq/shuttle-examples/pull/40#issuecomment-1703842094)

### DIFF
--- a/resources/persist/src/lib.rs
+++ b/resources/persist/src/lib.rs
@@ -11,15 +11,15 @@ use std::path::PathBuf;
 use thiserror::Error;
 
 #[derive(Error, Debug)]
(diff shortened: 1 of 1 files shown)


orhun:
This has definitely has a *small* impact on the memory usage since now we use allocated strings instead of `str` reference. Do we take this into account?

jonaro00:
It is an error that should never happen in the first place. This plugin is also aimed at light/beginner users :)

orhun (approved):

oddgrd (approved):
Thanks!

PR #1198, opened 2023-09-04
feat(cargo-shuttle): prompt for init path when not given, warn if init dir not empty

jonaro00:
## Description of change

- Giving template and proj name but no path used to assume path='.', but will now correctly see that you didn't provide path, and prompt you.
- Warning can let you quit the init:
  image

Added a test that tests one version of these two changes in combination.


chesedo (approved):
LGTM

orhun (approved):

PR #1199, opened 2023-09-05
Revert "feat(shuttle-axum) Make AxumService generic to be able to use axum::State with it (#924)"

jonaro00:
This reverts commit e6ade25efbca897a732ee3bb4f7de285b8e16d50.

## Description of change

An axum router with `<S != ()>` cannot be turned into a service. IT is only implemented for `<S = ()>` https://docs.rs/axum/latest/axum/struct.Router.html#impl-Router%3C(),+B%3E
The generic `<S>` means which state the router is missing. https://docs.rs/axum/latest/axum/routing/struct.Router.html#what-s-in-routers-means

The example that was created for "verifying" #924, https://github.com/shuttle-hq/shuttle-examples/pull/52, did not verify the functionality of the generic (it's a good example tho). If you try setting S to anything, your code will now compile.
Basically, #924 was never fully tested.

### DIFF
--- a/services/shuttle-axum/src/lib.rs
+++ b/services/shuttle-axum/src/lib.rs
@@ -18,7 +18,7 @@ use shuttle_runtime::{CustomError, Error};
 use std::net::SocketAddr;
 
 /// A wrapper type for [axum::Router] so we can implement [shuttle_runtime::Service] for it.
-pub struct AxumService<S = ()>(pub axum::Router<S>);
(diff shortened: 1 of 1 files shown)


jonaro00:
@morlinbrot I'm curios what you think. Am I right or wrong?

orhun (approved):

iulianbarbu (approved):

PR #1200, opened 2023-09-05
chore(changelog): add git-cliff configuration

orhun:
## Description of change

Add [`git-cliff`](https://git-cliff.org/) configuration.

Changelog preview:

preview

N/a


jonaro00 (approved):

oddgrd (approved):
Thanks @orhun, this looks great! :star_struck: 

I have two (non-blocking) questions since I'm not familiar with git-cliff yet:
- How will commits that don't match any of the regexes in commit_parsers be grouped?
- How is it determined which part of the codebase the commit belongs too (e.g. `(gateway) <commit message>`?

orhun:
> How will commits that don't match any of the regexes in commit_parsers be grouped?

If they are conventional, they will appear in their own section. For example, `foo(bar): xyz` will be:

```
### Foo
- (bar): xyz
```

If they are not conventional, they will be skipped. We can potentially set up `git-cliff` to utilize unconventional commits as well but the most of the codebase seems to be using conventional commits so I'd say there is no need for that. Conventional commits gud.

> How is it determined which part of the codebase the commit belongs too (e.g. (gateway) \<commit message\>?

From the scope of the commit message, e.g. `feat(gateway): <commit message>`

jonaro00:
Alright, so we'll have to make sure we modify names of commits when we squash them. Would be good to have non-matching ones get thrown into Misc. just in case.

orhun:
Done in b3bd5577ba103652f4a6945598d5070c8877ea4a

oddgrd:
Perfect, thanks!

PR #1209, opened 2023-09-07
fix(cargo-shuttle): break deploy logs when StartResponse is received

iulianbarbu:
## Description of change

Receiving `StartResponse` for a `cargo shuttle deploy` means the deployment logs ended and we can stop the deployment logs stream.

By locally deploying a project on a local deployer.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -13,7 +13,9 @@ use std::path::{Path, PathBuf};
 use std::process::exit;
 use std::str::FromStr;
 
-use shuttle_common::deployment::{DEPLOYER_END_MESSAGES_BAD, DEPLOYER_END_MESSAGES_GOOD};
+use shuttle_common::deployment::{
+    DEPLOYER_END_MESSAGES_BAD, DEPLOYER_END_MESSAGES_GOOD, DEPLOYER_START_RESPONSE,
+};
 use shuttle_common::models::deployment::CREATE_SERVICE_BODY_LIMIT;
 use shuttle_common::{
     claims::{ClaimService, InjectPropagation},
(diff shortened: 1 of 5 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -1157,13 +1159,19 @@ impl Shuttle {
...
+
                     if DEPLOYER_END_MESSAGES_GOOD
                         .iter()
                         .any(|m| log_item.line.contains(m))
                     {
                         debug!("received end message, breaking deployment stream");
                         break;
                     }
+
+                    if log_item.line.contains(DEPLOYER_START_RESPONSE) {
```
chesedo, line 1171: Do we still need the `DEPLOYER_END_MESSAGES_GOOD` line with this?

chesedo (commented):
Small question

cargo-shuttle/src/lib.rs
iulianbarbu: I think so. Those are other cases in which we can break the loop. Maybe I can add the new message in that vector though?

iulianbarbu: Done.

common/src/deployment.rs
```diff
@@ -29,14 +29,16 @@ pub enum Environment {
 }
 
 pub const DEPLOYER_END_MSG_STARTUP_ERR: &str = "Service startup encountered an error";
+pub const DEPLOYER_END_MSG_BUILD_ERR: &str = "Is the Shuttle runtime missing?\"} service build encountered an error";
 pub const DEPLOYER_END_MSG_CRASHED: &str = "Service encountered an error and crashed";
 pub const DEPLOYER_END_MSG_STOPPED: &str = "Service was stopped by the user";
 pub const DEPLOYER_END_MSG_COMPLETED: &str = "Service finished running all on its own";
+pub const DEPLOYER_RUNTIME_START_RESPONSE: &str = "{response=\"StartResponse { success: ";
```
jonaro00, line 36: These should be used both for printing the message and for matching it. I'm updating the StartResponse message in my PR, so I can make these changes.

jonaro00: Have this on my PR now

iulianbarbu: No problem, will update its usage from `cargo-shuttle`.

iulianbarbu: NVM, as you pointed, they are already covered in the merged PR. Resolving this.

iulianbarbu:
Closing this since it was already covered.

PR #1210, opened 2023-09-07
refactor(runtime,codegen): avoid double timestamps problem

iulianbarbu:
## Description of change

Rebased on top of #1209.

* Runtime resource tracker emit logs prepended by a timestamp we add explicitly. This was removed.
* Our default tracing fmt layer formats events by prepending a timestamp, but we also emit the timestamp for each line from the runtime stdout collecting logic. I changed the default tracing layer to format events without a timestamp.

Note:
* users custom tracing is not touched and if they emit timestamps they will pop up in the deployment logs. I think this should be expected behavior and leave the users the choice to format their logs as wanted. 

Tested locally:
* with the default tracing
* without the default tracing

### DIFF
--- a/codegen/src/shuttle_main/mod.rs
+++ b/codegen/src/shuttle_main/mod.rs
@@ -16,8 +16,10 @@ pub(crate) fn r#impl(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let tracing_setup = if cfg!(feature = "setup-tracing") {
         Some(quote! {
                 use shuttle_runtime::colored::*;
+                use shuttle_runtime::tracing_subscriber::{registry, fmt, prelude::*};
                 shuttle_runtime::colored::control::set_override(true);
-                shuttle_runtime::tracing_subscriber::fmt::init();
+                registry().with(fmt::layer().without_time()).init();
+
                 println!(
                     "{}\n{}\nTo disable tracing, remove the default features from {}:\n{}\n{}",
                     "Shuttle's default tracing subscriber is initialized!".yellow().bold(),
--- a/runtime/src/resource_tracker.rs
+++ b/runtime/src/resource_tracker.rs
@@ -56,7 +56,7 @@ impl ResourceTracker {
 
 macro_rules! log {
     ($msg:expr) => {
-        println!("{} [Resource][{}] {}", chrono::Utc::now(), B::TYPE, $msg);
+        println!("[Resource][{}] {}", B::TYPE, $msg);
     };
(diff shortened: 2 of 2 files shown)


chesedo (commented):
Seems good to go. Just need to remove the changes from #1209

codegen/src/shuttle_main/mod.rs
```diff
@@ -16,8 +16,10 @@ pub(crate) fn r#impl(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let tracing_setup = if cfg!(feature = "setup-tracing") {
         Some(quote! {
                 use shuttle_runtime::colored::*;
+                use shuttle_runtime::tracing_subscriber::{registry, fmt, prelude::*};
                 shuttle_runtime::colored::control::set_override(true);
-                shuttle_runtime::tracing_subscriber::fmt::init();
+                registry().with(fmt::layer().without_time()).init();
```
oddgrd, line 21: We'll need to document that users that set up their own tracing need to do this too, then? Could we rather remove our timestamps, and then they won't have timestamps unless they enable them? :thinking: (you may have already discussed this, sorry :sweat_smile:)

jonaro00: User logs should not show up with our timestamps. Me and Iuilan came up with a rough solution in our heads: Add `UserService` to the Backend enum, and filter accordingly. Logs will also need to be sorted in deployer somehow.

iulianbarbu: > We'll need to document that users that set up their own tracing need to do this too, then?

I wouldn't say they need to do the same for their own tracing. I am probably of the unpopular opinion that for now having two timestamps in the logs is fine as long as the inner timestamp comes from users own tracing setup. 

This will be done better in the future when we find a way to separate between codegen/runtime logs and user service logs, as @jonaro00 mentioned.

> Could we rather remove our timestamps, and then they won't have timestamps unless they enable them?

I think this was briefly discussed. I don't remember exploring it too much. It sounds like a good option, but I am thinking that we'll have runtime logs without timestamps and possible user logs with timestamps (depending on the configs).. which still don't look great. What do you think?

> Ah, cool! So we can revert this then, right?

Not sure what you mean by reverting here.

oddgrd: Good point, I agree, I think we can start without removing our timestamp, and then consider changing it later.

chesedo (approved):
LGTM

PR #1213, opened 2023-09-08
fix: span names, log levels and messages

jonaro00:


chesedo (commented):
Looks good

jonaro00:
Missed a detail on colorize. We can probably remove tracing feature from codegen now because if the move.

runtime/src/alpha/mod.rs
```diff
@@ -64,6 +64,43 @@ pub async fn start(loader: impl Loader<ProvisionerFactory> + Send + 'static) {
         }
     };
 
+    // this is handled after arg parsing to not interfere with --version above
+    if cfg!(feature = "setup-tracing") {
```
orhun, line 68: Why this was moved to here from codegen?

jonaro00: It is not needed in codegen, and it interfered with the arg parsing of runtime, which the comment states.

cargo-shuttle/src/lib.rs
```diff
@@ -843,12 +843,9 @@ impl Shuttle {
             while let Ok(message) = rx.recv() {
                 match message {
                     Message::TextLine(line) => println!("{line}"),
-                    Message::CompilerMessage(message) => {
-                        if let Some(rendered) = message.message.rendered {
-                            println!("{rendered}");
-                        }
+                    message => {
+                        trace!("skipping cargo line: {message:?}")
```
oddgrd, line 847: Is this to avoid the super verbose cargo messages we say during builds?

common/src/claims.rs
```diff
@@ -219,7 +219,7 @@ impl Claim {
         let mut validation = Validation::new(jsonwebtoken::Algorithm::EdDSA);
         validation.set_issuer(&[ISS]);
 
-        trace!(token, "converting token to claim");
+        // trace!(token, "converting token to claim");
```
oddgrd, line 222:
```suggestion
        trace!("converting token to claim");
```

iulianbarbu:
> Missed a detail on colorize. We can probably remove tracing feature from codegen now because of the move.

It might be readded because of `shuttle-next`. I am still working through the default tracing for it, blocked by some wasmtime related impl details.

deployer/src/deployment/run.rs
```diff
@@ -423,13 +416,12 @@ async fn run(
...
         .into_inner();
 
-    info!("starting service");
     let response = runtime_client.start(start_request).await;
 
     match response {
         Ok(response) => {
             if response.into_inner().success {
-                info!(DEPLOYER_RUNTIME_START_RESPONSE);
+                info!("{}", DEPLOYER_RUNTIME_START_RESPONSE);
```
oddgrd, line 424: Just curious, what is the benefit of this change?

oddgrd (commented):
Looks good! Just left some questions.

cargo-shuttle/src/lib.rs
jonaro00: Yes

common/src/claims.rs
jonaro00: @iulianbarbu How about this?

deployer/src/deployment/run.rs
jonaro00: Logging `string` vs logging `{VARIABLE_NAME="string"}`

iulianbarbu (approved):
LGTM! resolved my comments

PR #1218, opened 2023-09-11
fix(otel): restore honeycomb and dd exporters

oddgrd:
## Description of change

Restores the exporters we disabled to ease local testing when we used the otel collector to export traces to our logger. This is a rollback to the previous code.

Only tested that the local environment runs as normal with this change. The otel collector fails with missing api-keys, as it always has for local environments.

### DIFF
--- a/extras/otel/otel-collector-config.yaml
+++ b/extras/otel/otel-collector-config.yaml
@@ -50,32 +50,25 @@ processors:
         action: insert
 
 exporters:
-  logging:
-    loglevel: error
-#   datadog:
-#     api:
-#       site: datadoghq.eu
-#       key: ${env:DD_API_KEY}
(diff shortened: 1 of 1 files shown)


extras/otel/otel-collector-config.yaml
```diff
@@ -50,32 +50,25 @@ processors:
...
     traces:
       receivers: [otlp]
       processors: [attributes, batch]
-      exporters: [logging, otlp]
+      exporters: [datadog, otlp]
     logs:
       receivers: [otlp]
       processors: [attributes, batch]
-      exporters: [logging]
+      exporters: [datadog, otlp]
```
chesedo, line 70: I'm thinking honey will crash when it receives logs :man_shrugging:

PR #1221, opened 2023-09-11
refactor: improve stream logs

chesedo:
## Description of change
Only sends logs for requested deployment id, and end closed streams.

Na


iulianbarbu (approved):
lgtm!

PR #1222, opened 2023-09-11
batch in 1 sec intervals

jonaro00:

### DIFF
--- a/proto/src/lib.rs
+++ b/proto/src/lib.rs
@@ -379,11 +379,9 @@ pub mod logger {
         }
 
         /// Create a batcher around inner. It will send a batch of items to inner if a capacity of 2048 is reached
-        /// or if an interval of 5 seconds are reached.
-        ///
-        /// These are the same defaults used by the otel batcher
+        /// or if an interval of 1 second is reached.
         pub fn wrap(inner: I) -> Self {
-            Self::new(inner, 2048, Duration::from_secs(5))
+            Self::new(inner, 2048, Duration::from_secs(1))
         }
 
         /// Send a single item into this batcher


orhun (approved):

orhun:
Can you add a type/scope to the commit message before merge?

jonaro00:
It won't make it to any release note tho. Feature branch ;)
Or are we not squashing it? O_o idk

proto/src/lib.rs
```diff
@@ -379,11 +379,9 @@ pub mod logger {
         }
 
         /// Create a batcher around inner. It will send a batch of items to inner if a capacity of 2048 is reached
-        /// or if an interval of 5 seconds are reached.
-        ///
-        /// These are the same defaults used by the otel batcher
+        /// or if an interval of 1 second is reached.
         pub fn wrap(inner: I) -> Self {
-            Self::new(inner, 2048, Duration::from_secs(5))
+            Self::new(inner, 2048, Duration::from_secs(1))
```
oddgrd, line 384: I was thinking we should set this to 1 in deployer, since we have some control over the volume of logs there and it's not that much. I think the best way forward might be to set it to 1 by default, and then manually set it with Batcher::new in the runtimes. Wdyt?

oddgrd:
We usually merge feature branches.

proto/src/lib.rs
chesedo: I was thinking we only change it for the recorder being send down here
https://github.com/shuttle-hq/shuttle/blob/91e9239f05f937f6c8f736a2012818f930197d21/deployer/src/lib.rs#L42

oddgrd: Makes sense, I was just thinking of setting it to 1 by default in case we wanted that in other services too (they should have similar characteristics to deployer).

PR #1227, opened 2023-09-12
fix: truncate log item strings

jonaro00:

### DIFF
--- a/common/src/log.rs
+++ b/common/src/log.rs
@@ -54,7 +54,8 @@ pub struct LogItem {
     pub line: String,
 }
 
-const LOG_LINE_MAX_CHARS: usize = 2048;
+const LOGLINE_MAX_CHARS: usize = 2048;
+const TRUNC_MSG: &str = "... (truncated)";
 
 impl LogItem {
     pub fn new(id: Uuid, internal_origin: Backend, line: impl Into<String>) -> Self {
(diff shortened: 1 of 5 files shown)


chesedo (approved):
LGTM

orhun (approved):

common/src/log.rs
```diff
@@ -54,8 +54,15 @@ pub struct LogItem {
     pub line: String,
 }
 
+const LOG_LINE_MAX_CHARS: usize = 2048;
+
 impl LogItem {
-    pub fn new(id: Uuid, internal_origin: Backend, line: String) -> Self {
+    pub fn new(id: Uuid, internal_origin: Backend, line: impl Into<String>) -> Self {
+        let mut line = line.into();
+        if line.chars().count() > LOG_LINE_MAX_CHARS {
```
oddgrd, line 62: Unless I am reading this incorrectly we now iterate over all the chars each time we create a new LogItem. Should we rather check the `len()`, which is stored with the string?

common/src/log.rs
```diff
@@ -54,8 +54,15 @@ pub struct LogItem {
     pub line: String,
 }
 
+const LOG_LINE_MAX_CHARS: usize = 2048;
+
 impl LogItem {
-    pub fn new(id: Uuid, internal_origin: Backend, line: String) -> Self {
+    pub fn new(id: Uuid, internal_origin: Backend, line: impl Into<String>) -> Self {
```
oddgrd, line 60: Can't this be just `line: String`?

common/src/log.rs
oddgrd: And if the line is over the limit, we still need to iterate over the entire string before we can know the count is over the limit.

common/src/log.rs
```diff
@@ -54,8 +54,15 @@ pub struct LogItem {
...
+        let mut line = line.into();
+        if line.chars().count() > LOG_LINE_MAX_CHARS {
+            line = line.chars().take(LOG_LINE_MAX_CHARS).collect()
```
oddgrd, line 63: Should we also add something to make it clear that this is truncated, like `...`?

iulianbarbu: I would add a suffix with `... [ truncated ]`.

common/src/log.rs
jonaro00: impl into is generally friendlier in an API that consumes String.

PR #1231, opened 2023-09-13
fix(deployer): added runtime error handling

iulianbarbu:
## Description of change

* Handled errors around runtime unexpected behavior by logging the errors and killing the runtime process.

When user services are CPU bound the runtime logs stream handled by the deployer can break with a broken pipe error (I haven't confirmed the root cause yet because it is not that straightforward - big stack based on tonic, hyper, and h2, but I expect it to happen because the runtime becomes CPU bound and it can not respond in time to the client keep-alive messages, which will be followed by closing the corresponding end of the tonic channel), exiting the `run` task with panic.

N/A

### DIFF
--- a/deployer/src/deployment/run.rs
+++ b/deployer/src/deployment/run.rs
@@ -74,6 +74,7 @@ pub async fn task(
                     active_deployment_getter.clone(),
                     runtime_manager.clone(),
                 );
+                let runtime_manager_clone = runtime_manager.clone();
                 let cleanup = move |response: Option<SubscribeStopResponse>| {
                     debug!(response = ?response,  "stop client response: ");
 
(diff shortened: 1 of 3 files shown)


deployer/src/runtime_manager.rs
```diff
@@ -168,6 +175,12 @@ impl RuntimeManager {
         Ok(runtime_client)
     }
 
+    pub fn kill_process(&mut self, id: Uuid) {
+        if let Some((mut process, _)) = self.runtimes.lock().unwrap().remove(&id) {
+            let _ = process.start_kill();
```
orhun, line 180: Two questions here:
- can we also handle the error here?
- is there a way to ensure that the child is sending `wait` call to avoid creating a zombie process?

iulianbarbu: > can we also handle the error here?

Yep, handling should be ok. I thought we didn't necessarily need it because there is also the aspect of `kill_on_drop()`, which will happen when the process is dropped, so we're somewhat confident the process will be killed in the end. I am doing this here just as an extra beside the log that will let us know the process should be killed.

> is there a way to ensure that the child is sending `wait` call to avoid creating a zombie process?

not sure what you mean by having the child send a `wait` call. Do you mean calling `wait` on the child before exiting the parent process (meaning the deployer service)? If yes, I think that if the deployer exits gracefully, the processes are killed because they have `kill_on_drop` set to true, so no need for waiting. In case of a non-graceful exit I am not sure we can do much if the runtime stays alive as a zombie process, but either way, the project is unusable at that point, and the runtime might still do some background work as a rogue process.

jonaro00 (approved):

deployer/src/deployment/run.rs
```diff
@@ -263,11 +277,11 @@ impl Built {
         let runtime_client = runtime_manager
             .lock()
             .await
-            .get_runtime_client(
+            .create_runtime_client(
                 self.id,
                 project_path.as_path(),
                 self.service_name.clone(),
-                alpha_runtime_path,
+                alpha_runtime_path.clone(),
```
oddgrd, line 284:
```suggestion
                alpha_runtime_path,
```

oddgrd (approved):
Nice! We had a lot of unwraps here :sweat_smile:

PR #1234, opened 2023-09-13
fix: ignore non matching logs earlier

jonaro00:

### DIFF
--- a/logger/src/lib.rs
+++ b/logger/src/lib.rs
@@ -124,17 +124,18 @@ where
             loop {
                 match logs_rx.recv().await {
                     Ok(logs) => {
+                        // only process this vec if not empty and logs was for this id
+                        // all logs in the vec should have the same id
                         if !logs
                             .first()
                             .is_some_and(|l| l.deployment_id == deployment_id)
                         {
                             continue;
                         }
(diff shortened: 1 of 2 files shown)


Cargo.lock
orhun, line 1: Are these `Cargo.lock` changes related?

jonaro00: Probably leftovers from a hasty merge of another PR

logger/src/lib.rs
```diff
@@ -124,10 +124,14 @@ where
...
-                            debug!("stream receiver queue size {}", logs_rx.len())
+                        if !logs
+                            .first()
+                            .is_some_and(|l| l.deployment_id == deployment_id)
+                        {
+                            continue;
                         }
-
+                        // only if logs was for this recipient, to reduce spam.
+                        debug!("stream receiver queue size {}", logs_rx.len());
```
oddgrd, line 134: We should still check if the rx queue is empty, I think.

jonaro00: Wait.. haha the merge goofed everything

jonaro00:
Can't prove that this will work, since deployer batchers could send mixed ids at intersection of deployments.

PR #1236, opened 2023-09-13
refactor(common): fix project name validation

### LINKED ISSUE #1223: [Improvement]: Fix ProjectName validation
Continuation of the task described in https://discord.com/channels/803236282088161321/1052364653373702235/1150824517548134531

1. Merge the ProjectName types so that it is only defined in common. Use the good parts from gateway (no uppercase, sqlx derive) and the good parts from #418 (reserved names, doc string, tests, etc) to common.
2. Gateway's imports of ProjectName can be redirected to common, and should be a clean replacement.
3. Cargo-shuttle can still use the type, but should never validate a string wrapped in it. Same for other user-facing crates.
   -  Beware that #1050 is getting merged soon (probably before this), and there are some touches on its uses there, so don't get too ahead of yourself.
4. Check that gateway validates the project name where necessary.

Let me know if I forgot something.
GLHF💪

d4ckard:
## Description of change

- Merge the two implementations of `ProjectName` in `gateway` and `common`.
- Improve project name validation with changes from PR #418.

Closes #1223

The existing tests were updated to reflect the new constraints added to the implementation in `common` from the deleted implementation in `gateway` and from PR #418.


common/src/project.rs
```diff
@@ -7,62 +7,65 @@ use std::error::Error;
...
-        !(hostname.bytes().any(|byte| !is_valid_char(byte))
-            || is_reserved(hostname)
-            || !is_profanity_free(hostname)
-            || hostname.ends_with(separators)
-            || hostname.starts_with(separators)
-            || hostname.is_empty())
+        !(label.bytes().any(|byte| !is_valid_char(byte))
+            || is_reserved(label)
+            || !is_profanity_free(label)
+            || is_too_long
```
jonaro00, line 65: nit: length check can be first to guard against unnecessary iteration.

common/src/project.rs
```diff
@@ -76,35 +79,46 @@ impl AsRef<String> for ProjectName {
     }
 }
 
+impl<'de> Deserialize<'de> for ProjectName {
```
jonaro00, line 82: Is it still reasonable to validate on deser? Does cargo-shuttle deser as this when reading Shuttle.toml for example?

common/src/project.rs
```diff
@@ -7,62 +7,65 @@ use std::error::Error;
...
-/// - It does not contain profanity.
-/// - It is not a reserved word.
+/// - It is shorter than 64 characters.
+///
+/// Additionaly, while host segments are technically case-insensitive, the filesystem isn't,
+/// so we restrict project names to be lower case. We also restrict the use of profanity,
+/// as well as a list of reserved words.
 #[derive(Clone, Serialize, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "backend", derive(sqlx::Type, Hash))]
+#[cfg_attr(feature = "backend", sqlx(transparent))]
```
jonaro00, line 24: nice

jonaro00 (commented):
I'm afraid we will wait a bit before merging this, but looking pretty good overall!

common/src/project.rs
d4ckard: Great idea, this adapts to different shapes of invalid inputs in a pretty neat way.

common/src/project.rs
d4ckard: That  I'd have to check.

I don't know how  bad the performance hit of always validating is. If we keep the validation, we can at least be certain that there are never going to be any `ProjectName`s floating around that are actually invalid. It really depends on how long it takes to re-validate correct project names.

jonaro00: We could stick to "only validate incoming names in gateway", but shouldn't hurt too much. I like the idea of thin client and having explicit validation points (deser calls are easily hidden).

jonaro00: Maybe worth considering that cargo-shuttle might want to do some validation so that the request URLs are valid (or handle the error if the resulting URL is invalid).

d4ckard: How would those validations impact the deserialization logic of `ProjectName`?

jonaro00: Probably not at all. Was a semi-unrelated thought. Feel free to choose the solution you like.

common/src/project.rs
jonaro00: Did you want to change this?
(P.S it can be: empty || too long || (other)... )

d4ckard: Yes, I would. Could do so by tomorrow. Don't hold back if you'd like to fix this now already 🙂

jonaro00:
I've thought about this change, and I must say that the `ProjectName` struct feels a bit over-engineered.

What we want to do:
- validate project name on incoming API endpoints in t he backend(s)

What we do:
- Wrap it in a struct, forcing us to derive and implement lots of traits + causing the sqlx derive issue
- Validate it on every Deser, which means it is re-validated every time it passes a backend border, but also...
- validate it when coming out of the gateway database (it should already be valid when being put in)
- Pass around this struct all over the codebase, when we already know it is a plain string that has already been validated
- Expose it in user-facing API (shuttle-metadata) for not much reason.

Therefore we might just:
- Replace all occurrences of it with String
- Make `is_valid` a pure function `valid_project_name`
- Explicitly validate it in each endpoint that takes it in

Or we make some sort of compromise.
Or we use `Deref`?

I realize that this is pretty substantial and can be carried over to another PR, if wanted.
What do you think @oddgrd?

jonaro00:
@d4ckard There is one conflict and then the ordering nitpick :)

jonaro00 (approved):
Nice, the boolean expression looks cleaner now :D

oddgrd:
> I've thought about this change, and I must say that the `ProjectName` struct feels a bit over-engineered.
> 
> What we want to do:
> 
>     * validate project name on incoming API endpoints in t he backend(s)
> 
> 
> What we do:
> 
>     * Wrap it in a struct, forcing us to derive and implement lots of traits + causing the sqlx derive issue
> 
>     * Validate it on every Deser, which means it is re-validated every time it passes a backend border, but also...
> 
>     * validate it when coming out of the gateway database (it should already be valid when being put in)
> 
>     * Pass around this struct all over the codebase, when we already know it is a plain string that has already been validated
> 
>     * Expose it in user-facing API (shuttle-metadata) for not much reason.
> 
> 
> Therefore we might just:
> 
>     * Replace all occurrences of it with String
> 
>     * Make `is_valid` a pure function `valid_project_name`
> 
>     * Explicitly validate it in each endpoint that takes it in
> 
> 
> Or we make some sort of compromise. Or we use `Deref`?
> 
> I realize that this is pretty substantial and can be carried over to another PR, if wanted. What do you think @oddgrd?

I think using the type system to ensure we can't accidentally use an invalid project name anywhere is a good design pattern, but I'm not sure about doing it on deserialize, it feels a bit obfuscated to me. But that may just be my personal preference. The way I've done this in the past (thanks zero2prod!) is to have a parse method on the struct, and that is the only way you can create an instance of the struct. It may seem like overkill now, but as the codebase grows it could prevent bugs. That being said, there might be reasons that strategy isn't a good fit here, so I'm very open to discussion.

By the way, am I mistaken in reading this as applying the stricter requirements we had in the gateway validation to the CLI as well? That will break older projects, we'd have to look into how many and how much it would take to migrate them, but it may block merging this for a bit.

jonaro00:
@oddgrd The only rule that was missing from gateway was the profanity check (which was checked in CLI only). Otherwise, all projects currently in gateway should be valid, depending on if anyone has created projects with manual requests. (They would be unusable via CLI though?) So there should not be a big blocker for merging this.

oddgrd:
> @oddgrd The only rule that was missing from gateway was the profanity check (which was checked in CLI only). Otherwise, all projects currently in gateway should be valid, depending on if anyone has created projects with manual requests. (They would be unusable via CLI though?) So there should not be a big blocker for merging this.

We also disallowed underscores in the gateway, but I'll try to make time this week and see how many projects would be affected by this.

d4ckard:
> By the way, am I mistaken in reading this as applying the stricter requirements we had in the gateway validation to the CLI as well? That will break older projects, we'd have to look into how many and how much it would take to migrate them, but it may block merging this for a bit.

While digging through the old implementation, to find out what to change, I found that it was quite hard to figure out what exactly the constraints where, and in which situations they applied. However, I believe that some of the checks were duplicated over multiple places, and that this new validation does not reject anything that would  have passed all checks before. The only thing that might have changed is where the check that filters out specific invalid requests fails.

d4ckard:
> [What we do:] Pass around this struct all over the codebase, when we already know it is a plain string that has already been validated

Regarding this point, I really don't think that this matters much. The code generated for the current iteration of the `ProjectName` struct is exactly the same that would be generated for a plain string (I disassembled a basic example, this  is just what the newtype pattern is supposed to work like). I agree that it does not make sense to give up the type safety that Rust gives us.

I don't know which one is the best way to go here.
1. Removing  `Deser` from `ProjectName` might eventually lead to strings creeping everywhere, because validation could be forgotten. It might work though, if there relevant functions (e.g. `create_project` all mandate  `ProjectName`s).
2. Allowing to construct `ProjectName` without validating it makes the entire domain model less secure. At least it keeps the semantic information attached to the type. That's reason enough to keep  some sort of `ProjectName` type, no matter how simple it is.
3. Keep `ProjectName` as is to get all the correctness guarantees attached to it. This makes all development down the line easier, because one doesn't need to worry about ever breaking the validation process: all `ProjectName`s *must* be valid.

I tend towards 3. Why give up the safety and correctness guarantees that Rust provides? Potentially, the validation procedure itself could be optimized a bit more, thus decreasing the performance hit (that might exist).

jonaro00:
Me and O did some digging, and found this:

- Axum extractors don't seem to call the custom Deser, so the only place gateway does validate the name is in `create_project` when creating a *new* project (which is kind of the only place it is needed). This realization kind of throws out the "safety" that the Deser was thought to give us. So, if it turns out that this is the only place where gateway ever validated names, then we should survive with keeping it that way, and maybe remove that type.
- 47 projects have `_` in their name, where at most one of them has recent activity. (remains to check other rule violations)

So, we will need to check some more things to make sure this change is not too breaking before it can be merged.

oddgrd:
Hey @d4ckard, I just wanted to update you on the state of things. In our investigation of which projects this would break, we determined that we can go ahead with this PR, since very few active projects are affected. However, we are heading to EuroRust next week, so we won't be able to get around to migrating those projects before that. We are happy with this PR, though, and we aim to get it merged as soon as we can after EuroRust. Apologies for the delay!

d4ckard:
No problem, I'm glad to hear that the changes are OK 😄.

oddgrd:
Hey @d4ckard, do you have time to address the failing CI and resolve the conflicts this week? If you're busy, I can also take care of it. :slightly_smiling_face:

d4ckard:
I'm a bit full this week so I'd be grateful if you could fix it.

oddgrd:
Sure thing!

common/src/project.rs
```diff
@@ -76,35 +76,46 @@ impl AsRef<String> for ProjectName {
...
 
+impl<'de> Deserialize<'de> for ProjectName {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s: String = String::deserialize(deserializer)?;
+        s.parse().map_err(DeError::custom)
+    }
+}
```
oddgrd, line 87: There is a problem with this, and that is that this will just return 400 and the message from the gateway via the axum path extractor, whereas we want it to return a `shuttle_gateway::Error`. This means if a user with an older `cargo-shuttle` calls the gateway after this change, it will be parsed to 400 status and "this request is invalid". It will not have the message regarding that this is an invalid project name and the rules for project names.

Ideally we would figure out a way to have this failure return a gateway `Error`, perhaps by creating a custom path extractor, or by not parsing the project name in deserialize, but rather a custom extractor for ProjectName. I'll look into it.

jonaro00: From what I remember, project name were not even validated in the extractor. Not sure though. I'm in favor of explicit validation in all places it is needed (not on deser).

oddgrd: I tested it and it did error in deserialize, but it didn't return an error that the CLI could parse. And making it return a gateway::Error seemed quite complicated. I am also in favor of moving it out of deserialize, and rather using a conventional newtype pattern where a ProjectName can only be created by parsing a valid project name. Perhaps a custom extractor in the API.

d4ckard: Removing the parse from deserialization sounds reasonable. I don't have an understanding of the gateway, so I think this up to you.

jonaro00:
I tried removing FromStr and Deser. It became very messy, as they are implicitly used in a lot of places.
Now I have removed all validation form cargo-shuttle.
Shall we remove the type from `DeploymentMetadata`, which is given to users in the metadata plugin?

common/src/project.rs
jonaro00: > user with an older cargo-shuttle

The older ones do client side validation, so is this an issue?

common/src/project.rs
```diff
@@ -3,146 +3,135 @@ use rustrict::{Censor, Type};
...
-        let s: String = String::deserialize(deserializer)?;
-        s.parse().map_err(DeError::custom)
-    }
-}
-
-impl std::fmt::Display for ProjectName {
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        self.0.fmt(f)
+impl ProjectName {
+    pub fn new(name: &str) -> Result<Self, InvalidProjectName> {
```
oddgrd, line 21: Should we rather have a method `parse` that combines both new and is_valid (returning either Self or an error)? Similarly to what is done for the api-key newtype.

oddgrd (commented):
Since we still parse in deserialize, will we still have the issue of not returning a parse-able error to the client if the name passed to the handlers from the `Path` is not valid?

common/src/models/error.rs
```diff
@@ -40,7 +42,7 @@ pub enum ErrorKind {
     UserNotFound,
     UserAlreadyExists,
     ProjectNotFound,
-    InvalidProjectName,
+    InvalidProjectName(InvalidProjectName),
```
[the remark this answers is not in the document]
jonaro00: Shouldn't be. It get converted to http code + message and CLI just prints it.

jonaro00:
@d4ckard I intended to get this merged alongside #1354, but after discussing this we want to keep the main branch clean and have #1354 with all changes to make it easier to review. Thanks for the work on this, your commits live on in #1354 :) (I hope it gets merged very soon!)

d4ckard:
Thanks for letting me know. I'm glad I could help 😊

PR #1243, opened 2023-09-14
feat: outdated log parse warning

jonaro00:
## Description of change

Improve error message caused by incompatibility between 0.25 & 0.26

EDIT: yes, I did test it. 🥺

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -15,7 +15,6 @@ use std::str::FromStr;
 
 use shuttle_common::deployment::{DEPLOYER_END_MESSAGES_BAD, DEPLOYER_END_MESSAGES_GOOD};
 use shuttle_common::models::deployment::CREATE_SERVICE_BODY_LIMIT;
-use shuttle_common::LogItem;
 use shuttle_common::{
     claims::{ClaimService, InjectPropagation},
     models::{
@@ -26,6 +25,7 @@ use shuttle_common::{
     },
 };
 use shuttle_common::{project::ProjectName, resource, ApiKey};
+use shuttle_common::{LogItem, API_URL_DEFAULT};
 use shuttle_proto::runtime::{
     self, runtime_client::RuntimeClient, LoadRequest, StartRequest, StopRequest, StorageManagerType,
 };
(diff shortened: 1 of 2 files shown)


orhun (approved):

cargo-shuttle/src/lib.rs
```diff
@@ -101,13 +101,16 @@ impl Shuttle {
...
-        trace!("running local client");
-
-        if args.api_url.as_ref().is_some_and(|s| s.ends_with('/')) {
-            println!(
-                "WARNING: API URL is probably incorrect. Ends with '/': {}",
-                args.api_url.clone().unwrap()
-            );
+        if let Some(ref s) = args.api_url {
+            if s != API_URL_DEFAULT {
+                eprintln!("INFO: Targetting non-standard API: {}", s);
```
iulianbarbu, line 106: Can we print at `stdout`? I would keep `stderr` exclusively for errors and this doesn't seem to be an error.

cargo-shuttle/src/lib.rs
```diff
@@ -101,13 +101,16 @@ impl Shuttle {
...
+            }
+            if s.ends_with('/') {
+                eprintln!(
+                    "WARNING: API URL is probably incorrect. Ends with '/': {}",
+                    args.api_url.clone().unwrap()
+                );
+            }
```
iulianbarbu, line 113: nit: the expectation is to not have a trailing slash so it's probably better to remove it if there and avoid printing this warning altogether. WDYT?

jonaro00: Don't want to obfuscate the parsing logic too much. This way is transparent in the sense that we are concatenating strings instead of "URL building".

cargo-shuttle/src/lib.rs
jonaro00: AFAIK the convention is to have only the main output on stdout and everything else on stderr. (think curl progress bar etc)

oddgrd: I'm leaning towards println here too, since this is more of a hint. The docs for eprintln! says to only use it for errors and progress messages. But I could be wrong here.

oddgrd (approved):
Two small comments, not really blocking. Thanks for this, good thinking!

PR #1244, opened 2023-09-15
feat: builder service

chesedo:
## Description of change
Extracts the builder service from the scaling feature branch

Not at all since it is not hooked up to deployer yet


oddgrd (approved):
LGTM, barring the discussed CI/Docker/Makefile changes, is the plan to do that in a separate PR? 

I also suppose we can't add the deploy layer yet, since that would cause the user to see duplicated build logs.

PR #1245, opened 2023-09-15
feat(cargo-shuttle): add suggestions in case of cmd failures

iulianbarbu:
## Description of change

Added suggestions in case of requests' errors when running `cargo-shuttle` commands. These suggestions do not consider the requests' status codes. This is a short-term solution for handling errors across CLI user flows (like ProjectNotReady 503s) by suggesting commands that verify if previous commands worked, repeating the command again and if the fialures persist, checking the status page or reaching out on Discord.

N/A


chesedo (approved):
Can we demo these somehow?

iulianbarbu:
> Can we demo these somehow?

Good question. Not very sure how to do it easier than simulating the error by interfering with the code, when running `cargo-shuttle` against a local Shuttle deployment.

cargo-shuttle/src/suggestions.rs
```diff
@@ -0,0 +1,225 @@
+//! Suggestions to be shown to users encountering errors while using cargo-shuttle.
+// TODO: Ideally, the suggestions would be inferred from the status codes returned by
+// the gateway in case of requests to it, or errors thrown by the client doing work
+// on the users machines. This is a naive way of handling the errors that should suggest
+// retrying common commands or reach out on our Discord server in case failures persist.
+
+use crossterm::style::Stylize;
+
+// --------------------------
+// API key related
```
orhun, line 10: Just one trivial comment: instead of separating different contexts in the same file with comments, wouldn't it be better if they have been in different modules? Not a big deal though :) LGTM

orhun (approved):

cargo-shuttle/src/suggestions.rs
iulianbarbu: Done.

PR #1247, opened 2023-09-18
feat: use smaller+newer images, script for patches, unique binary names

jonaro00:
## Description of change

- Building each image more specifically for what they need
- Patches are now applied from a template with text replacement
- Binaries in the images are now called their default name instead of `service`
- Removed duplicate ulid lib files, points to the path of gateway's copy instead

Stack runs locally. static files and postgres examples work. Worth testing more extensively on unstable


DEVELOPING.md
```diff
@@ -189,9 +178,11 @@ Finally, before gateway will be able to work with some projects, we need to crea
 The following command inserts a gateway user into the `auth` state with deployer privileges:
 
 ```bash
-docker compose -f docker-compose.rendered.yml -p shuttle-dev exec auth /usr/local/bin/service --state=/var/lib/shuttle-auth init-deployer --name gateway --key gateway4deployes
+docker compose -f docker-compose.rendered.yml -p shuttle-dev exec auth /usr/local/bin/shuttle-auth --state=/var/lib/shuttle-auth init-deployer --name gateway --key gateway4deployes
 ```
 
+### Deplying locally
```
orhun, line 184:
```suggestion
### Deploying locally
```

Containerfile
```diff
@@ -54,73 +54,91 @@ RUN cargo build \
...
-COPY --from=chef-builder /build/target/${CARGO_PROFILE}/shuttle-builder /usr/local/bin/service
+COPY --from=chef-builder /build/target/${CARGO_PROFILE}/shuttle-builder /usr/local/bin
+ENTRYPOINT ["/usr/local/bin/shuttle-builder"]
 FROM shuttle-builder AS shuttle-builder-dev
 
-FROM shuttle-crate-base AS shuttle-deployer
+
+#### DEPLOYER
+ARG RUSTUP_TOOLCHAIN
+FROM docker.io/library/rust:${RUSTUP_TOOLCHAIN}-bookworm AS shuttle-deployer
```
oddgrd, line 81: Will this be a possible breaking change for users who relied on dependencies installed in the buster image?

oddgrd (commented):
Thanks! Left one question, I'm also wondering if we should try deploying this to staging, just to double-check that it works?

Containerfile
oddgrd: It looks like they install the same dependencies :+1:

jonaro00: Just newer versions. There can be breaking changes though.

oddgrd: I think upgrades are okay, as long as we notify users in the release notes.

jonaro00:
Have tested deploying a logger, postgres, and mongo projects to unstable. Should we test it more?
There could be openssl-related build errors in newer debian image. :thinking:

oddgrd:
If you didn't see any openssl-related build errors in your testing, I think we should be good. Or is there anywhere in particular you suspet we may encounter this?

jonaro00:
I suspect there could be errors in the ocean of user projects, but shared-db using openssl worked, so hopefully that covers the majority of cases. Can test it with my personal projects once on staging. Will also be very heads up about this in the release notes.

PR #1248, opened 2023-09-18
feat(deployer): connect deployer to builder service

oddgrd:
## Description of change

This sets up the makefile/containerfile config and args needed for connecting the deployer to the builder service.

Tested with full local environment.


iulianbarbu (approved):
LGTM!

PR #1249, opened 2023-09-18
feat(ci): separation of tests that need docker

jonaro00:
## Description of change

Refactor CI for better speed and cost again ;)

Now happy with the results, CI now runs (cached) in 4m30s :D

Still sporadic fails on cargo-shuttle... at least it fails faster now :(


.circleci/config.yml
```diff
@@ -248,23 +254,26 @@ jobs:
...
-  test-workspace-member-integration-machine:
+  test-workspace-member-with-integration-xl:
     parameters:
       crate:
         description: "Crate in workspace to test"
         type: string
-    # Using a machine image since tests will start a docker container
-    executor: machine-ubuntu
+    executor: docker-rust
+    resource_class: xlarge
```
iulianbarbu, line 276: nit: we can create a docker-rust-xlarge executor.

.circleci/config.yml
```diff
@@ -248,23 +254,26 @@ jobs:
...
       - save-cargo-cache
-  test-workspace-member-integration-machine:
+  test-workspace-member-with-integration-xl:
     parameters:
       crate:
         description: "Crate in workspace to test"
         type: string
-    # Using a machine image since tests will start a docker container
-    executor: machine-ubuntu
+    executor: docker-rust
```
iulianbarbu, line 275: Do we need to have `install-rust` as a step?

cargo-shuttle/tests/integration/run.rs
```diff
@@ -396,3 +301,100 @@ async fn rocket_hello_world_with_router_ip() {
 
     assert_eq!(request_text, "Hello, world!");
 }
+
+// These examples use a shared Postgres/Mongo. Thus local runs should create a docker containers.
+mod needs_docker {
```
iulianbarbu, line 306: nit: rename to `docker_tests`.

.circleci/config.yml
jonaro00: I prefer it this way. I think this way is less DRY actually.

.circleci/config.yml
jonaro00: It is already there. The `-xl` job is a direct copy of the non-xl one

.circleci/config.yml
```diff
@@ -276,16 +285,46 @@ jobs:
...
+          command: |
+            SCCACHE_CACHE_SIZE=2G
+            set +o pipefail
+            path=$(echo "<< parameters.crate >>" | sed "s/shuttle-\(.*\)/\1/")
+            if [ -f "$path/integration_docker.sh" ]; then
+              bash "$path/integration_docker.sh"
+            else
+              echo "nothing to test here, adjust CI file or add an integration script"
+              exit 1
+            fi
```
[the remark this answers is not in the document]
jonaro00: It might be, but I've opted to do it this way so that
(1): You can easily run the same command locally.
(2): If you need to customize the ordering of tests execution, or anything else, per crate.

cargo-shuttle/tests/integration/init.rs
```diff
@@ -10,8 +10,14 @@ use tempfile::Builder;
 // quite high timeout since the template is being cloned over network
 const EXPECT_TIMEOUT_MS: u64 = 10000;
 
+/// So that cargo-generate does not crash in CI
```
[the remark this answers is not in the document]
jonaro00: They started crashing on this branch, and this solved it. Can be removed after #1269

cargo-shuttle/tests/integration/run.rs
jonaro00: We are not testing docker, we are running tests that need docker to run ;)

.circleci/config.yml
```diff
@@ -690,22 +729,28 @@ workflows:
             parameters:
               crate:
                 # - shuttle-admin # no tests
+                - shuttle-codegen
                 - shuttle-common
                 # - shuttle-common-tests # no tests
-      - test-workspace-member-integration:
+      - test-workspace-member-with-integration:
```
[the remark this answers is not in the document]
jonaro00: On it.

.circleci/config.yml
jonaro00: I found a better way.

.circleci/config.yml
jonaro00: I see what you mean now. The step does not install rust if it exists, but it does add wasm32-wasi target, which we need.

.circleci/config.yml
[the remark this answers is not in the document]
jonaro00: Hmm, it needs nix setup and massive nix build to complete... Should be separate ticket.

.circleci/config.yml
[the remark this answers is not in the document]
iulianbarbu: I would take simplicity anytime (a one-liner based on `cargo`) :), but I don't have a strong preference 

(1) I think we run locally just `cargo tests --all-features`. Also, the same command ran by the CI can be found in the config.yaml. 
(2) Sounds like an overoptimisation. Do we need to customize anything when running tests right now?

iulianbarbu (approved):

.circleci/config.yml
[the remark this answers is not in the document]
jonaro00: > Also, the same command ran by the CI can be found in the config.yaml.

That is the more annoying option. We want it to be easy for both dev and CI to run.

> Do we need to customize anything when running tests right now?

Only the builder exception right now, but I made this change because I suspect that we will need some sort of customization of test running to reduce spurious fails.

orhun (approved):

PR #1256, opened 2023-09-20
feat: added resource deletion

GugaGongadze:
## Description of change

Added service resources deletion.

N/A


resource-recorder/src/dal.rs
```diff
@@ -177,6 +180,18 @@ impl Dal for Sqlite {
...
+        sqlx::query("DELETE FROM resources WHERE project_id = ? AND service_id = ? AND type = ?")
+            .bind(resource.project_id.map(|u| u.to_string()))
+            .bind(resource.service_id.map(|u| u.to_string()))
+            .bind(resource.r#type)
+            .execute(&self.pool)
+            .await
+            .map(|_| ())?;
+
+        Ok(())
+    }
```
oddgrd, line 193: I think you missed a copy-paste here, we need to rewrite this to rather get a specific resource.

deployer/src/deployment/run.rs
```diff
@@ -574,20 +590,26 @@ mod tests {
...
+            _service_id: &ulid::Ulid,
+            _type: ResourceType,
+            _claim: Claim,
+        ) -> Result<ResourceResponse, Self::Err> {
+            Ok(ResourceResponse {
+                success: true,
+                message: "dummy impl".to_string(),
+                resource: None,
+            })
         }
```
oddgrd, line 605: We need to refactor this to match the type from the trait

deployer/src/handlers/mod.rs
```diff
@@ -320,6 +333,60 @@ pub async fn get_service_resources(
...
+        .ok_or_else(|| Error::NotFound("resource not found".to_string()))?;
+
+    // resource_recorder_client
+    // resource_manager
+    //     .delete_resource(&project_name, &resource_type)
+    //     .await
+    //     .map_err(|e| Error::Custom(e.into()))?;
+    persistence
+        .delete_resource(&service.id, persistence_resource_type, claim)
+        .await?;
```
oddgrd, line 385: I'm thinking we can try to roll the provisioner resource deletion in `persistence.delete_resource`.

deployer/src/main.rs
```diff
@@ -72,6 +76,10 @@ async fn main() {
         Backend::Deployer,
     );
 
+    let resource_manager = ResourceManager::new(&args.provisioner_address)
+        .await
+        .unwrap();
```
oddgrd, line 81: I think we can remove this ResourceManager struct, and rather pass the provisioner address to persistence, like we do with resource recorder at the start of this function (main). Then we put the provisioner client in the persistence struct, like we do for the resource manager client.

deployer/src/persistence/mod.rs
```diff
@@ -466,6 +466,35 @@ impl ResourceManager for Persistence {
...
+        claim: Claim,
+    ) -> Result<()> {
+        sqlx::query(r#"DELETE FROM resources WHERE service_id = ? AND type = ?"#)
+            .bind(service_id.to_string())
+            .bind(r#type)
+            .execute(&self.pool)
+            .await
+            .map(|_| ())
+            .map_err(Error::from)
+    }
```
oddgrd, line 497: We shouldn't need to delete from the sqlite database, resource deletion will only be available to new deployers, and they will use resource-recorder to record resources, not the sqlite database. So get_resource should use the resource-recorder client. Delete resource should first use the provisioner client to delete the resource, then if that succeeds it should use the resource-recorder client to delete the resource from the resource-recorder state.

oddgrd (commented):
Thanks Guga, this is getting there! I left some suggestions for how to solve some of the errors, as well as some suggestions for how we might re-structure this in deployer.

deployer/src/persistence/mod.rs
oddgrd: Hmm, we might also need to check that the resource is something the provisioner should delete. The provisioner only deals with databases currently. But we can probably start with the above comment and then tweak it after.

deployer/src/handlers/mod.rs
iulianbarbu: I think this should be fixed here: https://github.com/shuttle-hq/shuttle/pull/1256/commits/61eb0ebf6b1914f55ecc97e80a4676824ded85bf. LMK if you meant something else.

deployer/src/persistence/mod.rs
iulianbarbu: I think this commit reflects some of your suggestions: https://github.com/shuttle-hq/shuttle/pull/1256#discussion_r1338706181.

deployer/src/main.rs
iulianbarbu: Looking into this.

iulianbarbu: This should be covered in the latest changes.

deployer/src/handlers/mod.rs
```diff
@@ -157,6 +169,7 @@ impl RouterBuilder {
             )
             .layer(Extension(persistence))
             .layer(Extension(deployment_manager))
+            // .layer(Extension(resource_manager))
```
chesedo, line 172: Comment

deployer/src/persistence/mod.rs
```diff
@@ -459,13 +496,79 @@ impl ResourceManager for Persistence {
...
+            .await
+            .map_err(PersistenceError::ResourceRecorder)
+            .map(|res| res.into_inner());
+    }
+
+    async fn delete_resource(
+        &mut self,
+        project_name: String,
+        service_id: &Ulid,
+        resource_type: String,
```
chesedo, line 535: Shouldn't this type be the resource type (and not a string)?

deployer/src/persistence/mod.rs
```diff
@@ -459,13 +496,79 @@ impl ResourceManager for Persistence {
...
             }
         }
 
         Ok(res)
     }
+
+    async fn get_resource(
+        &mut self,
+        service_id: &Ulid,
+        r#type: String,
```
chesedo, line 510: Same here, this should be the resource enum to reduce errors

chesedo (commented):
Just minor notes

jonaro00 (commented):
Why are we deleting resources by type, and not by id?

deployer/src/persistence/mod.rs
iulianbarbu: Done.

deployer/src/handlers/mod.rs
iulianbarbu: Done.

oddgrd (approved):

PR #1265, opened 2023-09-21
ci: reduce shortest path in publish flow

jonaro00:
## Description of change

Shortest path is now 4 instead of 6. Should save about 2 mins on release.
Don't even know if referring to matrix member is valid...

We'll find out


oddgrd (approved):
This is great, thanks!

PR #1268, opened 2023-09-22
feat(builder): improve the nix build capturing of stdout/stderr

orhun:
## Description of change

Utilize `os_pipe` to both capture stdout and stderr of `nix build` command.

In progress.

### DIFF
--- a/builder/src/lib.rs
+++ b/builder/src/lib.rs
@@ -61,7 +61,8 @@ impl Service {
         let (reader, writer) = os_pipe::pipe()?;
         let writer_clone = writer.try_clone()?;
         let output_path = path.join("_archive");
-        let mut child = Command::new("nix")
+        let mut command = Command::new("nix");
+        command
             .args([
                 "build",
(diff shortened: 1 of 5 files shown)


builder/src/lib.rs
```diff
@@ -74,37 +73,24 @@ impl Service {
...
-                            error!(deployment_id, "unexpected stdout stream close: {}", err);
-                            stdout_ended = true;
-                        }
-                    }
+            .stdout(writer)
+            .stderr(writer_clone);
+        let mut child = command.spawn()?;
+
+        // Avoid a deadlock.
+        drop(command);
```
iulianbarbu, line 81: How can this happen if we're not dropping here?

iulianbarbu (approved):

builder/src/lib.rs
oddgrd: It's documented in the os_pipe docs here: <https://docs.rs/os_pipe/latest/os_pipe/#examples>

PR #1270, opened 2023-09-24
fix(cargo-shuttle): prompt for new port if port is taken

### LINKED ISSUE #1259: [Bug]: thread panic in `cargo_shuttle` while using `cargo shuttle run`
### What happened?

When executing cargo shuttle run the binary builds successfully, but the main thread then panics within the `cargo_shuttle` crate. I was running fine on the updated `v0.26.0` for a bit, but then this started happening every time out of nowhere.

### Version

v0.26.0

Linux

x86_64

### Relevant log output

```sh
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: status: Unknown, message: "transport error", details: [], metadata: MetadataMap { headers: {} }

Caused by:
    0: transport error
    1: http2 error: stream error received: stream no longer needed
    2: stream error received: stream no longer needed', /home/rhedgeco/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-shuttle-0.26.0/src/lib.rs:911:51
stack backtrace:
   0: rust_begin_unwind
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/panicking.rs:67:14
   2: core::result::unwrap_failed
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/result.rs:1651:5
   3: cargo_shuttle::Shuttle::local_run::{{closure}}
   4: cargo_shuttle::Shuttle::run::{{closure}}
   5: tokio::runtime::park::CachedParkThread::block_on
   6: tokio::runtime::runtime::Runtime::block_on
   7: cargo_shuttle::main
```

### Duplicate declaration

- [x] I have searched the issues and there are none like this.

BadgerBloke:
## Description of change
`cargo shuttle run` panic when the default port i.e., 8000 was already in use.
This fix is checking that port 8000 is available or not?
If it finds 8000 free then it continues with earlier implementation but if it found the as not free then it is checking for available 2 consecutive ports using `find_available_port` function and `find_available_port` function returns the higher one. So that, rest of the implementation like `runtime::start` will work as expected.

```rust
runtime::start(
            service.is_wasm,
            Environment::Local,
            &format!("http://localhost:{provisioner_port}"),
            None,
            run_args.port - idx - 1,
            runtime_executable,
            service.workspace_path.as_path(),
        )
```

The ports range has been set as `default port - 1` to  65535 i.e., `7999 to 65535`

Closes #1259

Test performed by spawning multiple instances of shuttle one after another and all of them was able to run properly on newer ports.


cargo-shuttle/src/lib.rs
```diff
@@ -152,7 +167,23 @@ impl Shuttle {
             Command::Login(login_args) => self.login(login_args).await,
             Command::Logout(logout_args) => self.logout(logout_args).await,
             Command::Feedback => self.feedback().await,
-            Command::Run(run_args) => self.local_run(run_args).await,
+            Command::Run(mut run_args) => {
+                let range_end = 65535;
+                if !portpicker::is_free_tcp(run_args.port) {
+                    let available_port = Shuttle::find_available_port(run_args.port - 1, range_end);
```
orhun, line 173: Why do we start checking from `port - 1`? Would it be better if we check from `port + 1` instead and remove +1 from `available_port` in `find_available_port` function?

cargo-shuttle/src/lib.rs
```diff
@@ -152,7 +167,23 @@ impl Shuttle {
             Command::Login(login_args) => self.login(login_args).await,
             Command::Logout(logout_args) => self.logout(logout_args).await,
             Command::Feedback => self.feedback().await,
-            Command::Run(run_args) => self.local_run(run_args).await,
+            Command::Run(mut run_args) => {
+                let range_end = 65535;
```
orhun, line 171: We can probably use `std::u16::MAX` here.

cargo-shuttle/src/lib.rs
```diff
@@ -104,6 +104,21 @@ impl Shuttle {
         self.run(args, provided_path_to_init).await
     }
 
+    fn find_available_port(range_start: u16, range_end: u16) -> u16 {
+        let mut available_port = range_start;
+
```
orhun, line 109: nit: we can remove this unnecessary newline.

```suggestion
```

cargo-shuttle/src/lib.rs
```diff
@@ -104,6 +104,21 @@ impl Shuttle {
         self.run(args, provided_path_to_init).await
     }
 
+    fn find_available_port(range_start: u16, range_end: u16) -> u16 {
```
orhun, line 107: Can you add unit tests for this function?

orhun (commented):
Overall looks good to me, just had some comments. Let's see what other maintainers think.

cargo-shuttle/src/lib.rs
BadgerBloke: sure

cargo-shuttle/src/lib.rs
BadgerBloke: updated the logic and incorporated @jonaro00's suggestion too.

BadgerBloke: to increase the base port by 10

cargo-shuttle/src/lib.rs
```diff
@@ -104,6 +104,29 @@ impl Shuttle {
...
+        if !Confirm::with_theme(&ColorfulTheme::default())
+            .with_prompt(format!(
+                "Default port {} is already in use. Would you like to continue on port {}?",
+                run_args.port, available_port
+            ))
+            .default(true)
+            .interact()
+            .unwrap()
+        {
+            exit(1);
```
jonaro00, line 125: Can be a 0 exit code, since the program did what the user wanted (exit early).

cargo-shuttle/src/lib.rs
BadgerBloke: It seems the tests of `local_run` also need to be updated!

cargo-shuttle/src/lib.rs
```diff
@@ -950,6 +973,11 @@ impl Shuttle {
             Child,
             RuntimeClient<ClaimService<InjectPropagation<Channel>>>,
         )> = Vec::new();
+
+        if !portpicker::is_free_tcp(run_args.port) {
+            Shuttle::find_available_port(&mut run_args);
```
jonaro00, line 978: I thought we might need to send in `services.len()` so that it can check the same number of adjacent port numbers as there are services, but I want to double check on those plans. Let me get back on Tuesday with an answer.

cargo-shuttle/src/lib.rs
```diff
@@ -104,6 +104,29 @@ impl Shuttle {
...
+        for port in (available_port..=std::u16::MAX).step_by(10) {
+            if portpicker::is_free_tcp(port) {
+                available_port = port;
+                break;
+            }
+        }
+
+        if !Confirm::with_theme(&ColorfulTheme::default())
+            .with_prompt(format!(
+                "Default port {} is already in use. Would you like to continue on port {}?",
```
jonaro00, line 118:
```suggestion
                "Port {} is already in use. Would you like to continue on port {}?",
```

jonaro00 (commented):
Looking pretty good now!
Writing a test for this is probably a bit cumbersome. I want to do some manual testing when it's ready anyways.
In the mean time, you could take a look at the task of timing out the confirmation if there is no response within, say, 10 seconds. In that case, we could exit(1) actually.

BadgerBloke:
> Looking pretty good now!

👍🏻

> Writing a test for this is probably a bit cumbersome. I want to do some manual testing when it's ready anyways.

👍🏻

> In the mean time, you could take a look at the task of timing out the confirmation if there is no response within, say, 10 seconds. In that case, we could exit(1) actually.

Sure, but instead of exit(1) can we continue with suggested port?

jonaro00:
Yeah, we can do that. We'll see what others think as well.

cargo-shuttle/src/lib.rs
BadgerBloke: There is one use case which I encountered, if a port is occupied by another service like NodeJS and if we provide the same port to the **shuttle** then it will start properly but we will get only NodeJS service on that port.

BadgerBloke: But let me check this once again

cargo-shuttle/src/lib.rs
BadgerBloke: may I mark this resolved now?

oddgrd (approved):
LGTM, thanks!

BadgerBloke:
> LGTM, thanks!

Thanks 😊, I'm trying for the timeout feature but that doesn't seems easy one, I'll be able to invest more time on weekend

jonaro00:
I double checked, and we do need to send in `services.len()` so that it can check N adjacent port numbers.
So, if there is only one service, we do like we do now. If there are two, we need to check base and base+1. If any of those were taken, check base+10 and base+11 etc. Do you have time to work on this this week @BadgerBloke? I can jump in and help if needed.

BadgerBloke:
Sure, I'll work and complete this port checking at least.

jonaro00:
The timeout is not needed to complete this PR. I might take a look at it if I have time.

jonaro00:
@BadgerBloke How is progress on this? We aim to get this into a release this week, so I might hop in and finish any remaining work unless you are already on it ;)

BadgerBloke:
Yes @jonaro00 I'm on it and push the changes by today itself.

BadgerBloke:
@jonaro00 I think just by adding `&& portpicker::is_free_tcp(port + 1)` inside for loop we will be able to achieve that. Do we still have to use **services.len()**.

Please let me know if I'm still missing something

```rust
if portpicker::is_free_tcp(port) && portpicker::is_free_tcp(port + 1) {
    available_port = port;
    break;
}
```

jonaro00:
Checking port+1 would not be enough if N > 2.
Inside the loop that takes steps of 10, there needs to be an inner loop that checks if all N adjacent ports starting from 8000/8010/etc are free.

So, if there are 3 services, and port 8000 is taken, it checks ports 8010, 8011, and 8012 in the next loop iteration. 

There can be issues with services not stopping and remaining open on port 8000. This is one of the reasons we are adding this feature.

BadgerBloke:
Ok, i think now I got the point, let me make the adjustment

cargo-shuttle/src/lib.rs
```diff
@@ -112,12 +112,21 @@ impl Shuttle {
         self.run(args, provided_path_to_init).await
     }
 
-    fn find_available_port(run_args: &mut RunArgs) {
+    fn find_available_port(run_args: &mut RunArgs, services_len: usize) {
         let mut available_port = 8010;
```
jonaro00, line 116: We need to start the search from `run_args.port`, we can't assume its value is 8000.

cargo-shuttle/src/lib.rs
```diff
@@ -112,12 +112,21 @@ impl Shuttle {
...
-        for port in (available_port..=std::u16::MAX).step_by(10) {
-            if portpicker::is_free_tcp(port) && portpicker::is_free_tcp(port + 1) {
-                available_port = port;
-                break;
+        for port in (available_port..=std::u16::MAX).step_by(services_len.max(10)) {
+            if portpicker::is_free_tcp(port) {
```
jonaro00, line 118: This `port` is already checked in the inner loop, so we shouldn't need the if-clause

cargo-shuttle/src/lib.rs
```diff
@@ -112,12 +112,21 @@ impl Shuttle {
...
+                let mut is_inner_ports_free = true;
+                for inner_port in port..(port + services_len as u16) {
+                    if !portpicker::is_free_tcp(inner_port) {
+                        is_inner_ports_free = false;
```
jonaro00, line 122: Tip: You can use a [loop label](https://doc.rust-lang.org/reference/expressions/loop-expr.html#loop-labels) to `continue` the outer loop instead of using a boolean flag.

cargo-shuttle/src/lib.rs
```diff
@@ -1042,7 +1051,7 @@ impl Shuttle {
         )> = Vec::new();
 
         if !portpicker::is_free_tcp(run_args.port) {
-            Shuttle::find_available_port(&mut run_args);
+            Shuttle::find_available_port(&mut run_args, services.len());
         }
```
jonaro00, line 1055: We should let the function do all the port checking, since it needs to check multiple ports.

jonaro00 (commented):
Some comments

BadgerBloke:
> Some comments

Thanks, let me fix that

cargo-shuttle/src/lib.rs
BadgerBloke: I am using `is_inner_ports_free` to be sure that all the inner loop port is free. please let me know if there is a way around

BadgerBloke: got it

jonaro00 (approved):
Nice job!
I tested it locally, and while it is not bulletproof yet, I think this is good for merging now.

BadgerBloke:
> Nice job!
> I tested it locally, and while it is not bulletproof yet, I think this is good for merging now.

Thanks 😊

I've noted the timeout feature, please let me know other points I'll look into that

jonaro00:
We might not need it. For example, we could just make it take the new ports without asking, but print a warning about it instead.
Which one would you prefer as a user?
I will also ask what others think.

BadgerBloke:
It sounds really good to me. There isn't much difference from UX point of view. In case do not want to run on suggested port then I can stop the started service. So, it's completely fine for me.

Just a rough comparision:
| Ability | With Timeout ⏳ | With Warning Print ⚠️ |
| ------------ | ---------------- | --------------------- |
| Continue on Suggested port | After timeout | Immediately |
| Cancel the process on suggested port | Immediately  | Hard close after starting `ctrl + c` |
| Start on custom specific port | after cancelling (immediately)* | after stopping the started service* |

**provided we supports `-p custom_port` or `--port custom_port` arg*

oddgrd (approved):
Looks good to me! I'm curious, why do we need a timeout here?

jonaro00:
This was the idea behind it. Not strictly necessary ofc. https://github.com/shuttle-hq/shuttle/pull/1270#pullrequestreview-1641110064

BadgerBloke:
Thank you so much guys 😍
It was a great learning for me

PR #1272, opened 2023-09-25
chore(docker): set up a local shared postgres for development

orhun:
## Description of change

This PR updates Makefile/`docker-compose` file for creating a local/shared Postgres database called "control DB" for development.

I also realized we already have a shared Postgres at `/dev/tcp/postgres/5432` which begs the question do we still need this control DB?

N/a


docker-compose.dev.yml
```diff
@@ -30,6 +30,22 @@ services:
   resource-recorder:
     ports:
       - 8007:8000
+    depends_on:
+      - auth
```
oddgrd, line 34: Nit: The .dev.yml file will be combined with the docker-compose.yml into docker-compose.rendered.yml for local development, so we don't need to specify the auth dependency here. I see we did the same thing for logger in this file.

oddgrd (approved):
Nice! Just one nit.

docker-compose.dev.yml
```diff
@@ -30,6 +30,22 @@ services:
   resource-recorder:
```
iulianbarbu, line 30: I think we can add a similar entry for the `auth` service too, with its dependency on the `control-db`.

iulianbarbu (approved):

docker-compose.dev.yml
orhun: 9efbed5845cb18cdfd3071a203abf99a5549ceda

docker-compose.dev.yml
orhun: 5862f38af4b0c5b5fd28b6d8c51adc3025b59134

Makefile
```diff
@@ -73,7 +73,12 @@ DEV_SUFFIX=-dev
 DEPLOYS_API_KEY?=gateway4deployes
 
 # this should use the same version as our prod RDS database
-LOGGER_POSTGRES_TAG?=15
+CONTROL_DB_POSTGRES_TAG?=15
+CONTROL_DB_POSTGRES_PASSWORD?=postgres
+CONTROL_DB_POSTGRES_URI?=postgres://postgres:${CONTROL_DB_POSTGRES_PASSWORD}@control-postgres:5432/postgres
+
+# use the same version as control DB
+LOGGER_POSTGRES_TAG?=${CONTROL_DB_POSTGRES_TAG}
```
oddgrd, line 81: While it's highly likely, this shouldn't necessarily be the same version as the control DB.
```suggestion
# this should use the same version as our prod RDS logger database
LOGGER_POSTGRES_TAG?=15
```

PR #1273, opened 2023-09-25
ci: separate ci and unstable jobs, better caching

jonaro00:
## Description of change

image

- unstable is no longer part of the ci job
  - makes approving/cancelling/rerunning unstable job much cleaner
- less filters for jobs that already depend on filters
- faster

Works with and without the filter commented

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -248,7 +248,7 @@ jobs:
   test-workspace-member:
     parameters:
       crate:
-        description: "Crate in workspace to test"
+        description: Crate in workspace to test
         type: string
     executor: docker-rust
     steps:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -55,13 +64,23 @@ commands:
...
-            - "~/.cache/sccache"
+            - ~/.cache/sccache
+  # This is only performed by the workspace clippy job, since it does the largest cargo fetch.
+  # Other jobs might restore a slightly older copy of this cache.
+  save-cargo-cache:
+    steps:
+      # Discard crates.io patches so that it produces a re-usable cache key checksum.
+      - run:
+          name: Restore Cargo.lock
+          command: git restore Cargo.lock
```
oddgrd, line 75: Will this cause failures when we have unreleased versions of our crates running in CI?

oddgrd (approved):
LGTM, just left one question.

.circleci/config.yml
jonaro00: This is only for ensuring proper caching in CI. Should never disturb.

PR #1275, opened 2023-09-26
feat: version checks between cli, gateway, deployer, runtime

### LINKED ISSUE #1110: [Improvement]: Version checks along the entire deployment chain
### Describe the improvement

From #1053. The solution can build upon #1088.

It would be nice if the versions of
- cargo-shuttle
- the API (think shuttle-gateway)
- (upcoming) builder
- shuttle-deployer
- shuttle-runtime of the user's project

were all logged and compatibility-checked throughout a deployment, with warnings for incompatible versions.

### Duplicate declaration

### LINKED ISSUE #1131: [Feature]: Could cargo-shuttle auto-check if it's out of date?
### Describe the feature

Right now, it's kinda easy to miss a new release of the shuttle runtime (especially since they're coming hot and fast 🔥 ).

Would the team merge a PR that (somehow) checks the latest version of `cargo-shuttle` and if it's greater than the current `cargo-shuttle` version, recommends an update? We'd probably only want to do this when a network call is already being made.

So if the user types in a command that makes a network call (which would be most of them) `cargo-shuttle` would attempt to complete that command and then print out a message saying something like "Run this command to update shuttle":

```
$ cargo shuttle status

Service Name:  [******]
Deployment ID: [******]
Status:        running
Last Updated:  [******]
URI:           [******]

Run this command to update `cargo-shuttle` from v0.20.0 to v0.22.0:
cargo binstall cargo-shuttle
Or set SHUTTLE_NO_UPDATE_CHECK to disable these checks in the future.
```

Quite a few tools have something like this. Pip comes to mind:

Screenshot 2023-08-04 at 18 56 35

But I think [`atuin`](https://atuin.sh/docs/config#update_check) also has something similar.

I've also been bitten before when the runtime got an update, and all my deployment attempts failed with weird errors until I realized I needed to update.

I'm happy to work on the PR if the team would be interested in this.

### Duplicate declaration

jonaro00:
## Description of change

Closes #1110, closes #1131 

- API has endpoint with the latest supported versions
- cargo-shuttle warns about outdated or too new versions of self, deployer, and runtime.
  - it knows the version of self,
  - versions of deployer and runtime are parsed from the log stream and warned about if old ones are detected.
  - This is slightly different from the plan, where deployer would check the runtime version of the incoming project on deploy, but warnings in the middle of deployment logs will be hard to notice. Therefore, cargo-shuttle handles them and prints them at the end.
    - main disadvantage: the runtime warning will not reach user in some cases (i.e. 0.26 -> 0.27 changed runtime args, which crashes it too early)
- Refactoring of cargo-shuttle commands to be more flexible.

All warnings have been tested by hardcoding older/newer versions in various places and verifying them.


cargo-shuttle/src/client.rs
```diff
@@ -19,20 +19,42 @@ use uuid::Uuid;
...
+        let url = format!("{}/versions", self.api_url);
+
+        self.client
+            .get(url)
+            .send()
+            .await?
+            .json()
+            .await
+            .context("parsing API version info")
+    }
```
iulianbarbu, line 56: Should we drop the client from the struct and use the `retry_client` instead? Looks like this is the only place where we use it, but correct me if I am wrong.

cargo-shuttle/src/lib.rs
```diff
@@ -153,44 +179,81 @@ impl Shuttle {
...
+                            Warning:
+                                A newer version of cargo-shuttle exists ({latest_version}).
+                                It is recommended to upgrade.
+                                Refer to the upgrading docs: https://docs.shuttle.rs/configuration/shuttle-versions#upgrading-shuttle-version"
+                        }
+                    ).or_else(||
+                        Some(formatdoc! {"
+                            Warning:
+                                Your version of cargo-shuttle ({my_version}) is newer than what the API expects ({latest_version}).
+                                Things might not work until the API is updated."
```
iulianbarbu, line 244:
```suggestion
                                Unexpected behavior can occur and this indicates a maintenance operation is underway. 
                                Please contact us on the Shuttle discord server if you want to find more."
```

iulianbarbu: I also think about being clearer here and blocking any command while the API is updated, but the message above can also serve as an intermediate step.

cargo-shuttle/src/lib.rs
```diff
@@ -1639,8 +1768,7 @@ impl Shuttle {
 }
 
 fn check_version(runtime_path: &Path) -> Result<()> {
-    let valid_version = semver::Version::from_str(VERSION)
-        .context("failed to convert cargo-shuttle version to semver")?;
+    let my_version = semver::Version::from_str(VERSION).unwrap();
```
iulianbarbu, line 1771: Why drop the `.context(` call? I think it is better than failing with a panic.

gateway/src/api/latest.rs
```diff
@@ -851,6 +851,19 @@ impl ApiBuilder {
...
+            .route(
+                "/versions",
+                get(|| async {
+                    axum::Json(VersionInfo {
+                        gateway: env!("CARGO_PKG_VERSION").parse().unwrap(),
+                        // For now, these use the same version as gateway.
+                        // Only one version is officially compatible, but more are in reality.
+                        cargo_shuttle: env!("CARGO_PKG_VERSION").parse().unwrap(),
+                        deployer: env!("CARGO_PKG_VERSION").parse().unwrap(),
+                        runtime: env!("CARGO_PKG_VERSION").parse().unwrap(),
```
iulianbarbu, line 863: This is not accurate if we don't release in lockstep. Maybe we can add a comment also about releasing in lockstep to link the compatible versions more easily?

cargo-shuttle/src/client.rs
jonaro00: I don't use retry client here because it could cause a 10 second wait at command invocation, before anything happens. I.e. this check should fail fast.

cargo-shuttle/src/lib.rs
jonaro00: I don't think we should use version mismatches as reasons to deny requests yet.

cargo-shuttle/src/lib.rs
jonaro00: the constant is always a valid semver.

cargo-shuttle/src/client.rs
iulianbarbu: Got it, thanks!

iulianbarbu (approved):
LGTM! Thanks!

PR #1276, opened 2023-09-26
feat(deployer): rate limit runtime stdout consumption

iulianbarbu:
## Description of change

Rate limits the stdout consumption of the runtimes. This prevents the issue of the slow receiver on the deployer side, which can result in a deployer being killed because of OOM. We have a 100 max log lines / s refill rate for each project stdout logging consumption, meaning we can consume from stdout <= 100 max log lines / s. This rate can be adjusted but this should provide plenty of space for reasonable projects.

In case of projects that need more, they will be IO-bound in case of logging from the request handlers or instrumenting the handlers.

Tested locally by starting the deployer with a project that logs inside a background task without any sleep. The deployer crashed after a few minutes. After adding the rate limiter, this doesn't happen anymore.


orhun (approved):

oddgrd:
In your description, you say this crashes the deployer after a few minutes, will it be clear to the user that it crashed because it was writing too much to stdout?

jonaro00:
Would be good if we injected log line "Dropping log lines. Too many lines per second." when the limit is reached (without spamming it ofc)

iulianbarbu:
> In your description, you say this crashes the deployer after a few minutes, will it be clear to the user that it crashed because it was writing too much to stdout?

I miswrote what the testing focused on. The crash happens before adding the rate limiter. Adding the rate limiter will not make it crash.

iulianbarbu:
> Would be good if we injected log line "Dropping log lines. Too many lines per second." when the limit is reached (without spamming it ofc)

Nice, lemme see what I can do, will ping

oddgrd:
I think we'd need to test this further before we merge it, and ideally we would base the limit on usage data. The first step here should be to gather data on the volume of logs our current users send to the logger, and if that is not possible/simple, we should first focus on setting up the required instrumentation.

We could merge this with a high limit to prevent obviously malicious uses, but we'd need to be clear on how this fails first, and how it represents the failure to the users.

oddgrd:
I think we can close this, but we may return to it at some point in the future, we'll see!

PR #1278, opened 2023-09-26
fix: cleanup for 0.28.0

jonaro00:


resources/static-folder/src/lib.rs
```diff
@@ -5,6 +5,10 @@ use serde::Serialize;
 use shuttle_service::{Error, Factory, ResourceBuilder, Type};
 
 #[derive(Serialize)]
+#[deprecated(
+    since = "0.27.0",
+    note = "Folder names can now be hard coded. https://docs.shuttle.rs/configuration/files"
```
iulianbarbu, line 10: nit: 
```suggestion
    note = "Folder names can now be hard coded: https://docs.shuttle.rs/configuration/files".
```

iulianbarbu (approved):

orhun (approved):

jonaro00:
Waiting with merge until release is closer

resources/static-folder/Cargo.toml
```diff
@@ -9,4 +10,4 @@ keywords = ["shuttle-service", "static-folder"]
 [dependencies]
 async-trait = "0.1.56"
 serde = { version = "1.0.148", features = ["derive"] }
-shuttle-service = { path = "../../service", version = "0.27.0" }
+shuttle-service = { path = "../../service", version = ">=0.27.0, <0.31.0" }
```
oddgrd, line 13: What is this change for?

oddgrd (approved):
LGTM! We're certain that nothing depends on openssl in cargo-shuttle now, right? We could run the QA jobs for mac and windows on this PR to ensure it works, but it's probably overkill.

resources/static-folder/Cargo.toml
jonaro00: This way, the final release 0.28 of static folder will stay compatible (but with deprecation compile warning) until 0.32 OR resource breaking change, whichever happen first.

PR #1279, opened 2023-09-27
feat(cargo-shuttle): check project name available

jonaro00:
## Description of change

image
image

What still remains:
- *Should we allow user to use the name even though it is taken? Prompt y/n for the rename?*
  - It's a bit tricky and not a killer. Using `--name` gets you around this check.

Bonus:
image

All 3 scenarios tested locally
Updated tests

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -297,12 +297,39 @@ impl Shuttle {
                 It will be hosted at ${{project_name}}.shuttleapp.rs, so choose something unique!
                 "
             );
-            // TODO: Check whether the project name is still available
-            project_args.name = Some(
-                Input::with_theme(&theme)
+            loop {
+                // not using validate_with due to being blocking
+                let p: ProjectName = Input::with_theme(&theme)
                     .with_prompt("Project name")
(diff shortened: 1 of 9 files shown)


chesedo (approved):
LGTM

cargo-shuttle/src/lib.rs
```diff
@@ -392,9 +416,21 @@ impl Shuttle {
...
+                        .expect("to have a project name provided")
+                ))
                 .default(true)
                 .interact()?;
+            if !should_create {
+                println!(
+                    "Note: The project name will not be claimed until \
+                    you start the project with `cargo shuttle project start`."
+                )
+            }
```
oddgrd, line 433: Will it be clear by this point that the name is globally unique?

oddgrd (approved):
LGTM! Just one nit/question, but feel free to merge if you think it's clear as is.

cargo-shuttle/src/lib.rs
jonaro00: The prompt at name selection does tell you that. The note here is to clarify for those that do not claim the name. The word "claim" also helps.

PR #1280, opened 2023-09-27
feat(installer): add installer script

orhun:
## Description of change

This PR adds the installer script for easily installing `cargo-shuttle` on different platforms.

Currently, it supports:

- Native installation on Arch Linux (and derivaties) via `pacman`
- Installing via `cargo-binstall` (if `cargo-binstall` is installed)
- Installing pre-built binaries from GitHub (only Linux/MacOS for now)
- Installing directly via `cargo install` (installs `cargo` via `rustup` if it is not found)

On my local machine by running `./install.sh`

We need to test this on MacOS as well.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -101,7 +101,7 @@ _install_cargo() {
     echo "rustup installed! Attempting cargo install"
   fi
 
-  cargo install cargo-shuttle
+  cargo install --locked cargo-shuttle
 }
 
 _install_unsupported() {
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,152 @@
+#! /usr/bin/env bash
+
+set -euo pipefail
+
+cat <<EOF
+     _           _   _   _      
+ ___| |__  _   _| |_| |_| | ___ 
+/ __| '_ \\| | | | __| __| |/ _ \\
+\__ \\ | | | |_| | |_| |_| |  __/
+|___/_| |_|\\__,_|\\__|\\__|_|\\___|
+                                
+https://www.shuttle.rs
+https://github.com/shuttle-hq/shuttle
+
+Please file an issue if you encounter any problems!
+===================================================
+EOF
+
+if ! command -v curl &>/dev/null; then
+  echo "curl not installed. Please install curl."
+  exit
+fi
+
+REPO_URL="https://github.com/shuttle-hq/shuttle"
+LATEST_RELEASE=$(curl -L -s -H 'Accept: application/json' "$REPO_URL/releases/latest")
+LATEST_VERSION="${LATEST_RELEASE#*\"tag_name\":\"}"
+LATEST_VERSION="${LATEST_VERSION%%\"*}"
+
+_install_linux() {
+  echo "Detected Linux!"
+  echo "Checking distro..."
+  if (uname -a | grep -qi "Microsoft"); then
+    OS="ubuntuwsl"
+  elif ! command -v lsb_release &>/dev/null; then
+    echo "lsb_release could not be found. Falling back to /etc/os-release"
+    OS="$(grep -Po '(?<=^ID=).*$' /etc/os-release | tr '[:upper:]' '[:lower:]')" 2>/dev/null
+  else
+    OS=$(lsb_release -i | awk '{ print $3 }' | tr '[:upper:]' '[:lower:]')
+  fi
+  case "$OS" in
+  "arch" | "manjarolinux" | "endeavouros")
+    _install_arch_linux
+    ;;
+  "ubuntu" | "ubuntuwsl" | "debian" | "linuxmint" | "parrot" | "kali" | "elementary" | "pop")
+    # TODO: distribute .deb packages via `cargo-deb` and install them here
+    _install_unsupported
+    ;;
+  *)
+    _install_unsupported
+    ;;
+  esac
+}
+
+_install_arch_linux() {
+  echo "Arch Linux detected!"
+  if command -v pacman &>/dev/null; then
+    echo "Installing with pacman"
+    sudo pacman -S cargo-shuttle
+  else
+    echo "Pacman not found"
+    exit 1
+  fi
+}
+
+# TODO: package cargo-shuttle for Homebrew
+_install_mac() {
+  _install_unsupported
+}
+
+_install_binary() {
+  echo "Installing pre-built binary..."
+  case "$OSTYPE" in
+  linux*) target="x86_64-unknown-linux-musl" ;;
+  darwin*) target="x86_64-apple-darwin" ;;
+  *)
+    echo "Cannot determine the target to install"
+    exit 1
+    ;;
+  esac
+  temp_dir=$(mktemp -d)
+  pushd "$temp_dir" >/dev/null || exit 1
+  curl -LO "$REPO_URL/releases/download/$LATEST_VERSION/cargo-shuttle-$LATEST_VERSION-$target.tar.gz"
+  tar -xzf "cargo-shuttle-$LATEST_VERSION-$target.tar.gz"
+  echo "Installing to $HOME/.cargo/bin/cargo-shuttle"
+  mv "cargo-shuttle-$target-$LATEST_VERSION/cargo-shuttle" "$HOME/.cargo/bin/"
+  popd >/dev/null || exit 1
+  if [[ ":$PATH:" != *":$HOME/.cargo/bin:"* ]]; then
+    echo "Add $HOME/.cargo/bin to PATH to run cargo-shuttle"
+  fi
+}
+
+_install_cargo() {
+  echo "Attempting install with cargo"
+  if ! command -v cargo &>/dev/null; then
+    echo "cargo not found! Attempting to install rustup"
+    if command -v rustup &>/dev/null; then
+      echo "rustup was found, but cargo wasn't. Something is up with your install"
+      exit 1
+    fi
+    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -q
+    echo "rustup installed! Attempting cargo install"
+  fi
+
+  cargo install cargo-shuttle
+}
+
+_install_unsupported() {
+  echo "Unknown or unsupported OS or architecture"
+
+  if command -v cargo-binstall &>/dev/null; then
+    echo "Installing with cargo-binstall"
+    cargo binstall cargo-shuttle
+    return 0
+  fi
+
+  while true; do
+    read -r -p "Do you wish to attempt to install the pre-built binary? [Y/N] " yn
+    case $yn in
+    [Yy]*)
+      _install_binary
+      break
+      ;;
+    [Nn]*)
+      read -r -p "Do you wish to attempt an install with 'cargo'? [Y/N] " yn
+      case $yn in
+      [Yy]*)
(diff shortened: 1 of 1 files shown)


orhun:
Yup, it is not really needed. At first I thought we didn't have binary releases but we have so we can simply install them with this script.

install.sh
```diff
@@ -0,0 +1,152 @@
...
+    OS="$(grep -Po '(?<=^ID=).*$' /etc/os-release | tr '[:upper:]' '[:lower:]')" 2>/dev/null
+  else
+    OS=$(lsb_release -i | awk '{ print $3 }' | tr '[:upper:]' '[:lower:]')
+  fi
+  case "$OS" in
+  "arch" | "manjarolinux" | "endeavouros")
+    _install_arch_linux
+    ;;
+  "ubuntu" | "ubuntuwsl" | "debian" | "linuxmint" | "parrot" | "kali" | "elementary" | "pop")
+    # TODO: distribute .deb packages via `cargo-deb` and install them here
```
jonaro00, line 45: Why? We just need to download one binary.

install.sh
```diff
@@ -0,0 +1,152 @@
...
+REPO_URL="https://github.com/shuttle-hq/shuttle"
+LATEST_RELEASE=$(curl -L -s -H 'Accept: application/json' "$REPO_URL/releases/latest")
+LATEST_VERSION="${LATEST_RELEASE#*\"tag_name\":\"}"
+LATEST_VERSION="${LATEST_VERSION%%\"*}"
+
+_install_linux() {
+  echo "Detected Linux!"
+  echo "Checking distro..."
+  if (uname -a | grep -qi "Microsoft"); then
+    OS="ubuntuwsl"
```
jonaro00, line 33: WSL can be other distros than ubuntu, but ubuntu is probably 90%+

install.sh
```diff
@@ -0,0 +1,152 @@
...
+  tar -xzf "cargo-shuttle-$LATEST_VERSION-$target.tar.gz"
+  echo "Installing to $HOME/.cargo/bin/cargo-shuttle"
+  mv "cargo-shuttle-$target-$LATEST_VERSION/cargo-shuttle" "$HOME/.cargo/bin/"
+  popd >/dev/null || exit 1
+  if [[ ":$PATH:" != *":$HOME/.cargo/bin:"* ]]; then
+    echo "Add $HOME/.cargo/bin to PATH to run cargo-shuttle"
+  fi
+}
+
+_install_cargo() {
```
jonaro00, line 92: Do we really want to install Rust for them?

install.sh
```diff
@@ -0,0 +1,152 @@
...
+  echo "Attempting install with cargo"
+  if ! command -v cargo &>/dev/null; then
+    echo "cargo not found! Attempting to install rustup"
```
jonaro00, line 95: ...install Rust and cargo via rustup

install.sh
```diff
@@ -0,0 +1,152 @@
...
+    if command -v rustup &>/dev/null; then
+      echo "rustup was found, but cargo wasn't. Something is up with your install"
+      exit 1
+    fi
+    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -q
+    echo "rustup installed! Attempting cargo install"
```
jonaro00, line 101: You need `source ~/.cargo/bin/env` or similar (rustup prints it) after this to find cargo

install.sh
```diff
@@ -0,0 +1,152 @@
...
+
+  cargo install cargo-shuttle
+}
+
+_install_unsupported() {
+  echo "Unknown or unsupported OS or architecture"
+
+  if command -v cargo-binstall &>/dev/null; then
+    echo "Installing with cargo-binstall"
+    cargo binstall cargo-shuttle
```
jonaro00, line 112:
```suggestion
    cargo binstall -y --locked cargo-shuttle
```

install.sh
```diff
@@ -0,0 +1,152 @@
...
+    fi
+    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -q
+    echo "rustup installed! Attempting cargo install"
+  fi
+
+  cargo install cargo-shuttle
+}
+
+_install_unsupported() {
+  echo "Unknown or unsupported OS or architecture"
```
jonaro00, line 108: I wouldn't call these platforms unsupported... maybe "installing with package manager not supported".
Btw, going via package manager might get you an outdated version, so this route is the "most" reliable.

install.sh
```diff
@@ -0,0 +1,152 @@
...
+
+_install_cargo() {
+  echo "Attempting install with cargo"
+  if ! command -v cargo &>/dev/null; then
+    echo "cargo not found! Attempting to install rustup"
+    if command -v rustup &>/dev/null; then
+      echo "rustup was found, but cargo wasn't. Something is up with your install"
+      exit 1
+    fi
+    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -q
```
jonaro00, line 100: Do we want `-q`? Don't want to obfuscate it too much.

install.sh
```diff
@@ -0,0 +1,152 @@
...
+# TODO: package cargo-shuttle for Homebrew
+_install_mac() {
+  _install_unsupported
+}
+
+_install_binary() {
+  echo "Installing pre-built binary..."
+  case "$OSTYPE" in
+  linux*) target="x86_64-unknown-linux-musl" ;;
+  darwin*) target="x86_64-apple-darwin" ;;
```
jonaro00, line 74: There is an `aarch64` linux binary. Just me or is there no binary for M1/M2? I thought we built binaries cause they were struggling.

install.sh
jonaro00: Imagine on a new release: "Please run the install script to upgrade!" -> package manager not updated -> wow

install.sh
```diff
@@ -0,0 +1,152 @@
...
+
+if ! command -v curl &>/dev/null; then
+  echo "curl not installed. Please install curl."
+  exit
+fi
+
+REPO_URL="https://github.com/shuttle-hq/shuttle"
+LATEST_RELEASE=$(curl -L -s -H 'Accept: application/json' "$REPO_URL/releases/latest")
+LATEST_VERSION="${LATEST_RELEASE#*\"tag_name\":\"}"
+LATEST_VERSION="${LATEST_VERSION%%\"*}"
```
jonaro00, line 27: Hmm, sometimes we release a patch version of cargo-shuttle without bumping everything... Need to think this through for #1275 too.

install.sh
```diff
@@ -0,0 +1,152 @@
...
+
+Please file an issue if you encounter any problems!
+===================================================
+EOF
+
+if ! command -v curl &>/dev/null; then
+  echo "curl not installed. Please install curl."
+  exit
+fi
+
```
jonaro00, line 23: Check if `cargo-shuttle` is already install and if it is already the latest version.

jonaro00 (changes requested):
We also need a redirect from `shuttle.rs/install` to this file on main :) I can fix that if you want.

install.sh
```diff
@@ -0,0 +1,152 @@
...
+    echo "cargo not found! Attempting to install rustup"
+    if command -v rustup &>/dev/null; then
+      echo "rustup was found, but cargo wasn't. Something is up with your install"
+      exit 1
+    fi
+    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -q
+    echo "rustup installed! Attempting cargo install"
+  fi
+
+  cargo install cargo-shuttle
```
jonaro00, line 104:
```suggestion
  cargo install --locked cargo-shuttle
```

install.sh
orhun: Let's stick to what distributions use to keep things native.

install.sh
orhun: Why not?

install.sh
jonaro00: Ah, it is considered for in #1275. But here it can be trickier.

install.sh
orhun: 48bba2485df1a41a37765d85cdbeeb7c57bd5640

install.sh
jonaro00: I can see if some prefer such install scripts to be well-contained, aka "install shuttle script" does nothing else than installing shuttle. Telling ppl to go to rustup.rs in case of missing rust is a valid alternative. At least we should prompt for it, like we do for the binary/cargo installs.

install.sh
orhun: 477aeb3cd83e0f988cecaf0c1e6123d017a8a1da

install.sh
orhun: > Btw, going via package manager might get you an outdated version, so this route is the "most" reliable.

Although it is true that you might get an outdated version, I disagree with the reliability statement. Using package managers are usually more reliable than installing pre-built binaries since the packages are built specifically for that distro (and sometimes even have necessary fix/security patches for that package to work).

> maybe "installing with package manager not supported".

eff50729efb869548708e96712804395e349896c

install.sh
orhun: 452c1c22f70b49d23aaa1a021d004aaa14b89471

install.sh
orhun: 810975d64e8fc072caca27668741147a041473e1

install.sh
orhun: a27d05534805febfab302fdcac76f25f9b79d4bb

install.sh
jonaro00: Agree with your point, but since we are lowkey requiring the latest version to be installed (for now), the scenario "don't work cause outdated" feels more likely than "don't work cause not installed properly with package manager" in Shuttle's case.
Maybe we can introduce a check+prompt "latest version not yet available through <package manager>, install with binary/cargo instead?". My main concern is
> Imagine on a new release: "Please run the install script to upgrade!" -> package manager not updated -> wow

install.sh
orhun: d48ae8a74f81700f0baa742cd2954847a86eaaa0

install.sh
orhun: d48ae8a74f81700f0baa742cd2954847a86eaaa0

install.sh
jonaro00: We are adding a warning if pacman does not have latest version yet.

install.sh
oddgrd: We didn't implement a binary build for M1/M2 yet, because they can use the x86_64 binary, and it was a bit tricky to do at the time with CircleCI. However, we now use an M1 machine to build the x86_64 binary, so adding an m1 binary should be quite simple, essentially just copying the current mac build, removing the rosetta install step and changing the target.

jonaro00: Or do both in one go to save on that juicy mac CI cost :+1:

orhun:
ready for re-review

we also need to test this on MacOS before merge.

jonaro00 (approved):

oddgrd:
We should also update the readme in this repo with the fact that you can install with this script now, and we should add it to the installation guide in the docs repo. We should be clear about the fact that it doesn't work on windows yet.

jonaro00:
@oddgrd I can make those changes. (separate PRs)

PR #1281, opened 2023-09-28
refactor(cargo-shuttle): remove `cargo-generate` dependency

### LINKED ISSUE #1269: Refactor to lighter implementation of cargo-generate cloning
### What?
`cargo-shuttle` uses the library part of `cargo-generate` for cloning repos in `cargo shuttle init`.
Get rid of `cargo-generate` and re-implement its core functionality.

### Why?
- Reduce dependency tree weight from cargo-generate, which has more features than we need.
  - It also panics sometimes
  - It is not written to be used as a library
- Possibly get rid of openssl/native-tls dependency

### Requirements
- All documented features and options of the `cargo shuttle init` command should remain the same and work the same way. (https://github.com/shuttle-hq/shuttle-examples#readme)
- The solution should mimic `cargo-generate`, which includes, but not limited to:
  - Create a temp dir with `tempdir`
  - Parse the `--from` parameter and determine if the arg is a repo or path:
    - `gh:aaa/bbb` => clone from GitHub user aaa repo bbb, same for `gl:` and `bb:`
    - `aaa/bbb` => same as above
    - `aaa` & `aaa/bbb/ccc` etc => could be seen as relative paths, but should error, to not get confused with the previous bullet
    - `./aaa` => from local path `aaa/`
    - `../../aaa/bbb` => also local path
    - `/aaa/bbb` => local absolute path
    - `C:/aaa/bbb` => local absolute path (windows)
  - Clone/Copy from the source into the temp dir
  - Take only the subfolder param (if provided) and copy all of those files into target folder
  - Do a git init if relevant (same way it does now)
  - Do the final modifications with `set_crate_name` , `edit_shuttle_toml`, `create_gitignore_file`. (basically no change)
#### Bonus requirements (optional)
- Get rid of openssl/native-tls dependency. NOTE: We only need to support cloning over HTTP, which should hopefully simplify this a bit. Investigate the suitable crate for this (`gix`?)
- Update unit/integration tests to cover any missing parts of the `init` API
- Allow passing args or syntax for specifying a branch/tag/commit to clone from. A `--rev` arg??

d4ckard:
## Description of change

`cargo-generate` has been removed as a dependency of `cargo-shuttle`, and all functionality of `cargo-shuttle` that was used by the `cargo shuttle init` command has been re-implemented.

The `openssl` dependency introduced by cloning the template Git repositories using the `git2` crate has also been removed. Instead, `gix` is used to clone repositories, which uses `rustls`.

Closes #1269

/claim #1269

The applicable tests have been updated where required. Additional tests have been added to ensure that the new parsing-logic accepts many relevant examples (e.g. those in [the `shuttle-examples` README](https://github.com/shuttle-hq/shuttle-examples#how-to-clone-run-and-deploy-an-example). Also, some previously ignored tests now pass.

### DIFF
--- a/cargo-shuttle/src/init.rs
+++ b/cargo-shuttle/src/init.rs
@@ -8,7 +8,7 @@ use std::{
 use anyhow::{Context, Result};
 use regex::Regex;
 use shuttle_common::project::ProjectName;
-use tempfile::Builder;
+use tempfile::{Builder, TempDir};
 use toml_edit::{value, Document};
 use url::Url;
 
@@ -29,18 +29,39 @@ pub fn generate_project(
     name: &ProjectName,
     temp_loc: TemplateLocation,
 ) -> Result<()> {
-    let gen = GenerateFrom::from(temp_loc).context("Failed to parse template location")?;
-
     println!(r#"Creating project "{name}" in "{}""#, dest.display());
 
-    gen.generate(&dest, |path: &Path| {
-        // Modifications applied to the content of the template:
-        set_crate_name(path, name.as_str())
-            .with_context(|| "Failed to set crate name. No Cargo.toml in template?")?;
-        edit_shuttle_toml(path).with_context(|| "Failed to edit Shuttle.toml.")?;
-        create_gitignore_file(path).with_context(|| "Failed to create .gitignore file.")?;
-        Ok(())
-    })?;
+    let temp_dir: TempDir = setup_template(&temp_loc.auto_path)
+        .context("Failed to setup template generation directory")?;
+
+    let path = match temp_loc.subfolder {
+        Some(subfolder) => {
+            let path = temp_dir.path().join(&subfolder);
+            if path.exists() {
+                path
+            } else {
+                anyhow::bail!(format!(
+                    r#"There is no sub-folder "{}" in the template found at "{}""#,
+                    subfolder, temp_loc.auto_path
+                ))
+            }
+        }
+        None => temp_dir.path().to_owned(),
+    };
+
+    // Prepare the template by changing its default contents.
+    set_crate_name(&path, name.as_str())
+        .context("Failed to set crate name. No Cargo.toml in template?")?;
+    edit_shuttle_toml(&path).context("Failed to edit Shuttle.toml")?;
+    create_gitignore_file(&path).context("Failed to create .gitignore file")?;
+
+    copy_dirs(&path, &dest, GitDir::Ignore)
+        .context("Failed to copy the prepared template to the destination")?;
+
+    // `close` is kind of optional, but the dtor will ignore potential errors.
+    temp_dir
+        .close()
+        .context("Failed to close the temporary directory that the project was generated into")?;
 
     // Initialize a Git repository in the destination directory if there
     // is no existing Git repository present in the surrounding folders.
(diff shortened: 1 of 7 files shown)


cargo-shuttle/src/init.rs
```diff
@@ -1,56 +1,291 @@
...
+                path.display()
+            ))
+        }
+    }
+}
+
+/// Mimic the behavior of `git clone`, cloning the Git repository found at
+/// `from_url` into a directory `to_path`, using the API exposed by `gix`.
+fn gix_clone(from_url: &str, to_path: &Path) -> Result<()> {
+    let mut fetch = PrepareFetch::new(
```
jonaro00, line 198: If it's possible to clone with depth 1, that could save some time.

cargo-shuttle/src/init.rs
```diff
@@ -108,3 +343,100 @@ fn create_gitignore_file(path: &Path) -> Result<()> {
...
+        ($auto_path:literal, $subfolder:expr) => {
+            GenerateFrom::from(TemplateLocation {
+                auto_path: $auto_path.to_owned(),
+                subfolder: $subfolder,
+            })
+        };
+    }
+
+    #[test]
+    fn parse_valid_template_location() {
```
jonaro00, line 370: We are testing that these do parse, but not if the result is equal to the expected.
Adding that to all of these seems excessive, but adding it for a few of these can do. Good for verifying the path concatenation, for example.

jonaro00 (commented):
Looking great! Clean code.

Other things that can be added:
- Use gix instead of git2 for dirty check
- Allow passing args or syntax for specifying a branch/tag/commit to clone from

Up to you if you want to add them :)

d4ckard:
Yes, I want to fully replace `git2` with `gix`. I am still working on the dirty check.

cargo-shuttle/src/init.rs
```diff
@@ -1,56 +1,291 @@
...
+            anyhow::bail!("template location is invalid")
+        }
+    }
+
+    fn generate<F>(self, dest: &Path, prepare: F) -> Result<()>
+    where
+        F: FnOnce(&Path) -> Result<()>,
+    {
+        let temp_dir =
+            tempdir().context("Failed to create a temporary directory to generate project into")?;
```
jonaro00, line 132: Can you use a prefix, like we do in the integration tests?

cargo-shuttle/src/init.rs
```diff
@@ -1,56 +1,291 @@
...
+                owner: caps["owner"].to_owned(),
+                name: caps["name"].to_owned(),
+                subfolder: loc.subfolder.map(PathBuf::from),
+            })
+        } else if Path::new(&loc.auto_path).is_absolute() || loc.auto_path.starts_with('.') {
+            // Local paths to template locations must be absolute or start with a
+            // dot pattern. This way, conflicts between relative path in the form
+            // `foo/bar` and repository name in the form `owner/name` are avoided.
+            match loc.subfolder {
+                Some(subfolder) => Ok(Self::LocalPath(Path::new(&loc.auto_path).join(subfolder))),
```
jonaro00, line 101: Just a thought: if we are to add branch/rev selection to the clone operation, we would need to require a local path to be handled like the other alternatives, so that the clone and checkout can be done correctly.

d4ckard: Do you mean in cases where there is a Git repository at the local path? Yes, that's a good idea.

The `generate` function ought to be divided into a *setup* which ends in the same generic state, no matter what the initial form of the origin was. This state could be the raw template places in a temporary directory. After the *setup*, the rest of the code would be the same for local folders and repository clones. This code always uses `subfolder`.

I actually prefer this design over the current one. It separates the different stages and the information required at each one of them much better.

cargo-shuttle/src/init.rs
d4ckard: What prefix would you suggest? I'm torn between `shuttle-init`, which is shorter, and `cargo-shuttle-init`, which makes the context more clear.

cargo-shuttle/src/init.rs
jonaro00: Yeah it's a bit more streamlined.
I wonder if the `depth 1` and potential branch/rev parameter clash in some way :thinking:

cargo-shuttle/src/init.rs
```diff
@@ -29,18 +29,39 @@ pub fn generate_project(
...
+    edit_shuttle_toml(&path).context("Failed to edit Shuttle.toml")?;
+    create_gitignore_file(&path).context("Failed to create .gitignore file")?;
+
+    copy_dirs(&path, &dest, GitDir::Ignore)
+        .context("Failed to copy the prepared template to the destination")?;
+
+    // `close` is kind of optional, but the dtor will ignore potential errors.
+    temp_dir
+        .close()
+        .context("Failed to close the temporary directory that the project was generated into")?;
```
jonaro00, line 64: I think we can use dtor and ignore failure to remove tempdir.

cargo-shuttle/src/init.rs
```diff
@@ -248,12 +188,12 @@ fn copy_template(src: &Path, dest: &Path) -> Result<()> {
...
 
         if entry_type.is_dir() {
-            if entry_name == ".git" {
+            if git_policy == GitDir::Ignore && entry_name == ".git" {
                 continue;
             }
 
             // Recursion!
-            copy_template(&entry.path(), &entry_dest)?;
+            copy_dirs(&entry.path(), &entry_dest, git_policy)?;
```
jonaro00, line 196: Recursive copy could go infinite with a symlink? :thinking:

cargo-shuttle/src/init.rs
```diff
@@ -271,26 +211,9 @@ fn copy_template(src: &Path, dest: &Path) -> Result<()> {
...
-            Self::GitLab => "https://gitlab.com/",
-        }
-    }
-}
-
-impl std::fmt::Display for GitVendor {
-    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "{}", self.as_str())
-    }
+enum GitDir {
```
jonaro00, line 214: I think many of the derives are not needed

cargo-shuttle/src/init.rs
d4ckard: Yes, I think `cargo-generate` filtered out symlinks beforehand. I'll add a check.

cargo-shuttle/src/init.rs
d4ckard: `Copy` depends on `Clone`, and `Eq` also needs `PartialEq`. `Debug` is optional, but nice to have.

cargo-shuttle/src/init.rs
d4ckard: `is_dir` and `is_file` don't return true for symlinks (as described [here](https://doc.rust-lang.org/std/fs/struct.FileType.html#method.is_symlink)). A warning would be nice though.

cargo-shuttle/src/init.rs
jonaro00: Ah now I see where Copy is used. But Eq is not needed. Sorry for nitpicking.

d4ckard: No worries : )
`Eq` has no methods, and it's only used to highlight that the comparisons are equivalence relations. The actual implementation is still in `PartialEq`. It's described here: https://doc.rust-lang.org/std/cmp/trait.Eq.html.

d4ckard:
By replacing `git2` with `gix`, the number of crates used has gone up again. `gix` is very modular and divided up into many small crates, so that users can choose which features they need very precisely. Compile times have improved compared to the previous implementation that uses `cargo-generate` and `git2` with `openssl`. At the current state of this PR, `cargo-shuttle` takes an average 76.28 s to compile. In comparison, the version of `cargo-shuttle` that's currently in use compiles in an average of 103.37 s (both benchmarks were conducted using `hyperfine`. Each run was prepared by running `cargo clean` first).

jonaro00:
Splendid! I'll check out the changes later. Is this PR "done" now?

d4ckard:
I'm still figuring out how a `--rev` flag would best be implemented using `gix`. Since adding the `--rev` flag as a whole also requires changes all over `cargo-shuttle`, I suggest to put it in another issue/PR.

So yes, besides the `--rev` flag, this PR should now  close everything discussed in #1269 :grinning:

jonaro00 (commented):
Wow, gix was a bit tricky it seems.

I tried a local deploy, but the commit ID was malformed. Maybe the hash you get from gix is the binary data, not encoded to a hex string?

image

Also, the dirty flag did not turn true when having untracked files or staged files, only when having a removed (changed) file. We don't have strict requirements for these, but the former two seem like reasonable triggers, since they cause non-commited files to get deployed.

-----

On my 20 thread CPU.
- `cargo b -r -p cargo-shuttle`
- `cargo clean` between every build

test | main | pr 1281
-- | -- | --
first test | 58 s | 54 s (should have been less due to partially missing crate downloads and sccache)
hot sccache (wow!!!) | 15 s | 13 s
registry cached, no sccache | 60 s | 56 s (ignore the above I guess)

Haha I guess it don't matter much on such a powerful CPU. Not sure how fair my tests were either. The improvement will be bigger for weaker CPUs tho :D

d4ckard:
What are the specific commands you used to get the image you linked? I'd like to know, so I can debug the issue. I'll also take a look at how to determine the strictness of the dirty check.

d4ckard:
I just found the tracking issue on `gix status` while searching for documentation on how to use `gix` to also set the `dirty` if changes are staged but not committed or untracked. It says that determining changes of staged and untracked files is not yet implemented :fearful:.

I guess a reasonable way to go would be to remove my last commit and keep using `git2` for now. This should also remove the other `gix`-related issues mentioned by @jonaro00 in their last comment.

jonaro00:
Oh no :(
Yeah, let's keep the changes well isolated and wait with using gix for that part.

If you want, you can move the last commit to a new draft PR, so that the work done can be reused if gix becomes more feature complete.

jonaro00 (approved):
:partying_face:

orhun (approved):

PR #1284, opened 2023-09-30
fix: gateway container startup on apple m1

AlexCzar:
## Description of change
Allows shuttle to start in docker on M1 Pro MacBooks, there might be additional change needed for other Apple CPUs, I have no way to test.
This change achieves this by adding the `TARGETPLATFORM` value used by Docker on the mentioned MacBook to the list of platforms for which an arm64 version of the `ulid0.so` should be used.

Before this change gateway container wouldn't start on an M1 Pro MacBook during the `make up` step of the [running-locally](https://github.com/shuttle-hq/shuttle/blob/main/DEVELOPING.md#running-locally) guide, now it does.
I was able to successfully deploy an example locally after the change.


jonaro00 (approved):
Huh, it surprises me that M1 is "linux" arm v8.
Maybe we end up just checking just the TARGET variable, but we will have to see how this works for other archs.

AlexCzar:
The failed check is rustup timing out on rust install in CI. Doesn't look related to my changes at all.

Issue #1000, opened 2023-06-13
Allow using a different linker or allow the compiled binary to be used independently instead of `cargo shuttle run`


that-ambuj:
### Describe the improvement

Hi, In my rust projects, I use a different linker like [`lld`](https://lld.llvm.org/) or [`mold`](https://github.com/rui314/mold) for faster compile times because I use a budget laptop for coding and rust is notorious for it's slow default compile times.

Lately, I have started using https://shuttle.rs for personal projects using their cli tool `cargo-shuttle`. The issue I'm facing here that when running `cargo run` or `cargo build`, cargo uses my `.cargo/config.toml` file to select a linker and that works out just fine and I get great compile times but when I use `cargo shuttle run` to run the shuttle project locally, the project gets rebuilt using the slow, default `gold` linker(this is just an assumption) and it takes ~15s to rebuild after every change(I use `cargo watch -cx "shuttle run"` to reload on change detection) and it is very annoying to debug sessions or just run a dev server while working in general.

The default binary built using `cargo build` or `cargo run` is unusable because it requires some options to be set and still doesn't work the same as `cargo shuttle run`.

There might be two reasons for long compile times that I can think of:

- `cargo shuttle run` does not respect the `$HOME/.cargo/config.toml` or the `project/.cargo/config.toml` file and uses the default linker every time.
- `cargo shuttle run` builds the binary in release mode and release binaries are known for taking a long time to compile.

The improvements that can be done here are:

- Respect the linker of choice of the users from the configuration.
- Allow which build mode to use for locally run binaries.
- Allow users to use the binary as is with some default options set and not requiring them to use the cli for executing the binary.

Above are the best ideas that I could think of, but I'd really like to hear everyone's opinion and ideas about this.

### Duplicate declaration

- [X] I have searched the issues and this improvement has not been requested before.

jonaro00:
Since the last release, #922 has been merged, which means that the locally installed cargo will be used instead of a bundled one. I think that might solve the issue you're facing. We can wait until 0.19 and see if it works.

that-ambuj:
Thanks @jonaro00, I'll wait until the 0.19 and if the issue still persists, I'll let you know; otherwise, I'll close this issue.

chesedo:
v0.19.0 should be released on Monday (19 Jun) :crossed_fingers:

jonaro00:
Sounds like this will need #928, which I'm addressing in #1050

that-ambuj:
[...]

PR #1286, opened 2023-10-01
feat: add lld and mold linkers

jonaro00:
## Description of change

!! Wait until #1247 before merging this, mold is not in Debian 10's apt.

- Add lld and mold linkers, for people that want to use them (#1000)

Let ppl test it in production :tf:


oddgrd (approved):
If we add these now, we will also need to support them in the builder, so as to not break old projects. But I suppose adding them there should be the same as adding the other ones we add here. So I think we can do this now.

jonaro00:
Yeah I am casually raising the requirements bar for builder all the time :D

PR #1287, opened 2023-10-01
fix(cargo-shuttle): spam less requests when waiting for project ready

jonaro00:
## Description of change

Still looks very smooth, but now does not spam 100s of requests while waiting.

Works locally.


iulianbarbu (approved):

oddgrd (approved):
Nice, thanks!

PR #1289, opened 2023-10-02
feat(auth): added billing backend support

iulianbarbu:
## Description of change

Added backend billing support.

Integration tests were added.

TODO:
* add integration tests for a trial subscription?
* simplify integration test mocking (checkout sessions and subscriptions objects can be parameterised and changed based on the fields of interest).


auth/src/lib.rs
```diff
@@ -29,7 +29,7 @@ pub async fn start(pool: SqlitePool, args: StartArgs) -> io::Result<()> {
     let router = api::ApiBuilder::new()
         .with_sqlite_pool(pool)
         .with_sessions()
-        .into_router();
+        .into_router(stripe::Client::new(args.stripe_secret_key));
```
oddgrd, line 32: Should we rather add the stripe client with a builder method?

iulianbarbu: I don't have a strong preference for this. I haven't used a builder method for it because I wanted to pass it directly to the scope which will use it, which will attach it to the UserManager. Didn't feel the need to store the client in both ApiBuilder and UserManager. Either way should be fine though, lmk if you have a strong preference for having a builder method for it.

iulianbarbu: I agree. Will improve this a bit on the lines you suggested.

iulianbarbu: Done.

DEVELOPING.md
```diff
@@ -287,6 +287,13 @@ Finally, configure Docker Compose. You can either
 
 If you are using `nftables`, even with `iptables-nft`, it may be necessary to install and configure the [nftables CNI plugins](https://github.com/greenpau/cni-plugins)
 
+## Testing the Pro tier
+
+We use Stripe to start Pro subscriptions and verify them with a Stripe client that needs a secret key. The `STRIPE_SECRET_KEY` environment variable
+should be set to test upgrading a user to Pro tier, or to use a Pro tier feature with cargo-shuttle CLI. On a local environment, that requires
+setting up a Stripe account and generating a test API key. Auth can still be initialised and used without a Stripe secret key, but it will fail if
+when retrieving a user, and when we'll verify the subscription validity.
```
oddgrd, line 295: Thanks for the docs!
```suggestion
## Testing the Pro tier

We use Stripe to start Pro subscriptions and verify them with a Stripe client that needs a secret key. The `STRIPE_SECRET_KEY` environment variable
should be set to test upgrading a user to Pro tier, or to use a Pro tier feature with cargo-shuttle CLI. On a local environment, that requires
setting up a Stripe account and generating a test API key. Auth can still be initialised and used without a Stripe secret key, but it will fail
when retrieving a user, and when we'll verify the subscription validity.
```

oddgrd (approved):
LGTM! Just two nits

auth/src/user.rs
```diff
@@ -94,20 +175,78 @@ pub struct User {
...
+
+        Ok(false)
+    }
+
+    // Synchronize the tiers with the subscription validity.
+    async fn sync_tier(&mut self, user_manager: &UserManager) -> Result<bool, Error> {
+        let subscription_is_valid = self
+            .subscription_is_valid(&user_manager.stripe_client)
+            .await?;
+        debug!(%subscription_is_valid, "to be or not to be");
```
oddgrd, line 217: To keep or not to keep this message, that is the question :smile:

auth/src/user.rs
```diff
@@ -94,20 +175,78 @@ pub struct User {
...
+        }
+
+        Ok(false)
+    }
+}
+
+impl FromRow<'_, SqliteRow> for User {
+    fn from_row(row: &SqliteRow) -> Result<Self, sqlx::Error> {
+        let subscription_id =
+            SubscriptionId::from_str(row.try_get("subscription_id").unwrap()).ok();
```
oddgrd, line 241: Will this fail if the user doesn't have a subscription id?

iulianbarbu: Good catch. We have a test that covers it: https://github.com/shuttle-hq/shuttle/blob/main/auth/tests/api/users.rs#L59. It looks like the `try_get(<column_name>)` will do as best as it can and will return an empty string after the `.unwrap()` in case the value is `NULL`, which would be an invalid subscription id and will cause a `None`, so that's why it wasn't an issue previously.

However, I will adjust this and avoid using the `unwrap()`.

auth/src/user.rs
```diff
@@ -94,20 +175,78 @@ pub struct User {
...
     pub fn is_admin(&self) -> bool {
         self.account_tier == AccountTier::Admin
     }
 
-    pub fn new(name: AccountName, key: ApiKey, account_tier: AccountTier) -> Self {
+    pub fn new(
+        name: AccountName,
+        key: ApiKey,
+        account_tier: AccountTier,
+        session_id: Option<SubscriptionId>,
```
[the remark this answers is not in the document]
iulianbarbu: Done.

DEVELOPING.md
iulianbarbu: Done.

auth/src/user.rs
iulianbarbu: Removed it 😆

auth/src/user.rs
iulianbarbu: Adjusted this.

oddgrd (approved):

PR #1301, opened 2023-10-05
ci: use smaller machines for docker tests

jonaro00:
## Description of change

The xlarge machine is 5x more expensive than large, and cargo-shuttle barely utilized its resources, so downgrading to see its performance. The other two use even less.
All of them use much less sccache after the docker split of jobs, checking if this is enough.


orhun (approved):

oddgrd (approved):
There isn't a noticeable change in duration as far as I can tell, so LGTM, thanks!

PR #1307, opened 2023-10-05
feat: project delete

jonaro00:
## Description of change

TODO: Rebase on main when #1256 is merged.


gateway/src/service.rs
```diff
@@ -596,6 +591,15 @@ impl GatewayService {
...
     }
 
+    pub async fn delete_project(&self, project_name: &ProjectName) -> Result<(), Error> {
+        query("DELETE FROM projects WHERE project_name = ?1")
+            .bind(project_name)
+            .execute(&self.db)
+            .await?;
+
+        Ok(())
+    }
```
oddgrd, line 601: Do we need to delete related rows from custom domains first?

cargo-shuttle/src/lib.rs
oddgrd, line 1: Perhaps the spinner refactor should have been a separate pr? :sweat_smile:

cargo-shuttle/src/client.rs
```diff
@@ -182,12 +182,22 @@ impl Client {
...
         let path = format!("/projects/{}", project.as_str());
 
         self.delete(path).await
     }
 
+    pub async fn delete_project(&self, project: &ProjectName, dry_run: bool) -> Result<String> {
+        let path = format!(
+            "/projects/{}/delete{}",
+            project.as_str(),
+            if dry_run { "?dry_run=true" } else { "" }
```
oddgrd, line 195: Good idea! Don't we need to add this to the clap args as well?

oddgrd (commented):
Awesome! I left a few comments, but this seems very close!

oddgrd:
Could we also expand some of the gateway integration tests to also test project deletion?

cargo-shuttle/src/lib.rs
jonaro00: Yeah, but it was hard to get anything done on this branch when gateway was being spammed to the point of freezing

cargo-shuttle/src/client.rs
jonaro00: c-s does a dry run before the actual call to check that business rules are fulfilled before the final deletion prompt.

orhun:
General +1 from me, will check again once @oddgrd's review points are addressed.

gateway/src/api/latest.rs
```diff
@@ -280,6 +280,108 @@ async fn destroy_project(
...
+        .uri(
+            format!("/projects/{project_name}/services/{project_name}/resources")
+                .parse::<Uri>()
+                .unwrap(),
+        )
+        .method("GET")
+        .body(hyper::Body::empty())
+        .unwrap();
+    let res = route_project(State(state.clone()), scoped_user, resource_req).await?;
+    if res.status() != StatusCode::OK {
```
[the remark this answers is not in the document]
jonaro00: Yeah I wonder how new deployer will need to be in order for this to work... I guess we tell ppl to restart if delete does not work, and contact us if there's still trouble.

deployer/src/deployment/queue.rs
```diff
@@ -453,13 +453,13 @@ async fn run_pre_deploy_tests(
 /// This will store the path to the executable for each runtime, which will be the users project with
 /// an embedded runtime for alpha, and a .wasm file for shuttle-next.
 #[instrument(skip(executable_path, to_directory, new_filename))]
-async fn move_executable(
+async fn copy_executable(
```
oddgrd, line 456: What is this change addressing?

jonaro00: Adhoc redeployment speed improvement. Should cherry pick it to another branch tbh.

cargo-shuttle/src/client.rs
```diff
@@ -182,12 +182,22 @@ impl Client {
...
     }
 
-    pub async fn delete_project(&self, project: &ProjectName) -> Result<project::Response> {
+    pub async fn stop_project(&self, project: &ProjectName) -> Result<project::Response> {
         let path = format!("/projects/{}", project.as_str());
 
         self.delete(path).await
     }
 
+    pub async fn delete_project(&self, project: &ProjectName, dry_run: bool) -> Result<String> {
```
Kazy, line 191: **suggestion**: use an enum instead of a bool for the dry run. `delete_projects("foo", DryRun::True)` is clearer that `delete_projects("foo", true)`.

cargo-shuttle/src/lib.rs
```diff
@@ -1718,59 +1750,91 @@ impl Shuttle {
...
+                    This will...
+                    - Delete all Secrets and Persist data in this project.
+                    - Release the project name from your account.
+                    This action is permanent."#,
+                self.ctx.project_name()
+            )
+            .bold()
+            .red()
+        );
+        if !Confirm::with_theme(&ColorfulTheme::default())
```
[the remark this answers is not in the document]
jonaro00: Tbh, deleting a database is a bigger deal, and in that case, we don't have a unique name for the user to type.

cargo-shuttle/src/client.rs
jonaro00: Too verbose for my taste, considering it would bloat the codebase if applied everywhere.

cargo-shuttle/src/lib.rs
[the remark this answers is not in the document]
Kazy: I don't think the fact that we don't have that for databases should exempt us from doing it for projects. But we can tackle both of them later on, to make sure it's consistent.

cargo-shuttle/src/client.rs
Kazy: My point is not to do it everywhere, just here, since it has quite the impact on the operation.

gateway/src/service.rs
```diff
@@ -596,6 +591,26 @@ impl GatewayService {
...
+
+        query("DELETE FROM custom_domains WHERE project_id = ?1")
+            .bind(project_id)
+            .execute(&self.db)
+            .await?;
+
+        query("DELETE FROM projects WHERE project_name = ?1")
+            .bind(project_name)
+            .execute(&self.db)
+            .await?;
```
oddgrd, line 609: Should we make this a transaction to be on the safe side?

jonaro00: I couldn't figure out how to do it :joy: . Feel free to take the steering wheel.

gateway/src/main.rs
```diff
@@ -64,12 +64,12 @@ async fn main() -> io::Result<()> {
...
 
-    let worker = Worker::new();
+    let ambulance_worker = Worker::new();
 
-    let sender = worker.sender();
+    let sender = ambulance_worker.sender();
 
     let worker_handle = tokio::spawn(
-        worker
+        ambulance_worker
```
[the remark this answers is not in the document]
jonaro00: Because it is a "Worker" receiving things from the ambulance thread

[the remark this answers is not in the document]
iulianbarbu: It is used for other `orchestration` related code paths too (e.g. each gateway API handler might make use of the `RouterState`, who's using the worker (by sending tasks to it) for various tasks (e.g. project creation). I think we shouldn't change the name.

cargo-shuttle/src/lib.rs
```diff
@@ -570,29 +572,35 @@ impl Shuttle {
...
-            .stop_service(proj_name)
-            .await
-            .map_err(suggestions::deployment::stop_deployment_failure)?;
+        let p = self.ctx.project_name();
+        wait_with_spinner(|i, pb| async move {
+            let service = if i == 0 {
+                client.stop_service(p).await?
+            } else {
+                client.get_service(p).await?
+            };
```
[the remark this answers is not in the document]
jonaro00: I like to apply DRY at 3

gateway/src/api/latest.rs
```diff
@@ -280,6 +280,112 @@ async fn destroy_project(
...
+    let service_req = rb
+        .uri(
+            format!("/projects/{project_name}/services/{project_name}")
+                .parse::<Uri>()
+                .unwrap(),
+        )
+        .method("GET")
+        .body(hyper::Body::empty())
+        .unwrap();
+    let res = route_project(State(state.clone()), scoped_user.clone(), service_req).await?;
```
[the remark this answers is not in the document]
jonaro00: It correctly wakes the project up an proceeds

cargo-shuttle/src/client.rs
jonaro00: I'll skip it since it is only being used in one place.

gateway/src/main.rs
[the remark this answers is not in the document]
oddgrd: Good point, reverted now.

oddgrd (approved):

PR #1309, opened 2023-10-09
fix(cargo-shuttle): fix init login bugs

oddgrd:
## Description of change

`cargo-shuttle` init was not authorizing the client in the event that the api-key was already set in the config. It was also the case that if passing certain args, the interactive flow was disabled, which meant if the user wasn't logged in and didn't pass --api-key in the command, they would get an error about trying to login but failing. I fixed that too, but it does mean the path prompt will be interactive even if given. The init logic is quite complicated, so I'm very open to suggestions here.

Tested running `cargo run -p cargo-shuttle -- --wd ./  init --from https://github.com/shuttle-hq/eurorust-demo --name unique-name --create-env
 ./unique-name` when logged in with config file, with --api-key, and neither.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -319,16 +319,14 @@ impl Shuttle {
             };
 
             self.login(login_args).await?;
-        } else {
-            if interactive {
-                println!("First, let's log in to your Shuttle account.");
-                self.login(args.login_args.clone()).await?;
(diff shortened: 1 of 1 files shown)


jonaro00 (approved):
That scenario sounds very obscure, so it's fine. "When I have time," I want to rewrite the core logic on this, so it is less bug prone.

PR #1313, opened 2023-10-09
chore: bump and refactor images, code cleanup

jonaro00:
## Description of change

- Bump stack's rust version to 1.73, CI and MSRV remains 1.72
- Some refactorings to allow use of smaller images
- Copy executable instead of move for faster redeploys

stack went up locally


Containerfile
```diff
@@ -54,14 +54,16 @@ RUN cargo build \
     --bin shuttle-next -F next
 
 
+####### Helper step
+
+FROM docker.io/library/debian:bookworm-20230904-slim AS bookworm-20230904-slim-plus
+RUN apt update && apt install -y curl ca-certificates; rm -rf /var/lib/apt/lists/*
```
chesedo, line 60: Why are we doing this for all projects and not just `auth`?

runtime/Cargo.toml
```diff
@@ -27,7 +27,7 @@ tokio = { workspace = true, features = ["full"] }
 tokio-stream = { workspace = true }
 tonic = { workspace = true }
 tower = { workspace = true }
-cap-std = { workspace = true, optional = true }
+cap-std = { version = "1.0.15", optional = true }
```
chesedo: Why the downgrade from v2 (in the workspace)?

chesedo (commented):
Just one comment

runtime/Cargo.toml
chesedo: Nvm, figured this out

Containerfile
jonaro00: Common for auth, gw, and prov was that they can run in the slim image only if they have ca-certs and/or curl. Therefore, this middle step was made.

chesedo: Ohh, I see 👍

chesedo (approved):
LGTM

PR #1314, opened 2023-10-10
fix(resources/shared-db): retry initial connection a few times

Kazy:
## Description of change

Fixes a bug where running a project with the shared DB for the first time could result in a failure to connect, which could be fixed simply by running it again.

The connection attempt is made too quickly (some would say it's blazingly fast) and the Postgres database doesn't have the time to start accepting connection. A simple wait is enough.

 With the Eurorust demo, 2nd step, and simply doing `cargo shuttle run`. I'm using Podman, and I can consistently hit the bug simply by removing the running PSQL container and trying again.

### DIFF
--- a/cargo-shuttle/src/provisioner_server.rs
+++ b/cargo-shuttle/src/provisioner_server.rs
@@ -155,6 +155,12 @@ impl LocalProvisioner {
                 .expect("failed to start none running container");
         }
 
+        self.wait_for_ready(&container_name, is_ready_cmd.clone())
+            .await?;
+
+        // The container enters the ready state, runs an init script and then reboots, so we sleep
+        // a little and then check if it's ready again afterwards.
+        sleep(Duration::from_millis(450)).await;
         self.wait_for_ready(&container_name, is_ready_cmd).await?;
 
         let res = DatabaseResponse {


resources/shared-db/src/postgres.rs
```diff
@@ -59,14 +60,30 @@ impl ResourceBuilder<sqlx::PgPool> for Postgres {
...
+                .max_connections(5)
+                .connect(&connection_string)
+                .await
+            {
+                Ok(pool) => return Ok(pool),
+                Err(err) => {
+                    if attempt >= max_attempt {
+                        Err(CustomError::new(err))?;
+                    } else {
+                        println!("Failed to connect to Postgres database, retrying in {timeout_in_seconds} seconds");
```
Kazy, line 79: Do we want to print something ?

resources/shared-db/src/postgres.rs
```diff
@@ -59,14 +60,30 @@ impl ResourceBuilder<sqlx::PgPool> for Postgres {
...
+            match sqlx::postgres::PgPoolOptions::new()
+                .min_connections(1)
+                .max_connections(5)
+                .connect(&connection_string)
+                .await
+            {
+                Ok(pool) => return Ok(pool),
+                Err(err) => {
+                    if attempt >= max_attempt {
+                        Err(CustomError::new(err))?;
```
Kazy, line 77: Not a big fan of relying on `?` for returning that way, I would be happier with a proper `return`, but this means using the following form, and I'm not sure that's better.
```rust
return Err(<Error as From<CustomError>>::from(CustomError::new(err)));
```

resources/shared-db/src/postgres.rs
```diff
@@ -59,14 +60,30 @@ impl ResourceBuilder<sqlx::PgPool> for Postgres {
...
 
-        let pool = sqlx::postgres::PgPoolOptions::new()
-            .min_connections(1)
-            .max_connections(5)
-            .connect(&connection_string)
-            .await
-            .map_err(CustomError::new)?;
+        let mut attempt = 0;
+        let mut timeout_in_seconds = 1;
+        let max_attempt = 3;
```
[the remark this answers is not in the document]
Kazy: I wanted to avoid adding a dependency when it's just a couple of lines of code.

resources/shared-db/Cargo.toml
```diff
@@ -12,6 +12,7 @@ mongodb = { version = "2.3.0", optional = true }
 serde = { version = "1.0.148", features = ["derive"] }
 shuttle-service = { path = "../../service", version = "0.29.0" }
 sqlx = { version = "0.7.1", optional = true }
+tokio = { version = "1", features = ["time"] }
```
Kazy, line 15: I've added tokio. but we might not want to force users to use tokio. I'm not sure how to handle that otherwise.

jonaro00: Tokio is already forced from the runtime.

iulianbarbu (commented):
I am wondering why this doesn't help: https://github.com/shuttle-hq/shuttle/blob/main/cargo-shuttle/src/provisioner_server.rs#L158. We should be already waiting for the db to be ready accepting connections IIUC. I am expecting that the connection will be started sometime at the end of the function `provision_database`.

oddgrd:
We can change this to use the same fix we use in the provisioner and logger tests. So the problem is that the local databases, on first start, will start and run an init script. They will then reboot, so a local run might try to connect to the DB before it's rebooted and ready. The fix from provisioner/logger is here: https://github.com/shuttle-hq/shuttle/blob/main/logger/tests/helpers.rs#L54-L59

For more information on the postgres init scripts: https://hub.docker.com/_/postgres (under initialization scripts).

oddgrd (approved):

PR #1320, opened 2023-10-14
feat(cargo-shuttle): adaptive page hints

### LINKED ISSUE #1316: [Improvement]: Remove redundant hint based on row count
### Describe the improvement

Currently, the table listing commands `cargo shuttle project list` and `cargo shuttle deployment list` show this notice at the end:
`More projects might be available on the next page using --page.`
It would be nice if this message was only displayed if there actually *can* be more items in the list, i.e. when the number of returned rows is equal to the "limit" argument for the request. (if there are fewer items than the limit, we know that we are viewing the last page)

### Duplicate declaration

x04:
## Description of change
Changed the project and deployment list endpoints to return a `Page` struct that contains the original Vec of data, but also a `has_next_page` variable to indicate to the client whether there is another page that can be fetched. This value is determined by simply extending the `limit` by 1 and checking if the number of returned values matches that increased limit, this means we do not have to use a `COUNT` query or even any other queries at all. However, to ensure we return the proper number of values to the client we call `pop` on the Vec so that the extra value we requested is removed. There are a couple other small changes but they are all more self-explanatory than what I discussed.

Closes #1316 

Only tested on a local setup, should be fine as there are no breaking changes besides the data returned from the 2 relevant endpoints, however those endpoints seem to only be used by the 2 relevant commands so that should not be an issue.

### DIFF
--- a/deployer/src/handlers/mod.rs
+++ b/deployer/src/handlers/mod.rs
@@ -450,6 +450,7 @@ pub async fn create_service(
             .git_branch
             .map(|s| s.chars().take(GIT_STRINGS_MAX_LENGTH).collect()),
         git_dirty: deployment_req.git_dirty,
+        count: None,
     };
 
     persistence.insert_deployment(deployment.clone()).await?;
@@ -515,7 +516,7 @@ pub async fn stop_service(
     get,
     path = "/projects/{project_name}/deployments",
     responses(
-        (status = 200, description = "Gets deployments information associated to a specific project.", body = shuttle_common::models::deployment::Response),
+        (status = 200, description = "Gets deployments information associated to a specific project.", body = shuttle_common::models::deployment::Page),
         (status = 500, description = "Database error.", body = String),
         (status = 404, description = "Record could not be found.", body = String),
     ),
@@ -530,23 +531,19 @@ pub async fn get_deployments(
     Query(PaginationDetails { page, limit }): Query<PaginationDetails>,
 ) -> Result<Json<deployment::Page>> {
     if let Some(service) = persistence.get_service_by_name(&project_name).await? {
-        let limit = limit.unwrap_or(u32::MAX - 1) + 1;
+        let limit = limit.unwrap_or(u32::MAX);
         let page = page.unwrap_or(0);
-        let mut deployments: Vec<deployment::Response> = persistence
+        tracing::info!("test l = {} p = {}", limit, page);
+        let deployments = persistence
             .get_deployments(&service.id, page * limit, limit)
-            .await?
-            .into_iter()
-            .map(Into::into)
-            .collect();
+            .await?;
+        let count = deployments.first().and_then(|d| d.count).unwrap_or(0);
 
-        let has_next_page = deployments.len() == limit as usize;
-        if has_next_page {
-            deployments.pop();
-        }
+        let deployments = deployments.into_iter().map(Into::into).collect();
 
         let page = deployment::Page {
             data: deployments,
-            has_next_page,
+            count,
         };
 
         Ok(Json(page))
@@ -561,7 +558,7 @@ pub async fn get_deployments(
     get,
     path = "/projects/{project_name}/deployments/{deployment_id}",
     responses(
-        (status = 200, description = "Gets a specific deployment information.", body = shuttle_common::models::deployment::Response),
+        (status = 200, description = "Gets a specific deployment information.", body = shuttle_common::models::deployment::Page),
         (status = 500, description = "Database or streaming error.", body = String),
         (status = 404, description = "Record could not be found.", body = String),
     ),
@@ -573,7 +570,7 @@ pub async fn get_deployments(
 pub async fn get_deployment(
     Extension(persistence): Extension<Persistence>,
     Path((project_name, deployment_id)): Path<(String, Uuid)>,
-) -> Result<Json<shuttle_common::models::deployment::Response>> {
+) -> Result<Json<deployment::Response>> {
     if let Some(deployment) = persistence.get_deployment(&deployment_id).await? {
         Ok(Json(deployment.into()))
     } else {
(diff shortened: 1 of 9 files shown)


x04:
Tests passed however I will say an issue I thought about as I was about to go for the night:
`let limit = limit.unwrap_or(u32::MAX - 1) + 1;`
I did not consider the case where someone could still pass `limit` as `u32::MAX` and that would cause an overflow, which isn't ideal. I could add a check to ensure `limit < u32::MAX` but I'll just leave this here for now and wait for a second opinion.

jonaro00 (commented):
Instead of modifying the limit and trimming the vector, we could add a `COUNT(*)` to the SQL query and determine if we are at the end based on that. Then the response could just contain the count instead of "has next", and the receiver (cargo-shuttle) can make the calculation.

FYI the API response model change is a breaking change, but we are fine that.

jonaro00:
Also, the u32::MAX concern can be ignored if the limit passed to the query is optional, and the query is mutated based on it.

x04:
> Instead of modifying the limit and trimming the vector, we could add a `COUNT(*)` to the SQL query and determine if we are at the end based on that. Then the response could just contain the count instead of "has next", and the receiver (cargo-shuttle) can make the calculation.
> 
> FYI the API response model change is a breaking change, but we are fine that.

Only reason I didn't use `COUNT(*)` is because it has to get duplicated on every row returned unless we use 2 separate queries for counting and getting the data. Simply fetching an extra row is a lot "cleaner" to check for than having a lot of rows with a duplicated field. I'll make the changes but just wanted to give my reasoning.

Also, just for my understanding, is the API change breaking because of some internal usage in the shuttle backend or did I miss some usage somewhere within this repo?

> Also, the u32::MAX concern can be ignored if the limit passed to the query is optional, and the query is mutated based on it.

To clarify what I meant by this is if `limit` is passed to that function with a value of `Some(u32::MAX)` then it would cause an overflow due to my adding 1 to limit for the page check, however the changes you mentioned above should fix that because there is no longer a need to modify the limit.

jonaro00:
cargo-shuttle calls the endpoints of the backends, so a change in the models makes it breaking. Changes in internally used models are not breaking since we update all backends to the same version at the same time.

x04:
> cargo-shuttle calls the endpoints of the backends, so a change in the models makes it breaking. Changes in internally used models are not breaking since we update all backends to the same version at the same time.

Ohh yes I see what you're saying now, it would break old versions of cargo shuttle, I was only thinking in the context of everything being up to date.

x04:
Made the changes to use row count instead of an extra row, however imo it looks worse and adds some unnecessary bloat but if it looks ok then it's fine. Everything was tested locally same as I did when I initially made the PR.

I can fix the 2 tests that failed but I'd like to have confirmation this is the route we want to go down as I think there is a chance you will probably prefer the original method, one other method would be cursor based pagination instead of a simple limit/offset and returning a bool for if there is a next page or not.

To note the reason the 2 tests failed, it is because I am technically getting the count in a subquery that is returned as a column and since deployments have their types a bit more tightly integrated it's a bit more work to filter out that extra column. The alternative is using 2 separate queries but I was trying to avoid that initially, although with a subquery I am pretty sure it is basically the same as 2 separate queries at this point unless the planner can optimize the subquery away.

deployer/src/persistence/mod.rs
```diff
@@ -248,10 +248,14 @@ impl Persistence {
...
+            QueryBuilder::new("SELECT *, (SELECT COUNT(*) FROM deployments WHERE service_id = ");
 
+        let service_id = service_id.to_string();
         query
-            .push_bind(service_id.to_string())
+            .push_bind(&service_id)
+            .push(") as cnt FROM deployments WHERE service_id = ")
+            .push_bind(&service_id)
             .push(" ORDER BY last_update DESC LIMIT ")
             .push_bind(limit);
```
jonaro00, line 260: LIMIT can be behind an if-case if the option is not unwrap_or'ed.

deployer/src/handlers/mod.rs
```diff
@@ -530,23 +531,19 @@ pub async fn get_deployments(
     Query(PaginationDetails { page, limit }): Query<PaginationDetails>,
 ) -> Result<Json<deployment::Page>> {
     if let Some(service) = persistence.get_service_by_name(&project_name).await? {
-        let limit = limit.unwrap_or(u32::MAX - 1) + 1;
+        let limit = limit.unwrap_or(u32::MAX);
         let page = page.unwrap_or(0);
-        let mut deployments: Vec<deployment::Response> = persistence
+        tracing::info!("test l = {} p = {}", limit, page);
```
jonaro00, line 536: reminder to remove.

jonaro00 (commented):
We might wait a while before merging this since it is a breaking change.

deployer/src/persistence/mod.rs
x04: > LIMIT can be behind an if-case if the option is not unwrap_or'ed.

Left this the same as it was not something I wrote, but what would we use for offset then? Because offset is calculated as `page * limit` so that logic would have to be changed, or is having no offset if limit is 0 fine?

jonaro00: Leaving it is fine. I didn't see that there would be more changes involved.

jonaro00:
If you want to get a non-breaking change merged sooner than this can be merged, implementing the original idea in #1316 is good enough until then.

x04:
> If you want to get a non-breaking change merged sooner than this can be merged, implementing the original idea in #1316 is good enough until then. (New PR)

That original idea doesn't really work because lets say the limit is 5 and we have exactly 5 records, it would then think we have extra pages. However one idea I did have that shouldn't technically be breaking is adding a field to the return type that contains the record count, that is to say I would still return just a Vec of objects as expected it would just have an extra field `count` or something that represents how many rows are left. This should not be breaking technically because extra fields are ignored during deserialization so although it would look really weird to have that data duplicated a lot it would technically work.

jonaro00:
The original idea works in >90% of cases, but will of course still arrive at "there might be more" when viewing the last page and `count % limit == 0`. It is a non-breaking imperfect compromise.

x04:
> The original idea works in >90% of cases, but will of course still arrive at "there might be more" when viewing the last page and `count % limit == 0`. It is a non-breaking imperfect compromise.

Alright sounds good! Will get this implemented and make a PR later tonight or sometime tomorrow.

oddgrd:
Superseded by #1357

PR #1325, opened 2023-10-18
build: update dependencies

Kazy:
Update packages to remove several `cargo audit` issues. There is only one left with `wasmtime`, but the current version constraint is 7.0.0 while the last one is 13, and I need more time to check this one. Went through the updates done to the lockfile, there doesn't seem to be any breaking change we care about.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,9 +40,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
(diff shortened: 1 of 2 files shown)


oddgrd (approved):
Thanks! I think we can do the wasmtime upgrade in a separate PR/ticket, since it may some require code changes as well.

PR #1331, opened 2023-10-20
feat(gateway): project limit

jonaro00:
## Description of change

Temporary limits for the first release.

Works locally 
image


common/src/models/error.rs
```diff
@@ -80,7 +81,10 @@ impl From<ErrorKind> for ApiError {
             ),
             ErrorKind::ProjectUnavailable => {
                 (StatusCode::BAD_GATEWAY, "project returned invalid response")
-            }
+            },
+            ErrorKind::TooManyProjects => {
+                (StatusCode::FORBIDDEN, "you cannot create more projects. delete some first.")
```
orhun, line 86:
```suggestion
                (StatusCode::FORBIDDEN, "You cannot create more projects. Delete some projects first.")
```

gateway/src/service.rs
```diff
@@ -549,6 +552,35 @@ impl GatewayService {
...
+        &self,
+        account_name: &AccountName,
+        is_admin: bool,
+        is_pro: bool,
+    ) -> Result<(), Error> {
+        if is_admin {
+            return Ok(());
+        }
+
+        let proj_count: u32 = query("SELECT COUNT(*) FROM projects WHERE account_name = ?1")
```
oddgrd, line 565: Nit: would it be more efficient to count the primary key?
```suggestion
        let proj_count: u32 = query("SELECT COUNT(project_name) FROM projects WHERE account_name = ?1")
```

common/src/claims.rs
```diff
@@ -85,6 +85,9 @@ pub enum Scope {
     /// of DNS-01 challenge completion.
     GatewayCertificateRenew,
 
+    /// Perform Pro tier action
+    Pro,
+
```
oddgrd, line 90: Hmm, I'm not sure if this should be a scope. Shouldn't we rather have scopes for actions a pro user is allowed to take, that are only assigned to users with the pro account_tier? E.g. create up to 15 projects, create custom domains and provision RDS (when those are implemented).

oddgrd: Also, are we creating a breaking change here, or is adding new fields fine? :thinking:

oddgrd (commented):
Nice! I think we just need to discuss what these initial limits will be. Having them higher at first makes sense, I think, since some users might need support to be able to delete projects.

common/src/claims.rs
jonaro00: I tried going down that road, but all our claims logic is based on checking if a Vec contains a scope, which means we can't parameterize this enum easily (i.e. `ProjectCreate(u32)`). The admin scope is done in the same way. New field is not breaking.

gateway/src/service.rs
jonaro00: Shouldn't matter. It just counts the matching rows AFAIK.

common/src/claims.rs
jonaro00: As we discovered, breaking changes here are not preferred :hot_face: 
I'm all for reworking claims to be a struct with bools and options, but that would be a huge breaking change for another time.

oddgrd: Could there be something like an ExtraProjects variant, only assigned to users with the pro tier? And then if anyone doesn't have that scope, they get 3 projects.

jonaro00: This does not scale to the optimal case, where the limit is fully adjustable for any group of users or individuals.
> struct with bools and options

This would allow things of that sort.

oddgrd (approved):

common/src/claims.rs
```diff
@@ -53,6 +53,9 @@ pub enum Scope {
     /// Create and delete projects
     ProjectWrite,
 
+    /// Perform Pro tier action
```
oddgrd, line 56: Perhaps something like this?
```suggestion
    /// Create more projects than the free tier default
```

gateway/src/service.rs
```diff
@@ -549,6 +551,28 @@ impl GatewayService {
         }
     }
 
+    pub async fn check_project_count(
+        &self,
+        account_name: &AccountName,
+        limit: Option<u32>,
+    ) -> Result<(), Error> {
+        let Some(limit) = limit else {
```
oddgrd, line 559: Shall we add a comment here clarifying that an admin user will have no limit?

oddgrd (approved):
Nice! I just left some comment suggestions.

PR #1332, opened 2023-10-21
refactor(runtime): hide internals from public-facing API, export tokio

jonaro00:
## Description of change

*finally a well isolated PR...*

- Move internals into hidden module, which hides them from the cargo doc page (docs.rs).
- Re-export tokio, allows user code to not have tokio in Cargo.toml, unless they need it for something ofc :)

### DIFF
--- a/codegen/src/shuttle_main/mod.rs
+++ b/codegen/src/shuttle_main/mod.rs
@@ -232,7 +232,7 @@ impl ToTokens for Loader {
             None
         } else {
             Some(parse_quote!(
-                use ::shuttle_runtime::__internals::{Factory, ResourceBuilder};
+                use ::shuttle_runtime::{Factory, ResourceBuilder};
             ))
         };
 
@@ -399,7 +399,7 @@ mod tests {
                 mut resource_tracker: ::shuttle_runtime::__internals::ResourceTracker,
             ) -> ShuttleComplex {
                 use ::shuttle_runtime::__internals::Context;
-                use ::shuttle_runtime::__internals::{Factory, ResourceBuilder};
+                use ::shuttle_runtime::{Factory, ResourceBuilder};
                 let pool = ::shuttle_runtime::__internals::get_resource(
                     shuttle_shared_db::Postgres::new(),
                     &mut factory,
@@ -510,7 +510,7 @@ mod tests {
                 mut resource_tracker: ::shuttle_runtime::__internals::ResourceTracker,
             ) -> ShuttleComplex {
                 use ::shuttle_runtime::__internals::Context;
-                use ::shuttle_runtime::__internals::{Factory, ResourceBuilder};
+                use ::shuttle_runtime::{Factory, ResourceBuilder};
                 let vars = std::collections::HashMap::from_iter(factory.get_secrets().await?.into_iter().map(|(key, value)| (format!("secrets.{}", key), value)));
                 let pool = ::shuttle_runtime::__internals::get_resource (
                     shuttle_shared_db::Postgres::new().size(&::shuttle_runtime::__internals::strfmt("10Gb", &vars)?).public(false),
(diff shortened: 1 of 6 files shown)


oddgrd (approved):

oddgrd:
Cool! If the user wants a different version of tokio, cargo should use the one they specify, right (as long as it's not a major version change, but that isn't planned for tokio)?

jonaro00:
Yes, the min version remains what we have in our Cargo.toml (1.22), and creating or bumping the lock file will choose the latest compatible one.

orhun (approved):

codegen/src/shuttle_main/mod.rs
```diff
@@ -14,9 +14,13 @@ pub(crate) fn r#impl(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let loader = Loader::from_item_fn(&mut fn_decl);
 
     quote! {
-        #[tokio::main]
-        async fn main() {
-            shuttle_runtime::start(loader).await;
+        fn main() {
+            use ::shuttle_runtime::tokio;
+            #[tokio::main]
+            async fn main() {
```
chesedo, line 20: Why is there a `main()` inside `main`?

services/shuttle-poise/src/lib.rs
```diff
@@ -22,7 +21,7 @@
 //!     // Get the discord token set in `Secrets.toml`
 //!     let discord_token = secret_store
 //!         .get("DISCORD_TOKEN")
-//!         .context("'DISCORD_TOKEN' was not found")?;
```
chesedo, line 25: Why this change?

chesedo (commented):
Looking good. It would w great to run a `cargo check` on all our examples in case we missed anything with the review.

codegen/src/shuttle_main/mod.rs
jonaro00: To scope off our tokio import and not leak it to the end user's main.rs

services/shuttle-poise/src/lib.rs
jonaro00: It depends on runtime re-exporting anyhow, which is a bit unnecessary

chesedo: It just seems that the old error would have been a better practise

codegen/src/shuttle_main/mod.rs
chesedo: Maybe we should not call the `tokio::main` macro then and just expand it ourselves.
https://docs.rs/tokio/latest/tokio/attr.main.html#using-the-multi-thread-runtime

chesedo: I'm trying to reduce the confusion here for the future

codegen/src/shuttle_main/mod.rs
```diff
@@ -14,9 +14,15 @@ pub(crate) fn r#impl(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let loader = Loader::from_item_fn(&mut fn_decl);
 
     quote! {
-        #[tokio::main]
-        async fn main() {
-            shuttle_runtime::start(loader).await;
+        fn main() {
+            // manual expansion of #[tokio::main]
+            ::shuttle_runtime::tokio::runtime::Builder::new_multi_thread()
```
chesedo, line 19: Can we turn off some tokio defaults because of this? I'm thinking of turning off the `macros` feature to reduce the dependencies on users.

chesedo (approved):
We can merge if the dependencies improvement is not possible 😄

codegen/src/shuttle_main/mod.rs
chesedo: We can also do it in a separate PR

jonaro00: It seems like other dependencies are enabling it. It's a messy task.

PR #1334, opened 2023-10-22
feat(cargo-shuttle): adaptive page hints, non-breaking 

x04:
## Description of change
Changed the project and deployment commands to only return page hint if the items returned == limit.

Mostly closes https://github.com/shuttle-hq/shuttle/issues/1316

Tested with local setup

### DIFF
--- a/common/src/models/deployment.rs
+++ b/common/src/models/deployment.rs
@@ -67,7 +67,12 @@ impl State {
     }
 }
 
-pub fn get_deployments_table(deployments: &Vec<Response>, service_name: &str, page: u32, limit: u32) -> String {
+pub fn get_deployments_table(
+    deployments: &Vec<Response>,
+    service_name: &str,
+    page: u32,
+    limit: u32,
+) -> String {
     if deployments.is_empty() {
         if page <= 1 {
             format!(
@@ -159,8 +164,9 @@ pub fn get_deployments_table(deployments: &Vec<Response>, service_name: &str, pa
 
         if deployments.len() == limit as usize {
             format!(
-                "{formatted_table}\n\n{page_hint}", 
-                page_hint = "More projects might be available on the next page using --page.".bold()
+                "{formatted_table}\n\n{page_hint}",
+                page_hint =
+                    "More projects might be available on the next page using --page.".bold()
             )
         } else {
             formatted_table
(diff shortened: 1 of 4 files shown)


common/src/models/deployment.rs
```diff
@@ -150,18 +150,21 @@ pub fn get_deployments_table(deployments: &Vec<Response>, service_name: &str, pa
...
+        let formatted_table = format!(
+            "Most recent {} for {}\n{}",
             "deployments".bold(),
             service_name,
             table,
-            "More projects might be available on the next page using --page.".bold()
-        )
+        );
+
+        if deployments.len() == limit as usize {
```
oddgrd, line 160: I like it! But I guess this will still be wrong when there are exactly `limit` projects, right? Could we change the limit logic to be `limit + 1`, then only display `limit` projects. So say there are 15 projects, we have a limit of 10 and we show those 10, but we got 11, so we know there are more projects on the next page.

oddgrd: Oh, and a `cargo fmt` should fix CI

x04: > I like it! But I guess this will still be wrong when there are exactly `limit` projects, right? Could we change the limit logic to be `limit + 1`, then only display `limit` projects. So say there are 15 projects, we have a limit of 10 and we show those 10, but we got 11, so we know there are more projects on the next page.

Yes I did something similar to what you described in a separate PR in this commit:
https://github.com/shuttle-hq/shuttle/pull/1320/commits/25b382d608de82f65ce8d3162ccf19a6423afd5c

However I was instructed to change it to just return a record count vs a bool for the next page so I made that change and that PR still remains open here:
https://github.com/shuttle-hq/shuttle/pull/1320

However the changes are breaking so I made this that can be merged without breaking.

x04: Oh I just reread this as I was checking up on the PR and I realized you meant doing the limit + 1 on the client-side, I can definitely change it to do that and as you mentioned it's 100% accurate whereas this can falsely hint at a next page, I was just doing it on he server-side not the client-side which caused it to be breaking.

oddgrd: Yes, sorry for not making that clear! :smile: It would be great if you could implement this and resolve the conflicts, then I think we can get this merged for the next release.

oddgrd:
Superseded by #1357

PR #1336, opened 2023-10-23
feat(logger): add basic instrumentation to the API

oddgrd:
## Description of change

Add some basic instrumentation to the logger API. This should also allow us to observe the volume of logs sent from specific deployments.

Tested locally and with honeycomb test environment.


orhun (approved):

logger/src/lib.rs
```diff
@@ -45,6 +45,7 @@ where
     }
 
     async fn get_logs(&self, deployment_id: String) -> Result<Vec<LogLine>, Error> {
+        tracing::info!(deployment_id = %deployment_id, "fetching logs for deployment");
```
chesedo, line 48: I think we will be fine removing all the info traces in this PR

logger/src/lib.rs
```diff
@@ -56,13 +57,19 @@ impl<D> Logger for Service<D>
...
     async fn store_logs(
         &self,
         request: Request<StoreLogsRequest>,
     ) -> Result<Response<StoreLogsResponse>, Status> {
         let request = request.into_inner();
         let logs = request.logs;
+
         if !logs.is_empty() {
+            Span::current().record("deployment_id", &logs[0].deployment_id);
+            Span::current().record("batch_size", logs.len());
```
chesedo, line 70: Maybe store the result of `Span::current()` so that we don't call it multiple times.

chesedo (commented):
Some small comments :smile:

logger/src/lib.rs
oddgrd: Hmm, will we see anything if there are no events?

oddgrd: Ah, we will see it in Honeycomb, but not in the logger itself. I'll remove the events then, since they are quite verbose.

chesedo (approved):
LGTM

PR #1338, opened 2023-10-23
feat(gateway): block project deletion on secrets deletion

oddgrd:
## Description of change

This is based on #1307.

This forces uses to delete secrets from resource-recorder before they can delete a project. 

With a local run of a project with secrets and postgres.


gateway/src/api/latest.rs
```diff
@@ -361,11 +362,23 @@ async fn delete_project(
...
-            return Err(Error::from_kind(ErrorKind::ProjectHasDatabase));
+            .filter(|resource| {
+                matches!(
+                    resource.r#type,
+                    shuttle_common::resource::Type::Database(_)
+                        | shuttle_common::resource::Type::Secrets
+                )
+            })
+            .map(|resource| resource.r#type.to_string())
+            .collect::<HashSet<_>>()
```
oddgrd, line 376: This may be a bit overkill (it needs to add a few `Derive(Hash)` annotations), there shouldn't be more than one of each resource, but we have seen that it can happen.

jonaro00: In that case, it might be better to not dedup them here, so that the issue is not obfuscated.

jonaro00 (approved):

gateway/src/api/latest.rs
iulianbarbu: > This may be a bit overkill (it needs to add a few `Derive(Hash)` annotations), there shouldn't be more than one of each resource, but we have seen that it can happen.

Not sure why it matters if there is more than on resource attached for checking if the project has resources.

iulianbarbu (approved):

gateway/src/api/latest.rs
oddgrd: I'll drop it, thanks!

PR #1345, opened 2023-10-24
ci: add cargo-audit step

Kazy:
Add a `cargo audit` step to our CI. We probably don't want it to be blocking, to avoid any PR "just" because of a security issue, but this should still be looked at. Not clear how to make it non blocking though, probably from an admin in Github.


.circleci/config.yml
```diff
@@ -10,7 +10,7 @@ executors:
   docker-rust:
     docker:
       - image: cimg/rust:1.72.0
-    resource_class: small
+    resource_class: medium
```
jonaro00, line 13: We do not need to bump this

.circleci/config.yml
```diff
@@ -188,6 +188,26 @@ jobs:
...
+    steps:
+      - checkout
+      - run: |
+          echo "$(cargo search cargo-audit | grep '^cargo-audit =' | sha256sum)" > /tmp/cargo-audit-cache-key
+      - restore_cache:
+          name: Restore Cargo audit cache
+          keys: 
+            - cargo-audit-{{ checksum "/tmp/cargo-audit-cache-key" }}
+      - run: |
+          cargo install --force cargo-audit
```
jonaro00: If we download a binary instead, we don't need caching, see for example how sccache is done

jonaro00 (changes requested):
A job failing will not block merging. If red, we just review the output and fix it if needed.

.circleci/config.yml
Kazy: Oops mistake, didn't want to bump this one.

PR #1349, opened 2023-10-27
fix(deployer): handle cargo fetch without blocking logs, use async channels

### LINKED ISSUE #1322: [Bug]: Deployment logs at the start of deployment are delayed
### What happened?

Whenever we start a deployment it appears to hang for several seconds, because no logs are output.

image

In the image above, we can see that the first lines (until build start) were correctly produced many seconds earlier, but they were kept back from being sent over the websocket for some reason. Only when the build logs from cargo starts printing do we receive them all at once.

I think the wait could be related to cargo downloading crates, but it is unclear why the logs are blocked during that time.

### Version

v0.29.0

In deployment

In deployment

### Relevant log output

_No response_

### Duplicate declaration

jonaro00:
## Description of change

Fixes #1322 and some more

- Using blocking crossbeam channels was causing logs to be delayed. Switched over to tokio channels everywhere possible.
  - Initial logs from deployment now appear instantly, and one line is printed if the fetching takes time.
- Do a manual cargo fetch when building, and don't let metadata commands fetch. (this is causing CI fail, will look for compromise.)

Local run works as usual, with and without crates cached.
Deployment works for axum hello world with and without --no-test, with and without failing test.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -39,7 +39,6 @@ use shuttle_service::{
 };
 
 use anyhow::{anyhow, bail, Context, Result};
-use cargo_metadata::Message;
 use clap::{parser::ValueSource, CommandFactory, FromArgMatches};
 use clap_complete::{generate, Shell};
 use config::RequestContext;
@@ -1100,15 +1099,10 @@ impl Shuttle {
     async fn pre_local_run(&self, run_args: &RunArgs) -> Result<Vec<BuiltService>> {
         trace!("starting a local run for a service: {run_args:?}");
 
-        let (tx, rx): (crossbeam_channel::Sender<Message>, _) = crossbeam_channel::bounded(0);
-        tokio::task::spawn_blocking(move || {
-            while let Ok(message) = rx.recv() {
-                match message {
-                    Message::TextLine(line) => println!("{line}"),
-                    message => {
-                        trace!("skipping cargo line: {message:?}")
-                    }
-                }
+        let (tx, mut rx) = tokio::sync::mpsc::channel::<String>(256);
+        tokio::task::spawn(async move {
+            while let Some(line) = rx.recv().await {
+                println!("{line}");
             }
         });
 
@@ -1140,6 +1134,7 @@ impl Shuttle {
 
     #[cfg(target_family = "unix")]
     async fn local_run(&self, mut run_args: RunArgs) -> Result<CommandOutcome> {
+        debug!("starting local run");
         let services = self.pre_local_run(&run_args).await?;
         let (provisioner_server, provisioner_port) = Shuttle::setup_local_provisioner().await?;
         let mut sigterm_notif =
--- a/deployer/src/deployment/queue.rs
+++ b/deployer/src/deployment/queue.rs
@@ -1,12 +1,10 @@
 use std::collections::{BTreeMap, HashMap};
 use std::fmt;
-use std::io::{BufRead, BufReader, Read};
+use std::io::Read;
 use std::path::{Path, PathBuf};
-use std::process::{Command, Stdio};
+use std::process::Stdio;
 use std::time::Duration;
 
-use cargo_metadata::Message;
-use crossbeam_channel::Sender;
 use flate2::read::GzDecoder;
 use opentelemetry::global;
 use shuttle_common::{
@@ -22,6 +20,7 @@ use shuttle_service::builder::{build_workspace, BuiltService};
 use tar::Archive;
 use tokio::{
     fs,
+    io::AsyncBufReadExt,
     task::JoinSet,
     time::{sleep, timeout},
 };
@@ -147,10 +146,7 @@ pub async fn task(
 
 #[instrument(name = "Build failed", skip(_id), fields(deployment_id = %_id, state = %State::Crashed))]
 fn build_failed(_id: &Uuid, error: impl std::error::Error + 'static) {
-    error!(
-        error = &error as &dyn std::error::Error,
-        "{}", DEPLOYER_END_MSG_BUILD_ERR,
-    );
+    error!("{DEPLOYER_END_MSG_BUILD_ERR}: {error}");
 }
 
 #[instrument(name = "Waiting for queue slot", skip(queue_client), fields(deployment_id = %id, state = %State::Queued))]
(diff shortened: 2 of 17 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -1100,15 +1099,10 @@ impl Shuttle {
...
-                match message {
-                    Message::TextLine(line) => println!("{line}"),
-                    message => {
-                        trace!("skipping cargo line: {message:?}")
-                    }
-                }
+        let (tx, mut rx) = tokio::sync::mpsc::channel::<String>(256);
+        tokio::task::spawn(async move {
+            while let Some(line) = rx.recv().await {
+                println!("{line}");
```
oddgrd, line 1105: Will this add very verbose output (printing any message, not just Message::TextLine), or just the missing downloading crates output?

service/src/builder.rs
```diff
@@ -209,64 +240,52 @@ async fn compile(
...
-                error!("Failed to read Cargo log messages");
-            };
+    cmd.stderr(Stdio::piped());
+    cmd.stdout(Stdio::null());
+    let mut handle = cmd.spawn()?;
+    let reader = tokio::io::BufReader::new(handle.stderr.take().unwrap());
+    tokio::spawn(async move {
+        let mut lines = reader.lines();
+        while let Some(line) = lines.next_line().await.unwrap() {
+            let _ = tx.send(line).await.map_err(|e| error!("{e}"));
```
oddgrd, line 284: Would you mind explaining this change? It seems to me like we are first ignoring the stderr stream, and then we start reading it (`handle.stderr.take()`)? Or is this a better way to combine stderr and stdout into one stream?

deployer/src/deployment/queue.rs
```diff
@@ -432,18 +409,27 @@ async fn run_pre_deploy_tests(
...
-    }
+    });
+    let reader = tokio::io::BufReader::new(handle.stderr.take().unwrap());
+    tokio::spawn(async move {
+        let mut lines = reader.lines();
+        while let Some(line) = lines.next_line().await.unwrap() {
+            let _ = tx2.send(line).await.map_err(|e| error!("{e}"));
+        }
+    });
+    let status = handle.wait().await.map_err(TestError::Run)?;
```
oddgrd, line 430: Is opting not to use os_pipe here relevant to the bug fix (missing crates download output)? It was a pretty good and simple way of combining stderr and stdout, what is the advantage of doing it this way?

oddgrd (commented):
Thanks for getting to the bottom of this! I left some questions.

deployer/src/deployment/queue.rs
oddgrd: Oh, nevermind, it seems we didn't actually use os_pipe here, and we ignored stderr. So this change should be an improvement. We might still consider using os_pipe to combine stderr and stdout, WDYT?

cargo-shuttle/src/lib.rs
jonaro00: We were not sending any non-TextMessage messages on this channel, so it was a pointless wrapper. Probably an artifact of using cargo as a lib.

deployer/src/deployment/queue.rs
jonaro00: Main thing is that os_pipe and pipe and crossbeam are blocking, so I changed them all out for tokio async readers, which solved the bug, and should be more sound in an async app.

service/src/builder.rs
jonaro00: I've changed this implementation to only read stderr, as I discovered that cargo build seems to only print to stderr. `deployer/src/deployment/queue.rs` is an example where I open two readers for reading both streams.

oddgrd: Alright, thanks for the explanation!

deployer/src/handlers/mod.rs
```diff
@@ -795,6 +794,7 @@ pub async fn get_secrets(
     }
 }
 
+#[instrument(skip_all, fields(%project_name))]
```
chesedo, line 797: Fyi, `project_name` is always automatically traced for all deployer routes currently.

https://github.com/shuttle-hq/shuttle/blob/ffb760abb92f644cde9c777b6973e412b46c8ebe/deployer/src/handlers/mod.rs#L215-L221

service/src/builder.rs
```diff
@@ -69,22 +67,63 @@ fn extract_shuttle_toml_name(path: PathBuf) -> anyhow::Result<String> {
...
+        .arg("fetch")
+        .arg("--manifest-path")
+        .arg(&manifest_path)
+        .arg("--color=always")
+        .stdout(Stdio::null())
+        .stderr(Stdio::null());
+    let notification = tokio::spawn({
+        let tx = tx.clone();
+        async move {
+            tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
```
chesedo, line 92: Why the wait?

service/src/builder.rs
```diff
@@ -69,22 +67,63 @@ fn extract_shuttle_toml_name(path: PathBuf) -> anyhow::Result<String> {
...
+    if !command.status().await?.success() {
+        tx.send("      Failed to fetch crates".into())
+            .await
+            .expect("log receiver to exist");
+    }
+    notification.abort();
+
+    let metadata = {
+        // Modified implementaion of `MetadataCommand::exec` (v0.15.3).
+        // Uses tokio command instead of std
```
chesedo, line 107: Just add another comment for why this decision is made

service/src/builder.rs
chesedo, line 284: Just add an `error` field to the logged error

chesedo (commented):
Looking good. Comments are mostly small

deployer/src/handlers/mod.rs
jonaro00: Good to know! It was mainly for consistency in relation to the others.

service/src/builder.rs
jonaro00: To mimic cargo fetch: skip printing this line if there is nothing to download. Only if it takes more than 500ms do we assume a download is happening and print this.

deployer/src/persistence/mod.rs
```diff
@@ -691,7 +689,11 @@ impl StateRecorder for Persistence {
     type Err = Error;
 
     fn record_state(&self, state: DeploymentState) -> Result<()> {
-        self.state_send.send(state).map_err(Error::from)
+        let s = self.state_send.clone();
+        // move out of async context to be able to send
+        std::thread::spawn(move || s.blocking_send(state).map_err(Error::from))
```
iulianbarbu, line 694: why do we need a thread and call `join` and not call `blocking_send` directly?

jonaro00: A blocking send can't be done in a tokio thread, so that's why it needs to move out to a std thread. A `tokio::task::spawn_blocking` did not work.

jonaro00: Solution: use unbounded channel for instant sends https://docs.rs/tokio/latest/tokio/sync/mpsc/index.html#communicating-between-sync-and-async-code

PR #1351, opened 2023-10-27
feat(logger): rate limit based on peer address

oddgrd:
## Description of change

Add rate limiting by peer address to the logger server. This sets the sustained load LPS limit for projects to 512, but allowing bursts of up to 256 * 6. The rate limiter will regenerate a slot every 0.5s.

Tested by running projects locally that went over the limit. I also created an integration test to verify the limits work as expected, and the returned tonic status and headers are as expected when rate limiting is applied.


cargo-shuttle/src/lib.rs
```diff
@@ -707,9 +707,22 @@ impl Shuttle {
...
-                        .context("Failed parsing logs. Is your cargo-shuttle outdated?")?;
-                    println!("{log_item}")
+                    match serde_json::from_str::<shuttle_common::LogItem>(&line) {
+                        Ok(log_item) => {
+                            println!("{log_item}")
+                        }
+                        Err(err) => {
+                            debug!(error = %err, "failed to parse logs");
+
+                            let message = if line.contains("rate limit") {
```
oddgrd, line 717: I'm not too happy about how I handle this case (string matching), but it should be rare (it will only happen when the application is being rate limited and the user tries to request logs, and it won't always fail). Suggestions on how to improve the error handling here is most welcome.

Kazy: What about having the error be an actual JSON as well ? You could deserialize this as an enum (log line from the app or something else), and match on that instead.

deployer/src/handlers/error.rs
```diff
@@ -26,6 +26,8 @@ pub enum Error {
     Internal(#[from] anyhow::Error),
     #[error("Missing header: {0}")]
     MissingHeader(String),
+    #[error("{0}. Retry the command in a few minutes")]
```
chesedo, line 29: Why is rate limiting telling the user to retry a command?

deployer/src/handlers/mod.rs
```diff
@@ -660,16 +660,27 @@ pub async fn get_logs(
...
-        ))
-    } else {
-        Err(Error::NotFound("deployment not found".to_string()))
+        )),
+        Err(error) => {
+            if error.code() == tonic::Code::Unavailable
+                && error.metadata().get("x-ratelimit-limit").is_some()
+            {
+                Err(Error::RateLimited(
+                    "your application is producing too many logs, any interaction with the shuttle logger service will be rate limited"
```
chesedo, line 677:
```suggestion
                    "your application is producing too many logs. Interactions with the shuttle logger service will be rate limited for a while"
```

deployer/src/handlers/mod.rs
```diff
@@ -660,16 +660,27 @@ pub async fn get_logs(
...
+    match client.get_logs(logs_request).await {
+        Ok(logs) => Ok(Json(
             logs.into_inner()
                 .log_items
                 .into_iter()
                 .map(|l| l.to_log_item_with_id(deployment_id))
                 .collect(),
-        ))
-    } else {
-        Err(Error::NotFound("deployment not found".to_string()))
```
chesedo, line 672: Is the not found case no longer relevant?

deployer/src/handlers/mod.rs
```diff
@@ -719,9 +730,22 @@ async fn logs_websocket_handler(
...
-            let _ = s
-                .send(ws::Message::Text("failed to get logs".to_string()))
-                .await;
+            if error.code() == tonic::Code::Unavailable
+                && error.metadata().get("x-ratelimit-limit").is_some()
+            {
+                let _ = s
+                    .send(ws::Message::Text(
+                        Error::RateLimited(
+                            "your application is producing too many logs, any interaction with the shuttle logger service will be rate limited"
```
chesedo, line 739: We should improve this wording in the same way we decide to improve the earlier wording

logger/tests/integration_tests.rs
```diff
@@ -199,10 +301,27 @@ mod needs_docker {
 
         exec_psql(&format!(r#"CREATE DATABASE "{}";"#, &db_name));
 
+        let governor_config = GovernorConfigBuilder::default()
+            .per_millisecond(500)
+            .burst_size(6)
+            .use_headers()
+            .key_extractor(TonicPeerIpKeyExtractor)
+            .finish()
+            .unwrap();
```
chesedo, line 310: In theory we could make changes to the main code (rate limiter) and not catch it in the test because of this repeat. Maybe there should be a helper function in the main code that returns the governor? Or something that adds it to the server?

proto/src/lib.rs
```diff
@@ -344,14 +344,72 @@ pub mod logger {
...
+                                        nanos: timestamp.timestamp_subsec_nanos() as i32,
+                                    }),
+                                    service_name: Backend::Runtime(service_name.clone())
+                                        .to_string(),
+                                    data: "your application is producing too many logs, log recording is being rate limited".into(),
+                                }),
+                            };
+
+                            // Give the rate limiter time to refresh.
+                            tokio::time::sleep(Duration::from_millis(1500)).await;
```
chesedo, line 386: We should somehow make it clear how to calculate this number (1500ms) in the future. Maybe a basic explanation of how the rate limiter clears up after a second and where this clearup is configured

proto/src/lib.rs
```diff
@@ -379,7 +437,7 @@ pub mod logger {
         /// Create a batcher around inner. It will send a batch of items to inner if a capacity of 2048 is reached
         /// or if an interval of 1 second is reached.
         pub fn wrap(inner: I) -> Self {
-            Self::new(inner, 2048, Duration::from_secs(1))
+            Self::new(inner, 256, Duration::from_secs(1))
```
chesedo, line 440: Just need to update the function comment too :smile:

chesedo (commented):
Left some notes :smile:

deployer/src/handlers/mod.rs
oddgrd: No, I don't think it ever was. The https://github.com/shuttle-hq/shuttle/blob/main/logger/src/lib.rs#L89-L100 call can fail in the following ways:
- can't extract the claim from extensions, this shouldn't happen since we check the claim in the deployer scopedlayers. This would return internal error
- claim is missing the logs scope, shouldn't happen because we check it in deployer scopedlayer. This would return permission denied, I can add a match for this to the deployer handler, even though it is currently impossible, if we make changes in the deployer that may change.
- the db query fails, this would return internal error.

logger/tests/integration_tests.rs
oddgrd: Ah, yes, I tried to extract the service builder into a utility function, but I couldn't get it to compile with the complicated types. I tried again now to just extract the config creation into a util, but that was also difficult due to some private types. I will add constants for these values, it's not the cleanest solution, but it should do the trick.

proto/src/lib.rs
oddgrd: :see_no_evil:

deployer/src/handlers/mod.rs
oddgrd: Hm, after thinking about this some more, if the caller has access to this endpoint, but the logger client doesn't have access to get_logs, I think it makes sense to return an internal error from the deployer as well.

proto/src/lib.rs
oddgrd: Done, thanks!

proto/src/lib.rs
```diff
@@ -344,14 +344,79 @@ pub mod logger {
...
 
         async fn receive(&mut self, items: Vec<Self::Item>) {
+            // A log vector should never be received without any items. We clone the first item
+            // here so we can use IDs to generate a rate limiting logline, which we will send to
+            // the logger as a warning to the user that they are being rate limited.
+            let Some(item) = items.first().cloned() else {
+                error!("received log vector without any items");
+
+                return;
+            };
```
jonaro00, line 354: `items::is_empty()`?

logger/src/rate_limiting.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn key_name(&self, key: &Self::Key) -> Option<String> {
+        Some(key.to_string())
+    }
+}
+
+/// Convert errors from the Governor rate limiter layer to tonic statuses.
+pub fn tonic_error(e: BoxError) -> tonic::Status {
+    if e.is::<GovernorError>() {
+        // It shouldn't be possible for this to panic, since we already know it's a GovernorError
+        let error = e.downcast_ref::<GovernorError>().unwrap().to_owned();
```
jonaro00, line 43: You can use downcast_ref in the if clause instead

jonaro00 (approved):
LGTM

proto/src/lib.rs
oddgrd: We need the item here (`let Some(item)`) in case we want to send an item to the logger about the caller being rate limited.

logger/src/rate_limiting.rs
oddgrd: Thanks!

chesedo (approved):
LGTM

PR #1354, opened 2023-10-28
refactor: fix ProjectName validation, custom Path extractor for parsing it

jonaro00:
## Description of change

Custom axum extractor that prints the parse error message for the custom deser.

Supersedes of #1236, Check only the last commits.
- [ ] TODO?: Move axum module somewhere else?
- [x] Q: Deployer's endpoints can skip using it due to ProjectNameGuard. Should they?

On normal API calls
image
On interaction in init (should this be red?)
image

+ added tests


oddgrd (approved):
This looks great to me, thanks! And thanks for the tests and thorough comments!

common/src/project.rs
```diff
@@ -3,146 +3,137 @@ use rustrict::{Censor, Type};
...
-/// - It does not contain profanity.
-/// - It is not a reserved word.
+/// Initially we'll implement a strict subset of the IETF RFC 1123.
+/// Additionaly, while host segments are technically case-insensitive, the filesystem isn't,
+/// so we restrict project names to be lower case. We also restrict the use of profanity,
+/// as well as a list of reserved words.
 #[derive(Clone, Serialize, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "backend", derive(sqlx::Type, Hash))]
+#[cfg_attr(feature = "backend", sqlx(transparent))]
 pub struct ProjectName(String);
```
chesedo, line 18: Does c-s use `ProjectName` in any way now? It might be worth moving it to the `backend` feature entirely?

chesedo (commented):
One suggestion. Btw, nicely done with the new tests

common/src/project.rs
jonaro00: It is exposed in `shuttle_metadata::Metadata`. I'm down to strip that away (make it String) and make ProjectName backend-only. Does that sound good?

chesedo (approved):
LGTM

PR #1357, opened 2023-10-30
feat: adapative page hints (client-side only)

x04:
## Description of change

Adds adaptive page hints to cargo-shuttle by adding 1 to the specified limit and checking if we reach that adjusted limit. This is not breaking as it requires no API changes and is just a small change to client-side logic. Only thing worth mentioning is technically you can only specify a limit up `u32::MAX - 1` because we always adjust the limit to be `limit + 1` however I cannot really see a case where someone would ever need that high of a limit so it is basically irrelevant.

Closes issue #1316 
Also can close my other 2 PRs #1334 and #1320 as this resolves all the issues addressed in those.

Tested with local setup

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -734,13 +734,20 @@ impl Shuttle {
             println!();
             return Ok(CommandOutcome::Ok);
         }
+        let limit = limit + 1;
 
         let proj_name = self.ctx.project_name();
-        let deployments = client
+        let mut deployments = client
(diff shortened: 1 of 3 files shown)


x04:
Only check that failed was an out of date crate and I can update that but will wait for confirmation that is the path to take.

jonaro00 (approved):
Thanks for being patient and arriving at this nice solution! Much cleaner than the others :)
I added some small changes.

oddgrd (approved):
Great, thank you! :partying_face:

PR #1359, opened 2023-10-31
feat(deployer): delete secrets from deployer persistence on secrets resource delete

oddgrd:
## Description of change

This ensures that when deleting the secrets resource, the secrets are also deleted from the deployer persistence secrets table.

I still want to explore whether secrets should still be persisted in the deployer, and perhaps if secrets should be deletable by key, but that's for a future PR.

I first reproduced the bug locally by: deploying with secrets, deleting secrets and removing Secrets.toml, re-deploying and secrets returned.

After my change, this was no longer the case, secrets remained deleted on re-deploy.


chesedo (approved):
LGTM

orhun (approved):

PR #1360, opened 2023-10-31
fix typos found in codebase

GlenDC:
Fix typos found while reading the codebase.

### DIFF
--- a/common/src/deployment.rs
+++ b/common/src/deployment.rs
@@ -56,7 +56,7 @@ pub const DEPLOYER_END_MSG_BUILD_ERR: &str = "Service build encountered an error
 pub const DEPLOYER_END_MSG_CRASHED: &str = "Service encountered an error and crashed";
 pub const DEPLOYER_END_MSG_STOPPED: &str = "Service was stopped by the user"; // don't include this in end messages so that logs are not stopped too early
 pub const DEPLOYER_END_MSG_COMPLETED: &str = "Service finished running all on its own";
-pub const DEPLOYER_RUNTIME_START_RESPONSE: &str = "Runtime started successully";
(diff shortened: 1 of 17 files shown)


orhun (approved):

common/src/deployment.rs
```diff
@@ -56,7 +56,7 @@ pub const DEPLOYER_END_MSG_BUILD_ERR: &str = "Service build encountered an error
 pub const DEPLOYER_END_MSG_CRASHED: &str = "Service encountered an error and crashed";
 pub const DEPLOYER_END_MSG_STOPPED: &str = "Service was stopped by the user"; // don't include this in end messages so that logs are not stopped too early
 pub const DEPLOYER_END_MSG_COMPLETED: &str = "Service finished running all on its own";
-pub const DEPLOYER_RUNTIME_START_RESPONSE: &str = "Runtime started successully";
+pub const DEPLOYER_RUNTIME_START_RESPONSE: &str = "Runtime started successfully";
```
jonaro00, line 59: Sadly, fixing this one is a breaking change. Will come back when I know what to do about it.

GlenDC: One of those horror stories where someone relies on a string match of an error or other kind of log? :O

GlenDC: Either way you should have edit rights, feel fee to revert this change and keep it for another time.

jonaro00: I might do that and attempt at making this more robust. Context about this design: When we reworked our logging system, we dumbed down some parts, for better or for worse.

jonaro00 (approved):
I've noted down the remaining typo. Will try to get that fixed in some other way.

PR #1364, opened 2023-11-02
feat: use proto-gen for generating proto code

jonaro00:
## Description of change

- Use new method for generating proto code that does not break when tonic or prost has breaking changes.
- Clippy warnings fixed in other PR

- Local run works with project on 0.31 and c-s on this branch, aka nothing in the protobufs on wire changed, only the rust lib code.
- added CI


.gitattributes
oddgrd, line 1: TIL: https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github, cool!

oddgrd (approved):
Cool, thanks for breaking this out, and thanks for making a note that you verified this isn't breaking in the testing :heart:

PR #1366, opened 2023-11-03
feat: suggest project restart when trying to delete

jonaro00:
## Description of change

Additional context when project is not ready on delete command.
image

See image above

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -1827,10 +1827,12 @@ impl Shuttle {
                     println!();
                     println!("{}", "Try restarting the project with `cargo shuttle project restart` first.".yellow());
                     println!("{}", "This is needed to check for any resources linked to it.".yellow());
+                    println!("{}", "For more help with deleting projects, visit https://docs.shuttle.rs/support/delete-project".yellow());
                     println!();
                     return err;
                 }
             }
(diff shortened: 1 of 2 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -1819,7 +1819,25 @@ impl Shuttle {
         let client = self.client.as_ref().unwrap();
 
         // If a check fails, print the returned error
-        client.delete_project(self.ctx.project_name(), true).await?;
+        client.delete_project(self.ctx.project_name(), true).await.map_err(|err| {
+            if let Some(api_error) = err.downcast_ref::<ApiError>() {
+                // If the returned error string changes, this could break
+                if api_error.message.contains("not ready") {
```
orhun, line 1825: Not sure if we should use plain string match or have a constant somewhere for the message and check that.

e.g.

```rs
const PROJECT_NOT_READY_MSG: &str =  "project not ready. Try running `cargo shuttle project restart`.";
```

It is something trivial but worth considering I think.

jonaro00: I used a small string for matching, because if an entire string is used like that, a single change to the string will be breaking, like we saw in #1360. In this scenario, it wouldn't break functionality, just the extra hint, but still.

oddgrd (approved):

PR #1369, opened 2023-11-07
ci: audit on main, build release stack sooner, release crates faster

jonaro00:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -196,7 +196,7 @@ jobs:
     executor: docker-rust
     steps:
       - checkout
-      - run: 
+      - run:
           name: Install cargo-audit
           command: |
             CARGO_AUDIT_VERSION='v0.18.3'
@@ -682,7 +682,6 @@ workflows:
   ci:
     jobs:
       - workspace-fmt-clippy
-      - cargo-audit
       - test-standalone:
           name: << matrix.path >>
           matrix:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -668,21 +668,19 @@ jobs:
...
-    resource_class: medium
     steps:
       - checkout
       - run:
-          name: Crate publishing in order
-          command: |
-            cargo publish --manifest-path << parameters.path >>/Cargo.toml
+          name: Publish crate
+          # Don't build the crate, we know that the crate should build based on other workflows
+          command: cargo publish --no-verify --manifest-path << parameters.path >>/Cargo.toml
```
oddgrd, line 679: I know this should be safe, but it still makes me nervous. :sweat_smile: I can't think of any edge cases where this might be a problem at the moment, so I think it's fine.

.circleci/config.yml
```diff
@@ -870,18 +872,21 @@ workflows:
             - build-binaries-aarch64
             - build-binaries-windows
             - build-binaries-mac
-      - approve-push-production:
-          type: approval
-          filters:
-            branches:
-              only: production
+      # time save: build and push without approval since a merge to production should want to deploy.
+      # deploy requires approval.
```
oddgrd, line 876: Nice!

oddgrd (approved):

.circleci/config.yml
jonaro00: If you want faster releases, we need to release faster :blush:

oddgrd: Indeed 😂

PR #1375, opened 2023-11-09
fix(cargo-shuttle): revert shuttle-common-tests to path dep

oddgrd:
## Description of change

`cargo-shuttle` failed to publish with `shuttle-common-tests` as a workspace dependency.

Tested cargo publish with --dry-run, then published the crate on this branch.


orhun (approved):

jonaro00:
Just add a comment about why, so we don't repeat the mistake :D

PR #1376, opened 2023-11-09
refactor: clean up deployer db resources, delete after sync to r-r, delete secrets command

jonaro00:
## Description of change

- Resources are deleted from deployer after they have been successfully synced to r-r.
- Deleted the `secrets` CLI command since it only targeted the deployer cache. The new way is to use `resource list`
- Drop secrets table. Is possible due to the above.

Tested with inserting a resource in deployer, and the resource list operation correctly synced them up to r-r. The project could then be deleted without hassle.


cargo-shuttle/src/args.rs
```diff
@@ -119,12 +119,6 @@ pub enum Command {
     /// Manage resources of a Shuttle project
     #[command(subcommand)]
     Resource(ResourceCommand),
-    /// Manage secrets for this Shuttle service
-    Secrets {
```
Kazy, line 123: **suggestion(non-blocking)**: And actually, are we sure we want to remove this command ? I know we're treating secrets as a resource now, but isn't that an implementation detail ? We could instead ask the r-r for just the secrets.

jonaro00: Could do that, yeah. I think it's more clear if they are only in one place where they currently logically belong. This command has been broken for quite some time.

jonaro00:
> Have you tested it through an update of Shuttle ?

My test with inserting an old record mimics the upgrade from an old version (based on what I observed during support week). I would have to go back several months to do it "naturally". To me, this is sufficient.

iulianbarbu (approved):
LGTM 🥳 ! Super happy about the simplifications, good job

cargo-shuttle/src/args.rs
iulianbarbu: I think it is worth simplifying this and removing the cmd, given the costs coming from maintaining, bugs fixing, and the support specifically for this cmd (and maybe feature requests?).

However, it is worth saying we remove it without a deprecation notice, and backward compat support e.g. previous versions of `cargo-shuttle` can call `secrets` cmd on a recent deployer, and I think it will fail, right? We would need to instruct users to use `resource list` instead, which shouldn't be that bad.

Also, it is worth saying secret retrieval is not extremely useful, given we're not displaying the actual values, so having a dedicated command just for them doesn't make sense to me. Down the line, secrets displaying (along other resources displaying) should be further refined to maybe display hints and data about when it was last modified (and possibly accessed?), but still, not sure if it is worth a dedicated cmd.

oddgrd: It seems to me like the command is currently broken, like Johan mentioned, so I think we can safely remove it. :smile:

Kazy: My idea was more to keep the `secrets` part but in a workable state :p But fine if you think resources make more sense :)

PR #1382, opened 2023-11-13
feat: add limits and tier to jwt claim

oddgrd:
## Description of change

Add a `Limits` struct that will hold all the account limits to the JWT `Claim`, and also add the `AccountTier` (moved to shuttle-common) to the `Claim`, so we can use it in instrumentation. I left adding it to instrumentation out of this PR, that will be done in a follow-up.

I used an extension trait for the methods to extract the limits from the `Claim`. This allows us to have the methods for extracting limits from the claim only available where they are needed. An alternative would be to implement the methods on Claim directly, or make the fields of `Claim` public, but that would make us more exposed to [breaking changes](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md#structs). I'm not sure that breaking changes will be an issue here, though, since we create the limits from the tier. Thoughts on this subject are very welcome!

Updated the `auth` test to verify that the limits and tier are correctly set for an admin and for a basic user, and verified that the gateway tests that test project limit still pass. 

Additionally, I tested deploying a hello-world and postgres project to staging on an old deployer, as well as a new deployer (after deploying this branch to staging).

### DIFF
--- a/auth/src/api/handlers.rs
+++ b/auth/src/api/handlers.rs
@@ -105,7 +105,12 @@ pub(crate) async fn convert_cookie(
         .get::<AccountTier>("account_tier")
         .ok_or(StatusCode::UNAUTHORIZED)?;
 
-    let claim = Claim::new(account_name, account_tier.into(), account_tier);
+    let claim = Claim::new(
+        account_name,
+        account_tier.into(),
+        account_tier,
+        account_tier,
+    );
 
     let token = claim.into_token(key_manager.private_key())?;
 
@@ -129,7 +134,12 @@ pub(crate) async fn convert_key(
         .await
         .map_err(|_| StatusCode::UNAUTHORIZED)?;
 
-    let claim = Claim::new(name.to_string(), account_tier.into(), account_tier);
+    let claim = Claim::new(
+        name.to_string(),
+        account_tier.into(),
+        account_tier,
+        account_tier,
+    );
 
     let token = claim.into_token(key_manager.private_key())?;
 
--- a/auth/src/api/handlers.rs
+++ b/auth/src/api/handlers.rs
@@ -1,6 +1,6 @@
 use crate::{
     error::Error,
-    user::{AccountName, AccountTier, Admin, Key, User},
+    user::{AccountName, Admin, Key, User},
 };
 use axum::{
     extract::{Path, State},
@@ -9,7 +9,10 @@ use axum::{
 use axum_sessions::extractors::{ReadableSession, WritableSession};
 use http::StatusCode;
 use serde::{Deserialize, Serialize};
-use shuttle_common::{claims::Claim, models::user};
+use shuttle_common::{
+    claims::{AccountTier, Claim},
+    models::user,
+};
 use stripe::CheckoutSession;
 use tracing::instrument;
 
@@ -102,7 +105,12 @@ pub(crate) async fn convert_cookie(
         .get::<AccountTier>("account_tier")
         .ok_or(StatusCode::UNAUTHORIZED)?;
 
-    let claim = Claim::new(account_name, account_tier.into(), Some(account_tier.into()));
+    let claim = Claim::new(
+        account_name,
+        account_tier.into(),
+        Some(account_tier.into()),
+        Some(account_tier),
+    );
 
     let token = claim.into_token(key_manager.private_key())?;
 
@@ -130,6 +138,7 @@ pub(crate) async fn convert_key(
         name.to_string(),
         account_tier.into(),
         Some(account_tier.into()),
+        Some(account_tier),
     );
 
     let token = claim.into_token(key_manager.private_key())?;
--- a/auth/src/api/handlers.rs
+++ b/auth/src/api/handlers.rs
@@ -102,7 +102,7 @@ pub(crate) async fn convert_cookie(
         .get::<AccountTier>("account_tier")
         .ok_or(StatusCode::UNAUTHORIZED)?;
 
-    let claim = Claim::new(account_name, account_tier.into());
+    let claim = Claim::new(account_name, account_tier.into(), Some(account_tier.into()));
 
     let token = claim.into_token(key_manager.private_key())?;
 
@@ -126,7 +126,11 @@ pub(crate) async fn convert_key(
         .await
         .map_err(|_| StatusCode::UNAUTHORIZED)?;
 
-    let claim = Claim::new(name.to_string(), account_tier.into());
+    let claim = Claim::new(
+        name.to_string(),
+        account_tier.into(),
+        Some(account_tier.into()),
+    );
 
     let token = claim.into_token(key_manager.private_key())?;
 
--- a/auth/src/api/handlers.rs
+++ b/auth/src/api/handlers.rs
@@ -105,12 +105,7 @@ pub(crate) async fn convert_cookie(
         .get::<AccountTier>("account_tier")
         .ok_or(StatusCode::UNAUTHORIZED)?;
 
-    let claim = Claim::new(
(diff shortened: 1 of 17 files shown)


auth/src/api/handlers.rs
```diff
@@ -102,7 +105,7 @@ pub(crate) async fn convert_cookie(
         .get::<AccountTier>("account_tier")
         .ok_or(StatusCode::UNAUTHORIZED)?;
 
-    let claim = Claim::new(account_name, account_tier.into());
+    let claim = Claim::new(account_name, account_tier.into(), account_tier);
```
oddgrd, line 108: We could have only two args here, subject and account tier, then we create the scopes from the tier in the body of the new method as well. That would not allow us to manually add single scopes for testing, which [we do currently](https://github.com/shuttle-hq/shuttle/pull/1382/files#diff-37cffdd9fbd9d2426e519e0d75b4bf560cc755d5a11ee15b8977a338747253c3), but I'm not sure how useful it is, since those tests only add the minimum required scopes (basic would work too).

common/Cargo.toml
```diff
@@ -64,7 +64,8 @@ backend = [
     "tracing-subscriber/fmt",
     "ttl_cache",
     "sqlx",
-    "sqlx/postgres",                 # Fix for derive macro error when different backends enable sqlite & postgres (mainly for grouped compilation in clippy CI and Docker build)
+    "sqlx/postgres",            # Fix for derive macro error when different backends enable sqlite & postgres (mainly for grouped compilation in clippy CI and Docker build)
+    "sqlx/sqlite",
```
oddgrd, line 68: This is needed to move AccountTier from Auth to common.

common/src/claims.rs
```diff
@@ -166,6 +169,55 @@ impl Default for ScopeBuilder {
...
+}
+
+impl From<AccountTier> for Limits {
+    fn from(value: AccountTier) -> Self {
+        match value {
+            AccountTier::Basic | AccountTier::PendingPaymentPro | AccountTier::Deployer => {
+                Self::default()
+            }
+            AccountTier::Pro | AccountTier::Team => Self::new(MAX_PROJECTS_EXTRA),
+            AccountTier::Admin => Self::new(100),
```
oddgrd, line 196: This is a bit arbitrary, since we currently ignore this limit if the claim has the `Scope::Admin` variant. Maybe this one should be default as well?

common/src/claims.rs
```diff
@@ -182,11 +234,45 @@ pub struct Claim {
...
+}
+
+impl Limits {
+    pub fn new(project_limit: u32) -> Self {
+        Self { project_limit }
+    }
+}
+
+pub trait ClaimExt {
+    fn project_limit(&self) -> u32;
```
chesedo, line 264: It would be better to have this method as `can_create_project()` to prevent the logic from spilling into the handlers (controllers).

chesedo: It would also be good to put this trait in a separate file.

common/src/claims.rs
```diff
@@ -182,11 +234,45 @@ pub struct Claim {
...
+    fn project_limit(&self) -> u32 {
+        self.limits.project_limit
+    }
 }
 
 impl Claim {
-    /// Create a new claim for a user with the given scopes
-    pub fn new(sub: String, scopes: Vec<Scope>) -> Self {
+    /// Create a new claim for a user with the given scopes and limits.
+    pub fn new(sub: String, scopes: Vec<Scope>, tier: AccountTier) -> Self {
```
chesedo, line 275: Maybe we should add a `limits: Into<Limits>` to communicate the intent better?

common/src/claims.rs
chesedo: Yip, we can make it default too

chesedo (commented):
Looking good, thanks O!

common/src/claims.rs
oddgrd: I originally passed in Limits to this constructor as well, but it seemed redundant, since it's always created from the tier. That may change in the future though, and not having it in the signature does kinda hide it. I can add it back. :+1:

auth/src/api/handlers.rs
oddgrd: I'll leave it in for readability.

common/src/claims.rs
jonaro00: > It would also be good to put this trait in a separate file.

Why?

> can_create_project

Are you saying that the business logic should live in that function? Then it has to take in `current_project_count` or something.

jonaro00 (commented):
You could mark some of the now-outdated scopes as "deprecated but keep for compatibility"

common/src/claims.rs
oddgrd: Hmm, I started implementing this now, but I'm not sure about this. We will essentially be passing `account_tier` to Claim::new three times: 
```
let claim = Claim::new(name.to_string(), account_tier.into(), account_tier, account_tier);
```
Maybe we should just update the comment to make it clear that this will take a subject, a scope and a tier, and it will create a limits struct from the tier? I don't love that either, but maybe it's good enough.

common/src/claims.rs
oddgrd: I implemented these changes last night: [f900d1a](https://github.com/shuttle-hq/shuttle/pull/1382/commits/f900d1a379e792e7bf50eaafdf5b354869650c94)

It feels more organized, but it meant needing to make some changes to the test code in gateway.

common/src/claims.rs
jonaro00: Don't overthink it :)

oddgrd: Added it :+1:

common/src/claims.rs
chesedo: @jonaro00 to have the business logic in one file. Thereby being easier to update :smile:

gateway/src/service.rs
```diff
@@ -437,7 +437,7 @@ impl GatewayService {
         project_name: ProjectName,
         account_name: AccountName,
         is_admin: bool,
-        limit: Option<u32>,
+        can_create_project: bool,
         idle_minutes: u64,
     ) -> Result<FindProjectPayload, Error> {
```
chesedo, line 442: Technically this function should not take in an argument to check if a project can be created. I think it is fine for now since I  don't see an easy way to move the logic to the caller.

gateway/src/service.rs
```diff
@@ -966,13 +955,32 @@ pub struct FindProjectPayload {
...
+            return true;
+        };
+
+        let project_count = service
+            .get_project_count(account_name)
+            .await
+            .expect("to get the account's project count");
+
+        project_count < MAX_PROJECTS_DEFAULT
+    }
```
chesedo, line 982: Ideally tests should not have logic that replicates the code being tested. Can we not just pass `true` and `false` into the test arguments?

chesedo: Replicating the logic means your test won't identify any errors in the replicated code.

chesedo (commented):
Just one comment about the function in the test

gateway/src/service.rs
oddgrd: I didn't want to remove our only test of project limits, but I guess replicating the logic here does not resolve that. I think this should rather be tested at a higher level (handler tests), since it's now checked there. Thanks! I will remove this change.

oddgrd: I added a new test for project limits. These tests could use some more work I think, but they're not run in CI, and we may be rewriting them for the new orchestrator anyway.

chesedo (approved):
LGTM

PR #1387, opened 2023-11-15
fix(gateway): use project_id argument when inserting project

Kazy:
## Description of change

This should only fix the lowercase Ulid issue, but not the duplicate insertion.

Not at all yet.


jonaro00 (approved):
It should be self-evident to work. If you want to test it, it would be: Start a new project, deploy something with a resource. Restart project. Deploy again with resource. Check that r-r does not have duplicate row.

PR #1388, opened 2023-11-16
feat(cargo-shuttle): generate manpage

### LINKED ISSUE #1377: [Improvement]: Generate manpage with `clap-mangen` for `cargo-shuttle`
### Describe the improvement

We use [`clap`](https://crates.io/crates/clap) for parsing the command-line arguments for `cargo-shuttle`. We can also integrate [`clap-mangen`](https://crates.io/crates/clap-mangen) for generating a manpage. I see two ways of doing this:

- have a separate binary (`cargo-shuttle-mangen`) (see [this](https://github.com/orhun/halp/blob/main/src/bin/mangen.rs) example)
- use the build script (see [this](https://github.com/orhun/kmon/blob/master/build.rs) example)

Here's why having a manpage is good:

> Having a man page is advantageous because it serves as local and offline documentation for commands or programs, offering users quick and standardized access to crucial information. Its consistency, comprehensive details, and integration with the command line make it a valuable resource for understanding and utilizing software effectively. The accessibility of man pages promotes inclusivity and facilitates scripting and automation, contributing to an enhanced user experience in various computing environments.

### Duplicate declaration

selectiveduplicate:
Add the ability to generate man page for `cargo-shuttle`.

## Description of change
Introduces a new option `--manpage` for the `generate` sub-command to generate man page with [clap-mangen](https://crates.io/crates/clap-mangen).

To verify the changes, we can run the following from the project root `shuttle`:
```sh
cargo run --package cargo-shuttle -- generate --manpage
```
This generates the man page in the current working directory by default. We can set a custom directory by setting the `OUT_DIR` environment variable:
```sh
OUT_DIR=PATH_TO_CUSTOM_DIR cargo run --package cargo-shuttle -- generate --manpage
```
Closes #1377.

Please see the change description above.

### DIFF
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -131,7 +131,7 @@ pub enum Command {
     Login(LoginArgs),
     /// Log out of the Shuttle platform
     Logout(LogoutArgs),
-    /// Generate shell completions
+    /// Generate shell completions and manpage
     Generate {
         /// Which shell
         #[arg(short, long, env, default_value_t = Shell::Bash)]
@@ -139,6 +139,10 @@ pub enum Command {
         /// Output to a file (stdout by default)
         #[arg(short, long, env)]
         output: Option<PathBuf>,
+        /// Generate manpage in the directory specified by the `OUT_DIR`
+        /// environment variable (current working directory by default)
+        #[arg(short, long)]
+        manpage: bool,
     },
     /// Open an issue on GitHub and provide feedback
     Feedback,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1131,6 +1131,7 @@ dependencies = [
  "chrono",
  "clap",
  "clap_complete",
+ "clap_mangen",
  "crossterm 0.27.0",
  "dialoguer",
  "dirs 5.0.1",
@@ -1304,6 +1305,16 @@ name = "clap_lex"
 version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "702fc72eb24e5a1e48ce58027a675bc24edd52096d5397d4aea7c6dd9eca0bd1"
+
+[[package]]
+name = "clap_mangen"
+version = "0.2.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d3be86020147691e1d2ef58f75346a3d4d94807bfc473e377d52f09f0f7d77f7"
+dependencies = [
+ "clap",
+ "roff",
+]
 
 [[package]]
 name = "clru"
@@ -5124,6 +5135,12 @@ dependencies = [
  "rmp",
  "serde",
 ]
+
+[[package]]
+name = "roff"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b833d8d034ea094b1ea68aa6d5c740e0d04bad9d16568d08ba6f76823a114316"
 
 [[package]]
 name = "rsa"
--- a/cargo-shuttle/Cargo.toml
+++ b/cargo-shuttle/Cargo.toml
@@ -20,6 +20,7 @@ cargo_metadata = { workspace = true }
 chrono = { workspace = true }
 clap = { workspace = true, features = ["env"] }
 clap_complete = "4.3.1"
+clap_mangen = "0.2.15"
 crossterm = { workspace = true }
 dialoguer = { version = "0.11.0", features = ["fuzzy-select"] }
 dirs = { workspace = true }
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -14,6 +14,8 @@ use std::path::{Path, PathBuf};
 use std::process::exit;
 use std::str::FromStr;
 
+use clap_mangen::Man;
+
 use shuttle_common::{
     claims::{ClaimService, InjectPropagation},
     constants::{
@@ -205,7 +207,11 @@ impl Shuttle {
                 self.init(init_args, args.project_args, provided_path_to_init)
                     .await
             }
-            Command::Generate { shell, output } => self.complete(shell, output),
+            Command::Generate {
+                shell,
+                output,
+                manpage,
+            } => self.complete(shell, output, manpage),
             Command::Login(login_args) => self.login(login_args).await,
             Command::Logout(logout_args) => self.logout(logout_args).await,
             Command::Feedback => self.feedback(),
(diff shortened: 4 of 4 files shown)


cargo-shuttle/src/args.rs
```diff
@@ -131,14 +131,18 @@ pub enum Command {
     Login(LoginArgs),
     /// Log out of the Shuttle platform
     Logout(LogoutArgs),
-    /// Generate shell completions
+    /// Generate shell completions and manpage
     Generate {
         /// Which shell
         #[arg(short, long, env, default_value_t = Shell::Bash)]
         shell: Shell,
```
orhun, line 138: I think we need to make this variable optional. Otherwise:

```sh
$ cargo shuttle generate --manpage

error: invalid value '/bin/bash' for '--shell <SHELL>'
  [possible values: bash, elvish, fish, powershell, zsh]

For more information, try '--help'.
```

cargo-shuttle/src/args.rs
```diff
@@ -131,14 +131,18 @@ pub enum Command {
     Login(LoginArgs),
     /// Log out of the Shuttle platform
     Logout(LogoutArgs),
-    /// Generate shell completions
+    /// Generate shell completions and manpage
```
orhun, line 134: Can you also update `README.md` about this description change?

cargo-shuttle/src/args.rs
```diff
@@ -131,14 +131,18 @@ pub enum Command {
...
         /// Which shell
         #[arg(short, long, env, default_value_t = Shell::Bash)]
         shell: Shell,
         /// Output to a file (stdout by default)
         #[arg(short, long, env)]
         output: Option<PathBuf>,
+        /// Generate manpage in the directory specified by the `OUT_DIR`
+        /// environment variable (current working directory by default)
+        #[arg(short, long)]
+        manpage: bool,
```
orhun, line 145: How about we don't use `OUT_DIR` and print it out to `stdout` instead? (similar to shell completions)

orhun (commented):
Looks good, just needs a bit of tweaking 🐻

cargo-shuttle/src/args.rs
selectiveduplicate: > I think we need to make this variable optional. Otherwise:

You mean the `shell` option?

jonaro00: When running `generate`, the `--shell <shell>` OR `--manpage` flag has to be given, but not both at the same time. There are flags to set in the clap macros for making this behavior (I wonder if the env loading on the shell arg messes this up?). The description could say "Generate shell completions or manpage".

It's almost as if they could be turned into subcommands rather than flags :thinking:, but that would be a breaking change.

selectiveduplicate: From my testing, the `--shell` option *must* be given, whether or not `--manpage` is there or not. Just specifying `--manpage` seems to error out like @orhun mentioned. :confused:

orhun: Yeah, I would say let's make both `--shell` and `--manpage` optional and have a condition to generate them accordingly.

selectiveduplicate: As far as I understand, we may need to remove the `env` attribute from the `shell` option. Because otherwise I think it's picked up automatically by clap through the `SHELL` environment variable. And since we're making it `Option<Shell>`, we may also need to remove the `default_value_t` attribute.

orhun: > As far as I understand, we may need to remove the env attribute from the shell option.

Oooh, that's why I got "error: invalid value '/bin/bash'" (I have `SHELL` set to `bash`) Good catch!

Alternatively we can rename it to `SHELL_ARG` or something.

> And since we're making it Option<Shell>, we may also need to remove the default_value_t attribute.

I think this is fine.

jonaro00: I think removing the env parse on SHELL and default value is fine. Explicit arg makes it more fool proof.

cargo-shuttle/src/args.rs
jonaro00: Wait a minute... `--shell` supports out dir... I gotta say though, it feels cleaner to only allow stdout, and let the user pipe to a file.

cargo-shuttle/src/args.rs
```diff
@@ -131,14 +131,17 @@ pub enum Command {
...
-        #[arg(short, long, env, default_value_t = Shell::Bash)]
-        shell: Shell,
+        #[arg(short, long)]
+        shell: Option<Shell>,
         /// Output to a file (stdout by default)
         #[arg(short, long, env)]
         output: Option<PathBuf>,
+        /// Generate manpage to the standard output
+        #[arg(short, long)]
+        manpage: bool,
```
jonaro00, line 144: Since they are mutually exclusive, it feels cleaner if they were separate subcommands: `cargo shuttle generate shell <SHELL>` and `cargo shuttle generate manpage`. Although a breaking change, could this be acceptable? @oddgrd

jonaro00: The `OUTPUT` env var usage feels weird as well. Could be a risk of collision with other programs using it. How do we feel about it?

cargo-shuttle/src/args.rs
selectiveduplicate: Agree. File I/O can be expensive, and imo the logic inside `complete` would seem simpler if we were to only support stdout for shell completions as well.

cargo-shuttle/src/args.rs
orhun: Yeah using `OUTPUT` is a bit risky. OTOH although the subcommand approach might look cleaner it would add unnecessary complexity to the code IMO. But I'm fine with either way - and I would say let's have this ready soon and not get caught up into discussion too much 🐻

oddgrd: > Since they are mutually exclusive, it feels cleaner if they were separate subcommands: `cargo shuttle generate shell <SHELL>` and `cargo shuttle generate manpage`. Although a breaking change, could this be acceptable?

I think having them as separate subcommands makes the most sense. Regarding the `output` arg, we could remove the env. That makes two breaking changes, but I doubt it will affect many (if any) users.

selectiveduplicate:
I've added the changes and updated the PR description. Apologies for any mistakes.

jonaro00 (commented):
It works, but the manpage only contains the global args and references other manpages for each subcommand (not very useful in our case), and I can find any way to generate proper manpages for subcommands or make the default manpage contain the manuals for all subcommands :thinking:

selectiveduplicate:
> It works, but the manpage only contains the global args and references other manpages for each subcommand (not very useful in our case), and I can find any way to generate proper manpages for subcommands or make the default manpage contain the manuals for all subcommands 🤔

Yeah you're right; it doesn't have much :sweat_smile: . I think to have manual for the sub-commands as well, we need to iterate over them using `get_subcommands`.

selectiveduplicate:
And since we're outputting to stdout, I wonder how we can do separate man pages for the sub-commands. As for dumping everything into one manual, maybe we can do this:

```rust
fn generate_manpage(&self) -> Result<CommandOutcome> {
    let app = Command::command();
    let output = std::io::stdout();
    let mut output_handle = output.lock();
    Man::new(app.clone()).render(&mut output_handle)?;
    for subcommand in app.get_subcommands().cloned() {
        Man::new(subcommand.clone()).render(&mut output_handle)?;
        if subcommand.has_subcommands() {
            for sb in subcommand.get_subcommands().cloned() {
                Man::new(sb).render(&mut output_handle)?;
            }
        }
    }

    Ok(CommandOutcome::Ok)
}
```

Just wanted to discuss the approach without bloating the commit history. :sweat_smile:

jonaro00:
I quickly tested changing the input Command to Man to be a subcommand, but the result was an empty manpage, so not sure what was happening there.

jonaro00:
If you find a way to get everything into one manual, go for it.

jonaro00:
Cool. I see that a lot of parts are repeated (title and footer for each manpage), so we could customize which parts of the manpage are printed for each section to make it more unified.

jonaro00:
They do have `render_*` methods for custom rendering: https://docs.rs/clap_mangen/latest/clap_mangen/struct.Man.html#method.render_title

selectiveduplicate:
We should have less repetitions now.

jonaro00 (approved):

PR #1394, opened 2023-11-16
misc(auth): added service healthcheck

iulianbarbu:
## Description of change

Use an auth service health-check when deploying the stack.

Locally, by querying the container state before and after the health check addition.
E.g. before the health-check, this `docker inspect --format='{{json .State.Health}}' <container_id>`, returns `null`,  while after adding the healthcheck, it shows something along:
```
{"Status":"healthy","FailingStreak":0,"Log":[{"Start":"2023-11-16T15:10:22.278007509Z","End":"2023-11-16T15:10:22.362298051Z","ExitCode":0,"Output":"  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n                                 Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\n"},{"Start":"2023-11-16T15:11:22.371958801Z","End":"2023-11-16T15:11:22.447879968Z","ExitCode":0,"Output":"  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n                                 Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\n"},{"Start":"2023-11-16T15:12:22.456275676Z","End":"2023-11-16T15:12:22.543789593Z","ExitCode":0,"Output":"  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n                                 Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\n"},{"Start":"2023-11-16T15:14:17.750973049Z","End":"2023-11-16T15:14:17.832457424Z","ExitCode":0,"Output":"  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n                                 Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\n"}]}
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -55,8 +55,8 @@ services:
       - "--stripe-secret-key=${STRIPE_SECRET_KEY}"
     healthcheck:
       test: curl --fail http://localhost:8000/ || exit 1
-      interval: 1m
-      timeout: 15s
+      interval: 15s
+      timeout: 10s
       retries: 15
   builder:
     image: "${CONTAINER_REGISTRY}/builder:${BUILDER_TAG}"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -55,9 +55,11 @@ services:
       - "--stripe-secret-key=${STRIPE_SECRET_KEY}"
     healthcheck:
       test: curl --fail http://localhost:8000/ || exit 1
-      interval: 15s
+      interval: 1m
       timeout: 10s
-      retries: 15
+      retries: 5
+      start_period: 10s
+      start_interval: 2s
   builder:
     image: "${CONTAINER_REGISTRY}/builder:${BUILDER_TAG}"
     depends_on:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,6 +53,11 @@ services:
       - "start"
       - "--address=0.0.0.0:8000"
       - "--stripe-secret-key=${STRIPE_SECRET_KEY}"
+    healthcheck:
+      test: curl --fail http://localhost:8000/ || exit 1
+      interval: 1m
+      timeout: 15s
+      retries: 15
   builder:
     image: "${CONTAINER_REGISTRY}/builder:${BUILDER_TAG}"
     depends_on:


jonaro00 (approved):

docker-compose.yml
```diff
@@ -53,6 +53,11 @@ services:
       - "start"
       - "--address=0.0.0.0:8000"
       - "--stripe-secret-key=${STRIPE_SECRET_KEY}"
+    healthcheck:
+      test: curl --fail http://localhost:8000/ || exit 1
+      interval: 1m
```
Kazy, line 58: Why use an interval ? If it fails it's going to be delayed it by 1 full minute, but I would expect the service to start much faster than that.

iulianbarbu: Haven't thought about a particular reason for `1m` and I don't think we need it to be that long, but we need it for the retries. Do you have a suggestion for how to set it or do you suggest removing it?

> If it fails it's going to be delayed by 1 full minute

Delayed in what sense?

jonaro00: > Delayed in what sense?

Delayed up to 1 minute until the health check discovers the failure. This also applies to time from startup: https://stackoverflow.com/a/69449702

15 retires seems a bit much, maybe 2 or 3 can be considered an unhealthy auth service

iulianbarbu: Ah, you said shorten it. I think 15s with 10s timeout can also work.

Kazy: Sounds good, we can always see how it goes and adapt it if it's taking too much time :+1:

iulianbarbu: Sorry guys, just reread the docs. It looks like the health check is not run necessarily to detect that the service failed at the startup, but while the service is running. There are two portions: `start_period` considered for the container/service bootstrap, and `start_interval` available for >= Docker Engine 25.0 😞 (which we don't use on prod, but hey, ppl can use locally), that applies a health-check interval specifically for this portion, and then the rest, which is subject to `interval` and `retries`.

The failing probes during the `start_period` are not counted for the number of retries, unless a probe is successful and afterward it starts to not be anymore.

Hence, I suggest we change the health check back to 1min because otherwise, it is going to be a bit spammy.
In terms of retries, I think it is worth doing them for 5 at least.

orhun (approved):

PR #1402, opened 2023-11-18
auth: use a centrally stored jwt signing private key

iulianbarbu:
## Description of change

We want our auth service to not generate a new private key for JWT signing at each release. This change should help in not invalidating the cached JWTs on the gateway when auth is released in isolation.

* start auth service based on an environment secret passed through a flag, for JWT signing.
* local development will use a default secret set in the `docker-compose.dev.yml`, to not require setting it up in the environment before starting the stack locally (the secret is compromised since it is stored in version control and we shouldn't use it for anything outside local development).
* the JWT signing private key parsing logic will fail early in case the provided key is not valid (e.g when empty if forgetting to set it)

* checked the parsing locally by starting the auth service
* Tested on staging:
1. deployed all services
2. deployed a project
3. scaled auth service down to 0 and then up to 1
4. rerun immediately a command
5. checked auth logs and no `JWT conversion` happened in auth while doing 4 (meaning a cached JWT was used successfully).


auth/src/args.rs
```diff
@@ -28,6 +28,11 @@ pub struct StartArgs {
     /// Stripe client secret key
     #[arg(long, default_value = "")]
     pub stripe_secret_key: String,
+
+    /// Auth JWT signing private key, as a base64 encoding of
+    /// a PEM encoded PKCS#8 v1 formatted unencrypted private key.
+    #[arg(long, default_value = "")]
```
oddgrd, line 34: We might have discussed this in the past, but should we not have a default value for this?

oddgrd (commented):
Makes sense! I left one question.

auth/src/secrets.rs
```diff
@@ -15,16 +15,27 @@ pub struct EdDsaManager {
...
-            .expect("to create a PKCS8 for edDSA");
-        let encoding_key = EncodingKey::from_ed_der(doc.as_ref());
-        let pair = Ed25519KeyPair::from_pkcs8(doc.as_ref()).expect("to create a key pair");
-        let public_key = pair.public_key();
+    // We take the private key used for JWT signing as an auth argument. It is the base64 enconding of a PEM encoded
+    // private key. The private key was generated previously using `OpenSSL 3.1.2 1 Aug 2023 (Library: OpenSSL 3.1.2 1
+    // Aug 2023)`, by using `openssl genpkey -algorithm ED25519`. This generates a PEM encoded PKCS#8 v1 formatted
+    // unencrypted key. We want to first parse the PEM encoded key and get the PKCS#8 v1 formatted key to parse it with
+    // `ring::signature::Ed25519KeyPair`. This is needed because the public key is not contained within the openssl
+    // generated private key, and should be computed/stored before serving it for JWT validation.
```
Kazy, line 23: What do you think of providing both the public and private key ? 

I'm asking because in [the documentation](https://docs.rs/ring/latest/ring/signature/struct.Ed25519KeyPair.html#method.from_pkcs8_maybe_unchecked), they're talking about a missing consistency check when using a PKCS#8 v1 file, since it doesn't have the public key, and while I don't know what impact it can have, the fact that they talk about it and that there exists a newer version of the file that fixes this would guide me towards preferring this solution. It is also exactly what the existing implementation is doing.

chesedo (commented):
Really nice!! :partying_face: 

Left small comments.

auth/src/args.rs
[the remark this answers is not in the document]
Kazy: I agree it's less discoverable, but you get an error when you start it, and usually you only need to set it once, you don't change them across invocations. And by using a command line argument, this also means that anyone that can do a `ps` will be able to see the secrets in clear text, and that interfaces that shows the command line will also display the secret.

auth/src/secrets.rs
```diff
@@ -15,16 +16,29 @@ pub struct EdDsaManager {
...
-        let encoding_key = EncodingKey::from_ed_der(doc.as_ref());
-        let pair = Ed25519KeyPair::from_pkcs8(doc.as_ref()).expect("to create a key pair");
-        let public_key = pair.public_key();
+    // It is the base64 enconding of a PEM encoded private key. The private key was generated
+    // previously using `OpenSSL 3.1.2 1 Aug 2023 (Library: OpenSSL 3.1.2 1 Aug 2023)`, by using
+    // `openssl genpkey -algorithm ED25519`. This generates a PEM encoded PKCS#8 v1 formatted
+    // unencrypted key. We want to first parse the PEM encoded key and get the PKCS#8 v1 formatted
+    // key to parse it with `ring::signature::Ed25519KeyPair`. This is needed because the public key
+    // is not contained within the openssl generated private key, and should be computed/stored before
+    // serving it for JWT validation.
```
chesedo, line 25:
```suggestion
    /// Create a new manager from a base64 PEM encoded private key. This key can be generated using:
    /// ```bash
    /// openssl genpkey -algorithm ED25519 -out auth_jwtsigning_private_key.pem
    /// base64 < auth_jwtsigning_private_key.pem
    /// ```
```

Nit to make the comment simpler

chesedo (commented):
Small nit

PR #1405, opened 2023-11-19
refactor: better feature scoping, fix turso compilation, prune library dependency tree

jonaro00:
## Description of change

Only refactorings, no changes.

- Prunes the dependency weight by proto by a lot, which affects all resource libs
  - Fixes turso example compiling (no sqlx / sqlite crates in our tree exposed)
  - **Axum hello world dependency tree: 366 -> 266 !!**
  - **Cargo-shuttle release build tree: 605 -> 416 !!!!!!!** compile time (20 threads) 70s -> 51s (-27%)
- moved `runtime::start` out of proto :sunglasses: 

- [x] Clippy every crate individually
- [x] CI
- [x] Check all examples with patches
- anything else?


cargo-shuttle/Cargo.toml
```diff
@@ -9,9 +9,9 @@ homepage = "https://www.shuttle.rs"
 rust-version = "1.70"
 
 [dependencies]
-shuttle-common = { workspace = true, features = ["models"] }
-shuttle-proto = { workspace = true }
-shuttle-service = { workspace = true, features = ["builder"] }
+shuttle-common = { workspace = true, features = ["models", "backend"] }
```
oddgrd, line 12: Naming nit: should this feature still be named backend if it's used in the CLI?

oddgrd (approved):
Super nice reduction in dependencies exposed to users! :partying_face:  It would also be nice to know what the reduction in build time will be from this PR.

cargo-shuttle/Cargo.toml
chesedo: Hmm, what inside of `backend` does the cli need here?

jonaro00: I was mistaken. A small refactor solved it, and reduced the dep tree by an extra 10 :D

jonaro00:
cargo-shuttle release compile time (20 threads) 70s -> 51s (-27%)

PR #1410, opened 2023-11-21
 feat(cargo-shuttle): ability to force a name to be used in init

jonaro00:
## Description of change

Useful if you are initing a project with a name that you already own.

`--force-name` arg disabled name validation and taken check.
In interactive, "reply" with the same name again to a failed validation to use the name anyways.

All combinations of with/without arg and with/without interactive works as expected.

### DIFF
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -277,6 +277,9 @@ pub struct InitArgs {
     #[arg(default_value = ".", value_parser = OsStringValueParser::new().try_map(parse_init_path))]
     pub path: PathBuf,
 
+    /// Don't check the project name's validity or availability and use it anyways
+    #[arg(long)]
+    pub force_name: bool,
     /// Whether to start the container for this project on Shuttle, and claim the project name
     #[arg(long)]
     pub create_env: bool,
@@ -399,6 +402,7 @@ mod tests {
             from: None,
             subfolder: None,
             create_env: false,
+            force_name: false,
             login_args: LoginArgs { api_key: None },
             path: PathBuf::new(),
         };
@@ -416,6 +420,7 @@ mod tests {
             from: None,
             subfolder: None,
             create_env: false,
+            force_name: false,
             login_args: LoginArgs { api_key: None },
             path: PathBuf::new(),
         };
@@ -433,6 +438,7 @@ mod tests {
             from: None,
             subfolder: None,
             create_env: false,
+            force_name: false,
             login_args: LoginArgs { api_key: None },
             path: PathBuf::new(),
         };
@@ -450,6 +456,7 @@ mod tests {
             from: Some("https://github.com/some/repo".into()),
             subfolder: Some("some/path".into()),
             create_env: false,
+            force_name: false,
             login_args: LoginArgs { api_key: None },
             path: PathBuf::new(),
         };
@@ -467,6 +474,7 @@ mod tests {
             from: None,
             subfolder: None,
             create_env: false,
+            force_name: false,
             login_args: LoginArgs { api_key: None },
             path: PathBuf::new(),
         };
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -225,7 +225,7 @@ pub struct ProjectStartArgs {
     pub idle_minutes: u64,
 }
 
-#[derive(Parser, Clone, Debug)]
+#[derive(Parser, Clone, Debug, Default)]
 pub struct LoginArgs {
     /// API key for the Shuttle platform
     #[arg(long)]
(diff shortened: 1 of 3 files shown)


cargo-shuttle/src/args.rs
```diff
@@ -399,6 +402,7 @@ mod tests {
             from: None,
             subfolder: None,
             create_env: false,
+            force_name: false,
```
chesedo, line 405: It might be worth refactoring this to somehow create this struct so that we don't have to update a bunch of tests each time a new field is added.

cargo-shuttle/src/lib.rs
```diff
@@ -337,42 +337,53 @@ impl Shuttle {
...
-                            "{}",
-                            "Failed to check if project name is available.".yellow()
-                        );
-                        break;
-                    }
+                    .interact()?
+            };
+            // handle --force on retires
+            let name: String = if let Some(prev_name) = prev_name {
+                if project_args.name.is_none() && name == "--force" {
```
chesedo, line 354: I'm lost. :confused: Where does `--force` come from? The user's input?

cargo-shuttle/tests/integration/init.rs
```diff
@@ -48,11 +47,10 @@ async fn non_interactive_rocket_init() {
 
     let args = ShuttleArgs::parse_from([
         "cargo-shuttle",
-        "--api-url",
-        "http://shuttle.invalid:80",
         "init",
         "--api-key",
         "dh9z58jttoes3qvt",
+        "--force-name",
```
chesedo, line 53: Rather create a new test instead of updating the current ones since I think the old tests should still exist too.

chesedo (commented):
I'm a bit confused with this PR. Why do we need it? And why all the `force` variants?

cargo-shuttle/src/lib.rs
jonaro00: Yes, see the prompt that gets printed if the check fails.

cargo-shuttle/tests/integration/init.rs
jonaro00: The tests break without this, since it will try to verify the name without this flag. This flag also removes the need to use a dummy API url

cargo-shuttle/src/lib.rs
chesedo: Hmm, that is not good DX imo. The `force-name` argument should be enough and the error can suggest users use it instead.

jonaro00: Disagree. Forcing them to quit the command and add arguments instead of resolving it in interactive mode is worse.

chesedo: Else there should be a control key or force prompt instead. Having `--force` as a name input is just not going to cut it

jonaro00: A separate prompt hinders the normal flow of letting user write a different project name instead. The --force is the more rare use case.

chesedo: Then maybe a control key like "Ctrl+f" to use it anyway

chesedo: Or ask the user to type the same name if they want to force use it

jonaro00: I like it.

cargo-shuttle/src/lib.rs
```diff
@@ -337,42 +337,53 @@ impl Shuttle {
...
+            let name: String = if let Some(prev_name) = prev_name {
+                if project_args.name.is_none() && name == prev_name {
+                    project_args.name = Some(prev_name);
+                    break;
+                } else {
+                    name
                 }
+            } else {
+                name
+            };
```
chesedo, line 362: I'm thinking there should be a way to combine this with the next if since they share logic. Something like

Something like
```rust
let should_force = prev_name.map(|n| n == name).unwrap_or(args.force_name);
```

This is most likely not accurate, but you get the gist :smile:

chesedo (commented):
One extra comment

cargo-shuttle/src/lib.rs
jonaro00: Can't break a loop inside a closure. This anti-pattern is only because `if let  ... = ... && expr` is not stable yet.

cargo-shuttle/src/lib.rs
```diff
@@ -330,54 +331,59 @@ impl Shuttle {
...
+                    .interact()?
+            };
+            // handle forced name on retires
+            if needs_name && prev_name.as_ref().is_some_and(|prev| prev == &name) {
+                project_args.name = Some(prev_name.unwrap());
+                break;
+            }
+            // skip validation if forced
+            if args.force_name {
+                project_args.name = Some(name);
```
chesedo, line 360:
```suggestion
            let force_name = args.force_name || (needs_name && prev_name.as_ref().is_some_and(|prev| prev == &name));
            if force_name {
                project_args.name = Some(name);
                break;
            }
```

Shorter still :smile:

chesedo (commented):
Think it can still be combined

chesedo (approved):

PR #1413, opened 2023-11-22
bug: delete a project even if the current state is destroyed

chesedo:
## Description of change
Causes the delete command to start up destroyed (stopped) projects to be able to delete them.

By making new gateway tests. Use the following steps to run them

```
USE_PANAMAX=disable make up # To start up the other service on your local environment
SHUTTLE_TESTS_NETWORK=shuttle-dev_user-net cargo test --package shuttle-gateway # to run all the gateway tests
```

If your tests take too long to run and eventually errors with `503` then make sure your firewall is not blocking any requests. See comments in the code.

### DIFF
--- a/gateway/src/api/latest.rs
+++ b/gateway/src/api/latest.rs
@@ -1108,7 +1108,7 @@ pub mod tests {
 
     use super::*;
     use crate::service::GatewayService;
-    use crate::tests::{RequestBuilderExt, World};
+    use crate::tests::{RequestBuilderExt, RouterExt, World};
 
     #[tokio::test]
     async fn api_create_get_delete_projects() -> anyhow::Result<()> {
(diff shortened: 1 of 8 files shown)


gateway/src/api/latest.rs
```diff
@@ -1364,6 +1385,23 @@ pub mod tests {
         Ok(())
     }
 
+    #[test_context(TestProject)]
+    #[tokio::test]
+    async fn api_delete_project_that_is_ready(project: &mut TestProject) -> anyhow::Result<()> {
+        project.router_call(Method::DELETE, "/delete").await;
+
+        Ok(())
+    }
```
oddgrd, line 1394: Clean! :heart_eyes: Can we apply these utilities for the existing tests as well, and break them up into smaller pieces (looking at you `api_create_get_delete_projects`)? We can do it in a later PR to not block this for tomorrows release.

oddgrd (approved):
LGTM! And the new test setup looks great!

gateway/src/api/latest.rs
chesedo: Yip, getting to them in another PR is the plan.... and maybe even adding them to the end of CI :smile:

PR #1414, opened 2023-11-22
auth: comment healthcheck `start_period` & `start_interval`

iulianbarbu:
## Description of change

CircleCI staging/prod deployment will fail because the `start_period` and `start_interval` configs for `healthcheck` are not handled gracefully when deploying the stack. The deployment will fail due to using a Docker Engine version that does not support the configs.

This PR comments on the configs explaining why, and when we'll upgrade, we can uncomment them.

Releasing on staging for #1402 .

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -58,8 +58,10 @@ services:
       interval: 1m
       timeout: 10s
       retries: 5
-      start_period: 10s
-      start_interval: 2s
+      # These options are pending on an upgrade to Docker Engine version 25.0.
+      # Using them will result in error because they are not handled gracefully.
+      # start_period: 10s
+      # start_interval: 2s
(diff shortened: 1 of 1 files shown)


oddgrd (approved):

docker-compose.yml
```diff
@@ -58,8 +58,10 @@ services:
       interval: 1m
       timeout: 10s
       retries: 5
-      start_period: 10s
```
oddgrd, line 61: It looks like we should be able to use this, is it useful on its own: https://docs.docker.com/compose/compose-file/compose-versioning/#version-34?

iulianbarbu: Nope, because the interval is set to 1 minute, so no health-check will be applied during that period.

PR #1415, opened 2023-11-22
fix(common): type conversion from str for a custom resource

iulianbarbu:
## Description of change

We're missing the `Custom` resource type conversion from `str`.

Reproduced when I deployed the `custom-resource/pdo` example.
TODO: test the project deployment again after the fix.


oddgrd (approved):
This gives me a feeling of déjà vu, I thought we fixed this a while back :sweat_smile:

iulianbarbu:
I was surprised to see the error too. Not sure what we've fixed the last time, or if we introduced a regression in the meantime. I haven't digged into it, but it might not be worth doing it.

PR #1416, opened 2023-11-22
feat(shuttle-serenity): support serenity 0.12 and 0.11, optional native tls feature

jonaro00:
## Description of change

Serenity has entered RC stage for 0.12. We only use their `Client`, and there is no breaking changes in that interface.
Also moved rustls to a default feature to allow native tls.
Having this released ahead of time means that we will support serenity 0.12 from the same moment it is released.

Compiled serenity hello world with 0.11 and their 0.12-rc branch, both worked.


oddgrd (approved):
Oh, we can avoid forcing users to upgrade, cool!

PR #1418, opened 2023-11-23
feat(cargo-shuttle): better handling of runtime version checks

jonaro00:

Ran local run with a normal axum app, and the timeout worked.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -7,6 +7,7 @@ mod suggestions;
 
 use std::collections::{BTreeMap, HashMap};
 use std::ffi::OsString;
+use std::fmt::Write as FmtWrite;
 use std::fs::{read_to_string, File};
 use std::io::stdout;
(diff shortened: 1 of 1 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -2057,20 +2060,27 @@ fn check_version(runtime_path: &Path) -> Result<()> {
...
+        bail!("shuttle-runtime binary not found");
     }
 
     // Get runtime version from shuttle-runtime cli
-    let runtime_version = std::process::Command::new(runtime_path)
-        .arg("--version")
-        .output()
-        .context("failed to run the shuttle-runtime binary to check its version")?
-        .stdout;
+    // It should print the version and exit immediately, so a timeout is used to not launch servers with non-Shuttle setups
```
iulianbarbu, line 2067: Isn't the process exiting immediately already when running the runtime with `--version`?

iulianbarbu (approved):

PR #1420, opened 2023-11-23
auth: store state in postgres instance

iulianbarbu:
## Description of change

Store the auth service state inside Postgres.

**Note**: This PR should **NOT** be merged before we set up the migration logic for the existing SQLite-based state to Postgres.

- [x] it was tested locally by starting the entire stack and deploying a project with a newly created admin user.
- [x] test on staging on a brand new empty db.


auth/migrations/0000_init.sql
```diff
@@ -1,5 +1,7 @@
+CREATE TYPE tier AS ENUM ('basic', 'pendingpaymentpro', 'pro', 'team', 'admin', 'deployer');
```
Kazy, line 1: I would suggest to keep using `TEXT` for now. in case we need to revert. We can do the other migrations and use more of Postgres feature later on.

auth/src/args.rs
```diff
@@ -1,12 +1,12 @@
-use std::{net::SocketAddr, path::PathBuf};
+use std::net::SocketAddr;
 
 use clap::{Parser, Subcommand};
 
 #[derive(Parser, Debug)]
 pub struct Args {
     /// Where to store auth state (such as users)
     #[arg(long, default_value = "./")]
```
Kazy, line 8: Default value should be removed.

Kazy (commented):
Nice work, didn't expect to see this that fast :muscle: Maybe switch the PR as draft to make sure we don't inadvertently merge it ?

auth/migrations/0000_init.sql
iulianbarbu: What would it look like in the end? 

I had trouble in making sqlx-postgres work with Rust enums like our `AccountTier`. What's a bit off-putting is this part of the docs [1]. It requires the creation of a postgres type as well (so double tracking) when trying to work seamlessly with Rust enums <-> Postgres type, and be able to get the data from postgres coercing it to `AccountTier` Rust enum. I haven't verified specifically if sending data works with `TEXT` type (I believe it should), but coercing it to the `AccountTier` enum is subject to type errors.

If we want to work just with `TEXT`, then we need to get it from Postgres and then convert it to `AccountTier`, which should be fine too. Will try to look more into a seamless conversion between the two, but happy to hear other suggestions too.

[1] https://docs.rs/sqlx/latest/sqlx/postgres/types/index.html#enumerations

Kazy: It would look exactly like your last paragraph :) 

We currently have that issue with SQLite as well so it's fine to have it for the moment since it will make the migration easier.

iulianbarbu: Not sure what issue we have with SQLite currently. We don't need a new SQL type to exist for the conversion, or `From` impls. However, I don't mind using a `From` to avoid double tracking (if it works), at the expense of not using the recommendation from the docs.

I also think the migration concern is important, so will try to make the adjustments and ping afterward.

iulianbarbu: Done.

auth/src/args.rs
iulianbarbu: Done.

auth/migrations/0000_init.sql
Kazy: I meant the issue that we don't have any assurance that what's in the database matches the enum when using SQLite, since it's a text field.

iulianbarbu: Ah yes, indeed.

test-utils/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+        sleep(Duration::from_millis(350));
+        Self::_wait_ready(Duration::from_secs(120), &is_ready_cmd);
+
+        Self {
+            container_name,
+            base_uri: format!("{engine}://{engine}:password@localhost:{host_port}"),
+        }
+    }
+
+    fn _wait_ready(mut timeout: Duration, is_ready_cmd: &[&str]) {
```
oddgrd, line 90:
```suggestion
    fn wait_ready(mut timeout: Duration, is_ready_cmd: &[&str]) {
```

test-utils/Cargo.toml
oddgrd, line 1: Could this be a part of common-tests, rather than a new crate?

test-utils/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+///
+///  #[tokio::test]
+/// async fn test_case() {
+///     // Create a unique database name so we have a new database for each test.
+///     let db_name = Uuid::new_v4().to_string();
+///     let db_uri = PG.get_unique_uri(db_name.as_str());
+///     
+///     // Test logic below, which can use `db_uri` to connect to the postgres instance.
+/// }
+///
```
oddgrd, line 32: This will give code highlighting.

```suggestion
/// ```
/// static PG: Lazy<PostgresDockerInstance> = Lazy::new(PostgresDockerInstance::default);

/// #[dtor]
/// fn cleanup() {
///    PG.cleanup();
/// }
///
///  #[tokio::test]
/// async fn test_case() {
///     // Create a unique database name so we have a new database for each test.
///     let db_name = Uuid::new_v4().to_string();
///     let db_uri = PG.get_unique_uri(db_name.as_str());
///     
///     // Test logic below, which can use `db_uri` to connect to the postgres instance.
/// }
/// ```
```

test-utils/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+
+impl Default for PostgresDockerInstance {
+    fn default() -> Self {
+        let s: String = rand::thread_rng()
+            .sample_iter(&Alphanumeric)
+            .take(7)
+            .map(char::from)
+            .collect();
+
+        PostgresDockerInstance::new(s.as_str())
```
oddgrd, line 46: Nit: we could use UUID here too and get rid of the rand dependency.

oddgrd (commented):
LGTM! Just a couple of questions and minor comments.

test-utils/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+}
+
+impl PostgresDockerInstance {
```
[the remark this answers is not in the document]
iulianbarbu: I just saw it today, looks promising.

test-utils/src/lib.rs
iulianbarbu: Done.

test-utils/Cargo.toml
iulianbarbu: Done.

test-utils/src/lib.rs
iulianbarbu: Done.

test-utils/src/lib.rs
iulianbarbu: Done.

test-utils/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+/// #[dtor]
+/// fn cleanup() {
+///    PG.cleanup();
+/// }
+///
+///  #[tokio::test]
+/// async fn test_case() {
+///     // Create a unique database name so we have a new database for each test.
+///     let db_name = Uuid::new_v4().to_string();
+///     let db_uri = PG.get_unique_uri(db_name.as_str());
```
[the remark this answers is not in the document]
iulianbarbu: Done.

test-utils/src/lib.rs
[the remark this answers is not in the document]
iulianbarbu: Opened https://github.com/shuttle-hq/shuttle/issues/1479. It shouldn't be difficult and might be a nice OSS contribution. What do you think?

.circleci/config.yml
```diff
@@ -834,7 +843,7 @@ workflows:
           aws-secret-access-key: DEV_AWS_SECRET_ACCESS_KEY
           shuttle-env: staging
           requires:
-            - approve-push-unstable
+            - approve-build-images-unstable
```
oddgrd, line 846: Nit: it does both build and push.

oddgrd (approved):
LGTM! :partying_face:

test-utils/src/lib.rs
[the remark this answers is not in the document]
oddgrd: I haven't looked closely at test containers, but it does seem like it would simplify this a lot. For the GitHub issue we might want to expand a bit on what should change, and where, since we use this strategy in logger, provisioner and now auth. Provisioner does not use the code you moved to common-tests, but maybe we don't need to share code in common-tests if we use test containers.

.circleci/config.yml
iulianbarbu: Good catch! Done. I added an approve step for building the images on production, but we don't use it, since merging something on `production` is intentional, only for a release, so approving the images build & push isn't needed.

test-utils/src/lib.rs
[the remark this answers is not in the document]
oddgrd: We might want to suggest that the contributor breaks it up into separate PRs, as well.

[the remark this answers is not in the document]
iulianbarbu: Thanks @oddgrd ! I added more details under the definition of done for the associated issue. I missed that we have the docker tests in the `provisioner` too, and it will probably use `testcontainers` (not sure about the `common-tests` additional logic though). In `common-tests` we created API to create/get unique conn strings, referring to unique databases, that can be used in parallel by test cases. I think this is the outstanding part shared by `logger` and `auth` soon. Not sure if the `provisioner` needs this functionality for its integration tests, and if not, it can probably avoid using `common-tests`, integrating directly with `testcontainers`.

PR #1421, opened 2023-11-23
bug: remove project resources automatically when delete the project

chesedo:
## Description of change
This causes a delete project request to try and remove all the resources automatically if possible. Else return a helpful error message.

By adding new tests


jonaro00 (commented):
Dry run flag is not being respected in handler. Ideally, the dry run flag will instead of deleting resources, return a list of them and let the user confirm the deletion of those resources. But we can also just display "...will delete ALL resources!!!" and let it go through after just the initial confirmation in CLI.

common/src/models/error.rs
```diff
@@ -91,14 +91,14 @@ impl From<ErrorKind> for ApiError {
                 (StatusCode::FORBIDDEN, "You cannot create more projects. Delete some projects first.")
             },
             ErrorKind::ProjectHasRunningDeployment => (
-                StatusCode::FORBIDDEN,
-                "A deployment is running. Stop it with `cargo shuttle stop` first."
+                StatusCode::INTERNAL_SERVER_ERROR,
+                "Could not automatically stop the running deployment for the project. Reach out to our support and ask them for a helping hand."
```
oddgrd, line 95: Nit: should we rephrase the last sentence to: `Please reach out to Shuttle support for help.`. Same for the ProjectHasResources one.

gateway/src/api/latest.rs
```diff
@@ -341,74 +342,36 @@ async fn delete_project(
...
-        )
-        .method("GET")
-        .body(hyper::Body::empty())
-        .unwrap();
-    let res = route_project(State(state.clone()), scoped_user.clone(), service_req).await?;
-    // 404 == no service == no deployments
-    if res.status() != StatusCode::NOT_FOUND {
+    let summary = project_caller.get_service_summary().await?;
+    if summary.deployment.is_some() {
+        let res = project_caller.stop_active_deployment().await?;
```
oddgrd, line 351: Should we also handle a deployment that is `Loading`?

gateway/src/api/project_caller.rs
```diff
@@ -0,0 +1,129 @@
...
+            let body_bytes = hyper::body::to_bytes(res.into_body())
+                .await
+                .map_err(|e| Error::source(ErrorKind::Internal, e))?;
+            let body = serde_json::from_slice(&body_bytes)
+                .map_err(|e| Error::source(ErrorKind::Internal, e))?;
+
+            Ok(body)
+        } else {
+            Ok(Default::default())
+        }
```
oddgrd, line 83: Hmm, why do we return default if the status code is 404?

gateway/src/args.rs
```diff
@@ -62,6 +62,9 @@ pub struct ContextArgs {
     /// Address to reach the authentication service at
     #[arg(long, default_value = "http://127.0.0.1:8008")]
     pub auth_uri: Uri,
+    /// Address to reach the resource recorder service at
+    #[arg(long, default_value = "http://resource-recorder:8000")]
+    pub resource_recorder_uri: Uri,
```
oddgrd, line 67: If we have access to r-r, do we need projects to be running to delete them? I suppose database deletion also goes through the deployer provisioner client, but could that also be done in gateway?

oddgrd (commented):
Looks good! I just have a couple of questions.

gateway/src/args.rs
oddgrd: On second thought, we do need the deployer to have been started recently, to ensure any resources cached there have been sent to r-r.

gateway/src/api/project_caller.rs
chesedo: I need the `summary.deployment` to be `None`. Else I need to add more logic in the controller :sweat_smile: 
Happy to rework this if it is too confusing

gateway/src/args.rs
chesedo: Oh, gateway won't access r-r. This is in the context struct that we send to our Docker worker. This uri is then sent to a deployer when we start it up. But it is an interesting option though....

gateway/src/api/project_caller.rs
oddgrd: This was clearer now, thanks!

gateway/src/api/project_caller.rs
```diff
@@ -77,29 +79,36 @@ impl ProjectCaller {
...
+                &format!("/projects/{project_name}/deployments"),
+                Method::GET,
+            )
+            .await?;
+
+        if let Some(deployments) = deployments {
+            Ok(deployments)
+        } else {
+            Ok(Vec::new())
+        }
```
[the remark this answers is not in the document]
chesedo: We sure can :heart_eyes: Thanks O!

PR #1426, opened 2023-11-27
revert: rate limit based on peer address #1351

oddgrd:
## Description of change

Revert #1351. While testing locally, the rate limiting layer worked as expected. However, when deployed in swarm mode, Docker puts a load balancer in front of the services in the user network. Therefore, it appears to the rate limiter that all logger requests are coming from the same peer address.

Tested that logging works as expected locally (get_logs and get_logs --follow).


deployer/src/handlers/mod.rs
```diff
@@ -663,16 +663,8 @@ pub async fn get_logs(
...
-            {
-                Err(Error::RateLimited(
-                    "your application is producing too many logs. Interactions with the shuttle logger service will be rate limited"
-                        .to_string(),
-                ))
-            } else {
-                Err(anyhow!("failed to retrieve logs for deployment").into())
-            }
+            error!(error = %error, "failed to retrieve logs for deployment");
+            Err(anyhow!("failed to retrieve logs for deployment").into())
```
oddgrd, line 667: If this were to be reverted, it would be reverted to return a 404, but that would be [incorrect](https://github.com/shuttle-hq/shuttle/pull/1351#discussion_r1392180264), so I left this in.

oddgrd:
Note: I did not revert this change: <https://github.com/shuttle-hq/shuttle/pull/1351/files#diff-5d10b574e97438514650fd6f44664299baf20536aab73554ca2754906d2a72d4L382-R447>, since it would ease implementing rate limiting based on number of requests in the future if the log batches are smaller.

oddgrd:
> Can we also not revert the log error handling in c-s? to remain along with: #1429

Thanks, that's a good idea!

jonaro00 (approved):

PR #1427, opened 2023-11-27
feat: downgrade user to basic tier only after period end

GugaGongadze:
## Description of change
The status of subscription is still `active` after indicating  `cancel_at_period_end` in front-end. It will become `cancelled` after period end.

Console PR: https://github.com/shuttle-hq/shuttle-console/pull/82

## How has this been tested?
Tested following scenarios:
- Upgrading from `basic` to `pro`
- manually downgrading from `pro` to `cancelledpro`
- Auto-downgrading from `cancelledpro` to `basic` after period end.


auth/src/user.rs
```diff
@@ -212,6 +212,14 @@ impl User {
             .subscription_is_valid(&user_manager.stripe_client)
             .await?;
 
+        if self.account_tier == AccountTier::CancelledPro && !subscription_is_valid {
```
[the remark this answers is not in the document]
oddgrd: It is set via the auth update tier endpoint from the console: https://github.com/shuttle-hq/shuttle-console/pull/82/files#diff-c6a74f597ccda9d58321be32a292acdc96a2519fc6f16d87860b0633cbdcad12R157

common/src/limits.rs
```diff
@@ -45,6 +45,7 @@ impl From<AccountTier> for Limits {
             AccountTier::Admin
             | AccountTier::Basic
             | AccountTier::PendingPaymentPro
+            | AccountTier::CancelledPro
             | AccountTier::Deployer => Self::default(),
             AccountTier::Pro | AccountTier::Team => Self::new(MAX_PROJECTS_EXTRA, true),
```
oddgrd, line 50: I think it should be like this, right? The cancelled pro user should have access until it expires and they are made a Basic user.
```suggestion
            | AccountTier::Deployer => Self::default(),
            AccountTier::Pro | AccountTier::CancelledPro | AccountTier::Team => Self::new(MAX_PROJECTS_EXTRA, true),
```

oddgrd (commented):
Cool, thanks! I just have one small comment. 

We might need to make some further considerations for this implementation in the future, since currently the tier will only be changed from cancelled to basic if the user calls an action that triggers the subscription check. In the future we'd need something like a worker that regularly checks the subscriptions, I think.

auth/src/user.rs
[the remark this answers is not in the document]
iulianbarbu: Yep, confirmed with @GugaGongadze too. Resolving this.

auth/tests/api/helpers.rs
```diff
@@ -137,6 +135,20 @@ impl MockedStripeServer {
     async fn subscription_retrieve_handler(
         Path(subscription_id): Path<String>,
     ) -> axum::response::Response<String> {
+        static TOGGLE: Mutex<bool> = Mutex::new(false);
```
iulianbarbu, line 138: We can avoid using statics and store this toggle on the `MockedStripeServer`.
Also, naming it as `subscription_cancel_side_effect_toggle` would be great.
What do you think?

auth/tests/api/helpers.rs
```diff
@@ -137,6 +135,20 @@ impl MockedStripeServer {
...
+            let toggle_is_true = *TOGGLE.lock().unwrap();
+
+            if toggle_is_true {
+                return Response::new(MOCKED_SUBSCRIPTIONS[3].to_string());
+            } else {
+                let mut toggle = TOGGLE.lock().unwrap();
+                *toggle = true;
+                return Response::new(MOCKED_SUBSCRIPTIONS[2].to_string());
+            }
+        }
```
iulianbarbu, line 150: 👍

auth/tests/api/users.rs
```diff
@@ -169,6 +169,80 @@ async fn successful_upgrade_to_pro() {
...
+        .await;
+    assert_eq!(response.status(), StatusCode::OK);
+
+    // Cancel subscription
+    let response = app.put_user("test-user", "cancelledpro", "").await;
+    assert_eq!(response.status(), StatusCode::OK);
+
+    // Trigger status change to canceled
+    let response = app.get_user("test-user").await;
+    assert_eq!(response.status(), StatusCode::OK);
```
iulianbarbu, line 196: Can we check right after that the tier is still `cancelledpro`?

iulianbarbu: Ah, it looks like you're testing this in a dedicated test case. Just wondering, is it correct to assume for this test that the first `get_user` call will return the tier as `cancelledpro`? If yes, than I think it is worth checking (even if you check specifically in another test for it).

iulianbarbu (commented):
Looks great @GugaGongadze ! I left just a few comments which should be simple to resolve. let me know your thoughts.

iulianbarbu:
It seems that the `examples` dir is changed in this commit: [bc63399](https://github.com/shuttle-hq/shuttle/pull/1427/commits/bc63399e2ff8f530a55c47fd572e15e91d469221). Is this intentional?

iulianbarbu (approved):
LGTM! thanks 🥳

oddgrd (approved):
LGTM, thanks!

PR #1445, opened 2023-11-30
feat: protect pro tier projects and our services

chesedo:
## Description of change
Prevent the creation of new projects or the starting of idle projects that will overload our servers

Updating the tests

### DIFF
--- a/gateway/src/service.rs
+++ b/gateway/src/service.rs
@@ -244,8 +244,10 @@ pub struct GatewayService {
     task_router: TaskRouter<BoxedTask>,
     state_location: PathBuf,
 
-    /// Maximum number of containers the gateway can start
-    container_limit: u32,
+    /// Maximum number of containers the gateway can start before blocking cch projects
+    cch_container_limit: u32,
+    /// Maximum number of containers the gateway can start before blocking non-pro projects
+    soft_container_limit: u32,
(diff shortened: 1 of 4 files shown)


gateway/src/service.rs
```diff
@@ -940,9 +947,34 @@ impl GatewayService {
...
+
+        let has_capacity = if current_container_count < self.cch_container_limit {
+            true
+        } else if current_container_count < self.soft_container_limit {
+            !is_cch_project
+        } else if current_container_count < self.hard_container_limit {
+            matches!(account_tier, AccountTier::Pro)
+        } else {
+            false
+        };
```
oddgrd, line 971: Should we add tracing events with the count if we are over any of the limits?

oddgrd (approved):
LGTM! I left one comment about observability, but it's not blocking.

gateway/src/service.rs
chesedo: I'm honestly not sure. Because of the ambulance event (in the other PR) we can already setup a honeycomb alert. So I want to add a `warn!` in this function but don't see how it will be useful considering the ambulance event.

oddgrd: True, it's likely overkill. 👍

PR #1446, opened 2023-11-30
feat(gateway): allow manual blocking of cch project traffic at high load

jonaro00:

about to double check it manually locally


chesedo (approved):
LGTM

gateway/src/api/latest.rs
```diff
@@ -449,11 +449,17 @@ async fn route_project(
     let is_cch_project = project_name.is_cch_project();
 
     // Don't start cch projects if we will be going over the container limit
-    if is_cch_project {
+    if is_cch_project && !scoped_user.user.claim.is_admin() {
```
oddgrd, line 452: Is the idea behind this that we should stop these from waking? Should we also block the proxy, then?

jonaro00: Mainly to stop deploys. If deployments are blocked, proxy hits (manual tests or validation) become naturally discouraged.

oddgrd (approved):

PR #1452, opened 2023-12-01
fix(gateway): more descriptive project not found error

### LINKED ISSUE #1397: [Improvement]: Better error message when project is not found in gateway
### Describe the improvement

To improve the error message "project not found" to something like "project 'blablah' not found", it needs a little refactor.
Start with turning `ProjectNotFound` into a tuple struct `ProjectNotFound(ProjectName)`.

Then, go through all usages of this variant, and add the project name into it:
- In ScopedUser extractor, its local variable is called `scope`.
- The 3 usages of ProjectNotFound in gateway's `proxy.rs` can be handled like this:
  - 1st one: Replace with BadHost
  - 2nd one: Replace with InvalidProjectName
  - 3rd one: Replace with CustomDomainNotFound
- The rest should be straight forward.

After this change, calling `cargo r -p cargo-shuttle -- project status --name unusedprojectname` against the local stack should give the project name back in the error.

---

Feel free to comment on this issue if you want to work on it!

fatfingers23:
## Description of change
Added the project name that is attempted to be used when looking up the status of a project in the error message.

I added the feature flag `backend` to the cargo-shuttle/Cargo.toml, because it appears it needed it for a test. I debated removing the feature flag from project.rs for what was required but decided to leave it. If this is not the desired outcome, please let me know, and I will update the PR
 
Closes #1397 

Ran the command `cargo run --bin cargo-shuttle -- project status --name ibetthisnameisnttaken` as suggested in the issue and got the project name in the new error message.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -188,8 +188,8 @@ This will prompt your local gateway to start a deployer container.
 Then, deploy it.
 
 ```bash
-cargo run -p cargo-shuttle -- --wd examples/rocket/hello-world project start
-cargo run -p cargo-shuttle -- --wd examples/rocket/hello-world deploy
+cargo run -p cargo-shuttle -- --wd examples/rocket/hello-world shuttle project start
+cargo run -p cargo-shuttle -- --wd examples/rocket/hello-world shuttle deploy
 ```
 
 Test if the deployment is working:
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -188,8 +188,8 @@ This will prompt your local gateway to start a deployer container.
 Then, deploy it.
 
 ```bash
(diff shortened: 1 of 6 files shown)


jonaro00:
Thanks for the PR!
Having `backend` enabled from cargo-shuttle is not desired, so if there's an easy refactor you can try it. You can also compromise and turn the enum type into `ProjectNotFound(String)` as the newtype isn't needed in the context of errors.

fatfingers23:
Ended up taking the compromise and used String instead. Was the path with the least amount of changes and seemed what was appropriate since it's for an error message

jonaro00 (approved):
LGTM

oddgrd (approved):
Great, thanks!

PR #1457, opened 2023-12-02
feat: don't overload docker with requests

chesedo:
## Description of change
See saw a lot of the below calls that we suspect overloaded Docker yesterday (meaning about 1k happening in less that a second for the same container). So this causes us to backoff a bit between Docker calls

https://github.com/shuttle-hq/shuttle/blob/23ba41b197a8a8d1ff8a2d80494577cf601501bf/gateway/src/project.rs#L508

Local run

### DIFF
--- a/gateway/src/task.rs
+++ b/gateway/src/task.rs
@@ -1,3 +1,4 @@
+use std::cmp::min;
 use std::collections::{HashMap, VecDeque};
 use std::marker::PhantomData;
 use std::pin::Pin;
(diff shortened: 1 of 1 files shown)


gateway/src/task.rs
```diff
@@ -303,6 +307,14 @@ impl Task<ProjectContext> for RunUntilDone {
     type Error = Error;
 
     async fn poll(&mut self, ctx: ProjectContext) -> TaskResult<Self::Output, Self::Error> {
+        // Don't overload Docker with requests. Therefore backoff with each try up to 30 seconds
+        if self.tries > 0 {
+            let backoff = min(3_u64.pow(self.tries), 30_000);
+
+            sleep(Duration::from_millis(backoff)).await;
+        }
+        self.tries += 1;
```
oddgrd, line 316: Should there be a maximum amount of tries? :thinking:

oddgrd (approved):

PR #1460, opened 2023-12-02
refactor: remove generics

chesedo:
## Description of change
Removes a bunch of unused generics in the state machine so that it is easier to reason about it. It will be best to review this on a per commit basis

Clippy is still happy


oddgrd (approved):
LGTM, thanks for the cleanup! Just need that conflict resolved.

PR #1462, opened 2023-12-03
Setup datadog

Kazy:
## Description of change

The `SHUTTLE_ENV` thing isn't really required anymore. Started doing that because I wanted to switch the log based on the env, and also wanted to setup a different environment when using Datadog locally (`dev`) to be able to test it. I can revert or split it up in an other PR if needed. This is only the first commit.

The main changes are in the other ones.

- Log correlation + correctly named spans.
image

- Right resource, right service in the traces.
image

- Actual logs, correlated, with the right service.
image

Ran it locally, connected to Datadog.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4228,6 +4228,22 @@ checksum = "ff011a302c396a5197692431fc1948019154afc178baf7d8e37367442a4601cf"
 
 [[package]]
 name = "opentelemetry"
+version = "0.12.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "514d24875c140ed269eecc2d1b56d7b71b573716922a763c317fb1b1b4b58f15"
+dependencies = [
+ "async-trait",
+ "futures",
+ "js-sys",
+ "lazy_static",
+ "percent-encoding",
+ "pin-project",
+ "rand 0.8.5",
+ "thiserror",
+]
+
+[[package]]
+name = "opentelemetry"
 version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1e32339a5dc40459130b3bd269e9892439f55b33e772d2a9d402a789baaf4e8a"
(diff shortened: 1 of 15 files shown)


deployer/src/proxy.rs
```diff
@@ -74,7 +75,7 @@ pub async fn handle(
     };
 
     // Record current service for tracing purposes
-    span.record("service", &service);
+    span.record("shuttle.service", &service);
```
Kazy, line 78: `service` is already used to designated the Shuttle service (logger, gateway, ..). Might want to even rename it to something like `user.service`. But this is why it was renamed.

Cargo.lock
```diff
@@ -8622,3 +8697,75 @@ dependencies = [
...
+
+[[patch.unused]]
+name = "shuttle-rocket"
+version = "0.34.1"
+
+[[patch.unused]]
+name = "shuttle-salvo"
+version = "0.34.1"
+
+[[patch.unused]]
```
iulianbarbu, line 8741: Don't forget to remove the patched deps.

common/src/backends/otlp_tracing_bridge.rs
```diff
@@ -0,0 +1,154 @@
...
+    fn record_i64(&mut self, field: &tracing::field::Field, value: i64) {
+        if let Some(ref mut vec) = self.log_record.attributes {
+            vec.push((field.name().into(), value.into()));
+        } else {
+            let vec = vec![(field.name().into(), value.into())];
+            self.log_record.attributes = Some(vec);
+        }
+    }
+
+    // TODO: Remaining field types from AnyValue : Bytes, ListAny, Boolean
```
iulianbarbu, line 70: Do we want to cover this TODO before upgrading from draft PR? I guess there will be compile time errors in case the types need to be implemented, so not sure if it is worth it in case we don't have errors at this point.

extras/otel/otel-collector-config.yaml
```diff
@@ -27,10 +27,10 @@ receivers:
...
     config:
       scrape_configs:
-      - job_name: 'otelcol'
-        scrape_interval: 10s
-        static_configs:
-        - targets: ['0.0.0.0:8888']
+        - job_name: "otelcol"
+          scrape_interval: 10s
+          static_configs:
+            - targets: ["0.0.0.0:8888"]
```
iulianbarbu, line 33: is this indentation needed?

gateway/src/project.rs
```diff
@@ -499,13 +500,12 @@ where
...
-            .unwrap()
-            .container_id()
-            .map(|id| format!("{id}: "))
-            .unwrap_or_default();
-        debug!("{container_id}{previous_state} -> {new_state}");
+        let container_id = new.as_ref().unwrap().container_id().unwrap_or_default();
+
+        debug!(
+            shuttle.container.id = container_id,
+            "Moving projectt from {previous_state} to {new_state}"
```
iulianbarbu, line 507: nit: 
```suggestion
            "Moving project from {previous_state} to {new_state}"
```

common/src/backends/otlp_tracing_bridge.rs
```diff
@@ -0,0 +1,154 @@
...
+            log_record.trace_context = Some(TraceContext::from(&SpanContext::new(
+                trace_id,
+                span_id,
+                TraceFlags::default(),
+                false,
+                TraceState::default(),
+            )));
+        }
+
+        // add the `name` metadata to attributes
```
iulianbarbu, line 127: Is this referring to the span name? I don't see why we don't want to add it to `log_record.attributes`. The result, as I see, is to be able to look for the spans with a given name and check the logs emitted for it. 

Is this how we'll benefit from adding the name to the `log_record.attributes`?

common/src/backends/otlp_tracing_bridge.rs
```diff
@@ -0,0 +1,154 @@
...
+        // TBD - Propose this to be part of log_record metadata.
```
iulianbarbu, line 128: By metadata here you refer to `attributes`?

common/src/backends/otlp_tracing_bridge.rs
```diff
@@ -0,0 +1,154 @@
...
+                .get::<OtelData>()
+                .and_then(|ext| ext.builder.trace_id.zip(ext.builder.span_id))
+        }) {
+            log_record.trace_context = Some(TraceContext::from(&SpanContext::new(
+                trace_id,
+                span_id,
+                TraceFlags::default(),
+                false,
+                TraceState::default(),
+            )));
```
iulianbarbu, line 124: Are the logs attached to the spans (visible in datadog) because of attaching this `trace_context` to them, which contains the `trace_id` and `span_id`?

common/src/backends/otlp_tracing_bridge.rs
Kazy: Left over from the previous code that I removed because we don't actually need it.

oddgrd (approved):

PR #1464, opened 2023-12-04
refactor: get stats from cgroup file directly

oddgrd:
## Description of change

This cherry-picks this commit: https://github.com/shuttle-hq/shuttle/commit/0f99d4617b807649b4055e6c84bc9800b6dcf47a, minus the parts that were extracted in #1463

TODO: test that this doesn't cause deserialization errors in the old gateway when deploying to staging.


chesedo (approved):
LGTM

iulianbarbu (approved):
LGTM!

gateway/src/project.rs
```diff
@@ -1289,6 +1289,43 @@ impl ProjectReady {
...
+
+    let docker_stats_file: String =
+        std::fs::read_to_string(format!("/sys/fs/cgroup/cpuacct/docker/{id}/cpuacct.usage"))
+            .map_err(|e| {
+                error!(error = %e, "failed to access docker stats file for container");
+                ProjectError::internal("failed to access docker stats file for container")
+            })?;
+
+    let usage: u64 = docker_stats_file.parse().map_err(|e| {
+        error!(error = %e, "failed to parse cpu usage stat");
```
Kazy, line 1304:
```suggestion
        error!(error = %e, shuttle.container.id = id, "failed to parse cpu usage stat");
```

iulianbarbu (approved):
LGTM!

gateway/src/project.rs
```diff
@@ -1289,6 +1290,42 @@ impl ProjectReady {
     }
 }
 
+#[instrument(name = "getting container stats from the cgroup file", skip_all)]
+async fn get_container_stats(container: &ContainerInspectResponse) -> Result<u64, ProjectError> {
+    let id = safe_unwrap!(container.id);
+
+    let cpu_usage: u64 = std::fs::read_to_string(format!("{DOCKER_STATS_PATH}/{id}/cpuacct.usage"))
```
jonaro00, line 1297: Consider using tokio fs instead?

oddgrd: Done, thanks!

gateway/src/main.rs
```diff
@@ -37,6 +38,20 @@ async fn main() -> io::Result<()> {
...
+    // Return an error early if the docker stats path is not in the expected location.
+    if !docker_stats_path.exists() {
+        return Err(std::io::Error::new(
+            io::ErrorKind::NotFound,
+            format!(
+                "could not find docker stats at path: {:?}",
+                DOCKER_STATS_PATH
+            ),
+        ));
+    }
```
oddgrd, line 53: We should just ensure this works for local development, then we can merge it. I don't imagine it will work on Windows, but developing Shuttle locally on Windows is already limited. For Linux, we need to see which version of cgroups are available, or just check if either path exists. For mac, I'm not sure, wdyt @iulianbarbu? See the `get_container_stats` function for which paths we use on linux.

oddgrd: We'll release this without the local development fix, we can do that after.

PR #1465, opened 2023-12-04
feat(gateway/ambulance): parallelize work done by the ambulance

Kazy:
## Description of change

Uses a `FuturesUnordered` to keep track of 8 tasks in parallel. If we already have more than 8 in progress, it will wait for the first one to be done before moving on.

Testing in progress right now.

### DIFF
--- a/gateway/src/main.rs
+++ b/gateway/src/main.rs
@@ -124,12 +124,12 @@ async fn start(db: SqlitePool, fs: PathBuf, args: StartArgs) -> io::Result<()> {
                     let gateway = gateway.clone();
                     let sender = sender.clone();
                     async move {
-                        let mut work_set = futures::stream::FuturesUnordered::new();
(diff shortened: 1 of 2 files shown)


gateway/src/main.rs
```diff
@@ -101,20 +102,49 @@ async fn start(db: SqlitePool, fs: PathBuf, args: StartArgs) -> io::Result<()> {
...
+
                         for (project_name, _) in projects {
-                            if let Ok(handle) =
-                                gateway.new_task().project(project_name).send(&sender).await
-                            {
-                                // We wait for the check to be done before
-                                // queuing up the next one.
-                                handle.await
+                            // Wait for completion of next future before enqueuing a new one
+                            if work_set.len() >= 8 {
```
iulianbarbu, line 116: Can we extract `8` and add it somewhere as a constant?

iulianbarbu (approved):
LGTM!

PR #1466, opened 2023-12-05
fix(cargo-shuttle): remove newline from errored project state output

### LINKED ISSUE #1451: Project list table wraps for `errored` state
The table when running `cargo shuttle project list` breaks, I believe due to the `\t` in the line below:

https://github.com/shuttle-hq/shuttle/blob/5f58d283b00eba0ea6625d97c8590ed0167e47a5/common/src/models/project.rs#L135

Screenshot of the issue
image

Missed this in #1319 as I couldn't test the output for an errored project. I think removing the tab (`\t`) will centre the text correctly and avoid the miscount by the table which results in the line being too wide.

supleed2:
## Description of change

Updates `impl Display for State` to remove tab and newline from errored project state, which caused issues with the table output as shown in #1451.
Closes #1451.

## How has this been tested?

The project I would use to test has stopped now, so I don't have one in an errored state, but this change is very localised and does not affect control flow, only text output.

### Sidenote

In future it would be cool to have a testing API url which has a range of dummy projects / states, but that's for another issue :)

### DIFF
--- a/common/src/models/project.rs
+++ b/common/src/models/project.rs
@@ -132,7 +132,7 @@ impl Display for State {
             State::Destroying => write!(f, "destroying"),
             State::Destroyed => write!(f, "destroyed"),
             State::Errored { message } => {
-                write!(f, "errored\n\tmessage: {message}")
+                write!(f, "errored (message: {message})")
             }
             State::Deleted => write!(f, "deleted"),
         }


oddgrd (approved):
Thanks! That's a good idea, we'll make a note of it. We do have a staging environment, but it's not public. Once we add more in the future, we can consider something like this that would make it easier for contributors. :+1: For now, it should be possible to test this by manually editing the gateway project state and setting it to errored, I *think*.

jonaro00:
Modifying the project state to something invalid will cause an errored state IIRC

PR #1475, opened 2023-12-07
fix: cargo audit failures

oddgrd:
## Description of change

This PR removes updates our dependencies using `cargo update`, this resolves one of the cargo warnings, with pertains to the yanked wasm-bindegen version 0.2.88. 

The other warning is for https://rustsec.org/advisories/RUSTSEC-2023-0071.html, which is pulled in by `sqlx/mysql`, which we don't use, but it is pulled in due to a bug in cargo, see https://github.com/launchbadge/sqlx/issues/2911 and https://github.com/rust-lang/cargo/issues/10801. We should be safe to ignore the warning for `rsa`. A fix for the vulnerability is also [underway](https://github.com/RustCrypto/RSA/pull/394), but we'd need to wait for it to be finished and pulled into `sqlx`. I am following the `rsa` PR, so I will be notified when it is merged, we can then remove this ignore (when `sqlx` updates their `rsa` dependency).

CI.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -3,5 +3,8 @@ ignore = [
   # Advisory about the crate `mach` being unmaintained.
   # This is a dependency from `wasmtime`. A fork exists (`mach2`) but they haven't yet audited it.
   # Corresponding issue: https://github.com/bytecodealliance/wasmtime/issues/6000
-  "RUSTSEC-2020-0168"
+  "RUSTSEC-2020-0168",
+  # Advisory about a vulnerability in rsa, which we don't use, but it is pulled in via sqlx due
+  # to a bug in cargo. See https://github.com/shuttle-hq/shuttle/pull/1475 for context.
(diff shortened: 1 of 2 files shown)


oddgrd:
> One quick question before approving: will the fix here [RustCrypto/RSA#394](https://github.com/RustCrypto/RSA/pull/394) need to be followed by PR(s) in `cargo`, and a PR on our side, where we upgrade the rust version (including cargo), used to build the Shuttle crates that depend on `sqlx`? Just trying to understand the dependencies we'll need to get sorted out before removing the whitelisting of the advisory.

If a fix was released for `cargo` we would not rely on the `rsa` fix, since we would no longer depend on it via `sqlx/mysql`. But, I don't expect the `cargo` fix any time soon. The first fix we are likely to get is `sqlx` updating to a fixed version of `rsa`.

iulianbarbu (approved):

PR #1476, opened 2023-12-07
fix(gateway): dynamically pick docker stats source

fatfingers23:
## Description of change
This is a possible solution to dynamically get the cpu stats to return the ability to run the gateway on other systems without cgrpoup v1 that was lost in #1464 

This is a draft, the test suite has not been run and I don't think tests will even build. I wanted to see if this was a direction that the shuttle team felt like it should go and get feedback before polishing it. 

Ran locally on a Mac. Builds, runs, and can build and deploy new projects. I don't know how else to test it, but I assumed the service would get a fatal error during this process. 

I did not test on a Linux system but left the parser used for cgroup v1 and took the example that was commented out for v2


gateway/src/project.rs
```diff
@@ -1402,7 +1430,25 @@ where
...
+                        Some(StatsOptions {
+                            one_shot: true,
+                            stream: false,
+                        }),
+                    )
+                    .next()
+                    .await
+                    .unwrap()?;
+                new_stat.cpu_stats.cpu_usage.total_usage
+            }
```
Kazy, line 1450: What do you think of moving that to the `impl` of `DockerStatsSource`, or even of `DockerContext` itself (well, inside their implementation) ? And instead just have `ctx.get_stats(&container)` here. 

In addition to the encapsulation of the stats feature, it would help to have a single common span name for the instrumentation + an additional field that represents its version. Currently `get_container_stats_cgroup_v2` and `get_container_stats_cgroup_v1` have a different name, and if we call the `bollard` version, there is no span AFAIK. If you're going the route of having a single `get_stats` function, you'll just have one place to instrument.

gateway/src/main.rs
```diff
@@ -38,18 +40,16 @@ async fn main() -> io::Result<()> {
...
+
+    let docker_stats_path_v1 = PathBuf::from_str(DOCKER_STATS_PATH_CGROUP_V1)
+        .expect("to parse docker stats path for cgroup v1");
+    let docker_stats_path_v2 = PathBuf::from_str(DOCKER_STATS_PATH_CGROUP_V2)
+        .expect("to parse docker stats path for cgroup v2");
+    if docker_stats_path_v1.exists() {
+        docker_stats_source = DockerStatsSource::CgroupV1;
+    } else if docker_stats_path_v2.exists() {
+        docker_stats_source = DockerStatsSource::CgroupV2;
     }
```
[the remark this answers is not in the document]
Kazy: I would favor exiting early, considering how important this is.

gateway/src/main.rs
```diff
@@ -38,18 +40,16 @@ async fn main() -> io::Result<()> {
...
-    // Return an error early if the docker stats path is not in the expected location.
-    if !docker_stats_path.exists() {
-        return Err(std::io::Error::new(
-            io::ErrorKind::NotFound,
-            format!(
-                "could not find docker stats at path: {:?}",
-                DOCKER_STATS_PATH
-            ),
-        ));
+    let mut docker_stats_source = DockerStatsSource::Bollard;
```
Kazy, line 43: I would probably add a log statement to know which method was chosen for the stats.

Kazy (commented):
Good stuff, thanks a lot for taking a stab at it ! This feels like what I wanted us to do, so great :)

gateway/src/service.rs
```diff
@@ -1023,6 +1040,62 @@ impl DockerContext for GatewayContext {
...
+                .unwrap_or_default()
+                .lines()
+                .next()
+                .unwrap()
+                .split(' ')
+                .nth(1)
+                .unwrap_or_default()
+                .parse::<u64>()
+                .unwrap_or_default()
+                    * 1_000;
```
oddgrd, line 1077: We should handle these errors like we do for cgroup v1 above.

gateway/src/service.rs
```diff
@@ -1023,6 +1040,62 @@ impl DockerContext for GatewayContext {
     fn container_settings(&self) -> &ContainerSettings {
         &self.settings
     }
+
+    #[instrument(name = "getting container stats from the proper source", skip_all)]
```
oddgrd, line 1044: I think we should record these fields.
```suggestion
    #[instrument(name = "getting container stats from the proper source", skip_all, fields(docker_stats_source = %self.docker_stats_source, shuttle.container.id = container_id))]
```

oddgrd (commented):
This is great, thanks! I tested it on my machine which uses cgroupv2, and that works, except for a small path bug I left a comment about. I'll test it on our staging environment as well when the reviews are addressed, which uses cgroupv1. 

Regarding your testing with setting the source to a cgroup, that seems odd. :thinking: I would expect the projects to become errored when this fails, but perhaps it's due to the incomplete error handling of the cgroupv2 source? I left a comment about that.

fatfingers23:
Glad it worked! Ahh, I think it must have been me using cgroupv2 with the unwraps. I was able to get the errors I was expecting with these changes and hardcode it to a cgroup. Now, it should be good to go!

gateway/src/service.rs
```diff
@@ -1071,7 +1155,15 @@ pub mod tests {
     #[tokio::test]
     async fn service_create_find_stop_delete_project() -> anyhow::Result<()> {
         let world = World::new().await;
-        let svc = Arc::new(GatewayService::init(world.args(), world.pool(), "".into()).await);
+        let svc = Arc::new(
+            GatewayService::init(
+                world.args(),
+                world.pool(),
+                "".into(),
+                DockerStatsSource::Bollard,
```
oddgrd, line 1163: Hmm, rather than passing in a new arg and defaulting to Bollard, should we do the check we do in gateway main in `GatewayService::init`, and use that to decide the stats source? This way, tests ran locally on Linux and in CI (also Linux) will use the same source as our deployments.

oddgrd (commented):
This looks good to me! :pray: I have just one small comment on the test setup. Other than that, we will be a bit delayed in testing this in our staging environment, but we're planning to get that done and merge this by the end of the week. :+1:

gateway/src/service.rs
```diff
@@ -1023,6 +1040,73 @@ impl DockerContext for GatewayContext {
     fn container_settings(&self) -> &ContainerSettings {
         &self.settings
     }
+
+    #[instrument(name = "getting container stats from the proper source", skip_all, fields(docker_stats_source = %self.docker_stats_source, shuttle.container.id = container_id))]
```
Kazy, line 1044: Just a small nitpick left for me :)

```suggestion
    #[instrument(name = "get container stats", skip_all, fields(docker_stats_source = %self.docker_stats_source, shuttle.container.id = container_id))]
```

gateway/src/service.rs
fatfingers23: Sure, we can do that! I am still getting familiar with Rust and its error handling. I am currently searching for a solution, but do you happen to know where in the project I could find some inspiration on how to throw an error if Bollard is the source type and the environment is staging/production? Or should I consider changing the return type of init to a Result, then handle the error the same as before on `start` in `gateway/main` with a match or something similar?

fatfingers23:
I moved the Docker stats source from `main.rs` to `init` in `service.rs`. Additionally, I changed `init` to a result so that I could generate an error if the crgroup path was not found in staging or production.

The only other solution I had was to check the environment variables and file locations in `main.rs` so it could return as soon as possible, but I decided to go with this solution to avoid performing that check twice. Please let me know if you would like anything changed.

Thank you!

oddgrd (approved):
Thanks @fatfingers23, for figuring out the strategy, implementing and testing this! I tested it locally on cgroupv2 and in our staging environment which is v1, it works as expected. LGTM! :partying_face:

gateway/src/service.rs
oddgrd: Oh, sorry, I didn't see this comment until now. You handled it well, though!

PR #1478, opened 2023-12-08
feat(auth): add subscription items endpoint

oddgrd:
## Description of change

Add an endpoint to auth so we can add subscription items when we provision resources. This will add RDS (or other price_ids) as a product to the existing subscription, as a recurring item. If it is added in the middle of a subscription period, only the cost of the remaining days will be added to the invoice.

An endpoint to remove items from the subscription will follow in a subsequent PR.

I have tested this manually on staging, it deployed correctly with the new arg, it increased the subscription correctly and returned the expected message for the initial deployment of the resource. 

I also refactored the tests to make it a bit easier to mock stripe, and added a test for the new endpoint.


deployer/src/deployment/run.rs
```diff
@@ -379,23 +394,52 @@ async fn load(
...
                 })
                 .collect();
+
             resource_manager
                 .insert_resources(resources, &service_id, claim.clone())
                 .await
                 .expect("to add resource to persistence");
 
+            if !cached_resources_contains_rds && loaded_resources_contains_rds {
+                // TODO: this message should be generated by the resource type?
```
oddgrd, line 430: I'm thinking of refactoring to an extension trait to extract prices and price_ids from resources.

oddgrd: Although it's not ideal that the deployer knows the price/price_id of an item, that means we can't change it without the deployer being updated. I might rather hard-code a message for now, that the user should check the console to see the increased cost, and in the long run we can get it from a service we control, e.g. a call to auth to get billing details for a resource.

common/src/constants.rs
```diff
@@ -38,3 +38,9 @@ pub mod limits {
     pub const MAX_PROJECTS_DEFAULT: u32 = 3;
     pub const MAX_PROJECTS_EXTRA: u32 = 15;
 }
+
+pub mod stripe_price_ids {
+    /// The price ID of the recurring AWS RDS instance product.
+    /// TODO: this is set to the id of the test product, set to real product before releasing.
+    pub const AWS_RDS_INSTANCE_RECURRING: &str = "price_1OIS06FrN7EDaGOjaV0GXD7P";
```
chesedo, line 45: Maybe this should rather be passed as an arg to auth?

auth/src/api/builder.rs
```diff
@@ -73,6 +74,19 @@ impl ApiBuilder {
             .route("/auth/refresh", post(refresh_token))
             .route("/public-key", get(get_public_key))
             .route("/users/:account_name", get(get_user))
+            .route(
+                "/users/subscription/items",
+                post(
+                    add_subscription_items.layer(JwtAuthenticationLayer::new(move || {
+                        let public_key = public_key.clone();
+                        async move { public_key.clone() }
+                    })),
```
jonaro00, line 83: Why is this needed only on this endpoint?

auth/src/api/builder.rs
```diff
@@ -73,6 +74,19 @@ impl ApiBuilder {
...
+                ),
+            )
+            .route(
+                "/users/subscription/:account_name",
```
jonaro00, line 87: Path param is never used. Can it therefore be simplified?

deployer/src/deployment/run.rs
chesedo: Redirecting to the console sounds like a good idea for now

chesedo (commented):
Left a few comments :smile:

auth/src/api/builder.rs
```diff
@@ -73,6 +74,19 @@ impl ApiBuilder {
...
+                post(add_subscription_items),
```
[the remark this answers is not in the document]
oddgrd: Woops, I forgot to remove this. Thanks!

deployer/src/deployment/run.rs
jonaro00: Make deployer dumb again :tm: ;)

auth/src/api/builder.rs
oddgrd: The auth API has only dealt with API keys or cookies so far, but rather than creating an admin user for provisioner and using that api key to call auth, we decided to just pass on the JWT of the user. This will be the only endpoint that needs to take a JWT, so we're just putting the layer on this handler for now.

auth/src/api/builder.rs
oddgrd: This is a route I forgot to remove, so it can be simplified a lot (deleted :smile:)

auth/tests/api/users.rs
```diff
@@ -227,6 +229,74 @@ mod needs_docker {
...
+        let request = Request::builder()
+            .uri("/users/subscription/items")
+            .method("POST")
+            .header(CONTENT_TYPE, "application/json")
+            .body(Body::from(subscription_item.clone()))
+            .unwrap();
+
+        let response = app.send_request(request).await;
+
+        assert_eq!(response.status(), StatusCode::INTERNAL_SERVER_ERROR);
```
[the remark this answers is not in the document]
oddgrd: The layer is designed to return 400 (although it should be 401) for this, but since there is no token in the request axum returns a 500 due to failing to set the extension. When we pass an invalid token in the next test, however, the extension should not be set either, but that returns the expected 400. :thinking:

auth/src/api/builder.rs
jonaro00: So what happens if there is no cached JWT and the first request made is to this endpoint?

oddgrd: Hmm, what do you mean by that? The provisioner needs a JWT to provision RDS, and it then uses this JWT to call the subscription update. It could be problematic if the JWT expires while provisioning the RDS instance, though, but in that case (which should be rare) we may just delete the RDS instance and the user needs to try again.

auth/src/args.rs
```diff
@@ -31,7 +31,7 @@ pub struct StartArgs {
 
     /// Auth JWT signing private key, as a base64 encoding of
     /// a PEM encoded PKCS#8 v1 formatted unencrypted private key.
-    #[arg(long, default_value = "")]
+    #[arg(long)]
```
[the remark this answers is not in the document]
iulianbarbu: No particular reason, I wanted to force the program to fail early in case the JWT signing key is not (provided or valid). Removing the `default_value` here is the same as making the arg as `required` and simplifies the code. It is a must to provide a value and if the value is invalid (base64 of a PEM encoded private key), we fail early.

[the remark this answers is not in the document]
iulianbarbu: More context here: https://github.com/shuttle-hq/shuttle/pull/1402#discussion_r1401851215.

auth/src/user.rs
```diff
@@ -246,6 +280,34 @@ impl User {
 
         Ok(false)
     }
+
+    async fn add_subscription_items(
+        &self,
+        user_manager: &UserManager,
```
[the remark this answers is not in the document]
oddgrd: Gotcha, makes sense!

provisioner/src/lib.rs
```diff
@@ -338,6 +346,34 @@ impl MyProvisioner {
...
+            .map_err(|err| {
+                error!(error = %err, "failed to connect to auth service");
+                Error::Plain("failed to connect to auth service".to_string())
+            })?;
+
+        match response.status() {
+            StatusCode::OK => Ok(()),
+            status_code => {
+                error!(status_code = %status_code, "failed to update subscription");
+                Err(Error::Plain("failed to update subscription".to_string()))
```
oddgrd, line 372: That the deployment failed and that it was due to a failure to update the subscription will be visible to the user, but it will not be very useful. The most likely cause for this error will be that the JWT expired, but it should be very rare. We could consider having some suggestions to retry the command, and if it doesn't work reach out to support? :thinking:

oddgrd: The JWT layer should return a 499 if that is the case, actually, so I should be able to match on that here and return a more appropriate message. I'll look into that.

oddgrd: Addressed in https://github.com/shuttle-hq/shuttle/pull/1478/commits/4a500aedb1689b86dcdcddfab6b7ec5fb46024b9 and [e508559](https://github.com/shuttle-hq/shuttle/pull/1478/commits/e50855924b7dda0c957b2bd657953e026b95fd53).

auth/src/subscription.rs
```diff
@@ -0,0 +1,19 @@
...
+
+impl SubscriptionItemExt for shuttle_common::backends::subscription::SubscriptionItem {
+    fn price_id(&self) -> String {
+        match self {
+            shuttle_common::backends::subscription::SubscriptionItem::AwsRds => {
+                stripe_price_ids::AWS_RDS_INSTANCE_RECURRING.to_string()
+            }
+        }
+    }
+}
```
oddgrd, line 19: This needs to be dynamic, checking whether we are on prod and returning the correct price_id. I have created a ticket for this and will implement it in a follow-up after the production price ID has been created.

Makefile
```diff
@@ -47,6 +47,7 @@ MONGO_INITDB_ROOT_USERNAME?=mongodb
 MONGO_INITDB_ROOT_PASSWORD?=password
 STRIPE_SECRET_KEY?=""
 AUTH_JWTSIGNING_PRIVATE_KEY?=""
+STRIPE_RDS_PRICE_ID?=""
```
iulianbarbu, line 50: reminder: to update shuttle-aws deploying.md guide.

auth/src/api/handlers.rs
```diff
@@ -68,6 +65,20 @@ pub(crate) async fn update_user_tier(
...
+    Extension(claim): Extension<Claim>,
+    State(user_manager): State<UserManagerState>,
+    NewSubscriptionItemExtractor(update_subscription_items): NewSubscriptionItemExtractor,
+) -> Result<(), Error> {
+    user_manager
+        .add_subscription_items(AccountName::from(claim.sub), update_subscription_items)
+        .await?;
+
+    Ok(())
+}
```
iulianbarbu, line 80: Don't we check anything about the claim?

auth/src/user.rs
```diff
@@ -206,7 +235,7 @@ impl User {
         Ok(false)
     }
 
-    // Synchronize the tiers with the subscription validity.
+    /// Synchronize the tiers with the subscription validity.
```
iulianbarbu, line 238: nit: Why three slashes? Should this function be considered public-facing, needing doc comments? (https://doc.rust-lang.org/reference/comments.html#doc-comments_)

auth/tests/api/users.rs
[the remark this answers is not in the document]
iulianbarbu: > but that returns the expected 400.

You mean 401?

Should we open a ticket for this bug if it is unrelated to the code changes, and fix it separately? It is important for our monitoring. @oddgrd you can assign it to me if you do it.

iulianbarbu (commented):
First part of the review (auth and common realated). Continuing tomorrow with the deployer and provisioner parts.

auth/src/user.rs
```diff
@@ -227,15 +227,16 @@ impl User {
...
+
             return Ok(subscription.status == SubscriptionStatus::Active
                 || subscription.status == SubscriptionStatus::Trialing);
         }
 
         Ok(false)
     }
 
-    // Synchronize the tiers with the subscription validity.
+    // TODO: should this return account tier? It could lead to a downgrade.
```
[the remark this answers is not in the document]
oddgrd: This has been removed after you started your review. It's useful to see in the add sub items method if the user was downgraded from pro in the sync, but `sync_tier`mutates the user it's called on, so we already have the tier updated and available in following calls.

auth/src/api/handlers.rs
oddgrd: Hmm, what should we check? We check in the jwt auth layer that it isn't expired.

auth/src/user.rs
oddgrd: I think doc comments are always useful, for example in VsCode the comment shows on hover, so we can quickly see the comment wherever this function is used.

common/src/constants.rs
oddgrd: Implemented in latest commits.

auth/tests/api/users.rs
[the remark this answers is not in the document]
oddgrd: Expected in the sense that the current jwt layer implementation would return a 400 for a missing bearer token. I'll open a ticket for it, since I am a bit confused about the code path that leads to this behavior: https://github.com/shuttle-hq/shuttle/blob/main/common/src/backends/auth.rs#L351-L355.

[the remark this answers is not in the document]
oddgrd: It looks like we just proceed with the request, calling inner if there is no bearer token, which seems incorrect since that should return a 401.

auth/src/subscription.rs
oddgrd: Refactored to taking in an arg for the price id, so this is no longer relevant.

auth/src/user.rs
iulianbarbu: Ah, I see. sounds good then

auth/src/api/handlers.rs
iulianbarbu: I was thinking about checking if the claim has a scope for updating the subscription. Are we checking it downstream?

oddgrd: We technically are both upstream and downstream, this endpoint is currently only called when rds is provisioned in the provisioner, which is only done if the claim has access to RDS. Before we send the update request to stripe downstream, we also check that the account tier is still pro, which means they can provision RDS. I can add a check at the handler level either way, though, since the upstream and downstream may change in the future.

auth/tests/api/users.rs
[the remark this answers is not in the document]
oddgrd: This comment https://github.com/shuttle-hq/shuttle/pull/1478#discussion_r1430498392 prompted me to add a scopedlayer to the endpoint, and now it does return 401, since the jwt auth layer calls the next layer, which is the scopedlayer, and that correctly returns 401 for missing token.

auth/src/api/handlers.rs
oddgrd: I added a scopedlayer now, restricted to "ResourcesWrite". We can consider adding a new scope for updating subscription, but I'm not sure it's necessary yet.

Adding the scopedlayer fixed the unexpected 500 from missing token: https://github.com/shuttle-hq/shuttle/pull/1478#discussion_r1431223035

auth/src/api/handlers.rs
```diff
@@ -68,6 +65,20 @@ pub(crate) async fn update_user_tier(
     Ok(())
 }
 
+// #[debug_handler]
```
chesedo, line 68: Not sure what this is? :sweat_smile:

auth/src/user.rs
```diff
@@ -246,6 +275,32 @@ impl User {
...
+        // A pro or admin account tier is required to get to this point, but the ssubscription sync
+        // ahead of this call may lead to an account downgrade.
+        if !matches![self.account_tier, AccountTier::Pro | AccountTier::Admin] {
+            error!("account was downgraded from pro in sync, denying the addition of new items");
+            return Err(Error::Unauthorized);
+        }
+
+        let Some(ref subscription_id) = self.subscription_id else {
+            return Err(Error::MissingSubscriptionId);
+        };
```
chesedo, line 293: Since this is the data layer, it should not be doing business logic checks like these. Rather the handler should do these and this function should only update the needed "datastores"

chesedo (commented):
Looking good!

auth/src/user.rs
oddgrd: The difficulty with this is that the subscription is synced with stripe in the call before this, in `UserManager::add_subscription_items`, so the tier could have changed between this call and the handler checks. But I can investigate refactoring it so the subscription is synced earlier.

auth/src/api/handlers.rs
oddgrd: Remnants of past trauma. :joy: Will remove.

provisioner/src/lib.rs
```diff
@@ -470,9 +515,31 @@ impl Provisioner for MyProvisioner {
...
+                    // subscription expiring, delete the instance immediately.
+                    if let Err(err) = self
+                        .add_subscription_items(
+                            // The token should be set on the claim in the JWT auth layer.
+                            claim.token().expect("claim should have a token"),
+                            NewSubscriptionItem::new(SubscriptionItem::AwsRds, 1),
+                        )
+                        .await
+                    {
+                        self.delete_aws_rds(&request.project_name, &engine).await?;
```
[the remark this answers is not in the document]
oddgrd: Sure!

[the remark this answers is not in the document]
oddgrd: Hmm, there will be an error! event if add_subscription_items fails, so I'm not sure a debug event will be useful here.

chesedo (approved):
LGTM!

Makefile
oddgrd: PR sent.

auth/src/user.rs
oddgrd: Addressed in latest commits.

PR #1485, opened 2023-12-15
feat: track deployments

GugaGongadze:
## Description of change
The idea was to implement tracking for 2 user actions: Logging in via CLI and project deployment.
Deployment part is pretty straightforward and is included in this PR.

Logging in is a bit tricky: We already track login actions in console, and we have a restriction that we cannot add tracking actions to the CLI itself (maybe we could come around this idea later?).

Anyways, logging in works in 2 ways from the CLI: if you provide you api key upfront, the request does not go anywhere and we cannot track it since it all happens inside cargo-shuttle.
Otherwise user is redirected to console to log in, which we already track.

`shuttle_api_start_deployment` was ingested by PostHog client.


deployer/src/handlers/mod.rs
```diff
@@ -621,6 +623,12 @@ pub async fn start_deployment(
         };
         deployment_manager.run_push(built).await;
 
+        let event =
+            async_posthog::Event::new("shuttle_api_start_deployment".to_string(), account_name);
+        if let Err(err) = deployment_manager.posthog_client().capture(event).await {
+            error!(error = %err, "failed to send event to posthog")
+        };
```
jonaro00, line 630: nit: let this be done in a tokio task to not disrupt the execution of this thread and delay the deployment.

jonaro00 (commented):
check clippy as well

GugaGongadze:
@jonaro00 Is there s way to integrate clippy locally? cargo check did not yield any errors for that line.

deployer/Cargo.toml
```diff
@@ -57,6 +57,7 @@ utoipa = { workspace = true }
 utoipa-swagger-ui = { workspace = true }
 ulid = { workspace = true }
 uuid = { workspace = true, features = ["v4"] }
+ async-posthog = { git = "https://github.com/shuttle-hq/posthog-rs", branch = "main" }
```
oddgrd, line 60:
```suggestion
async-posthog = { git = "https://github.com/shuttle-hq/posthog-rs", branch = "main" }
```

deployer/src/args.rs
```diff
@@ -70,4 +70,8 @@ pub struct Args {
     /// Add an auth layer to deployer for local development
     #[arg(long)]
     pub local: bool,
+
+    // Posthog client key
+    #[clap(long)]
+    pub posthog_key: String,
```
oddgrd, line 76: We'll need to add this to the makefile too, as well as adding the key for both staging and production to aws secrets manager + circleci before release. We then need to pass it in when we create the deployer config in gateway: https://github.com/shuttle-hq/shuttle/blob/main/gateway/src/project.rs#L872-L913. To know more about how this is done from makefile to gateway, I would look at how some of the other deployer args are handled.

deployer/src/handlers/mod.rs
```diff
@@ -609,6 +609,8 @@ pub async fn start_deployment(
     CustomErrorPath((project_name, deployment_id)): CustomErrorPath<(String, Uuid)>,
 ) -> Result<()> {
     if let Some(deployment) = persistence.get_runnable_deployment(&deployment_id).await? {
+        let account_name = claim.sub.replace('|', "-");
```
oddgrd, line 612: Do we have account names with `|`?

deployer/src/handlers/mod.rs
```diff
@@ -621,6 +623,19 @@ pub async fn start_deployment(
...
 
+        tokio::spawn(async move {
+            let event =
+                async_posthog::Event::new("shuttle_api_start_deployment".to_string(), account_name);
+            match deployment_manager.posthog_client().capture(event).await {
+                Ok(_) => {
+                    info!("Event sent to posthog");
+                }
+                Err(err) => {
+                    error!(error = %err, "failed to send event to posthog");
```
oddgrd, line 634: I don't think we should log anything in the Ok path, we're covered with just the error event. So we can do `if let Err(err) = posthog_client()...` here.

oddgrd (commented):
Just left a couple of nits, and some thoughts on how we need to handle the new deployer arg. Otherwise LGTM!

deployer/src/handlers/mod.rs
```diff
@@ -621,6 +623,19 @@ pub async fn start_deployment(
         };
         deployment_manager.run_push(built).await;
 
+        tokio::spawn(async move {
+            let event =
+                async_posthog::Event::new("shuttle_api_start_deployment".to_string(), account_name);
```
chesedo, line 628:
```suggestion
                async_posthog::Event::new("shuttle.start_deployment".to_string(), account_name);
```
I'm not sure what the PostHog conventions are for naming events. So just a suggestion.

deployer/src/handlers/mod.rs
GugaGongadze: This `sub` field comes with `|` in them. That's how Auth0 stores users, not sure why though.

GugaGongadze: Just took a look at this, `sub` is indeed coming from auth0, and `|` is the delimiter they use. It goes like: `{social_provider|{unique_id}}.

oddgrd: Hmm, maybe I am missing something, but the claim is created here: https://github.com/shuttle-hq/shuttle/blob/main/auth/src/api/handlers.rs#L130-L142, and the sub is created from the account name that is stored in `auth`, and they are stored as `github-<id>` (at least all the ones in the staging database). Have you seen the format you speak of in your testing?

GugaGongadze: Not sure where it's set in backend, but it always comes with `|` from auth0.

When user first registers on Shuttle using Auth0 Github connection, this piece of code is run:

```
const shuttlified = session.user.sub.replace('|', '-');
// At this point no calls have been made to our backend, `|` is coming from Auth0.

// We then create new user using this replaced version.
const user = await Shuttle.create_user(shuttlified);

// Then we set user info to session and that's basically it
session.user.api_key = user.key;
session.user.userId = user.name;
session.user.account_tier = user.account_tier;
session.user.subscription_id = user.subscription_id;
```

GugaGongadze: Auth0 probably calls `/auth/session` endpoint suring it's oauth flow with account id having `|`.

deployer/src/handlers/mod.rs
GugaGongadze: This actually resembles the naming convention we already use. For example:
`console_project_back`

deployer/src/handlers/mod.rs
oddgrd: Ah, I see. So the claim subject will always be in the `github-<id>` format. Auth0 is not involved in the creation of the JWT, auth0 is just involved in the creation of a user, and from your codeblock it's clear that we change it to `github-<id>` before we send it to the backend, where it is persisted in the auth state. When the claim is created, it gets the account name from this state.

oddgrd:
This looks nearly there @GugaGongadze, you'll just need to resolve the compiler error from the clippy run in CI, and resolve the conflicts.

gateway/src/lib.rs
```diff
@@ -559,6 +559,9 @@ pub mod tests {
 
             let docker_host = "/var/run/docker.sock".to_string();
 
+            let posthog_key = env::var("POSTHOG_KEY")
+                .unwrap_or_else(|_| "public.ecr.aws/shuttle-dev/deployer:latest".to_string());
```
oddgrd, line 563: Perhaps we should just set this to "" for tests? We could also mock it for testing with wiremock, but this functionality is simple, so if adding mocking here will delay the release, it can wait.

deployer/src/handlers/mod.rs
oddgrd: We should be good to remove this line then.

.circleci/config.yml
```diff
@@ -420,6 +420,9 @@ jobs:
       stripe-secret-key:
         description: "Stripe secret key used to connect a client to Stripe backend"
         type: string
+      posthog-key:
+        description: "S Public key used for PostHog connection"
```
oddgrd, line 424:
```suggestion
        description: "Public key used for PostHog connection"
```

oddgrd (commented):
Hmm, I see some deployer tests are failing in CI, we should try to resolve those before we merge this. Pending that and the staging testing, this LGTM!

GugaGongadze:
@oddgrd Why is it ok to remove `.replace` line?

oddgrd:
> @oddgrd Why is it ok to remove `.replace` line?

The replacement of `|` with `-` is done in the console/frontend before it is sent to the `auth` service for persistence, and it's from the `auth` state the `claim.sub` is set. So we shouldn't see any claim subjects with `|` in the name.

Or did I misunderstand and the intention here is to change the account name to have a `|` before sending it to posthog? In that case, it should be `.replace("-", "|")`.

GugaGongadze:
> > @oddgrd Why is it ok to remove `.replace` line?
> 
> The replacement of `|` with `-` is done in the console/frontend before it is sent to the `auth` service for persistence, and it's from the `auth` state the `claim.sub` is set. So we shouldn't see any claim subjects with `|` in the name.
> 
> Or did I misunderstand and the intention here is to change the account name to have a `|` before sending it to posthog? In that case, it should be `.replace("-", "|")`.

Console replaces it so auth can set the account id in shuttle-specific format. ie, with `-`.
We are using `sub` field here, which is always with `|`.

oddgrd:
Why do you say that `claim.sub` uses `|`? The claim is created here: https://github.com/shuttle-hq/shuttle/blob/main/auth/src/api/handlers.rs#L155-L167, and it gets the account name from the `auth` state, which uses `-`. Have you seen it differently in your testing?

docker-compose.yml
```diff
@@ -153,6 +153,7 @@ services:
       - "--proxy-fqdn=${APPS_FQDN}"
       - "--use-tls=${USE_TLS}"
       - "--admin-key=${GATEWAY_ADMIN_KEY}"
+      - "--posthog-key=${POSTHOG_KEY}"
```
iulianbarbu, line 156: What should this look like when doing local development? We have some notes in DEVELOPING.md for Stripe and auth testing, and we might want to mention a few notes for Posthog too.

GugaGongadze: Yeah, already reference it in `DEVELOPING.md` file.

deployer/src/args.rs
jonaro00: Make sure that any use of the key is optional (wherever we move this to), and that nothing happens if it is unset

docker-compose.yml
jonaro00: Since the eu hostname is already hard-coded, can we consider hard-coding the key as well, since it is public?

GugaGongadze:
Closing is favor of: https://github.com/shuttle-hq/shuttle/pull/1508

PR #1486, opened 2023-12-15
chore: bump salvo

dalton-oliveira:
## Description of change
Current [Salvo](https://github.com/salvo-rs/salvo) dependency version doesn't have websocket support and current salvo version introduces a breaking change.

Testes locally only as with current version of Salvo it's not possible to deploy.


oddgrd:
Thank you for the PR @dalton-oliveira! We're a bit busy with the holidays, and we have some things we have to get to first, but we'll try to get this reviewed and merged as soon as we can.

oddgrd (approved):
Thanks! Our current Salvo example is indeed broken without this update, so this is a no-brainer. :smile:

PR #1487, opened 2023-12-19
auth: guard the `/auth/key` endpoint

iulianbarbu:
## Description of change

This PR is guarding the `/auth/key` endpoint.

Locally, when adding an admin secret that doesn't correspond to an admin key. This made the JWT-guarded API calls fail with an `unauthorized` status code.


docker-compose.dev.yml
```diff
@@ -68,6 +68,24 @@ services:
...
+      - --prefix=shuttle_
+      - --network-name=shuttle-dev_user-net
+      - --docker-host=/var/run/docker.sock
+      - --auth-uri=http://auth:8000
+      - --deploys-api-key=gateway4deployes
+      - --provisioner-host=provisioner
+      - --builder-host=builder
+      - --proxy-fqdn=unstable.shuttleapp.rs
+      - --use-tls=disable
+      - --convert-key-to-jwt-secret=dh9z58jttoes3qvt
```
jonaro00, line 88: looks like a lot of unnecessary hardcoding. Can we not just add the convert key as a default in makefile?

jonaro00 (commented):
nit(optional): the variable name might be a bit excessive/confusing. How about `gateway-admin-key`?

docker-compose.dev.yml
iulianbarbu: Yes, will do.

iulianbarbu:
> nit(optional): the variable name might be a bit excessive/confusing. How about `gateway-admin-key`?

Yes, that can work. Initially, I thought I could scope it down, but that can happen only for JWTs.

docker-compose.dev.yml
iulianbarbu: Covered both in [3cb7107](https://github.com/shuttle-hq/shuttle/pull/1487/commits/3cb71076febad054919b32b654b327bc3c642e14). Let me know if you see something else than needs to be addressed.

auth/tests/api/auth.rs
```diff
@@ -5,64 +5,89 @@ mod needs_docker {
...
+        // Create test basic user
+        let response = app.post_user("test-user-basic", "basic").await;
         assert_eq!(response.status(), StatusCode::OK);
+        // Extract the API key from the response so we can use it in a future request.
+        let body = hyper::body::to_bytes(response.into_body()).await.unwrap();
+        let user: Value = serde_json::from_slice(&body).unwrap();
+        let basic_user_key = user["key"].as_str().unwrap();
 
+        // Create test admin user
+        let response = app.post_user("test-user-admin", "admin").await;
```
oddgrd, line 23: We create an admin user when we initialize the app, for which the API key is available as a constant, could we use that to make the test simpler?

auth/tests/api/helpers.rs
```diff
@@ -117,6 +120,20 @@ impl TestApp {
         self.send_request(request).await
     }
 
+    pub async fn get_jwt_from_non_admin_api_key(
```
oddgrd, line 123: Should we refactor or remove the original "get_jwt_...` method, since it's no longer usable?

oddgrd (commented):
Just some small notes on the test, you'll also need to resolve the conflicts now, after which there should be another use of "get_jwt_from_api_key" method to fix.

auth/tests/api/auth.rs
iulianbarbu: Yes, definitely. Thanks for the suggestion, done!

auth/tests/api/helpers.rs
iulianbarbu: Thanks! Simplified the jwt from api key generation too. LMK if it looks better now.

auth/tests/api/helpers.rs
```diff
@@ -108,12 +111,22 @@ impl TestApp {
         self.send_request(request).await
     }
 
-    pub async fn get_jwt_from_api_key(&self, api_key: &str) -> Response {
-        let request = Request::builder()
+    /// If we don't provide a valid admin key, then the`basic_api_key` parameter
+    /// should be of an admin user.
+    pub async fn get_jwt_from_api_key(
+        &self,
+        basic_api_key: &str,
```
oddgrd, line 118: Should this rather be named something like "user_api_key"?

iulianbarbu: Yes, would be better indeed. Done.

auth/src/user.rs
```diff
@@ -408,12 +408,29 @@ where
...
+                if admin_user.is_admin() {
+                    Ok(Self { user: admin_user })
+                } else {
+                    Err(Error::Unauthorized)
+                }
+            }
+            Ok(_) => Err(Error::Unauthorized),
+            // Returning forbidden for the cases where we don't understand why we can not authorize.
+            Err(_) => Err(Error::Forbidden),
+        };
```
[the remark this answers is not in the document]
iulianbarbu: Haha, good point 😅 . Done this too.

PR #1492, opened 2023-12-22
refactor(gateway): renew gateway certificate returns more info about success

iulianbarbu:
## Description of change

Simplified the `renew_gateway` handler, and returned more info to differentiate between renewing the gateway vs not renewing it because there are more than 30 days of validity, which means we can wait for a little longer until we renew the certificate.

N/A


chesedo (approved):
LGTM

oddgrd (approved):

PR #1495, opened 2023-12-26
fix: return empty list when querying for project deployments

GugaGongadze:
## Description of change
This fixes `get_deployments` function, so that it retuns empty list instead of error if there are no deployments associated with the project yet.

Tested in local development -> Console and Postman.


oddgrd (commented):
How does this affect the CLI output? It currently returns this:
image

It is a bit confusing that this is an error about service not found, since it may not be clear to the user what is meant by service here. The first suggestion is also somewhat misleading, but the suggestions in the error message itself is useful. But yeah, does the change to return an empty list break the CLI? Or will it just be an empty table? :thinking:

jonaro00:
The CLI likely needs an update to show "no deployments found" instead of an empty table

cargo-shuttle/src/lib.rs
```diff
@@ -847,7 +847,12 @@ impl Shuttle {
         let table = get_deployments_table(&deployments, proj_name, page, raw, page_hint);
 
         println!("{table}");
-        println!("Run `cargo shuttle logs <id>` to get logs for a given deployment.");
+
+        if deployments.is_empty() {
+            println!("Run `cargo shuttle deploy` to deploy your project.");
+        } else {
+            println!("Run `cargo shuttle logs <id>` to get logs for a given deployment.");
+        }
```
jonaro00, line 855: I think the table call should be skipped as well if empty

GugaGongadze: @jonaro00 It returns this message: `No deployments are linked to this service` which is nice to have.

jonaro00 (approved):

oddgrd (approved):

PR #1500, opened 2023-12-28
fix: tracing fixes and nits

oddgrd:
## Description of change

Tracing instrument uses the debug impl by default, and for newtypes that means output like `AccountName("name")`. This PR fixes that by using the display impl instead. Additionally, I added some missing fields in the deployer proxy span, and set the status code to 500 in case of errors. 

For the deployer proxy I added the `shuttle.service` field to the span, since it is being recorded in the function body. We do already have the `service` field, however, which is the same thing, and that is being recorded correctly. I'd like to deduplicate it and have just `shuttle.service`, but I haven't figured out where `service` is actually recorded yet.

TODO: test this to verify these updates work as expected.

### DIFF
--- a/deployer/src/proxy.rs
+++ b/deployer/src/proxy.rs
@@ -23,7 +23,7 @@ static PROXY_CLIENT: Lazy<ReverseProxy<HttpConnector<GaiResolver>>> =
     Lazy::new(|| ReverseProxy::new(Client::new()));
 static SERVER_HEADER: Lazy<HeaderValue> = Lazy::new(|| "shuttle.rs".parse().unwrap());
 
-#[instrument(name = "proxy_request", skip_all, fields(http.method = %req.method(), http.uri = %req.uri(), http.status_code = field::Empty, service = field::Empty))]
+#[instrument(name = "proxy_request", skip_all, fields(http.method = %req.method(), http.uri = %req.uri(), http.status_code = field::Empty, http.host = field::Empty, shuttle.service = field::Empty, service = field::Empty))]
 pub async fn handle(
     remote_address: SocketAddr,
     fqdn: FQDN,
@@ -118,6 +118,12 @@ pub async fn handle(
                     "error while handling request needing upgrade in reverse proxy"
                 ),
             };
+
+            Span::current().record(
+                "http.status_code",
+                StatusCode::INTERNAL_SERVER_ERROR.as_u16(),
+            );
+
             Ok(Response::builder()
                 .status(StatusCode::INTERNAL_SERVER_ERROR)
                 .body(Body::empty())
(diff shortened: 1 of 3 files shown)


deployer/src/proxy.rs
```diff
@@ -81,6 +81,8 @@ pub async fn handle(
         Ok(Some(address)) => address,
         Ok(None) => {
             trace!(?host, service, "service not found on this server");
+            Span::current().record("http.status_code", StatusCode::NOT_FOUND.as_u16());
```
chesedo, line 84: I'm thinking the proxy should return a unique status code so that we know something went wrong in the proxy (and not the users app)

gateway/src/task.rs
```diff
@@ -491,7 +491,7 @@ impl Task<()> for ProjectTask {
 
         let span = info_span!(
             "polling project",
-            ctx.project = ?project_ctx.project_name.to_string(),
+            ctx.project = %project_ctx.project_name,
             ctx.state = project_ctx.state.state(),
             ctx.state_after = field::Empty
```
chesedo, line 496: Maybe we should rename these to `shuttle.project.name` or something?

chesedo (commented):
Nice, thanks O! Left some comments

deployer/src/proxy.rs
oddgrd: Yeah, that makes sense!

gateway/src/task.rs
oddgrd: We can do that, we currently have a lot of different names for the same thing, e.g. "project", "project_name", "ctx.project".

deployer/src/proxy.rs
oddgrd: I added a new field `proxy.status_code` for the status code from the proxy itself.

gateway/src/task.rs
oddgrd: I cleaned this up, so now all references to a project name should have they key `shuttle.project.name`. Tested in honeycomb that there are no traces where the old keys exist after running a bunch of project commands. I also changed `shuttle.service` to ` shuttle.service.name`, and replaced all instances of `service_name` keys with it.

gateway/src/proxy.rs
```diff
@@ -139,7 +139,7 @@ impl UserProxy {
             .await?;
 
         // Record current project for tracing purposes
-        span.record("project", &project_name.to_string());
+        span.record("shuttle.project.name", &project_name.to_string());
```
chesedo, line 142: Remember to also update the `field::Empty` name in the span / instrument

chesedo (commented):
Just on more comment

gateway/src/proxy.rs
oddgrd: Done, I missed on in task.rs as well, fixed now.

chesedo (approved):

PR #1502, opened 2023-12-29
docs: update CONTRIBUTING.md

sponkurtus2:
## Description of change
I  added a code style section so new users can understand how they should write code for PR


oddgrd (commented):
Thanks, @sponkurtus2! This is great, but I am a bit hesitant to merge it, since we already have some notes on running clippy and cargo fmt under the `Comitting` header. One could argue they rather belong in a section like this, though. So we could consider merging this if those were woven into this new section, maybe? I am a bit uncertain of what is the best way forward here, so I'm open to suggestions!

sponkurtus2:
Oh ok I understand, I'm a bit confused because this is my first time contributing, so is there any specific way where I can help?, maybe something you need

PR #1503, opened 2023-12-31
feat(installer): add windows installer script

### LINKED ISSUE #1496: [Feature]: Windows Powershell install script
### Describe the feature

Make a Windows cousin for the Linux/Mac [install script](https://github.com/shuttle-hq/shuttle/blob/main/install.sh).

You can take inspiration from the mentioned installer and other projects that have PS install scripts to create a script that does the same high level things:
- Attempt to install with cargo-binstall if it exists
- Attempt to download a binary from github
- Install with cargo install
  -  Offer to install Rust via rustup if cargo is not found

For example, `iwr https://[url-to-script] -useb | iex`

### Duplicate declaration

supleed2:
## Description of change

Adds Powershell script, following steps as in bash install script, closes #1496.

Install command added to readme will work once shuttle-hq/www#220 is merged.

Tested locally on multiple machines, including commenting out codepaths to check both sides of branches. The section for downloading a binary may need to be updated in the future if CI is updated to produce aarch64 windows binaries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,12 @@ On Linux and macOS, you can use this install script, which will automatically in
 curl -sSfL https://www.shuttle.rs/install | bash
 ```
 
+On Windows, you can use this install script to do the same:
+
+```powershell
+Set-ExecutionPolicy Unrestricted -Scope Process; iex (iwr "https://www.shuttle.rs/install-win").Content
+```
+
 Our binaries can also be installed using [cargo-binstall](https://github.com/cargo-bins/cargo-binstall).
 To install with `cargo-binstall`, run:
 
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ curl -sSfL https://www.shuttle.rs/install | bash
 On Windows, you can use this install script to do the same:
 
 ```powershell
-Set-ExecutionPolicy Unrestricted -Scope Process; iex (iwr "https://www.shuttle.rs/install-win").Content
+iwr "https://www.shuttle.rs/install-win" | iex
 ```
 
 Our binaries can also be installed using [cargo-binstall](https://github.com/cargo-bins/cargo-binstall).
--- a/install.ps1
+++ b/install.ps1
@@ -68,7 +68,7 @@ else {
 	if ($Arch -in "AMD64", "x86") {
 		Write-Host "Could not find cargo.exe, Rust may not be installed" -ForegroundColor Red
 		$Confirm = Read-Host -Prompt "Would you like to install Rust via Rustup? [y/N]"
-		if ($Confirm -notin "y", "yes") {
+		if ($Confirm -inotin "y", "yes") {
 			Write-Host "Skipping rustup install, cargo-shuttle not installed"
 			[Environment]::Exit(1)
 		}
--- a/install.ps1
+++ b/install.ps1
@@ -1,8 +1,7 @@
-$ErrorActionPreference = "Stop"
+function Install-Cargo-Shuttle {
+	$RepoUrl = "https://github.com/shuttle-hq/shuttle"
 
-$RepoUrl = "https://github.com/shuttle-hq/shuttle"
-
-Write-Host @"
+	Write-Host @"
      _           _   _   _
  ___| |__  _   _| |_| |_| | ___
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -87,6 +87,12 @@ On Linux and macOS, you can use this install script, which will automatically in
 curl -sSfL https://www.shuttle.rs/install | bash
 ```
 
+On Windows, you can use this install script to do the same:
+
+```powershell
+Set-ExecutionPolicy Unrestricted -Scope Process; iex (iwr "https://www.shuttle.rs/install-win").Content
```
jonaro00, line 93: Which step is the policy change needed for? Would be nice if this line is as short as possible. Is it possible to move the policy thing into the script?

Also, is piping not a thing in PS?

jonaro00 (commented):
Looks good. Will test on my windows machine later

README.md
supleed2: The execution policy is set for the current terminal session, which I believe was needed for Invoke-Expression (iex) to work. I don't think setting it inside the expression would work.

I can fix the pipe, oops.

README.md
```diff
@@ -87,6 +87,12 @@ On Linux and macOS, you can use this install script, which will automatically in
 curl -sSfL https://www.shuttle.rs/install | bash
 ```
 
+On Windows, you can use this install script to do the same:
+
+```powershell
+iwr "https://www.shuttle.rs/install-win" | iex
```
jonaro00, line 93: nit: the quotes are not needed

supleed2: I don't mind removing them, but I had added them because I think the highlighting looks much better with:

image

Vs without quotes:

image

jonaro00:
Tried it on a machine that has tar but not rust:

- Script assumes Home\.cargo\bin exists, and fails if it doesn't (which is not a blocker; we can add checking for sufficient rust install to both scripts later)

(added the bin folder manually, then)

- The exit call after successful install made my powershell window close itself, which is probably not desired. (is it due to not running in a subshell anymore?)

(opened and ran it again)

- The raw download route fails if the destination file already exists. It should overwrite an existing destiantion.

- Other caveat: Only if on non-AMD64 and with no tar.exe does it offer to install rust :thinking: . As said above, we will probably revisit this for both scripts so that the rust check+install can be the first step for both.

install.ps1
```diff
@@ -0,0 +1,95 @@
...
+	else {
+		Write-Host "Could not install cargo-shuttle using cargo" -ForegroundColor Red
+		[Environment]::Exit(1)
+	}
+}
+else {
+	if ($Arch -in "AMD64", "x86") {
+		Write-Host "Could not find cargo.exe, Rust may not be installed" -ForegroundColor Red
+		$Confirm = Read-Host -Prompt "Would you like to install Rust via Rustup? [y/N]"
+		if ($Confirm -notin "y", "yes") {
```
jonaro00, line 71: can have `"Y"` as well

supleed2:
1) Maybe the binary download section can go before `cargo.exe install cargo-shuttle --locked` inside the `cargo.exe` check, since this would indicate `cargo_home` should exist?

Should the script create the `bin` folder in `cargo_home` if it doesn't exist?

2) Yes, running in the same shell means it exits, 2 options, prompt user to press enter before exiting, or nest the code in if-else branches and remove all exit calls?

3) I can just add `-Force` to the `Move-Item` call?

jonaro00:
1. Not sure what you mean by that. The binary download is higher prio than cargo install. I don't think we should create "our own" cargo bin folder, but instead let the check (in later PR) for that cargo is installed guard against this error.
2. Hmm, ok. I guess that is due to `iex` not spawning a sub-process for evaluating it. Maybe it is possible to wrap it in a function and use `Return` instead? This isn't a major blocker, so it can be left to fix in an upcoming PR.
3. Sounds good.

After resolving (3), this is probably ready to be merged.

supleed2:
Added the `-Force` flag and wrapped it in a function, it also restores the `$ErrorActionPreference` since it no longer exits.

jonaro00 (approved):

oddgrd (approved):
Awesome, thanks so much @supleed2!

PR #1504, opened 2024-01-02
feat: delete rds subscription item on rds resource deletion

oddgrd:
## Description of change

- [x] Refactor the add_subscription_item flow to add a unique id to the metadata of the subscription item, so we can know which item to delete. 
- [x] Add a new endpoint to auth to delete an item from a users subscription.
- [x] Call the new endpoint when deleting an RDS database.
- [x] Code cleanups and refactoring.

I have added a new mocked integration test to the auth service, a new mocked test to the provisioner, and I have tested the full flow on staging, for basic users, pro users and lapsed pro users with RDS.

### DIFF
--- a/auth/src/api/handlers.rs
+++ b/auth/src/api/handlers.rs
@@ -107,8 +107,6 @@ pub(crate) async fn delete_subscription_items(
         return Err(Error::MissingSubscriptionId);
     };
 
-    // TODO: do we need to check account tier here, like we do when adding items?
-
     span::Span::current().record("account.subscription_id", subscription_id.as_str());
 
     user_manager
--- a/auth/src/api/handlers.rs
+++ b/auth/src/api/handlers.rs
@@ -14,7 +14,7 @@ use shuttle_common::{
     models::user,
 };
 use stripe::CheckoutSession;
-use tracing::{error, instrument};
+use tracing::{error, instrument, span};
 
 use super::{builder::KeyManagerState, RouterState, UserManagerState};
 
@@ -65,7 +65,7 @@ pub(crate) async fn update_user_tier(
     Ok(())
 }
 
-#[instrument(skip(claim, user_manager), fields(account.name = claim.sub, account.tier = %claim.tier))]
+#[instrument(skip(claim, user_manager), fields(account.name = claim.sub, account.tier = %claim.tier, account.subscription_id = tracing::field::Empty))]
 pub(crate) async fn add_subscription_items(
     Extension(claim): Extension<Claim>,
     State(user_manager): State<UserManagerState>,
@@ -79,6 +79,8 @@ pub(crate) async fn add_subscription_items(
         return Err(Error::MissingSubscriptionId);
     };
 
+    span::Span::current().record("account.subscription_id", subscription_id.as_str());
+
     if !matches![user.account_tier, AccountTier::Pro | AccountTier::Admin] {
         error!("account was downgraded from pro in sync, denying the addition of new items");
         return Err(Error::Unauthorized);
@@ -86,6 +88,31 @@ pub(crate) async fn add_subscription_items(
 
     user_manager
         .add_subscription_items(subscription_id, update_subscription_items)
+        .await?;
+
+    Ok(())
+}
+
+#[instrument(skip(claim, user_manager), fields(account.name = claim.sub, account.tier = %claim.tier, account.subscription_id = tracing::field::Empty))]
+pub(crate) async fn delete_subscription_items(
+    Extension(claim): Extension<Claim>,
+    State(user_manager): State<UserManagerState>,
+    Path(metadata_id): Path<String>,
+) -> Result<(), Error> {
+    // Fetching the user will also sync their subscription. This means we can verify that the
+    // caller still has the required tier after the sync.
+    let user = user_manager.get_user(AccountName::from(claim.sub)).await?;
+
+    let Some(ref subscription_id) = user.subscription_id else {
+        return Err(Error::MissingSubscriptionId);
+    };
+
+    // TODO: do we need to check account tier here, like we do when adding items?
+
+    span::Span::current().record("account.subscription_id", subscription_id.as_str());
+
+    user_manager
+        .delete_subscription_item(subscription_id, &metadata_id)
         .await?;
 
     Ok(())
--- a/auth/src/api/handlers.rs
+++ b/auth/src/api/handlers.rs
@@ -10,6 +10,7 @@ use axum::{
 use axum_sessions::extractors::{ReadableSession, WritableSession};
 use http::StatusCode;
 use shuttle_common::{
+    backends::subscription::SubscriptionResponse,
     claims::{AccountTier, Claim},
     models::user,
 };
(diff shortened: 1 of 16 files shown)


auth/src/api/handlers.rs
```diff
@@ -91,6 +94,45 @@ pub(crate) async fn add_subscription_items(
...
+    match user_manager
+        .delete_subscription_item(subscription_id, &metadata_id)
+        .await
+    {
+        Ok(_) => Ok(Json(SubscriptionResponse {
+            success: true,
+            message: "successfully deleted subscription item".to_string(),
+        })),
+        // If the subscription item is missing, we still return 200, since the rds instance could
+        // have been provisioned before we implemented billing. If the subscription is canceled,
```
Kazy, line 123: Can you explain that part about RDS a bit more ? I don't understand how this can happen.

auth/src/error.rs
```diff
@@ -27,8 +27,14 @@ pub enum Error {
...
     #[error("Interacting with stripe resulted in error: {0}.")]
     Stripe(#[from] StripeError),
-    #[error("Missing subscription ID from the checkout session.")]
+    #[error("Missing subscription ID.")]
     MissingSubscriptionId,
+    #[error("found more than one subscription items with the same metadata id: {0}")]
+    DuplicateSubscriptionItems(String),
+    #[error("found no subscription item with the given metadata id: {0}")]
+    MissingSubscriptionItem(String),
+    #[error("stripe subscription is canceled")]
```
Kazy, line 36:
```suggestion
    #[error("stripe subscription is already canceled")]
```

Kazy, line 36: Should we use a capital letter at the start to align with the rest of the message ?

auth/tests/api/users.rs
```diff
@@ -344,6 +328,181 @@ mod needs_docker {
...
+
+        assert_eq!(response.status(), StatusCode::OK);
+
+        let body = hyper::body::to_bytes(response.into_body()).await.unwrap();
+        let response: SubscriptionResponse = serde_json::from_slice(&body).unwrap();
+
+        assert!(!response.success);
+        assert_eq!(
+            response.message,
+            format!("found no subscription item with the given metadata id: non-existent-id")
```
Kazy, line 463: **nit**: Personally in those cases I prefer to construct the expected error and get back the error message from it instead of duplicating. This avoids having to fix the test if we change the wording. What matters is that we've return the right error, not the message itself.

provisioner/src/lib.rs
```diff
@@ -531,17 +579,23 @@ impl Provisioner for MyProvisioner {
                 // Skip updating subscriptions for admin users, and only update subscription if the
                 // rds instance is new.
                 if claim.tier != AccountTier::Admin && created_new_instance {
+                    let instance_name = format!("{}-{}", &request.project_name, engine);
+
                     // If the subscription update fails, e.g. due to a JWT expiring or the subject's
```
Kazy, line 584: Not part of your code, but this seems extremely risky from how it's defined. If the JWT expires (I don't have the full spec in mind, but I feel like this could happen for whatever reason), we're deleting a database with user's data. If we can't connect to the auth service, we're going to delete the database as well; if the subscription has expired (they forgot to pay, oops), we're deleting all of your data.

provisioner/src/lib.rs
```diff
@@ -573,8 +633,23 @@ impl Provisioner for MyProvisioner {
...
+                self.delete_subscription_item(
+                    claim
+                        .token()
+                        .expect("the token should be set on the claim in the jwt auth layer"),
+                    &instance_name,
+                )
+                .await
+                .map_err(|err| Status::internal(err.to_string()))?;
+
+                self.delete_aws_rds(&instance_name).await?
```
Kazy, line 652: It seems that if we fail to delete the subscription (which could be something like "we sent a request, it was executed, but we got an error along the way because of the network"), we're going to have deleted a subscription but the RDS instance will be left.

I'm not sure how problematic this is (it's late), but this might need more scrutiny.

Kazy (commented):
Good job :) I'm not sure I understand all of it (w.r.t. how we're handling subscriptions), but this looks good. 

There is one major warning though in the provisioner, where it seems like we can delete a database way too easily.

One addition that you could make is to add a test to make sure that we can't delete the subscription item of an other account. But not a blocker at all.

auth/src/error.rs
jonaro00: `cancelled`

oddgrd: It depends on if you go for American or British English, and Stripe uses the American spelling (one L). But it does feel like a typo to me too, so I'll change it to cancelled.

oddgrd: I pondered changing all of these to all lowercase, since I think in general error messages should be, but it felt like I was adding quite a few unnecessary changes to an already large PR. I'll keep them consistent for now and we can clean it up later if anyone feels strongly enough. :smile:

auth/src/api/handlers.rs
oddgrd: So in the past anyone could provision RDS, and currently only Pro tier users can provision it. That means we have quite a few users with an RDS instance, but without a subscription. We don't want to block them from deleting their RDS instance because of that.

There are also Pro tier users who have a subscription, but when they provisioned RDS it was still unpaid, no subscription item was added. We will be adding an item to their subscriptions manually following this release and an announcement with a warning, but there will be some time where we will have users with subscriptions and RDS but no RDS subscription item. I can expand the comment with a note that we should not return 200 for missing sub item after that point, since it would then be a bug.

auth/tests/api/users.rs
oddgrd: Yup, that is much better, thanks!

provisioner/src/lib.rs
oddgrd: The RDS provisioning is only called for new instances at present, if it has already been provisioned it is returned from resource-recorder. In the past it was called each time, however, and the password was cycled. Even so, to guard against deleting databases with data if this code changes or we start calling it after creation, there is a guard as well. When the rds instance is provisioned, it also returns a bool, the bool indicates whether a new instance was created. The instance is only deleted if it's created in the same call, to avoid creating instances that do not have an associated subscription item.

While I don't think it's high risk at the moment, it is high consequence, so I can add more comments to guard against code changes leading to accidental deletion in the future. Or more guards if you think it's necessary.

provisioner/src/lib.rs
oddgrd: If there is an error deleting the subscription, e.g. network error or failure to connect to Stripe, we will not try to delete the RDS instance. There are some cases where we will return 200, but a subscription item was not deleted, I explained that a bit more in another comment.

provisioner/src/lib.rs
Kazy: Ah right, I've missed the boolean part. Reviews late in the day aren't the best ideas. Thanks for the detailed explanation, all clear now :pray:

provisioner/src/lib.rs
Kazy: My point was the other way around this time. We've successfully deleted the subscription, but a network error made the `delete_subscription_item` call fail here. This is an edge case though, we can disregard that.

auth/src/api/handlers.rs
Kazy: Great explanation, all clear now, thanks !

provisioner/src/lib.rs
oddgrd: Ah, you mean the subscription deletion succeeds, but the database deletion fails? Yes, we do not cover that case. We could try to re-add the subscription item in that case, which should be fine since it's prorated to the second, it shouldn't incur any extra cost.

provisioner/src/lib.rs
oddgrd: No problem, I appreciate the sanity check!

provisioner/src/lib.rs
Kazy: That's one way to do it, but I think it's still prone to the network or the app shutting down mid function. Ideally we should make the operation idempotent, log (i.e. save it in a database) that we want to do the operation first, update it when done. That way in the case of an event, we can more easily recover since we know something was started. But again, this is just later, we're not there yet :p

oddgrd: Ah, I see what you mean. Yes, that should be implemented eventually, I'll make a note of it! 

I implemented the attempt to recreate the subscription item if it was deleted and the db deletion failed in the latest commit.

oddgrd: As discussed on Discord, I dropped the change to try to add back the item.

auth/src/error.rs
jonaro00: classic dilemma xD, TIL

chesedo (approved):
Remember to uncomment the CI filters

auth/src/subscription.rs
```diff
@@ -24,21 +24,28 @@ where
     async fn from_request(req: Request<B>, state: &S) -> Result<Self, Self::Rejection> {
         // Extract the NewSubscriptionItem, the struct that other services should use when calling
         // the endpoint to add subscription items.
-        let NewSubscriptionItem { quantity, item } = axum::Json::from_request(req, state)
+        let NewSubscriptionItem {
+            metadata_id,
```
iulianbarbu, line 28: as we've discussed, I believe we can still have this feature without using metadata, given we should be able to update the quantity and not care about the underlying RDS instances (given they are the same).

PR #1507, opened 2024-01-03
chore(shuttle-axum): use axum 0.7 by default

jonaro00:
## Description of change

Merge with https://github.com/shuttle-hq/shuttle-examples/pull/123

Tested both configurations locally


oddgrd (commented):
It looks like the c-s CI failure is due to being out of sync with new examples, but the shuttle-service failure I think needs an update to the test dependencies.

oddgrd (approved):

PR #1508, opened 2024-01-04
feat: track project deployments

GugaGongadze:
## Description of change
Moving tracking functionality in gateway, as opposed to doing it in deployer: https://github.com/shuttle-hq/shuttle/pull/1485

`shuttle_api_start_deployment` was ingested by PostHog client.


gateway/src/api/latest.rs
```diff
@@ -956,12 +979,14 @@ pub(crate) struct RouterState {
...
     pub sender: Sender<BoxedTask>,
     pub running_builds: Arc<Mutex<TtlCache<Uuid, ()>>>,
+    pub posthog_client: async_posthog::Client,
 }
 
 pub struct ApiBuilder {
     router: Router<RouterState>,
     service: Option<Arc<GatewayService>>,
     sender: Option<Sender<BoxedTask>>,
+    posthog_client: Option<async_posthog::Client>,
```
jonaro00, line 989: nit: Since the posthog client is essentially a hardcoded global value, it could be a static variable with OnceLock. That way we don't need to pass it through all of the functions and builder. Nothing wrong with having it this way tho.

jonaro00 (approved):
lgtm

gateway/src/api/latest.rs
```diff
@@ -436,11 +436,34 @@ async fn delete_project(
...
+
+    // Check if it matches route: "/projects/:project_name/services/:project_name"
+    if method == Method::POST
+        && uri_path == format!("/projects/{}/services/{}", project_name, project_name)
+    {
+        let account_name = scoped_user.user.claim.sub.clone().to_string();
+
+        tokio::spawn(async move {
+            let event =
+                async_posthog::Event::new("shuttle_api_start_deployment".to_string(), account_name);
```
oddgrd, line 459: Small nit:
```suggestion
        let account_name = scoped_user.user.claim.sub.clone();

        tokio::spawn(async move {
            let event =
                async_posthog::Event::new("shuttle_api_start_deployment", &account_name);
```

oddgrd (approved):
Thanks, LGTM! I just left a small nit.

gateway/src/api/latest.rs
GugaGongadze: Interesting, would it be globally available?

jonaro00: It is similar to a global variable, yeah. We can merge this as is tho.

gateway/src/api/latest.rs
jonaro00: oof, it violated rustfmt (:

PR #1516, opened 2024-01-10
feat: have otel collect our docker stats

chesedo:
## Description of change
Cause otel to push docker metrics to a S3 bucket

Will on staging

### DIFF
--- a/extras/otel/otel-collector-config.yaml
+++ b/extras/otel/otel-collector-config.yaml
@@ -32,12 +32,18 @@ receivers:
           static_configs:
             - targets: ["0.0.0.0:8888"]
   docker_stats:
-    endpoint: unix:///var/run/docker.sock
-    timeout: 20s
-    api_version: 1.41
+    collection_interval: 1m
+    container_labels_to_metric_labels:
+      shuttle.idle_minutes: shuttle.idle_minutes
+      shuttle.project: shuttle.project_name
+      shuttle.project_id: shuttle.project_id
+      shuttle.prefix: shuttle.prefix
 
 processors:
   batch:
+  resourcedetection:
+    detectors: [gcp, ecs, ec2, azure, system]
+  batch/datadog:
     # Make small enough to be processed by datadog
     # https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/datadogexporter#why-am-i-getting-errors-413---request-entity-too-large-how-do-i-fix-it
     send_batch_max_size: 100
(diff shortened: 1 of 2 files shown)


extras/otel/otel-collector-config.yaml
```diff
@@ -68,17 +74,28 @@ exporters:
     endpoint: "api.honeycomb.io:443"
     headers:
       "x-honeycomb-team": ${env:HONEYCOMB_API_KEY}
+  awss3/metrics:
```
Kazy, line 77: How is the access to this bucket handled ?

chesedo: On AWS it will just use the instance metadata endpoint. The [otel-s3-doc] says it uses the default from the [go-sdk] which fallbacks to the "EC2 Instance Role Credentials" [1]

[otel-s3-doc]: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/awss3exporter#aws-credential-configuration
[go-sdk]: https://docs.aws.amazon.com/sdk-for-go/api/#hdr-Configuring_Credentials
[1]: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html

extras/otel/otel-collector-config.yaml
```diff
@@ -32,12 +32,18 @@ receivers:
           static_configs:
             - targets: ["0.0.0.0:8888"]
   docker_stats:
-    endpoint: unix:///var/run/docker.sock
-    timeout: 20s
-    api_version: 1.41
+    collection_interval: 1m
+    container_labels_to_metric_labels:
+      shuttle.idle_minutes: shuttle.idle_minutes
+      shuttle.project: shuttle.project_name
```
oddgrd, line 38: Should this be `shuttle.project.name`? And perhaps the same convention for the project id and idle minutes, but we haven't implemented that elsewhere in our code.

extras/otel/otel-collector-config.yaml
Kazy: If it fallbacks to the EC2 role, then does it mean that the user application containers will inherit these permissions as well without having to do anything ?

oddgrd:
By the way, do we know that this will not cause issues similar to what we saw in the past when we enabled Datadog docker stats collection? IIRC it was previously collecting every 10s, and this is once every minute. Are there other differences here?

extras/otel/otel-collector-config.yaml
chesedo: Oh dang, yes!! I recommend that we then rather find a way to install the collector directly on the VM :smile:

extras/otel/otel-collector-config.yaml
chesedo: Agreed, I'll make this change

chesedo:
> By the way, do we know that this will not cause issues similar to what we saw in the past when we enabled Datadog docker stats collection? IIRC it was previously collecting every 10s, and this is once every minute. Are there other differences here?

We don't know :sweat_smile: There are also no other differences.

However, I just realized that the old collector of 10s might be active. Hopefully it is not active and the cause of our Docker issues :sweat_smile:

chesedo:
Btw, this is a bit on ice while I'm working on the RDS subscriptions

PR #1520, opened 2024-01-16
feat: support poem v2.0.0

### LINKED ISSUE #1519: [Feature]: Support poem v2.0.0
### Describe the feature

Poem has recently released version 2.0.0.
The poem [changelog](https://github.com/poem-web/poem/blob/master/poem/CHANGELOG.md) does not document a lot of changes but the diff between its previous version and the 2.0.0 release shows some internal changes that seemingly break the shuttle-poem service: [1.3.59…2.0.0](https://github.com/poem-web/poem/compare/7c9cff63afd54da62d637d49a91d7eb806b4a446..18c267d04075ab1d1e19ad5f548f5fd2dfa820b3)

I am now on poem version 1.3.59 which works great with shuttle!
But now I am looking to upgrade to move to newer versions of poem, as other parts of poem also rely on this newer version, such as [poem-openapi](https://github.com/poem-web/poem/tree/master/poem-openapi)

To be exact, I list below the output of `cargo check` of the default shuttle-poem example, but using poem 2.0.0

```
error[E0277]: the trait bound `Route: poem::endpoint::endpoint::Endpoint` is not satisfied
  --> src/main.rs:15:12
   |
15 |     Ok(app.into())
   |            ^^^^ the trait `poem::endpoint::endpoint::Endpoint` is not implemented for `Route`
   |
   = help: the following other types implement trait `poem::endpoint::endpoint::Endpoint`:
             poem::endpoint::after::After<E, F>
             poem::endpoint::and_then::AndThen<E, F>
             poem::endpoint::around::Around<E, F>
             poem::endpoint::before::Before<E, F>
             poem::endpoint::catch_all_error::CatchAllError<E, F, R>
             poem::endpoint::catch_error::CatchError<E, F, R, ErrType>
             poem::endpoint::endpoint::EitherEndpoint<A, B>
             poem::endpoint::inspect_all_err::InspectAllError<E, F>
           and 21 others
   = note: required for `PoemService<Route>` to implement `From<Route>`
   = note: required for `Route` to implement `Into<PoemService<Route>>`

error[E0277]: the trait bound `impl Endpoint: poem::endpoint::endpoint::Endpoint` is not satisfied
  --> src/main.rs:11:1
   |
11 | #[shuttle_runtime::main]
   | ^^^^^^^^^^^^^^^^^^^^^^^^ the trait `poem::endpoint::endpoint::Endpoint` is not implemented for `impl Endpoint`
   |
   = help: the following other types implement trait `poem::endpoint::endpoint::Endpoint`:
             poem::endpoint::after::After<E, F>
             poem::endpoint::and_then::AndThen<E, F>
             poem::endpoint::around::Around<E, F>
             poem::endpoint::before::Before<E, F>
             poem::endpoint::catch_all_error::CatchAllError<E, F, R>
             poem::endpoint::catch_error::CatchError<E, F, R, ErrType>
             poem::endpoint::endpoint::EitherEndpoint<A, B>
             poem::endpoint::inspect_all_err::InspectAllError<E, F>
           and 21 others
   = note: required for `PoemService<impl Endpoint>` to implement `Service`
   = note: required for `fn(ProvisionerFactory, ResourceTracker) -> impl Future<Output = Result<PoemService<impl Endpoint>, shuttle_runtime::Error>> {loader}` to implement `shuttle_runtime::alpha::Loader<ProvisionerFactory>`
note: required by a bound in `start`
  --> /home/mouwrice/.cargo/registry/src/index.crates.io-6f17d22bba15001f/shuttle-runtime-0.36.0/src/alpha/mod.rs:49:33
   |
49 | pub async fn start(loader: impl Loader<ProvisionerFactory> + Send + 'static) {
   |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `start`
   = note: this error originates in the attribute macro `shuttle_runtime::main` (in Nightly builds, run with -Z macro-backtrace for more info)
```

P.S. Working with shuttle has been a breeze, I truly appreciate what Shuttle has to offer :pray: 

_No response_

### Duplicate declaration

paulotten:
## Description of change

Closes #1519.

Adds support for poem v2.0.0.

Approach is based on how support for Axum 0.7 was added to shuttle-axum in 0.34.1

Using /examples/poem/hello-world/, I did cargo check with the following changes to dependencies

Poem 1 (regression test)
```
poem = "1.3.55"
shuttle-poem = { path = "../../../services/shuttle-poem/" }
```

Poem 2
```
poem = "2.0.0"
shuttle-poem = { path = "../../../services/shuttle-poem/", default-features = false, features = ["poem-2"] }
```

### DIFF
--- a/services/shuttle-poem/Cargo.toml
+++ b/services/shuttle-poem/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "shuttle-poem"
-version = "0.36.1"
+version = "0.37.0"
 edition = "2021"
 license = "Apache-2.0"
 description = "Service implementation to run a poem webserver on shuttle"
@@ -9,14 +9,8 @@ keywords = ["shuttle-service", "poem"]
 [workspace]
 
 [dependencies]
-poem-1 = { package = "poem", version = "1.3.55", optional = true }
-poem-2 = { package = "poem", version = "2.0.0", optional = true }
+poem = { version = "2.0.0" }
 shuttle-runtime = { path = "../../runtime", version = "0.36.0", default-features = false }
 
 [dev-dependencies]
 tokio = { version = "1.26.0", features = ["macros", "rt-multi-thread"] }
-
-[features]
-default = ["poem-1"]
-
-poem-2 = ["dep:poem-2"]
--- a/services/shuttle-poem/Cargo.toml
+++ b/services/shuttle-poem/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "shuttle-poem"
-version = "0.36.0"
+version = "0.36.1"
 edition = "2021"
 license = "Apache-2.0"
 description = "Service implementation to run a poem webserver on shuttle"
@@ -9,8 +9,14 @@ keywords = ["shuttle-service", "poem"]
 [workspace]
 
 [dependencies]
-poem = { version = "1.3.55" }
+poem-1 = { package = "poem", version = "1.3.55", optional = true }
+poem-2 = { package = "poem", version = "2.0.0", optional = true }
 shuttle-runtime = { path = "../../runtime", version = "0.36.0", default-features = false }
 
 [dev-dependencies]
 tokio = { version = "1.26.0", features = ["macros", "rt-multi-thread"] }
+
+[features]
+default = ["poem-1"]
+
+poem-2 = ["dep:poem-2"]
--- a/services/shuttle-poem/README.md
+++ b/services/shuttle-poem/README.md
@@ -1,5 +1,14 @@
 ## Shuttle service integration for the Poem web framework.
 
+Poem 1.* is used by default.
+
+Poem 2.* is supported by using these feature flags:
+
+```toml,ignore
+poem = "2.0.0"
+shuttle-poem = { version = "0.36.1", default-features = false, features = ["poem-2"] }
+```
+
 ### Example
 
 ```rust,no_run
--- a/services/shuttle-poem/README.md
+++ b/services/shuttle-poem/README.md
@@ -1,14 +1,5 @@
 ## Shuttle service integration for the Poem web framework.
 
-Poem 1.* is used by default.
-
-Poem 2.* is supported by using these feature flags:
-
-```toml,ignore
-poem = "2.0.0"
-shuttle-poem = { version = "0.36.1", default-features = false, features = ["poem-2"] }
-```
-
 ### Example
 
 ```rust,no_run
--- a/services/shuttle-poem/src/lib.rs
+++ b/services/shuttle-poem/src/lib.rs
@@ -2,11 +2,6 @@
 /// A wrapper type for [poem::Endpoint] so we can implement [shuttle_runtime::Service] for it.
 pub struct PoemService<T>(pub T);
 
-#[cfg(feature = "poem-1")]
-use poem_1 as poem;
-#[cfg(feature = "poem-2")]
-use poem_2 as poem;
-
 #[shuttle_runtime::async_trait]
 impl<T> shuttle_runtime::Service for PoemService<T>
 where
--- a/services/shuttle-poem/src/lib.rs
+++ b/services/shuttle-poem/src/lib.rs
@@ -2,6 +2,11 @@
 /// A wrapper type for [poem::Endpoint] so we can implement [shuttle_runtime::Service] for it.
 pub struct PoemService<T>(pub T);
 
+#[cfg(feature = "poem-1")]
+use poem_1 as poem;
+#[cfg(feature = "poem-2")]
+use poem_2 as poem;
+
 #[shuttle_runtime::async_trait]
 impl<T> shuttle_runtime::Service for PoemService<T>
 where


services/shuttle-poem/Cargo.toml
```diff
@@ -9,8 +9,14 @@ keywords = ["shuttle-service", "poem"]
...
-poem = { version = "1.3.55" }
+poem-1 = { package = "poem", version = "1.3.55", optional = true }
+poem-2 = { package = "poem", version = "2.0.0", optional = true }
 shuttle-runtime = { path = "../../runtime", version = "0.36.0", default-features = false }
 
 [dev-dependencies]
 tokio = { version = "1.26.0", features = ["macros", "rt-multi-thread"] }
+
+[features]
+default = ["poem-1"]
```
paulotten, line 20: This default can be changed to poem 2 when shuttle-poem 0.37 is released. It's a breaking change so it shouldn't happen before then.

The README should be updated at the same time.

oddgrd:
Cool, thanks a lot, Paul! I am wondering, though, if we need to take this step for Poem. From the changelog it seems like there are very few changes except for the upgrade of Hyper, so as far as I can tell the user won't need to make updates to their application, we just need to update the part that starts the server on our end in `shuttle-poem`. Oh, and they'll need to update their version of poem if they update their version of shuttle-poem. Or am I missing something here?

paulotten:
> Oh, and they'll need to update their version of poem if they update their version of shuttle-poem. Or am I missing something here?

The version of poem (`poem = "1.3.55"` vs `poem = "2.0.0"`) used in the example/user application needs to match the version of poem in `shuttle-poem`.

Otherwise you get the

```
#[shuttle_runtime::main]
^^^^^^^^^^^^^^^^^^^^^^^^ the trait `poem::endpoint::endpoint::Endpoint` is not implemented for `impl Endpoint`
```

error reported in #1519.

This appears to be the entirety of the issue.

(It's also as I mentioned a breaking change since the user must update their version of poem at the same time or their app will break.)

paulotten:
Re: failed CI check

```
cargo clippy --tests \
             --all-targets \
             --all-features \
             --manifest-path services/shuttle-poem/Cargo.toml \
             --no-deps -- \
             --D warnings
```

```
error[E0252]: the name `poem` is defined multiple times
 --> src/lib.rs:8:5
  |
6 | use poem_1 as poem;
  |     -------------- previous import of the module `poem` here
7 | #[cfg(feature = "poem-2")]
8 | use poem_2 as poem;
  |     ^^^^^^^^^^^^^^ `poem` reimported here
[...]
```

The issue is the `--all-features` flag. Features `poem-1` and `poem-2` are mutually exclusive.

(It looks like the current version of shuttle-axum would also fail such a test. https://docs.rs/shuttle-axum/latest/src/shuttle_axum/lib.rs.html#5-8)

oddgrd:
Yes, but if they are already bumping the version of `shuttle-poem`, if bumping `poem` as well does not require any changes to their application, do we need feature flags to avoid forcing people to upgrade? Axum 0.7 also came with [other breaking changes](https://github.com/tokio-rs/axum/releases/tag/axum-v0.7.0), so we didn't want to force people to upgrade if they weren't prepared to upgrade their application. I'm not sure we need to take the same approach here. :thinking:

paulotten:
> Yes, but if they are already bumping the version of `shuttle-poem`, if bumping `poem` as well does not require any changes to their application, do we need feature flags to avoid forcing people to upgrade? Axum 0.7 also came with [other breaking changes](https://github.com/tokio-rs/axum/releases/tag/axum-v0.7.0), so we didn't want to force people to upgrade if they weren't prepared to upgrade their application. I'm not sure we need to take the same approach here. 🤔

If you increase the version of `shuttle-poem` to `0.37.0` then that should be ok.

If you only increase the version of `shuttle-poem` to `0.36.1` then you are going to have problems. Someone will checkout an example that uses `shuttle-poem = 0.36.0` and `poem = "1.3.55"`. Cargo will select shuttle-poem 0.36.1 (with the change) as compatible with 0.36.0. The user will not be able to compile the example due to the above error.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

paulotten:
There is also the question of if Shuttle wants to provide backwards support for poem 1 in shuttle-poem 0.37 and later. (Which is not a question I can answer.)

jonaro00 (commented):
I agree that the changes in Poem 2 are too minor to justify having feature flag support for both versions at the same time.
A release for 0.36.1 is unlikely to happen, so I think we can simply bump poem to v2 and aim for 0.37.0 with no backwards compatibility.

jonaro00 (approved):
Thanks! A PR that does the same thing for the poem examples (onto the develop branch) would be greatly appreciated! (don't worry about the shuttle crate version numbers or failing CI due to them)

Mouwrice:
Hi hi, original feature requester here :wave: 
Thanks for the speedy resolution @paulotten !

Just wanted to say that I have tested these changes and everything works as it should. As mentioned in this thread, poem 2.0.0 actually has little changes and in my case no changes needed to be made to my (small) codebase.

oddgrd (approved):
Great, thanks Paul! And thanks @Mouwrice for confirming the upgrade is easy on the user side!

PR #1521, opened 2024-01-16
feat(shuttle-serenity): make serenity 0.12 default, support poise 0.6, deprecate shuttle-poise

supleed2:
## Description of change

- Adds `poise` and `poise-0-6` feature flags, with `poise` enabled by default
- Note similar to `shuttle-serenity` added to readme

- `cargo clippy`
- `cargo clippy --no-default-features -F poise-0-6`

### DIFF
--- a/services/shuttle-poise/src/lib.rs
+++ b/services/shuttle-poise/src/lib.rs
@@ -1,20 +1,20 @@
 #![doc = include_str!("../README.md")]
-#[cfg(feature = "poise-0-6")]
-use poise_0_6::serenity_prelude as serenity;
-use std::net::SocketAddr;
 #[cfg(feature = "poise")]
+use poise::serenity_prelude as serenity;
+use std::net::SocketAddr;
+#[cfg(feature = "poise-0-5")]
 use std::sync::Arc;
 
 /// A wrapper type for [poise::Framework] so we can implement [shuttle_runtime::Service] for it.
 #[cfg(feature = "poise")]
-pub struct PoiseService<T, E>(pub Arc<poise::Framework<T, E>>);
-/// A wrapper type for [poise::Framework] so we can implement [shuttle_runtime::Service] for it.
-#[cfg(feature = "poise-0-6")]
 pub struct PoiseService<T, E> {
-    pub framework: poise_0_6::Framework<T, E>,
+    pub framework: poise::Framework<T, E>,
     pub token: String,
     pub intents: serenity::GatewayIntents,
 }
+/// A wrapper type for [poise::Framework] so we can implement [shuttle_runtime::Service] for it.
+#[cfg(feature = "poise-0-5")]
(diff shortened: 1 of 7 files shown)


jonaro00:
Thanks for the PR! :D
I think it'll soon be time for switching the serenity plugin to use 0.12 by default (a PR that you can do if you want to), so I think having this released at the same time with 0.6 as the default feature will be good.

supleed2:
Sounds great, all that's needed for that is changing the defaults? So:

`default = ["poise"]` -> `default = ["poise-0-6"]`
`default = ["rustls_backend"]` -> `default = ["serenity-0-12-rustls_backend"]`

jonaro00:
Rather with a new feature for the older version, and no feature for the current one. See the shuttle-axum crate.

jonaro00:
Since the changes are so closely related, I don't mind having both changes in one PR. Either way works.

jonaro00 (approved):

services/shuttle-poise/src/lib.rs
```diff
@@ -12,6 +23,16 @@ where
...
 {
     async fn bind(mut self, _addr: SocketAddr) -> Result<(), shuttle_runtime::Error> {
+        #[cfg(feature = "poise")]
+        serenity::ClientBuilder::new(self.token, self.intents)
+            .framework(self.framework)
+            .await
+            .map_err(shuttle_runtime::CustomError::new)?
+            .start_autosharded()
+            .await
+            .map_err(shuttle_runtime::CustomError::new)?;
```
jonaro00, line 33: When poise 0.5 has aged, we can consider having the poise integration just be a part of shuttle-serenity. If I'm not mistaken, you would just make a serenity client, add the poise framework, and then use it just like plain shuttle-serenity.

supleed2: I didn't realise this but yeah, it totally should! Would be nice to significantly clean things up.

jonaro00: Hmm... why wait for poise 0.5 to age? We can simply deprecate shuttle-poise in the state before this PR, but with a deprecation warning and a small example on how to use shuttle-serenity + 0.12 + poise 0.6. Does this sound good @oddgrd ?

oddgrd: Sounds great to me!

jonaro00:
@supleed2 You can restore shuttle-poise to its state before this PR, change the requirement on `shuttle-runtime` to `<=0.40.0`, add a `#[deprecated = "..." ]` on the PoiseService. The readme and the deprecation message can have a link to the latest docs.rs/shuttle-serenity.

Then there is also an opportunity to update the examples to use the newer versions of serenity and poise. :wink:

jonaro00 (approved):

PR #1522, opened 2024-01-16
feat(resources)!: get db connection string from resources, refactor ResourceBuilder

jonaro00:
## Description of change

Supersedes #1515

Main function arg's type declaration determine which type is returned from a resource, allowing DB plugins to return the connection string.

BREAKING: Postgres projects with the sqlx connector need to add a feature flag to keep the same behaviour. Also breaking changes in the public API of shuttle-service.

TODO:
- [x] Approval of pattern
- [x] Update rest of plugins with this ResourceBuilder
- [x] Update examples https://github.com/shuttle-hq/shuttle-examples/pull/127
- [x] Docs

All examples build with the PR above

### DIFF
--- a/resources/shared-db/Cargo.toml
+++ b/resources/shared-db/Cargo.toml
@@ -18,8 +18,6 @@ default = []
 
 # Mongodb
 mongodb = ["dep:mongodb"]
-# Mongodb with a default client
-mongodb-client = ["mongodb"]
 
 # Postgres
 postgres = ["sqlx?/postgres", "sqlx?/runtime-tokio", "sqlx?/tls-native-tls"]
--- a/resources/shared-db/Cargo.toml
+++ b/resources/shared-db/Cargo.toml
@@ -14,5 +14,15 @@ shuttle-service = { path = "../../service", version = "0.36.0" }
 sqlx = { version = "0.7.1", optional = true }
 
 [features]
-postgres = ["sqlx/postgres", "sqlx/runtime-tokio-native-tls"]
-postgres-rustls = ["sqlx/postgres", "sqlx/runtime-tokio-rustls"]
+default = []
+
+# Mongodb
+mongodb = ["dep:mongodb"]
+# Mongodb with a default client
+mongodb-client = ["mongodb"]
+
+# Postgres
+postgres = ["sqlx?/postgres", "sqlx?/runtime-tokio", "sqlx?/tls-native-tls"]
+postgres-rustls = ["sqlx?/postgres", "sqlx?/runtime-tokio", "sqlx?/tls-rustls"]
+# Postgres with an sqlx PgPool
+sqlx = ["dep:sqlx"]
--- a/resources/shared-db/Cargo.toml
+++ b/resources/shared-db/Cargo.toml
@@ -20,7 +20,7 @@ default = []
 mongodb = ["dep:mongodb"]
 
 # Postgres
-postgres = ["sqlx?/postgres", "sqlx?/runtime-tokio", "sqlx?/tls-native-tls"]
-postgres-rustls = ["sqlx?/postgres", "sqlx?/runtime-tokio", "sqlx?/tls-rustls"]
+postgres = ["sqlx?/postgres"]
 # Postgres with an sqlx PgPool
-sqlx = ["dep:sqlx"]
+sqlx = ["dep:sqlx", "sqlx/runtime-tokio", "sqlx/tls-rustls"]
+sqlx-native-tls = ["dep:sqlx", "sqlx/runtime-tokio", "sqlx/tls-native-tls"]
--- a/resources/shared-db/src/mongo.rs
+++ b/resources/shared-db/src/mongo.rs
@@ -4,14 +4,29 @@ use shuttle_service::{
     database, error::CustomError, DbInput, DbOutput, Error, Factory, ResourceBuilder, Type,
 };
 
+// Return different things based on feature choice
+#[cfg(feature = "mongodb-client")]
+type ReturnType = mongodb::Database;
+#[cfg(not(feature = "mongodb-client"))]
+type ReturnType = String;
+
 #[derive(Serialize)]
 pub struct MongoDb {
     config: DbInput,
 }
 
-/// Get a `mongodb::Database` from any factory
+impl MongoDb {
+    /// Use a custom connection string for local runs
+    pub fn local_uri(mut self, local_uri: &str) -> Self {
+        self.config.local_uri = Some(local_uri.to_string());
+
+        self
+    }
+}
+
+/// Get a MongoDB database connection or connection string
 #[async_trait]
-impl ResourceBuilder<mongodb::Database> for MongoDb {
+impl ResourceBuilder<ReturnType> for MongoDb {
     const TYPE: Type = Type::Database(database::Type::Shared(database::SharedEngine::MongoDb));
 
     type Config = DbInput;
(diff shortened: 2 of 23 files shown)


resources/shared-db/Cargo.toml
```diff
@@ -14,5 +14,13 @@ shuttle-service = { path = "../../service", version = "0.36.0" }
 sqlx = { version = "0.7.1", optional = true }
 
 [features]
-postgres = ["sqlx/postgres", "sqlx/runtime-tokio-native-tls"]
-postgres-rustls = ["sqlx/postgres", "sqlx/runtime-tokio-rustls"]
+default = []
+
+# Mongodb
+mongodb = ["dep:mongodb"]
```
chesedo, line 20: This line should not be needed right? The dependency is already optional (which implicitly adds this line)

resources/shared-db/Cargo.toml
```diff
@@ -14,5 +14,13 @@ shuttle-service = { path = "../../service", version = "0.36.0" }
...
+
+# Postgres
+postgres = ["sqlx?/postgres", "sqlx?/runtime-tokio", "sqlx?/tls-native-tls"]
+postgres-rustls = ["sqlx?/postgres", "sqlx?/runtime-tokio", "sqlx?/tls-rustls"]
+# Postgres with an sqlx PgPool
+sqlx = ["dep:sqlx"]
```
chesedo, line 26: The same for this :smile:

runtime/src/resource_tracker.rs
```diff
@@ -63,67 +63,54 @@ macro_rules! log {
...
     builder: B,
     factory: &mut ProvisionerFactory,
     resource_tracker: &mut ResourceTracker,
-) -> Result<T, shuttle_service::Error>
+) -> Result<O, shuttle_service::Error>
 where
-    B: ResourceBuilder<T, Output = O>,
-    O: Serialize + DeserializeOwned,
+    B: ResourceBuilder<Output = O>,
+    O: Serialize + DeserializeOwned + IntoResource<R>,
```
chesedo, line 73: I'm fine with this generic, just want to double check if we can potentially remove it?

service/src/lib.rs
```diff
@@ -39,59 +35,50 @@ pub trait Factory: Send + Sync {
...
-/// ## Creating your own managed resource
-///
-/// You may want to create your own managed resource by implementing this trait for some builder `B` to construct resource `T`.
-/// [`Factory`] can be used to provision resources on Shuttle's servers if your service will need any.
-///
-/// Please refer to `shuttle-examples/custom-resource` for examples of how to create custom resource. For more advanced provisioning
-/// of custom resources, please [get in touch](https://discord.gg/shuttle) and detail your use case. We'll be interested to see what you
-/// want to provision and how to do it on your behalf on the fly.
-///
-/// ```
```
chesedo, line 55: Why are we removing the steps that document how the community can create their own plugins?

service/src/lib.rs
```diff
@@ -39,59 +35,50 @@ pub trait Factory: Send + Sync {
...
+    /// The input config to this resource.
+    type Config: Default + Serialize;
 
-    /// The output type used to build this resource later
+    /// The output from requesting this resource.
+    /// A cached copy of this will be used if the same [`ResourceBuilder::Config`] is found for this [`ResourceBuilder::TYPE`].
     type Output: Serialize + DeserializeOwned;
 
-    /// Create a new instance of this resource builder
-    fn new() -> Self;
```
chesedo, line 68: Why are we dropping `new()`? I originally imagined we would need it in the future to allow resources to be configurable (`local_uri = "..."`)

service/src/lib.rs
```diff
@@ -39,59 +35,50 @@ pub trait Factory: Send + Sync {
...
+    /// The output from this function is passed to [`IntoResource::init`].
     async fn output(self, factory: &mut dyn Factory) -> Result<Self::Output, crate::Error>;
+}
 
-    /// Build this resource from its config output
-    async fn build(build_data: &Self::Output) -> Result<T, crate::Error>;
+/// Implement this on an [`ResourceBuilder::Output`] type to turn the
+/// base resource into the end type exposed to the Shuttle main function.
+#[async_trait]
+pub trait IntoResource<R>: Serialize + DeserializeOwned {
```
chesedo, line 71: Can we drop the `Serialize` and `DeserializeOwned`? I'm thinking some final outputs might not always have these implemented. Also, don't see the need in having them implemented.

service/src/lib.rs
```diff
@@ -39,59 +35,50 @@ pub trait Factory: Send + Sync {
...
+    async fn init(self) -> Result<R, crate::Error>;
```
chesedo, line 72: Maybe this should rather be `into_resource()`.

chesedo (commented):
I left some initial notes, but think this can work :partying_face:

resources/shared-db/Cargo.toml
jonaro00: Yes, but I prefer having it this way since it makes the list of features much easier to read. (cargo feature docs are usually hard to find, and I often find myself looking up the manifest file)

service/src/lib.rs
jonaro00: Yeah I should add it back but not as "your own *managed* resource"

service/src/lib.rs
jonaro00: All of those args are builder pattern, and there is never a way to set args in `new`. I instead added `Default` trait bound, since `new` always just called or returned a default value. Did you have other plans for `new`?

service/src/lib.rs
jonaro00: `R` is the final output with no bounds. `Serialize + DeserializeOwned` are required for what we are converting from, which will always be a `ResourceBuilder::Output`.

runtime/src/resource_tracker.rs
jonaro00: I don't see a way of doing it. Open to suggestions. `R` is the final output type getting into the user's main function, so it feels like it must be generic.

service/src/lib.rs
```diff
@@ -39,59 +35,50 @@ pub trait Factory: Send + Sync {
...
 
-    /// The internal config being constructed by this builder. This will be used to find cached [Self::Output].
-    type Config: Serialize;
+    /// The input config to this resource.
+    type Config: Default + Serialize;
 
-    /// The output type used to build this resource later
+    /// The output from requesting this resource.
+    /// A cached copy of this will be used if the same [`ResourceBuilder::Config`] is found for this [`ResourceBuilder::TYPE`].
     type Output: Serialize + DeserializeOwned;
```
jonaro00, line 49: I'm still thinking `Output` might want `IntoResource<R>` bound as well. But that would add generic `<R>` back to `ResourceBuilder` and defeat the flexibility that we are adding here (one `ResourceBuilder<R>` impl per final output type?).

resources/shared-db/Cargo.toml
chesedo: Hmm, I understand. Don't think it is idiomatic Rusty though. Will personally prefer not to have it.

runtime/src/resource_tracker.rs
chesedo: You mean the `init()` call won't work in the user's main function right?

service/src/lib.rs
chesedo: But isn't `new()` a clearer indication that this is the builder pattern than `default()`? (The `ResourceBuilder` mimics the builder pattern)

resources/shared-db/Cargo.toml
jonaro00: An explicit list of features with comments is IMO always better than looking for the optional tags. 
Examples of explicit feature->dep mapping with docs: 
https://github.com/emilk/egui/blob/7733d1d87c31849d4ac8a3039131c8ad3e71b547/crates/epaint/Cargo.toml#L61
https://github.com/emilk/egui/blob/7733d1d87c31849d4ac8a3039131c8ad3e71b547/crates/egui/Cargo.toml#L27
Not sure what the idiom is tho :)

service/src/lib.rs
chesedo: I see how `default()` can fix my original question though. So am now happy having this either way.

(btw the `Default` trait might be missing on `ResourceBuilder`)

service/src/lib.rs
chesedo: Got it, thanks! :pray:

service/src/lib.rs
chesedo: Hmm, yeah that will break the flexibility.

chesedo: Some `Output` might not need `IntoResource` to other types. Maybe we should have a default `impl<T> IntoResource<T> for T`?

resources/shared-db/src/mongo.rs
```diff
@@ -1,46 +1,48 @@
...
+    database, error::CustomError, resource::Type, DatabaseResource, DbInput, Error, Factory,
+    IntoResource, ResourceBuilder,
 };
 
-#[derive(Serialize)]
-pub struct MongoDb {
-    config: DbInput,
+/// Shuttle managed MongoDB in a shared cluster
+#[derive(Default)]
+pub struct MongoDb(DbInput);
```
oddgrd, line 10: Nit: is there a good reason to switch to a tuple struct here, or could we keep it as a struct with a named field, for increased readability?

oddgrd (commented):
This is a really cool solution, I was actually smiling the whole time I reviewed this :heart_eyes: I had to try it out with the shared db resource and cargo expand to wrap my head around it, it's been a while since I've worked on the macros, and it works really well! It should be easy to extend the resources with more connectors in the future now, e.g. diesel for the shared db, by implementing IntoResource. 

I left a small nit, and I'll do another review when the remaining tasks are done, but this LGTM!

resources/shared-db/src/mongo.rs
jonaro00: Both `MongoDb` and `Postgres` were basically following the newtype pattern already, with just one named field. This felt more coherent, but doesn't matter at the end of the day.

oddgrd (approved):
Looks great to me, thanks! One minor suggestion, should we use `Wrapper` over `Wrap` for the name of the database resource wrapper?

service/src/lib.rs
```diff
@@ -19,79 +19,94 @@ pub mod error;
...
 
-    /// Get the config output of this builder
+    /// Construct this resource with the help of metadata and by calling provisioner methods in the [`Factory`].
     ///
     /// This method is where the actual resource provisioning should take place and is expected to take the longest. It
     /// can at times even take minutes. That is why the output of this method is cached and calling this method can be
-    /// skipped as explained in [Self::config()].
+    /// skipped as explained in [`ResourceBuilder::config`].
+    ///
+    /// The output from this function is passed to [`IntoResource::init`].
```
chesedo, line 79:
```suggestion
    /// The output from this function is passed to [`IntoResource::into_resource`].
```

chesedo (approved):
LGTM

chesedo (approved):

PR #1524, opened 2024-01-17
ci: use default ubuntu machine image

jonaro00:
## Description of change

Current image is being deprecated this year. Replacing it with `default` which uses latest ubuntu, as recommended here https://discuss.circleci.com/t/linux-image-deprecations-and-eol-for-2024/50177

Works for the ci runs


oddgrd (approved):
Thanks for the link!

PR #1525, opened 2024-01-18
feat: base api and gateway client

chesedo:
## Description of change
Make a base API client that can be used to build API clients for other services. We want this to allow `provisioner` to call `gateway` and `resource-recorder` soon.

Will add tests

### DIFF
--- a/common/src/backends/client/gateway.rs
+++ b/common/src/backends/client/gateway.rs
@@ -1,0 +1,55 @@
+use headers::Authorization;
+use http::{Method, Uri};
+
+use crate::models;
+
+use super::{Error, ServicesApiClient};
+
+#[derive(Clone)]
+pub struct GatewayClient {
+    public_client: ServicesApiClient,
+    private_client: ServicesApiClient,
+}
+
+impl GatewayClient {
+    pub fn new(public_uri: Uri, private_uri: Uri) -> Self {
+        Self {
+            public_client: ServicesApiClient::new(public_uri),
+            private_client: ServicesApiClient::new(private_uri),
+        }
+    }
+
+    pub fn public_client(&self) -> &ServicesApiClient {
+        &self.public_client
+    }
+
+    pub fn private_client(&self) -> &ServicesApiClient {
+        &self.private_client
+    }
+
+    /// Get the projects that belong to a user
+    pub async fn get_user_projects(
+        &self,
+        user_token: &str,
+    ) -> Result<Vec<models::project::Response>, Error> {
+        let projects = self
+            .public_client
+            .request(
+                Method::GET,
+                "projects",
+                None::<()>,
+                Some(Authorization::bearer(user_token).expect("to build an authorization bearer")),
+            )
+            .await?;
+
+        Ok(projects)
+    }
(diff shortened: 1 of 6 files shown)


common/Cargo.toml
```diff
@@ -107,6 +107,7 @@ base64 = { workspace = true }
 hyper = { workspace = true }
 proptest = "1.1.0"
 ring = { workspace = true }
+shuttle-common-tests = { workspace = true }
```
chesedo, line 110: I'm not sure if we would rather want to inverse this dependency :thinking:

common-tests/src/gateway.rs
```diff
@@ -0,0 +1,63 @@
...
+    ];
+
+    Mock::given(method(Method::GET))
+        .and(path("/projects"))
+        .respond_with(move |req: &Request| {
+            let Some(bearer) = req.headers.get(&AUTHORIZATION) else {
+                return ResponseTemplate::new(401);
+            };
+
+            let user = bearer.last().as_str().split_whitespace().nth(1).unwrap();
```
chesedo, line 43: I think doing a full convert to our JWT is too much to just test the clients. Hence, using the bearer value as is

common/src/backends/client/gateway.rs
```diff
@@ -0,0 +1,129 @@
...
+            .await?
+            .into_iter()
+            .map(|p| p.id)
+            .collect();
+
+        Ok(ids)
+    }
+}
+
+#[async_trait::async_trait]
```
oddgrd, line 57: Same as above
```suggestion
```

common/src/backends/client/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+
+/// `Hyper` wrapper to make request to RESTful Shuttle services easy
+#[derive(Clone)]
+pub struct ServicesApiClient {
+    client: Client<HttpConnector>,
+    base: Uri,
+}
+
+impl ServicesApiClient {
+    /// Make a new client that to connect to the given endpoint
```
oddgrd, line 35:
```suggestion
    /// Make a new client that connects to the given endpoint
```

oddgrd (approved):
I left some minor comments, but nothing blocking, this looks great to me!

common/Cargo.toml
chesedo: I'm worried we want access to `shuttle-common` in the tests :sweat_smile: I guess we can't just move things around if / when that is a problem

common/src/backends/client/gateway.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /// Get the client of private API calls
+    pub fn private_client(&self) -> &ServicesApiClient {
+        &self.private_client
+    }
+}
+
+/// Interact with all the data relating to projects
+#[async_trait::async_trait]
```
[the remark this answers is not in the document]
chesedo: Nice catch!

oddgrd (approved):

iulianbarbu (approved):
LGTM!

PR #1527, opened 2024-01-19
fix(proxy): record http.host after parsing to avoid Some(..) in the attr

Kazy:
## Description of change
Prevents the `Some(foo.bar)` in Datadog attributes.

Not tested.


oddgrd (approved):
Nice!

iulianbarbu (approved):

PR #1529, opened 2024-01-19
feat(auth): add subscriptions table to auth, add rds quota to claim limits

oddgrd:
## Description of change

Add a subscriptions table to auth to support multiple types of subscriptions, and update the claim limits with an rds quota.

Existing tests pass without modification, but new tests will be needed as we add functionality to add more than a pro subscription.

### DIFF
--- a/auth/migrations/0001_sync_updated_at.sql
+++ b/auth/migrations/0001_sync_updated_at.sql
@@ -1,4 +1,4 @@
--- Create a trigger to automatically set updated_at to current_timestamp
+-- Create a function (that can be registered on triggers) to automatically set updated_at to current_timestamp
 CREATE OR REPLACE FUNCTION sync_updated_at() 
    RETURNS TRIGGER 
    LANGUAGE PLPGSQL
--- a/auth/migrations/0001_sync_updated_at.sql
+++ b/auth/migrations/0001_sync_updated_at.sql
@@ -1,0 +1,10 @@
+-- Create a trigger to automatically set updated_at to current_timestamp
+CREATE OR REPLACE FUNCTION sync_updated_at() 
+   RETURNS TRIGGER 
+   LANGUAGE PLPGSQL
+AS $$
+BEGIN
+   NEW.updated_at = current_timestamp;
+   RETURN NEW;
+END;
+$$
--- a/auth/migrations/0002_subscriptions.sql
+++ b/auth/migrations/0002_subscriptions.sql
@@ -3,6 +3,7 @@ CREATE TABLE IF NOT EXISTS subscriptions (
   account_name TEXT NOT NULL,
   type TEXT NOT NULL,
   quantity INT DEFAULT 1,
+  created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
   updated_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP, 
   UNIQUE (account_name, type),
   FOREIGN KEY (account_name) REFERENCES users(account_name)
--- a/auth/migrations/0002_subscriptions.sql
+++ b/auth/migrations/0002_subscriptions.sql
@@ -1,0 +1,25 @@
+CREATE TABLE IF NOT EXISTS subscriptions (
+  subscription_id TEXT PRIMARY KEY,
+  account_name TEXT NOT NULL,
+  type TEXT NOT NULL,
+  updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP, 
+  UNIQUE (account_name, type),
+  FOREIGN KEY (account_name) REFERENCES users(account_name)
+);
+
+-- Create a trigger to automatically update updated_at
+CREATE TRIGGER sync_users_updated_at
+BEFORE UPDATE
+ON subscriptions
+FOR EACH ROW
+EXECUTE PROCEDURE sync_updated_at();
+
+-- Insert existing subscriptions into the new subscriptions table
+INSERT INTO subscriptions (subscription_id, account_name, type)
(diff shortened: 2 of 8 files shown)


auth/migrations/0001_sync_updated_at.sql
```diff
@@ -0,0 +1,10 @@
+-- Create a trigger to automatically set updated_at to current_timestamp
```
chesedo, line 1: Interesting! Will this affect all tables?

auth/migrations/0002_subscriptions.sql
```diff
@@ -0,0 +1,26 @@
...
+
+-- Insert existing subscriptions into the new subscriptions table, all of which are of the pro type
+INSERT INTO subscriptions (subscription_id, account_name, type)
+SELECT subscription_id, account_name, 'pro'
+FROM users
+WHERE subscription_id IS NOT NULL;
+
+-- Drop the subscription_id column from the users table
+ALTER TABLE users
+DROP COLUMN subscription_id;
```
chesedo, line 26: Hmm, should we drop the `tier` too?

common/src/limits.rs
```diff
@@ -11,22 +11,26 @@ pub struct Limits {
...
+    /// The quantity of RDS instances this user can provision.
+    rds_quota: u32,
 }
 
 impl Default for Limits {
     fn default() -> Self {
         Self {
             project_limit: MAX_PROJECTS_DEFAULT,
             rds_access: false,
+            rds_quota: 0,
```
chesedo, line 23: We should be able to remove `rds_access` right?

chesedo (commented):
I left some comments. Just confirm which changes are valid and I'll get to them :smile:

auth/migrations/0001_sync_updated_at.sql
oddgrd: My comment is a bit inaccurate, this creates a function to set updated_at. We can then create triggers for the various tables that want to use it to automatically update updated_at on any updates. See the subscriptions migration for where I set the trigger for that table. We can do this for any table we want.

auth/migrations/0002_subscriptions.sql
```diff
@@ -0,0 +1,26 @@
+CREATE TABLE IF NOT EXISTS subscriptions (
+  subscription_id TEXT PRIMARY KEY,
+  account_name TEXT NOT NULL,
+  type TEXT NOT NULL,
+  quantity INT DEFAULT 1,
+  updated_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
```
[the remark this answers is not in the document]
oddgrd: Sure! That will be the same, just named created_at.

common/src/limits.rs
oddgrd: I was worried we would break something if we remove a field, but perhaps that's not the case. :thinking: I'm thinking that the deployers/runtimes will deserialize a claim with a new field just fine, since the new field will be ignored, but if an existing field is removed, they will break.

auth/migrations/0002_subscriptions.sql
oddgrd: So the way I've thought about this is keeping the subscriptions separate, that is keeping the pro subscription as is, and creating more general logic for addons. But it would be a lot nicer if we could make the logic general already... The pro subscription has some additional logic that I didn't consider necessary for RDS, but perhaps it isn't necessary for the pro tier either, perhaps we can just set the pro tier limits on the claim based on the subscription like we do for RDS.

common/src/limits.rs
chesedo: I found that only provisioner uses this field via the `Claim` since provisioner and auth will update at the same time (and we expect little traffic on this route), I went ahead and removed it :sweat_smile:

oddgrd: Alright! This is the line I was worried about: https://github.com/shuttle-hq/shuttle/blob/d9ad0179a5d3b3ff0194b81c1558e181addcf592/common/src/claims.rs#L372. The claim layer is used in the runtimes, do you think this will break? It's not entirely clear to me, since it's checking the type id.

chesedo: Ohhhh right :man_facepalming:

auth/migrations/0002_subscriptions.sql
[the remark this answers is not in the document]
chesedo: Done

common/src/limits.rs
oddgrd: If it was using serde it would be breaking, but since it's using a type map, I'm not sure if it is breaking. I'm not sure how the TypeId is calculated. :thinking:

auth/migrations/0002_subscriptions.sql
chesedo: Ok, I can't currently remove this because that will screw with our admin and deployer tier accounts... something for another day.

common/src/limits.rs
```diff
@@ -10,32 +10,44 @@ pub struct Limits {
     /// The amount of projects this user can create.
     project_limit: u32,
     /// Whether this user has permission to provision RDS instances.
+    #[deprecated(
+        since = "0.38.0",
```
oddgrd, line 14: Nice! Should it be 0.37? The current version is 0.36.
```suggestion
        since = "0.37.0",
```

oddgrd (commented):
On minor comment, but LGTM!

common/src/limits.rs
chesedo: I assumed this PR will only make it into the next release (if we release today) :man_shrugging:

chesedo: We can update this if it is released earlier :smile:

PR #1531, opened 2024-01-21
feat(shuttle-turso)!: replace `libsql-client` crate with the new and improved `libsql` crate

### LINKED ISSUE #1406: [Feature]: libsql crate support for turso
### Describe the feature

This allows users to use the more beta `libsql` crate as opposed to the `libsql_client` crate which has some nicer features along with being Send+Sync (`Connection` at least)

Pass in a `libsql::Database` object instead of a `libsql_client::Client` and then it would be available

### Duplicate declaration

Mouwrice:
## Description of change
Replaces the - soon-to-be deprecated[^1] - `libsql-client` with the new `libsql` crate. Resolves #1406.

shuttle-turso has been stuck on libsql-client v0.31.0 for a while now due to https://github.com/tursodatabase/libsql-client-rs/issues/52 and should be resolved with the new crate.

This is obviously a breaking change but a necessary one. Most of the logic of the old crate has remained the same, but the new crate is more refined and provides some new and improved functionality. Replacing the crate is a tedious task for the end-user, but not a difficult one.

On the shuttle side of things, the biggest change is that the database creation is now more explicit. Where the `libsql_client::Client` made use of URL prefixes for its construction (https://docs.rs/libsql-client/latest/libsql_client/client/enum.Client.html#impl-Client-1), the `libsql::Database` (the new equivalent of `libsql_client::Client`) offers specific builders with a lot more functionality (https://docs.rs/libsql/0.2.0/libsql/struct.Database.html). shuttle-turso currently only makes use of https://docs.rs/libsql/0.2.0/libsql/struct.Database.html#method.open and https://docs.rs/libsql/0.2.0/libsql/struct.Database.html#method.open_remote but it might be interesting to support more of the features in the future, such as local replication of a remote database.

TODO's
- [x] await confirmation on possible url (prefix) changes
- [x] update examples
- [ ] update docs page (https://docs.shuttle.rs/resources/shuttle-turso)

Can be tested by executing the updated turso example (https://github.com/shuttle-hq/shuttle-examples/pull/130)

[^1]: https://github.com/tursodatabase/libsql-client-rs/issues/52#issuecomment-1847968384


resources/turso/src/lib.rs
```diff
@@ -120,20 +123,23 @@ impl ResourceBuilder<Client> for Turso {
...
-            url: config.conn_url.clone(),
-            auth_token: config.token.clone(),
-        })
-        .await?;
-        Ok(client)
+    async fn build(config: &Self::Output) -> Result<Database, shuttle_service::Error> {
+        let database = match config.environment {
+            Environment::Deployment => Database::open_remote(
+                config.conn_url.to_string(),
+                config.token.clone().unwrap().to_string(),
```
oddgrd, line 138: Should we rather return an error here if the token is missing?

resources/turso/Cargo.toml
```diff
@@ -9,11 +9,11 @@ keywords = ["shuttle-service", "turso"]
 [dependencies]
 async-trait = "0.1.56"
 dunce = "1.0.4"
-libsql-client = "0.31.0" # Stays on 0.31 until https://github.com/libsql/libsql-client-rs/issues/52
+libsql = { version = "0.2.0", default-features = false, features = ["core", "remote"] }
```
oddgrd, line 12: Is "core" needed for the local run with a local file database, or can we do without it?

oddgrd (commented):
Thanks a lot for this @Mouwrice! This looks great to me, we'll just have to discuss how we want to go about releasing it. As you say it is a necessary breaking change, and the old implementation has grown out of date, so I am leaning towards releasing this soon. We'll get some thoughts from the rest of the team as well. :+1:

resources/turso/Cargo.toml
Mouwrice: Yes, core is for local databases, remote for, well remote :smile: 
https://docs.rs/libsql/latest/libsql/struct.Database.html

Mouwrice: image
I have only omitted the replication feature from the default features.

Mouwrice:
The following URL operations are in essence not really needed anymore:
https://github.com/shuttle-hq/shuttle/blob/19440ec75fe404d6b339fbeba1902555dad40e44/resources/turso/src/lib.rs#L100-L104

As libsql adds https on its own
https://github.com/tursodatabase/libsql/blob/2c0d47db614e7ea395c149446f322e7dd050fa23/libsql/src/util/mod.rs#L9-L19

But I suppose keeping it makes it clear what the URL references to. Should probably wait until the official documentation pages show an example.

jonaro00:
I merged with recent refactors on main.

I think the url prefix logic can be removed if the library is handling it correctly.

Looking at the new library's API, is it worth considering returning a `Connection` instead? On the other hand, having a `Database` allow calling `connect()` multiple times. Not sure what is preferred. I think you might find out which one is more suitable if you also update the turso example to compile with your changes. I think this PR will need to go hand-in-hand with a PR for the updated example.

Mouwrice:
@jonaro00 I found out that libsql does not require absolute URL's but Url::parse does. So I have enforced the usage of the `libsql://` or `https://` prefixes.

Indeed, it might make more sense to use `Connection`. `Database` seems more to be a builder pattern. So I have replaced it. Will provide an updated examples PR accordingly.

oddgrd (approved):
This looks great to me, thanks @Mouwrice!

jonaro00 (approved):

PR #1540, opened 2024-01-24
chore: remove panamax and deck-chores

jonaro00:

Stack runs locally


oddgrd (approved):
Nice, LGTM! We should also remove the registries from terraform/pulumi. We should also remember to delete the volumes once this is deployed, since it takes up a bit of space.

PR #1543, opened 2024-01-25
fix: remove obsolete resources

jonaro00:
## Description of change

Removes static folder, turso, metadata, perist pseudo-resources from deployer state (on restart) and r-r state.

New deployer ignores runtime responses with old resources.

TODO?: r-r ignore hardcoded list of incoming resources, as old deployer+runtimes can still send string of old resources

TODO: drop "custom" as well, and ignore inserting them

---

This change should only cause errors when:
- A newer c-s gets a resource list from an older deployer
- A newer c-s runs an older runtime and gets a resource list in a load response

Both of these are quite rare and easy to fix.


deployer/src/persistence/resource/mod.rs
```diff
@@ -66,139 +62,9 @@ impl FromRow<'_, SqliteRow> for Resource {
...
-            Type::Turso,
-            Type::Metadata,
-            Type::Custom,
-        ];
-
-        for input in inputs {
-            let actual = Type::from_str(&input.to_string()).unwrap();
-            assert_eq!(input, actual, ":{} should map back to itself", input);
-        }
-    }
```
chesedo, line 203: Just confirm a test like this also exists in common

resource-recorder/src/dal.rs
```diff
@@ -255,7 +256,11 @@ impl TryFrom<record_request::Resource> for Resource {
     type Error = String;
 
     fn try_from(value: record_request::Resource) -> Result<Self, Self::Error> {
-        Ok(Self::new(value.r#type.parse()?, value.data, value.config))
+        let r#type = value.r#type.parse()?;
+        if let Type::Custom = r#type {
+            return Err("Custom resources can not be saved in resource-recorder".into());
+        }
```
chesedo, line 262: There is nothing upstream (in deployer) that will trigger this error right?

chesedo (commented):
Just small nits

resource-recorder/src/dal.rs
jonaro00: Deployers will send resources that trigger this, but they are filtered out here https://github.com/shuttle-hq/shuttle/pull/1543/files/ed61ad7423e45410df33936159cef20080204921#diff-7b2cbae17753959effc4bd23a9be282544990c7b84c18dc8f8e587282f0512a6R69-R71

chesedo: Great, thanks! :pray:

deployer/migrations/0006_delete_outdated_resources.sql
```diff
@@ -0,0 +1,5 @@
+-- delete any old records of these so that they are not sent to r-r
+DELETE FROM resources WHERE type = "static_folder";
+DELETE FROM resources WHERE type = "turso";
+DELETE FROM resources WHERE type = "metadata";
+DELETE FROM resources WHERE type = "custom";
```
oddgrd, line 5: Nit: this could be one line, `DELETE FROM resources WHERE type IN ("static_folder", ...)`.

PR #1585, opened 2024-01-25
feat: write a wrapper for the provisioner to call gw and r-r clients

chesedo:
## Description of change
Allows provisioner to check how many RDSs a user is already using before trying to provision a new one.

A whole lot of new tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6686,6 +6686,7 @@ dependencies = [
  "async-trait",
  "chrono",
  "futures-core",
+ "http",
  "prost 0.12.3",
  "prost-types",
  "serde_json",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,60 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "aead"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7fc95d1bdb8e6666b2b217308eeeb09f2d6728d104be3e31916cc74d15420331"
+dependencies = [
+ "generic-array",
+]
+
+[[package]]
+name = "aes"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "884391ef1066acaa41e766ba8f596341b96e93ce34f9a43e7d24bf0a0eaf0561"
+dependencies = [
+ "aes-soft",
+ "aesni",
+ "cipher",
+]
+
+[[package]]
+name = "aes-gcm"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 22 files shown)


oddgrd (commented):
Looks correct to me so far, thanks Pieter!

common-tests/Cargo.toml
```diff
@@ -10,6 +10,8 @@ publish = false
 cargo-shuttle = { path = "../cargo-shuttle" }
 shuttle-proto = { workspace = true, features = ["builder", "logger", "resource-recorder"] }
 
+http = "1"
+http-types = "2.12.0"
```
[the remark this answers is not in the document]
chesedo: It's been removed :partying_face:

provisioner/src/lib.rs
```diff
@@ -470,7 +486,33 @@ impl Provisioner for ShuttleProvisioner {
                     .await?
             }
             DbType::AwsRds(AwsRds { engine }) => {
-                can_provision_rds?;
+                {
+                    let mut rr_client = self.rr_client.lock().await;
+
+                    let can_provision = match claim
+                        .can_provision_rds(&self.gateway_client, &mut *rr_client)
+                        .await
```
oddgrd, line 494: It looks like the tracing context will be propagated in the clients so we can trace the whole flow from "provision_database" to r-r and gateway, but should we instrument this call so we can isolate its duration ? This function will also call the aws sdk, so it would be good to be able to separate the time it spends in that call and this one.

oddgrd (commented):
I left a question about observability, and I'm also curious to know why those tests are failing in CI. Otherwise, this lgtm!

provisioner/src/lib.rs
chesedo: Very good idea. I'll add that :+1:

Cargo.lock
```diff
@@ -6223,14 +6223,19 @@ name = "shuttle-proto"
...
  "anyhow",
+ "async-trait",
  "chrono",
  "futures-core",
+ "http 0.2.11",
  "prost 0.12.3",
  "prost-types",
  "serde_json",
  "shuttle-common",
+ "shuttle-common-tests",
```
jonaro00, line 6234: proto can't have common_tests as a dep

chesedo: Done :pray:

common/src/backends/client/resource_recorder.rs
```diff
@@ -41,6 +42,7 @@ impl<T> ResourceDal for &mut T
 where
     T: ResourceDal,
 {
+    #[instrument(skip_all, fields(shuttle.project_id = project_id))]
```
oddgrd, line 45: We haven't established a convention for this field yet, although we do call it `shuttle.project_id` in the deployer container labels. Should we perhaps go for `shuttle.project.id` instead?

oddgrd (approved):
Thanks, one small question, but LGTM :partying_face:

common/src/backends/client/resource_recorder.rs
chesedo: Ah, yes sorry missed that :+1:

chesedo, line 45:
```suggestion
    #[instrument(skip_all, fields(shuttle.project.id = project_id))]
```

PR #1588, opened 2024-01-26
chore: remove utopia openAPI docs

jonaro00:
## Description of change

These docs were not being actively used and they make the code harder to work with.
To help Console development we should instead add `typeshare` soon. I've tried it on a different project and it should work well.


oddgrd (approved):
Nice! :partying_face:

PR #1589, opened 2024-01-26
feat: limit concurrent active cch23 projects

jonaro00:
## Description of change

If 20 cch23 projects are active, all traffic to all cch23 projects is blocked. They will then proceed to start idling (still 5 minutes).
This is not a perfect solution (everyone get locked out), but this limit should be hard to reach anyways, and it should quickly recover.

Tested that admin can still excede the limit, and a normal user can not access the projects when the limit is reached.

lmk if a unit test must be written


chesedo (approved):
LGTM

PR #1595, opened 2024-01-30
wip: use cargo-make for tasks and ci

jonaro00:
## Description of change

Aims to make the common CI tasks runnable locally, ie `cargo make ci-workspace`.

Also `cargo make proto`, `cargo make changelog 0.x.0`, and `cargo make audit`

CI + local tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,7 +108,7 @@ commands:
   apply-patches:
     steps:
       - run:
-          name: Patch crates
+          name: Patch local crates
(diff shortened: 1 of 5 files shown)


chesedo (approved):
Nice one!!

I'm just not sure if we want to go with cargo make or something else or maybe something like xtasks. Ie, why this and not another option?

jonaro00:
This one is closer to a cross-platform replacement for cmake, and xtasks need to compile to run.
If using xtask becomes relevant for us, that can also be done.

Issue #1288, opened 2023-10-02
[Improvement]: Consider if migrating to `tower-sessions` is appropriate


maxcountryman:
### Describe the improvement

Hi folks,

I'm the author of `axum-sessions`, which is a crate I see you all are using--I'm glad to see you've found it useful!

Over the course of the last year or so we've hit some roadblocks with our key dependency, `async-session`. The long and short of that is in order to unblock that and address some problems with `axum-sessions`'s design, we've released a new crate which aims to replace `axum-sessions`: [`tower-sessions`](https://github.com/maxcountryman/tower-sessions).

`tower-sessions` no longer relies on a third-party crate for its session implementation and this has allowed us to change its design to better fit `tower` and the broader `tower` ecosystem (i.e. `axum`). For instance, we no longer need writable and readable sessions, and have simplified the interface as a result.

I'd be curious if there's interest in migrating and am happy to help if so.

### Duplicate declaration

- [X] I have searched the issues and this improvement has not been requested before.

oddgrd:
Hey @maxcountryman, thanks for opening this issue and for offering to help in implementing this change. I think the new crate looks good, and I think it's great that you designed it around `tower`! The reason we didn't accept this issue yet, is that we don't currently use the sessions flow in our `auth` service (which is what uses `axum-sessions`). We might do so in the future, though, we just need to discuss this a bit internally. We'll get back to you when we have an answer. Thanks again!

oddgrd:
Hello again @maxcountryman! We have decided to fully remove our sessions code from the backend, so we will close this ticket for now. If we start using sessions on the backend again in the future, we'll consider using your crate. Thanks!

PR #1596, opened 2024-01-30
misc: remove session

chesedo:
## Description of change
Remove session management code from auth, gateway and deployer since we do not use it.

By running the existing tests


auth/src/api/handlers.rs
```diff
@@ -69,56 +68,22 @@ pub(crate) async fn update_user_tier(
...
-    let account_name = match session.get::<String>("account_name") {
-        Some(account_name) => account_name.into(),
-        None => match key {
-            Some(key) => user_manager.get_user_by_key(key.into()).await?.name,
-            None => return Err(Error::Unauthorized),
-        },
+    let account_name = match key {
+        Some(key) => user_manager.get_user_by_key(key.into()).await?.name,
+        None => return Err(Error::Unauthorized),
     };
```
oddgrd, line 77: We should be able to switch the parameter to be just `key: Key`, since it's a mandatory argument now. We do the same in `convert_key`.

deployer/src/handlers/local.rs
```diff
@@ -32,17 +32,9 @@ pub async fn set_jwt_bearer<B>(
     mut request: Request<B>,
     next: Next<B>,
 ) -> Result<Response, StatusCode> {
-    let mut auth_details = None;
-
     if let Some(bearer) = request.headers().typed_get::<Authorization<Bearer>>() {
```
oddgrd, line 35: Should we update the error in the `else` clause to reflect that a cookie is not an option anymore?

gateway/src/api/auth_layer.rs
```diff
@@ -31,7 +31,7 @@ static PROXY_CLIENT: Lazy<ReverseProxy<HttpConnector<GaiResolver>>> =
 const CACHE_MINUTES: u64 = 5;
 
 /// The idea of this layer is to do two things:
-/// 1. Forward all user related routes (`/login`, `/logout`, `/users/*`, etc) to our auth service
+/// 1. Forward all user related routes (`/login`, `/users/*`, etc) to our auth service
```
oddgrd, line 34: `/login` should be removed too

gateway/src/api/auth_layer.rs
```diff
@@ -118,7 +118,7 @@ where
         }
 
         let forward_to_auth = match req.uri().path() {
-            "/login" | "/logout" => true,
+            "/login" => true,
```
oddgrd, line 121: Neither of these exist now

oddgrd (commented):
Nice! I left some minor comments

deployer/src/handlers/local.rs
chesedo: Yip, I updated the error and remove all other references to "cookie" :smile:

gateway/src/api/auth_layer.rs
chesedo: I removed all references to login

auth/src/api/builder.rs
```diff
@@ -4,10 +4,9 @@ use axum::{
     extract::FromRef,
     middleware::from_extractor,
     routing::{get, post, put},
-    Router, Server,
```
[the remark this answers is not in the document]
chesedo: Not sure, this is gone in newer versions of axum (the re-export)... :thinking: Let me quickly try to undo this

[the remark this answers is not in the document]
chesedo: I was able to undo it :partying_face:

oddgrd (approved):
Thanks for the cleanup!

jonaro00:
Do we close #1288 due to this?

oddgrd:
> Do we close #1288 due to this?

Good point, done!

PR #1597, opened 2024-01-30
feat: have `auth` handle new subscriptions

chesedo:
## Description of change
Allow auth to record subscriptions and set the JWT correctly

New tests added. Will also test on staging


gateway/src/api/auth_layer.rs
```diff
@@ -117,15 +119,15 @@ where
...
+        if req_path == "/users/reset-api-key" {
             if let Some((cache_key, _)) =
                 cache_key_and_token_req(&req, self.gateway_admin_key.as_str())
             {
                 self.cache_manager.invalidate(&cache_key);
             };
         }
 
-        if req.uri().path().starts_with("/users") {
+        if req_path.starts_with("/users") || req_path.starts_with("/subscribe") {
```
oddgrd, line 130: Could subscribe also be behind `/users`?

auth/src/user.rs
```diff
@@ -21,16 +23,43 @@ use stripe::{
...
+    /// Change the tier for a user
     async fn update_tier(&self, name: &AccountName, tier: AccountTier) -> Result<(), Error>;
+
+    /// Get a user by their account name
     async fn get_user(&self, name: AccountName) -> Result<User, Error>;
+
+    /// Get a user by their api key
     async fn get_user_by_key(&self, key: ApiKey) -> Result<User, Error>;
+
+    /// Reset (remove) the key that belongs to an account
```
oddgrd, line 45: Thanks for adding all these comments! I don't think this one is completely accurate, since we are replacing the existing api key with a new one. 
```suggestion
    /// Reset the key that belongs to an account
```

auth/src/user.rs
```diff
@@ -207,6 +236,50 @@ impl UserManagement for UserManager {
...
+    ) -> Result<(), Error> {
+        // Insert a new subscription. If the same type of subscription already exists, update the
+        // subscription id and quantity.
+        query(
+            r#"INSERT INTO subscriptions (subscription_id, account_name, type, quantity)
+            VALUES ($1, $2, $3, $4)
+            ON CONFLICT (account_name, type)
+            DO UPDATE SET subscription_id = EXCLUDED.subscription_id, quantity = EXCLUDED.quantity
+        "#,
+        )
```
oddgrd, line 255: Oh, this is clever! I was thinking we would need another handler to increment the subscription, but this should work, and it's a lot simpler!

auth/src/api/handlers.rs
```diff
@@ -75,6 +75,36 @@ pub(crate) async fn put_user_reset_key(
...
+            &payload.id,
+            &payload.r#type,
+            payload.quantity,
+        )
+        .await?;
+
+    Ok(())
+}
+
+pub(crate) async fn delete_subscription(
```
oddgrd, line 96: I am wondering about when this will be called. If a user cancels their RDS subscription, will it be cancelled immediately, or at the end of the period? If it's cancelled immediately from the console, it makes sense for the console to call this after. If it's at the end of the period, this would have to be called at the end of the period.

auth/src/user.rs
```diff
@@ -337,19 +403,33 @@ impl FromRow<'_, PgRow> for Subscription {
...
+        limit.rds_quota = user
+            .subscriptions
+            .iter()
+            .find(|sub| matches!(sub.r#type, models::user::SubscriptionType::Rds))
+            .map(|sub| sub.quantity as u32)
+            .unwrap_or(0);
+
+        limit
+    }
+}
```
oddgrd, line 431: I'm thinking that we need to sync the rds subscription with stripe before we do this. This way, if a user has not paid their subscription, they will not be permitted to provision new RDS instances.

We already sync the account tier when we fetch the user, perhaps we can do the same for RDS? The current sync call is a bit hidden, though, perhaps there is a clearer way. The current sync logic is also a bit complicated.

auth/src/user.rs
```diff
@@ -376,16 +456,25 @@ where
...
     fn from(user: User) -> Self {
         Self {
             name: user.name.to_string(),
             key: user.key.expose().as_ref().to_owned(),
             account_tier: user.account_tier.to_string(),
-            // TODO: this just returns what was always returned, the id of the pro subscription.
-            // We will need to update this when we want to also return the rds subscription. We
-            // can return a vec of IDs, but it will be a breaking change for the console (I don't believe it is used anywhere else).
-            subscription_id: user.pro_subscription_id().map(|id| id.to_string()),
+            subscriptions: user.subscriptions.into_iter().map(Into::into).collect(),
```
oddgrd, line 465: This ended up being a breaking change for the console, right? We need to ensure that it's updated over there as well, but you probably talked with Guga about this already.

auth/src/user.rs
oddgrd: We'd then also need to mock the call to Stripe in the test.

auth/tests/api/users.rs
```diff
@@ -222,6 +231,70 @@ mod needs_docker {
...
+            models::user::SubscriptionType::Rds
+        );
+        assert_eq!(response.subscriptions[0].quantity, 1);
+
+        // Make sure JWT has the quota
+        let claim = app.get_claim(basic_user_key).await;
+        assert_eq!(claim.limits.rds_quota, 1);
+
+        // Send another request to insert an RDS subscription for the user.
+        // This uses a different subscription id to make sure we only keep record of one
```
oddgrd, line 277: I think what's guaranteeing we only keep one is that this account already has a subscription of the RDS type, right? It could be the same subscription ID, which I imagine will be the case if you increase the quantity of the subscription in the console (I'm not up to speed on the implementation in the console, but I assume increasing the quantity doesn't change the subscription id)?

oddgrd (commented):
This looks great, thanks! I left some questions, as well as some comments on some things I think we'll need to cover.

auth/src/api/handlers.rs
chesedo: Based on the TR I was thinking the console just calls this immediately for the time being

auth/src/user.rs
chesedo: Yip, the console update is in the works

auth/src/user.rs
chesedo: I was considering this. But since we decided in the TR to leave the sync for a later time I decided to just trust the DB (cache) for now

gateway/src/api/auth_layer.rs
chesedo: Let me check...

chesedo: Done

auth/src/user.rs
oddgrd: Ah, I interpreted that decision as we won't refactor to a cron-based sync yet, but we could still sync on actions that create a JWT with an RDS quota. They would be limited to their quota, however, so I suppose the worst case is that if they delete an instance to go under their quota, they can still create a new one indefinitely after they stop paying their subscription. I'm not sure what the timeline is for the subscription sync refactor, but if it's not too distant in the future, I think we can leave it as is for now.

auth/src/api/handlers.rs
oddgrd: Ah, okay! We should just make sure that the console also cancels the subscription immediately then, and if the user has already paid for future dates, they should be refunded. cc @GugaGongadze

oddgrd: If we delete the subscription immediately we could also delete the RDS, but that's out of scope for this PR. And we might change to delete at period end in the future.

common/src/limits.rs
```diff
@@ -17,7 +17,7 @@ pub struct Limits {
     #[serde(skip_deserializing)]
     rds_access: bool,
     /// The quantity of RDS instances this user can provision.
-    pub(crate) rds_quota: u32,
+    pub rds_quota: u32,
```
oddgrd, line 20: BTW I created a method on Limits to set the quota, if we're setting the field directly now we could remove it, I don't think it's used anywhere.

chesedo: Ah, thanks!! I missed that. Will update the code to use the method instead

oddgrd: I see from the failing tests that we'd need a getter as well now, since we need to read the quota. So we either make it pub, or add a getter method, and the setter should actually be renamed `set_rds_quota` in accordance with Rust naming convention.  :sweat_smile:

chesedo: Haha, I just saw the same. Will have a proper setter in a bit...

oddgrd (commented):
One more thing, now that we set the rds quota based on rds subscription quantity, we should refactor the code in limits that sets the quota to 1 by default for pro users.

auth/src/user.rs
```diff
@@ -244,9 +172,20 @@ impl UserManagement for UserManager {
...
     ) -> Result<(), Error> {
+        let mut transaction = self.pool.begin().await?;
+
+        if *subscription_type == models::user::SubscriptionType::Pro {
+            query("UPDATE users SET account_tier = $1 WHERE account_name = $2")
+                .bind(AccountTier::Pro.to_string())
+                .bind(name)
+                .execute(&mut *transaction)
+                .await?
+                .rows_affected();
```
oddgrd, line 183: We don't need rows_affected here, right?

auth/tests/api/users.rs
```diff
@@ -186,14 +191,12 @@ mod needs_docker {
...
+        let response = app.post_subscription("test-user", "", "pro", 1).await;
+        assert_eq!(response.status(), StatusCode::OK);
 
         // Test upgrading to pro with an incomplete checkout session object.
-        let response = app
-            .put_user("test-user", "pro", MOCKED_INCOMPLETE_CHECKOUT_SESSION)
-            .await;
-        assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+        let response = app.post_subscription("test-user", "null", "pro", 1).await;
+        assert_eq!(response.status(), StatusCode::OK);
```
oddgrd, line 199: Shouldn't these tests that deal with a checkout session be removed?

chesedo: I kept them to ensure the checkout session did not change

auth/src/user.rs
chesedo: Yes 🤦‍♂️ (copy-and-pasting 🙈 )

auth/tests/api/users.rs
oddgrd: Hmm, what do you mean by that? We don't send the checkout session to the backend anymore, right? Maybe it's just the comments here confusing me :thinking:

chesedo: Oh, I see what you mean by "checkout session". Let me remove it quick...

oddgrd (approved):
Thanks!

PR #1607, opened 2024-02-01
fix: don't try to deserialize non-200 res to service summary

oddgrd:
## Description of change

We have a lot of errors like the below in the gateway, due to the `start_last_deploy` call trying to deserialize the response from deployer as a `service::Summary` no matter what the status is, but the deployer only returns a `service::Summary` for 200s.
```
{"level":"ERROR","target":"shuttle_gateway::project","exception.message":"EOF while parsing a value at line 1 column 0","exception.stacktrace":"[]","error":"EOF while parsing a value at line 1 column 0","error.chain":"[]","code.filepath":"gateway/src/project.rs","code.namespace":"shuttle_gateway::project","code.lineno":"1441"}}]
```

This PR resolves that. We might want to refactor the error handling here further, but this will resolve at lot of inaccurate errors.

Tested with local env Datadog, and this change removes the deserialization errors.

### DIFF
--- a/gateway/src/project.rs
+++ b/gateway/src/project.rs
@@ -1505,8 +1505,6 @@ impl Service {
         jwt: String,
         admin_secret: String,
     ) -> Result<(), Box<dyn std::error::Error>> {
-        trace!(jwt, "getting last deploy");
-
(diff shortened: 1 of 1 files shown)


chesedo (approved):
Thanks for this O!! :heart:

PR #1608, opened 2024-02-02
feat: disable trace_layer on_failure, only emit error event for 500s

oddgrd:
### Motivation
- We were emitting two events for every 5xx, one from the IntoResponse implementation, which can know more about what the error was, and one from the trace layer [on_failure default implementation](https://docs.rs/tower-http/0.4.4/src/tower_http/trace/on_failure.rs.html#84-100), which only knows the status code (and some metadata).
- We were emitting error events for 4xx in the IntoResponse implementation, but they don't really warrant an error event. Trying to fetch a project that doesn't exist isn't possible, so it's not something we need to highlight as an error. The span will still have the 404 status code.
- With the trace_layer on_failure disabled, we only get the error event in IntoResponse, and we still set the status code on the span so we can track 4xx's.

### Alternatives

- With the current solution, we have to create an event in the IntoResponse implementation for three crates, rather than having the same behavior for all creates defined in the trace layer. We could disable the events in the IntoResponse instead, and just use the trace layer on_failure, but we would have less data in the event about what the error was.

Tested with local env and Datadog.

### DIFF
--- a/auth/src/error.rs
+++ b/auth/src/error.rs
@@ -43,8 +43,15 @@ impl IntoResponse for Error {
         let code = match self {
             Error::Forbidden => StatusCode::FORBIDDEN,
             Error::Unauthorized | Error::KeyMissing => StatusCode::UNAUTHORIZED,
-            Error::Database(_) | Error::UserNotFound => StatusCode::NOT_FOUND,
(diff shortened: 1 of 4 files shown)


chesedo (approved):
LGTM

PR #1609, opened 2024-02-02
feat: quite flag implemented

### LINKED ISSUE #1602: [Improvement]: `--quiet` flag for local runs
### Describe the improvement

When starting a local run, there are lots of printouts from both cargo-shuttle and the runtime's startup logic.

A `-q / --quiet` flag can be added to the `run` command, that supresses non-error prints in cargo-shuttle. The flag can also be passed on as `--quiet` to the runtime binary, where it can also be used to supress the lines that are usually printed (the ones that show up with `[Runtime]` prepended).

A solution that puts the relevant prints inside if-statements should be sufficient.

---

<sub>This issue is open to contribution. Comment if you want to work on it. Feel free to ask for help and guidance here or in our Discord.</sub>

### Duplicate declaration

dhruvdabhi101:
## Description of change

Implemented `--quite` flag for `cargo-shuttle run` command.
I added extra flag in `RunArgs` struct, then in `spin_local_run` function I checked if flag is true or not and accordingly I let print statement do the work

Closes #1602 

Tested Via running locally.
used this command `cargo run -p cargo-shuttle -- run --wd ../temp-project --quite`

and output is something like this
Screenshot 2024-02-02 at 21 59 28

Screenshot 2024-02-02 at 22 02 25

> This is very first implementation that came to my mind, but if it needs to be improved please suggest : )

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -1064,6 +1064,8 @@ impl Shuttle {
         let service_name = service.service_name()?;
         let deployment_id: Uuid = Default::default();
 
+        let quite_mode = run_args.quite;
+
         // Clones to send to spawn
         let service_name_clone = service_name.clone().to_string();
 
@@ -1079,7 +1081,13 @@ impl Shuttle {
                     shuttle_common::log::Backend::Runtime(service_name_clone.clone()),
                     line,
                 );
-                println!("{log_item}");
+                if !quite_mode {
+                    println!("{log_item}");
+                } else {
+                    if log_item.line.contains("ERROR") {
+                        println!("{log_item}");
+                    }
+                }
             }
         });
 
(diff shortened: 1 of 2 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -1079,7 +1081,13 @@ impl Shuttle {
...
                     line,
                 );
-                println!("{log_item}");
+                if !quite_mode {
+                    println!("{log_item}");
+                } else {
+                    if log_item.line.contains("ERROR") {
+                        println!("{log_item}");
+                    }
+                }
```
jonaro00, line 1090: We can't filter the runtime's lines like this. The goal is to supress shuttle's prints in cargo-shuttle and runtime. No lines from the user's program should be silenced.

jonaro00 (changes requested):
Thanks for the PR!
1. It is spelled `quiet`.
2. The printing of resource lists should not happen. The `Starting on ...` is probably the only line from cargo-shuttle that should stay.
3. Read the issue description carefully. The flag also needs to be passed to the runtime to ignore prints on that side too.

cargo-shuttle/src/lib.rs
dhruvdabhi101: Okay.. Thanks : ), will try to do that.

PR #1613, opened 2024-02-06
ci: move more tasks to cargo make

jonaro00:
Reduces number of CI jobs 36 -> 30.

The CI cache pruning is a test that will need to run for a while until we see results.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -9,14 +9,6 @@ default_to_workspace = false
 CARGO_MAKE_EXTEND_WORKSPACE_MAKEFILE = true
 # TAG = { script = ["git describe --tags --abbrev=0"] }
 
-[tasks.ci-workspace]
-run_task = { name = [
-    "check-lockfile-patches",
-    "fmt",
-    "clippy",
-    "check-lockfile",
-] }
-
 [tasks.fmt]
 command = "cargo"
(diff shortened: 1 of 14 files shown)


Makefile.toml
```diff
@@ -83,3 +120,66 @@ proto-gen \
...
+end
+
+echo
+println -s bold "### Manifest:"
+println -s bold "    ${manifest}"
+echo
+
+run cargo fmt --all --check --manifest-path ${manifest}
+
+for f in ${feature_flags}
```
chesedo, line 172: Didn't the old, CI tests run in parallel? Ie isn't this slowing down the CI?

chesedo (commented):
Looking good! Just a small nit about speed :smile:

Makefile.toml
jonaro00: Yes, however, these crates are small and generally complete before everything else. Since 90%+ of the dep tree is being compiled for one feature flag set, it is wasteful to not re-use the build output for the other feature flag sets.  The previous approach would get out of hand quickly if the matrix of flag sets to support grows.
This approach can also be taken to test more than one crate per job, but I think we are at a good middle ground here.

chesedo: I see. Sounds good!

chesedo (approved):
LGTM

PR #1614, opened 2024-02-06
refactor: uniform client wrappers

chesedo:
## Description of change
During the RDS work a pattern was starting for client code to connect to our internal services. This PR extends that pattern to the clients of all our services.

Ie. the client of a service can be found at: `use service_mod::Client`.
And the client can be constructed using `service_mod::get_client().await`.

The compiler is still happy :smile:


jonaro00 (approved):
Nice to get those long type names out to the library :D

PR #1612, opened 2024-02-05
feat(deployer, gateway): remove deployer proxy

jonaro00:
## Description of change

- Removes the proxy layer (including project name guard and fqdn check) from deployer.
  - Reduces deployer binary's CPU usage a lot.
- Runtimes always run on `0.0.0.0:8000` in deployer. Incoming proxy traffic hits the user app instantly.
- Address field in deployments table is always NULL from now on.
- Cleanup/refactor

Things to verify:
- Do old deployers break due to potential missing arg?
  - I assume not, since the args are only used at creation of a container

## How has this been tested? (if applicable)

Local stack successfully deploys, stops, and gets deployment summary.
Request throughput from a deployed app increased significantly.
Server header being set by gateway proxy works.

PR #1616, opened 2024-02-07
feat(gateway): allow multiple hostnames, proxy caching

jonaro00:
## Description of change

Extends #1612.

- Allows shuttleapp subdomain and multiple custom domains to target a project at once.
- Attempts to "fix" x-forwarded-for header, BUT the remote IP is now the network gateway in the docker network. A change in the docker network is likely required to get real IPs.
- Refactors gateway's proxy and bouncer to use axum routers instead of a tower service mess.
- Decreased the amount of cloning that was being done per proxy request.
- Caching of project name lookup in proxy, reduces number of project table reads by up to 4000x
- Caching of custom domain lookup, same principle

- Performance seems unaffected by the router change
- ~25% less CPU usage on gateway after project state caching introduced. Still, the gateway is using tons more CPU than the user app when being spammed with small requests.
  - Caching the IP instead of the state did not seem to improve performance. The bulk of the gateway CPU per proxy hit is still there.
- Things that did not improve performance:
  - turning off the tracing sub
- The changes to the bouncer work on unstable for adding two custom domains to a project.

### DIFF
--- a/gateway/src/proxy.rs
+++ b/gateway/src/proxy.rs
@@ -96,7 +96,7 @@ async fn proxy(
             .ok_or_else(|| Error::from_kind(ErrorKind::ProjectNotReady))?;
         state.project_cache.insert(
             project_name.as_str(),
-            ip.clone(),
+            ip,
             std::time::Duration::from_millis(1000),
         );
         ip
--- a/gateway/src/proxy.rs
+++ b/gateway/src/proxy.rs
@@ -9,6 +9,7 @@ use axum::response::Response;
 use axum::routing::any;
 use axum_server::accept::DefaultAcceptor;
 use axum_server::tls_rustls::RustlsAcceptor;
+use cached::{Cached, TimedCache};
 use fqdn::{fqdn, FQDN};
 use futures::prelude::*;
 use http::header::SERVER;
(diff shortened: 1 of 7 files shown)


gateway/src/proxy.rs
```diff
@@ -41,6 +44,7 @@ pub struct ProxyState {
     gateway: Arc<GatewayService>,
     task_sender: Sender<BoxedTask>,
     public: FQDN,
+    project_cache: Mutex<TimedCache<ProjectName, FindProjectPayload>>,
```
oddgrd, line 47: We have [a cache](https://github.com/shuttle-hq/shuttle/blob/main/common/src/backends/cache.rs) implementation in common. We use it in the gateway with an rw lock, which makes sense for the jwt, but perhaps not for this short-lived cache. It would be worth testing, though. Should we use the existing cache here rather than this crate, or does this crate offer any advantages, e.g. increased performance?

jonaro00: The TTL cache supports an individual duration per item, which we don't need, but it does looks more lightweight. I did try a RwLock instead of Mutex, but the TimedCache has a hit/miss counter, which means that `get_cached()` is mutable, and needs write access.
Regardless, I don't think it should matter much in terms of performance.

oddgrd: IMO it's worth checking, given that this is a performance optimization, and you're also adding a new dependency by not using the existing cache.

jonaro00: No meaningful difference, but I'll use the existing one since it is lighter and no new dep.

oddgrd: Thanks for checking!

chesedo (approved):
LGTM. We just have to test this thoroughly

gateway/src/service.rs
```diff
@@ -1102,6 +1102,7 @@ impl GatewayContext {
     }
 }
 
+#[derive(Clone)]
```
oddgrd, line 1105: Is this needed?

oddgrd (approved):
LGTM, thanks!

Issue #1565, opened 2024-01-25
[Resources]: Blob storage (S3 etc)


jonaro00:
A blob storage integration such as AWS S3 would make it easy to store many and/or large files via Shuttle apps.
This would have a Free tier limit and then be billed per additional GB.

We might be able to provide OpenDAL as one of the resource output types https://github.com/shuttle-hq/shuttle/issues/1321

---
<sub>This issue is a part of the [Product Feature Poll](https://github.com/shuttle-hq/shuttle/issues/1546) ([all issues](https://github.com/shuttle-hq/shuttle/issues?q=label%3A%22Feature+Poll%22+sort%3Acreated-asc)). React with :+1: if you want this feature. Comment if you have suggestions related to this feature.</sub>

Xuanwo:
This feature must be cool. I'm keen on contributing to its development. 🥳

Xuanwo:
The resource design seems cool and clear. I'm working on this now.

jonaro00:
shuttle-opendal (#1617) and other libraries allow you to use your own S3 etc, but keeping this issue open since it tracks Shuttle-native blob storage.

jonaro00:
S3 buckets etc that we provision, like we do for AWS RDS currently.

PR #1617, opened 2024-02-08
feat: Add OpenDAL resource support

Xuanwo:
## Description of change

Close https://github.com/shuttle-hq/shuttle/issues/1565

This PR will add opendal sources support that users can connect 50+ storage services including s3, azblob, gcs, oss, .....

unit test.


deployer/src/persistence/resource/mod.rs
```diff
@@ -81,6 +81,7 @@ pub enum Type {
     Persist,
     Turso,
     Metadata,
+    Opendal,
```
Xuanwo, line 84: I'm not sure if `Opendal` is a good name, maybe we can use `Blob` or `Storage` to better describe what it does?

jonaro00: It would be better to not add a variant here and instead use `Custom`. The reason for this is that we are trying to use this enum for resources that Shuttle can manage, and all client plugins are separate.

resources/opendal/README.md
```diff
@@ -0,0 +1,37 @@
...
+) -> ShuttleAxum {}
+```
+
+### Parameters
+
+| Parameter | Type  | Default    | Description                                      |
+|-----------|-------|------------|--------------------------------------------------|
+| scheme    | `str` | `"memory"` | The scheme of the storage service to connect to. |
+
+All secrets are loaded from your `Secret.toml` file. Visit the [OpenDAL Documentation](https://opendal.apache.org/docs/rust/opendal/services/index.html) for more information on how to setup the secrets for the storage service you want to connect to.
```
jonaro00, line 37: An example of this would be helpful, such as "Example: if you are using S3, you need to set the secrets AAA, BBB, and CCC. Visit the docs for more details: ..."

jonaro00 (commented):
Looks like a great addition!
We have some major refactorings of the resource code planned, and I think we'll at least want https://github.com/shuttle-hq/shuttle/pull/1543 done before adding this. I'll let you know of the progress.
Some small comments:

resources/opendal/README.md
Xuanwo: Added.

Xuanwo:
Happy to see #1543 has been merge! Are there any work I need to do for this PR?

jonaro00 (commented):
It looks good to me. I will check with the team for further approvals.

resources/opendal/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Parameters
+
+| Parameter | Type  | Default    | Description                                      |
+|-----------|-------|------------|--------------------------------------------------|
+| scheme    | `str` | `"memory"` | The scheme of the storage service to connect to. |
+
+All secrets are loaded from your `Secret.toml` file. 
+
+For instance, when using `s3`, you can configure the scheme to `s3` and specify the secrets: `bucket`, `access_key_id`, and `secret_access_key`.
```
oddgrd, line 39: This is a clever way to load the config vars! However, there might be more secrets in `Secrets.toml` that are not related to the OpenDAL resource. Will that be a problem, or will they simply be ignored when initializing the operator?

oddgrd (commented):
This is cool, thank you! I left one question. Other than that, an example in https://github.com/shuttle-hq/shuttle-examples would be terrific, but if you don't have time, we can also get to it. :slightly_smiling_face:

resources/opendal/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    fn from(error: Error) -> Self {
+        let msg = format!("Failed to build opendal resource: {:?}", error.0);
+        ShuttleError::Custom(CustomError::msg(msg))
+    }
+}
+
+#[async_trait]
+impl ResourceBuilder for Opendal {
+    const TYPE: Type = Type::Custom;
+    type Config = Self;
```
jonaro00, line 50: For this type of plugin it is better to use `()` as the config type (see some of the other plugins), since there is nothing that the shuttle servers should know about this resource.

resources/opendal/README.md
jonaro00: Looks like that's the case.

resources/opendal/README.md
```diff
@@ -0,0 +1,41 @@
...
+Users can connect OpenDAL to access data from a variety of storage services, including: s3, azblob, gcs, oss and [so on](https://opendal.apache.org/docs/rust/opendal/services/index.html).
+
+## Usage
+
+**IMPORTANT**: Currently Shuttle isn't able to provision a storage for you (yet). This means you will have to create the storage service first and setup the secrets accordingly.
+
+Add `shuttle-opendal` to the dependencies for your service by running `cargo add shuttle-opendal`.
+This resource will be provided by adding the `shuttle_opendal::Opendal` attribute to your Shuttle `main` decorated function.
+
+It returns a `opendal::OpenDAL` for you to connect the storage service.
```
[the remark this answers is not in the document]
Xuanwo: Fixed.

resources/opendal/src/lib.rs
Xuanwo: Fixed.

Xuanwo:
> Other than that, an example in [shuttle-hq/shuttle-examples](https://github.com/shuttle-hq/shuttle-examples) would be terrific, but if you don't have time, we can also get to it. 🙂

I'm willing to add examples! If I understand correctly, we need to release opendal resource first?

jonaro00 (approved):

PR #1611, opened 2024-02-05
feat: ensure errors conform to Datadog's error format

oddgrd:
## Description of change

WIP

## How has this been tested? (if applicable)

WIP

PR #1626, opened 2024-02-14
feat: enable Datadog APM error tracking with a tracing layer

oddgrd:
## Description of change

Add an error layer to the tracing subscriber to capture any `dyn std::error::Error`, and set the fields required for Datadog APM error tracking in the otel data extension on the current span.

This also fixes the issue where the error message field was being overwritten by the `http.status_code` field, by setting the status code in the otel data extension rather than directly on the span.

### Todo

- [x] Go over our errors and set a `&(dyn std::error::Error)` field where possible. (Done in [fa8d1b5](https://github.com/shuttle-hq/shuttle/pull/1626/commits/fa8d1b597aa074a104ca924f34ebd9e27525d829))
- [x] Remove some error events that shouldn't be error events (e.g. 400s). (Done in [fa8d1b5](https://github.com/shuttle-hq/shuttle/pull/1626/commits/fa8d1b597aa074a104ca924f34ebd9e27525d829))

More remains for both points above, but I have tried to remove all obvious "errors that shouldn't be errors", and I've migrated the errors in all error events to `&dyn std::error::Error` where it was straight forward. I think the clean-up can continue in future PRs.

- [x] Address ApiError styling changes: https://github.com/shuttle-hq/shuttle/pull/1626/files#diff-8adfbdd437430afaab5b1ea9005f514578a64308b91ff9cf26bae04155e4e0ab

### Caveats
- This does not enable logs error tracking, for that we would need to set these fields on all events, which is a bit messy without valuable support.
- For APM error tracking, in addition to the fields requirement, it's also a requirement that the span is a service entry span.
- We have no good way of getting the error type from a `dyn Error`, but the errors will still be distinguished in error tracking by error.message and error.stack (Datadog generates a fingerprint for each error based on `error.message`, `error.type` and `error.stack`). 

### Alternatives considered
- Using valuable for something like the above. We currently expose our tracing dependencies to users in the shuttle-runtime, so using an unstable feature isn't viable. We can reconsider it when the common features that leak the valuable dependency are removed from the runtime, and they will be when provisioning is extracted from the runtime. Or if valuable has a stable release.
- Using a utility function to emit an error event with the correct fields, and also set the fields on the otel data extension in the span. This works, and we'll have both APM and Logs error tracking, but we will need to modify all our existing error events. Some of which also do not impl Error, and some of them are exposed to users via our deployment log layer. Regardless, this is a viable option, but it's my opinion that it's better to start with this simple layer, and we can iterate on it when we get a clearer picture of the error tracking feature and what we are missing. The work on this option has been started here: https://github.com/shuttle-hq/shuttle/tree/feat/datadog-error-tracking

Tested with a local environment against Datadog.


oddgrd:
> > For APM error tracking, in addition to the fields requirement, it's also a requirement that the span is a service entry span.
> 
> Why is this the case?

I'm not sure, but it's listed in the [requirements here](https://docs.datadoghq.com/tracing/error_tracking/#use-span-tags-to-track-error-spans).

common/src/backends/otlp_tracing_bridge.rs
```diff
@@ -152,3 +153,104 @@ const fn severity_of_level(level: &Level) -> Severity {
...
+                let error_fields = [
+                    KeyValue::new("error.message", message),
+                    KeyValue::new("error.type", r#type),
+                    KeyValue::new("error.stack", stack),
+                ];
+                let builder_attrs = otel_data
+                    .builder
+                    .attributes
+                    .get_or_insert(Vec::with_capacity(3));
+                builder_attrs.extend(error_fields);
```
iulianbarbu, line 206: nit: out of curiosity, after getting the `span` can't we set the same fields by using the `.set_attribute` function? Just wondering if it can be done simpler.

common/src/backends/otlp_tracing_bridge.rs
```diff
@@ -152,3 +153,104 @@ const fn severity_of_level(level: &Level) -> Severity {
...
+        // This visitor is only concerned with recording errors, do nothing for debug fields.
+    }
+    fn record_error(&mut self, _field: &Field, value: &(dyn std::error::Error + 'static)) {
+        // Create an error source chain, including the top-level error.
+        let source_chain = {
+            // Datadog expects there to be at least two lines in the stack field for the apm error
+            // tracking feature to work, so we ensure there always is.
+            let mut chain: String = format!("Error source chain:\n{}", value);
+            let mut next_err = value.source();
+            // TODO: skip the first error?
```
iulianbarbu, line 241: what do you mean?

common/src/backends/trace.rs
```diff
@@ -80,8 +80,9 @@ where
     subscriber
         .with(filter_layer)
         .with(fmt_layer)
-        .with(otel_layer)
         .with(appender_tracing_layer)
+        .with(otel_layer)
+        .with(ErrorTracingLayer::new())
```
iulianbarbu, line 85: will this end up being used by `shuttle-runtime`?

iulianbarbu (commented):
Super excited by what this enables! Thanks @oddgrd ! 😍

common/src/backends/otlp_tracing_bridge.rs
```diff
@@ -152,3 +153,104 @@ const fn severity_of_level(level: &Level) -> Severity {
...
+}
+#[derive(Default)]
+struct ErrorVisitor {
+    error: DatadogError,
+}
+
+impl ErrorVisitor {
+    /// We only care about error events with the [ID_IDENT] field.
+    fn is_valid(metadata: &Metadata) -> bool {
+        metadata.is_event() && metadata.level() == &Level::ERROR
```
chesedo, line 226: The doc here indicates that we check for an `ID_IDENT` fields, but I don't see that in the code

common/src/backends/otlp_tracing_bridge.rs
```diff
@@ -152,3 +153,104 @@ const fn severity_of_level(level: &Level) -> Severity {
...
+    }
+}
+
+impl Visit for ErrorVisitor {
+    fn record_debug(&mut self, _field: &Field, _value: &dyn std::fmt::Debug) {
+        // This visitor is only concerned with recording errors, do nothing for debug fields.
+    }
+    fn record_error(&mut self, _field: &Field, value: &(dyn std::error::Error + 'static)) {
+        // Create an error source chain, including the top-level error.
```
chesedo, line 235: Ohhhh, nice :partying_face:

common/src/backends/otlp_tracing_bridge.rs
```diff
@@ -152,3 +153,104 @@ const fn severity_of_level(level: &Level) -> Severity {
...
+        event.record(&mut visitor);
+
+        let DatadogError {
+            message,
+            r#type,
+            stack,
+        } = visitor.error;
+
+        if let Some(span) = ctx.lookup_current() {
+            if let Some(otel_data) = span.extensions_mut().get_mut::<OtelData>() {
```
chesedo, line 196: I can't find what sets the `OtelData` extension. Under which circumstance does it get set?

common/src/backends/otlp_tracing_bridge.rs
chesedo: Also why?

common/src/models/error.rs
```diff
@@ -19,13 +18,11 @@ impl ApiError {
 }
 
 impl Display for ApiError {
+    // TODO: styling was removed, create pretty print method in cargo-shuttle?
```
chesedo, line 21: Todo

gateway/src/proxy.rs
```diff
@@ -121,8 +121,8 @@ async fn proxy(
     let mut res = PROXY_CLIENT
         .call(addr.ip(), &target_url, req)
         .await
-        .map_err(|e| {
-            error!(error = ?e, "gateway proxy client error");
+        .map_err(|err| {
+            error!(error = ?err, "gateway proxy client error");
```
chesedo, line 125: Not `dyn StdError` this time?

chesedo (commented):
Left some notes. I think this PR warrants a demo to all the engineers as we should use the `dyn StdError` whenever we are logging an error

common/src/backends/otlp_tracing_bridge.rs
oddgrd: My thinking was that maybe we want to skip the first error, since it will be in the error message field. But I decided against it, I thought I had removed this todo actually, so I will do it now. :smile:

common/src/backends/trace.rs
oddgrd: The runtime doesn't use `setup_tracing`

common/src/backends/otlp_tracing_bridge.rs
oddgrd: Woops, I did in an earlier iteration, but forgot to remove the comment.

common/src/backends/otlp_tracing_bridge.rs
oddgrd: It's set by the otel_layer in `setup_tracing`, so the error layer needs to go after it. I can leave a comment about that, as Jocelyn did in his valuable POC: https://github.com/shuttle-hq/shuttle/pull/1611/files#diff-6b6299a4e2fbb65c112d0cfaacb9227d9f15b1e1002d01c968b530d7c65d094c

gateway/src/proxy.rs
oddgrd: This is one I tried to change, but it was tricky, due to `ProxyError` from the reverse proxy crate not implementing std Error. There are a few more similar cases we still need to tweak to get them covered by error tracking.

common/src/backends/otlp_tracing_bridge.rs
oddgrd: The trait I used in the utility function is [implemented for `tracing::Span`](https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/trait.OpenTelemetrySpanExt.html#impl-OpenTelemetrySpanExt-for-Span), but the type here is [SpanRef.](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/registry/struct.SpanRef.html)

common/src/backends/otlp_tracing_bridge.rs
oddgrd: I believe this is the source: https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/src/layer.rs#L911

oddgrd:
> Left some notes. I think this PR warrants a demo to all the engineers as we should use the `dyn StdError` whenever we are logging an error

Sure, I'm happy to do a demo! I'm also planning to document how we should format errors in the M&O Wiki.

common/src/models/error.rs
oddgrd: The reason I removed the styling is we need to convert to `ApiError` to check the status code in `gateway` `impl IntoResponse for Error`, and I was then recording the `ApiError` in the error event. However, we can clone the gateway error kind, create the `ApiError` from that, and still emit the original error in the error event. So I restored the styling on `ApiError`.

common/src/backends/otlp_tracing_bridge.rs
iulianbarbu: I see, thanks!

iulianbarbu (approved):

Kazy (approved):
Great job :tada:  I agree with your comments regarding `Valuable` and that we should work iteratively. If we get the errors from the logs there, it will already be highly valuable (no pun intended).

common/src/backends/otlp_tracing_bridge.rs
chesedo: Ahhh, I see :+1:

chesedo: Yes, please comment this somewhere

oddgrd: I added a comment in setup_tracing here: https://github.com/shuttle-hq/shuttle/pull/1626/commits/729eb6b1918b5ba8ee7ff13c488b63130e6c0672

chesedo:
> Sure, I'm happy to do a demo! I'm also planning to document how we should format errors in the M&O Wiki.

I'm hoping a demo will make it clear as to why this is valuable - the way we should be doing it :smile:

PR #1627, opened 2024-02-14
fix(gateway): custom domain followup improvements

jonaro00:
## Description of change

- don't restart project on custom domain add
- fix returned url

not tested

### DIFF
--- a/gateway/src/api/latest.rs
+++ b/gateway/src/api/latest.rs
@@ -398,7 +398,7 @@ async fn override_get_delete_service(
         .find_custom_domain_for_project(&project_name)
         .await
         .unwrap_or_default() // use project name if domain lookup fails
-        .map(|c| c.fqdn.to_string())
+        .map(|c| format!("https://{}", c.fqdn.to_string()))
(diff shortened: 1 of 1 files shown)


gateway/src/api/latest.rs
```diff
@@ -619,36 +616,6 @@ async fn request_custom_domain_acme_certificate(
...
-    let idle_minutes = container.idle_minutes();
-
-    // Destroy and recreate the project with the new domain.
-    service
-        .new_task()
-        .project(project_name.clone())
-        .and_then(task::destroy())
-        .and_then(task::run_until_done())
-        .and_then(task::run(move |ctx| async move {
-            let creating = ProjectCreating::new_with_random_initial_key(
```
chesedo, line 640: I have a feeling some of the methods called here (like `new_with_random_initial_key`) are now unused and their definitions can safely be removed

chesedo (approved):
LGTM :partying_face: 

Left a small nit

gateway/src/api/latest.rs
jonaro00: Ctrl+F disagrees

PR #1629, opened 2024-02-15
Issue 1600

biplab5464:
## Description of change
Addition of head and taill arguments in the Cargo shuttle log command

### DIFF
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -117,13 +117,13 @@ pub enum Command {
         /// Don't display timestamps and log origin tags
         raw: bool,
         /// Views First N line of the logs
-        #[arg(long)]
+        #[arg(long, group = "output_mode")]
         head: Option<u32>,
         /// Views Last N line of the logs
-        #[arg(long)]
+        #[arg(long, group = "output_mode")]
         tail: Option<u32>,
         /// Views all line of the logs
-        #[arg(long)]
+        #[arg(long, group = "output_mode")]
         all: bool,
     },
     /// List or manage projects on Shuttle
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -116,6 +116,15 @@ pub enum Command {
         #[arg(long)]
         /// Don't display timestamps and log origin tags
         raw: bool,
+        /// Views First N line of the logs
+        #[arg(long)]
+        head: Option<u32>,
+        /// Views Last N line of the logs
+        #[arg(long)]
+        tail: Option<u32>,
+        /// Views all line of the logs
+        #[arg(long)]
+        all: bool,
     },
     /// List or manage projects on Shuttle
     #[command(subcommand)]
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -220,9 +220,9 @@ impl Shuttle {
                 latest,
                 follow,
                 raw,
-                head ,
+                head,
                 tail,
-                all
+                all,
             } => self.logs(id, latest, follow, raw, head, tail, all).await,
             Command::Deployment(DeploymentCommand::List { page, limit, raw }) => {
                 self.deployments_list(page, limit, raw).await
@@ -741,19 +741,16 @@ impl Shuttle {
         latest: bool,
         follow: bool,
         raw: bool,
-        head : Option<u32>,
-        tail : Option<u32>,
-        all : bool
+        head: Option<u32>,
+        tail: Option<u32>,
+        all: bool,
     ) -> Result<CommandOutcome> {
-        
-        let mut count = 0;
-        let mut mode =("tail",1000);
-        if let Some(num) = head { count += 1; mode = ("head",num) }
-        if let Some(num) = tail { count += 1; mode = ("tail",num) }
-        if all { count += 1; mode = ("all",0) }
-
-        if count > 1 {
-            bail!("Error: Only one of 'head', 'tail', or 'all' can be used at a time.")
+        let mode: (&str, u32);
+        match (head, tail, all) {
+            (Some(num), _, _) => mode = ("head", num),
+            (_, Some(num), _) => mode = ("tail", num),
+            (_, _, all) if all => mode = ("all", 0),
+            _ => mode = ("tail", 1000),
         }
 
         let client = self.client.as_ref().unwrap();
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -220,7 +220,10 @@ impl Shuttle {
                 latest,
                 follow,
                 raw,
-            } => self.logs(id, latest, follow, raw).await,
+                head ,
+                tail,
+                all
+            } => self.logs(id, latest, follow, raw, head, tail, all).await,
             Command::Deployment(DeploymentCommand::List { page, limit, raw }) => {
                 self.deployments_list(page, limit, raw).await
             }
@@ -738,7 +741,21 @@ impl Shuttle {
         latest: bool,
         follow: bool,
         raw: bool,
+        head : Option<u32>,
+        tail : Option<u32>,
+        all : bool
     ) -> Result<CommandOutcome> {
+        
+        let mut count = 0;
+        let mut mode =("tail",1000);
+        if let Some(num) = head { count += 1; mode = ("head",num) }
+        if let Some(num) = tail { count += 1; mode = ("tail",num) }
+        if all { count += 1; mode = ("all",0) }
+
+        if count > 1 {
+            bail!("Error: Only one of 'head', 'tail', or 'all' can be used at a time.")
+        }
+
         let client = self.client.as_ref().unwrap();
         let id = if let Some(id) = id {
             id


biplab5464:
Hello, is my pr ready to be merged

proto/logger.proto
```diff
@@ -26,6 +26,12 @@ message LogsRequest {
   string deployment_id = 1;
 }
 
+message LogsRequestMode {
+  string deployment_id = 1;
+  string mode = 2;
+  uint32 len = 3;
+}
```
oddgrd, line 33: We should rather add the mode and len to the existing LogsRequest.

cargo-shuttle/src/args.rs
oddgrd, line 1: There seem to be a lot of unrelated changes in this file, I'm not quite sure what has happened. We can try to help you figure it out, if you don't know what happened.

cargo-shuttle/src/client.rs
```diff
@@ -174,8 +174,18 @@ impl Client {
...
+        project: &str,
+        deployment_id: &Uuid,
+        mode: &str,
+        len: u32,
+    ) -> Result<Vec<LogItem>> {
+        let path = match mode {
+            "head" => format!("/projects/{project}/deployments/{deployment_id}/logs?head={len}"),
+            "tail" => format!("/projects/{project}/deployments/{deployment_id}/logs?tail={len}"),
+            _ => format!("/projects/{project}/deployments/{deployment_id}/logs"),
+        };
```
oddgrd, line 188: You could use something like this here instead: https://stackoverflow.com/a/30154791. We should also consider creating an enum for `mode` and putting it in common, so we can use it everywhere we need it.

cargo-shuttle/src/lib.rs
oddgrd, line 1: There are a lot of unrelated changes here as well, so it's difficult to review. See comment on cargo-shuttle/src/args for more info.

deployer/src/handlers/mod.rs
```diff
@@ -55,6 +55,11 @@ pub struct PaginationDetails {
     pub limit: Option<u32>,
 }
 
+#[derive(Deserialize)]
+struct LogSize {
+    head: Option<u32>,
+    tail: Option<u32>,
```
oddgrd, line 61: I think we can use an enum for the mode here as well, see https://stackoverflow.com/a/74965172

oddgrd (commented):
Thanks for working on this @biplab5464! I left a first round of comments, but it is a bit difficult to review some of the files since there are a lot of unrelated changes. It seems like a merge error, where you have overwritten the git history and changed the author to yourself, but I'm not quite sure yet, as I don't see any force pushes. Let us know if you have questions, and we can try to help you figure this out. See comments on the files in question.

biplab5464:
i was not active for some days, i saw it today, i will work on the changes you mentioned 

and i haven't did anything to the git history, nothing of my knowledge

i was asked to just work on normal log and leave the follow log so i had improvised this, 

can we talk on discord for helping me a little

cargo-shuttle/src/args.rs
biplab5464: it seems their is some problem with the git 
```
Logs {
        /// Deployment ID to get logs for. Defaults to currently running deployment
        id: Option<Uuid>,
        #[arg(short, long)]
        /// View logs from the most recent deployment (which is not always the latest running one)
        latest: bool,
        #[arg(short, long)]
        /// Follow log output
        follow: bool,
        #[arg(long)]
        /// Don't display timestamps and log origin tags
        raw: bool,
        /// Views First N line of the logs
        #[arg(long, group = "output_mode")]
        head: Option<u32>,
        /// Views Last N line of the logs
        #[arg(long, group = "output_mode")]
        tail: Option<u32>,
        /// Views all line of the logs
        #[arg(long, group = "output_mode")]
        all: bool,
    },
```
i have added the field `head`, `tail` and `all` that is all the changes in the file

oddgrd:
> i was not active for some days, i saw it today, i will work on the changes you mentioned
> 
> and i haven't did anything to the git history, nothing of my knowledge
> 
> i was asked to just work on normal log and leave the follow log so i had improvised this,
> 
> can we talk on discord for helping me a little

Hey @biplab5464, feel free to reach out in the contributors channel if you want more synchronous communication! Regarding the large diffs in this PR, a colleague pointed out that it may be caused by how windows handles line breaks. I found this article on the subject: https://medium.com/bootdotdev/how-to-get-consistent-line-breaks-in-vs-code-lf-vs-crlf-e1583bf0f0b6, I would try this out on your system to see if it removes the extra changes in the cargo-shuttle args.rs and lib.rs files.

biplab5464:
It may be the cause but i do coding in WSL, so i think it shouldn't be a problem

biplab5464:
is my pr ready for merge?

deployer/src/handlers/mod.rs
```diff
@@ -459,9 +465,21 @@ pub async fn get_logs(
...
 ) -> Result<Json<Vec<LogItem>>> {
+    let len = len.unwrap_or(0);
+    let mode: String = match mode {
+        Some(temp) => match temp {
+            LogMode::Head => "head".to_string(),
+            LogMode::Tail => "tail".to_string(),
+            _ => "all".to_string(),
+        },
+        None => "all".to_string(),
+    };
```
oddgrd, line 478: We can derive [strum::Display](https://docs.rs/strum/latest/strum/derive.Display.html) for the LogMode enum, and we don't need to manually convert it to a string here. We can add a `#[strum(serialize_all = "lowercase")]` to make the string representation lower case.

common/src/log.rs
```diff
@@ -31,6 +31,13 @@ pub enum Backend {
     Runtime(String),
 }
 
+#[derive(Deserialize)]
+pub enum LogMode {
+    Head,
+    Tail,
+    All,
+}
```
oddgrd, line 39: I'm not sure we need the `All` variant. Can't we just default to all if no log mode or length is passed? Then in the CLI, we can have a default length set for the logs, and if someone passes `--all` we will send the request without any length or mode.

cargo-shuttle/src/client.rs
```diff
@@ -174,8 +175,22 @@ impl Client {
...
+    ) -> Result<Vec<LogItem>> {
+        let path = match mode {
+            LogMode::Head => {
+                format!("/projects/{project}/deployments/{deployment_id}/logs?mode=head&len={len}")
+            }
+            LogMode::Tail => {
+                format!("/projects/{project}/deployments/{deployment_id}/logs?mode=tail&len={len}")
+            }
+            LogMode::All => format!("/projects/{project}/deployments/{deployment_id}/logs"),
+        };
```
oddgrd, line 193: The first part of the string is the same for all of them, so we don't need to repeat that for each variant. So we can have a url string at the start which is `/projects/{project}/deployments/{deployment_id}/logs`, and then if the mode is tail or head, we add the query params at the end.

oddgrd (commented):
Thanks for addressing my review @biplab5464! Sorry about the delay in responding, we've had Easter holidays here. And sorry about this PR taking so long, it's turned out to be a bit larger than we expected. But it's getting closer!

I have left some more comments you need to address. There are conflicts with `main` as well, so I would suggest you first fetch the latest main, then merge it into your branch. Then, address my review and test the changes.

And since this turned out to be more substantial than expected for a good first issue, do let us know if you are not able to continue, or if it has become too time-consuming for you, and we can take it over.

proto/logger.proto
```diff
@@ -24,6 +24,8 @@ message StoreLogsResponse {
 
 message LogsRequest {
   string deployment_id = 1;
+  string mode = 2;
+  uint32 len =3;
```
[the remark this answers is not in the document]
jonaro00: We could also do
```protobuf
optional uint32 head = 2;
optional uint32 tail = 3;
```

common/src/log.rs
jonaro00: I prefer `All` over having `Option<LogMode>`. My other suggestion allows us to send nothing in the case of `all`.

biplab5464:
yeah it is bit challenging and i like it so much, i am learning a lot of thing. Please guide me as it is. It is a fun learning experience.

My laptop charger broke so i couldn't do anything for 2 weeks as it was getting dispatch. So sorry for the delay. Also i fell sick in between. Also i am have job so mostly i able to work on it on weekends mostly. I mean i able to give less time on weekday I will be consistent form now.

common/src/log.rs
oddgrd: That makes sense, we can go for that!

proto/logger.proto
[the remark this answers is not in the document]
biplab5464: so i am moving ahead with @jonaro00  idea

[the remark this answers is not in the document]
oddgrd: Yes, I think that's a good suggestion.

common/src/log.rs
biplab5464: i have to remove Option from both LogMode and len if i understand it correctly?

jonaro00 (approved):
Did some final fixes, tested and works as expected. Great job @biplab5464

oddgrd (approved):
LGTM, thanks @biplab5464!

biplab5464:
Nice changes @jonaro00, i don't know why it didn't strike my mind. Thanks for teaching me
Also thanks to @oddgrd for guiding me 
I learn a lot.
Hope i will work with you again in future.

PR #1630, opened 2024-02-16
misc(provisioner): check project ownership in APIs

iulianbarbu:
## Description of change

Check the project ownership for provisioner requests.

Testing is done locally, projects are still deployed successfully (e.g. postgres shared db).


chesedo (approved):
LGTM

provisioner/src/lib.rs
```diff
@@ -478,6 +478,18 @@ impl Provisioner for ShuttleProvisioner {
...
+        // Check project ownership.
+        if !claim
+            .owns_project(&self.gateway_client, &request.project_name)
+            .await
+            .map_err(|_| Status::internal("can not verify project ownership"))?
+        {
+            return Err(Status::unauthenticated(
+                "the request lacks the authorizations",
+            ));
+        }
```
oddgrd, line 491: Should we emit an error event here if this is false, since if this happens, it's likely an attempt at something malicious? It seems like `Status` implements Error, so we could do `error = &status as &dyn std::error::Error`.

provisioner/src/lib.rs
```diff
@@ -527,11 +539,24 @@ impl Provisioner for ShuttleProvisioner {
...
```
oddgrd, line 558: Same suggestion as above comment.

oddgrd (commented):
LGTM! Just left a suggestion about a potentially useful error event.

provisioner/src/lib.rs
iulianbarbu: good idea, done!

provisioner/src/lib.rs
```diff
@@ -478,6 +478,19 @@ impl Provisioner for ShuttleProvisioner {
...
+
+        // Check project ownership.
+        if !claim
+            .owns_project(&self.gateway_client, &request.project_name)
+            .await
+            .map_err(|_| Status::internal("can not verify project ownership"))?
+        {
+            let msg = "the request lacks the authorizations";
+            let status = Status::permission_denied(msg);
+            error!(error = &status as &dyn std::error::Error, msg);
```
oddgrd, line 490: Thanks! Could we write it a bit simpler, like the suggestion below?
```suggestion
            let status = Status::permission_denied("the request is unauthorized");
            error!(error = &status as &dyn std::error::Error);
```

oddgrd: Perhaps a comment is good here too, since we wouldn't usually emit an error for a 4xx. We could mention that we emit an error event here because it indicates an attempt at something malicious.

iulianbarbu: Done

oddgrd (approved):

PR #1637, opened 2024-02-27
chore: remove builder

oddgrd:
## Description of change

Completely remove the builder service, as we have not used it since 0.37, and we do not plan to start using it again.

CI, local clippy runs.


chesedo (approved):
LGTM :partying_face:

PR #1640, opened 2024-02-28
feat(deployer): load phase caching, automatic startup

jonaro00:
## Description of change

- Make deployer always start its last running deploy on startup (restart / wakeup) instead of waiting for gateway's request for doing it manually
- Save resources when a service is started with a claim (deployment request with contact with backends), and read it on subsequent wakeups/restarts to not need any backend request.
- Fix incorrect deployment id of 'running' service being returned by `status`

- On staging, no 503/502 was observed on first idle hit against a hello world, (same test gave 502 on prod)


iulianbarbu (approved):
LGTM!

PR #1650, opened 2024-03-01
ci: remove buildx cache, bump versions

jonaro00:

Can test the buildx bump on next unstable build


oddgrd (approved):
I looked at the duration of the build & push before and after the buildx removal, and it looks like a ~1 minute duration reduction. Awesome!

PR #1652, opened 2024-03-02
misc: remove shuttle-next

jonaro00:
removing it from examples in https://github.com/shuttle-hq/shuttle-examples/pull/148


oddgrd (approved):
That's a great looking diff, thanks! :heart_eyes:

PR #1654, opened 2024-03-02
misc: add project name to the default directory

christos-h:
## Description of change
Add project name to the default dir.

`cargo-shuttle` integration tests.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -399,12 +399,15 @@ impl Shuttle {
             let path = args
                 .path
                 .to_str()
-                .context("path arg should always be set")?;
+                .context("path arg should always be set")?
+                .parse::<PathBuf>()?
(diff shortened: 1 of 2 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -398,31 +398,34 @@ impl Shuttle {
         let path = if needs_path {
             let path = args
                 .path
-                .to_str()
-                .context("path arg should always be set")?;
+                .join(project_args.name.as_ref().expect("name should be set"));
```
oddgrd, line 401: Could we keep the context, rather than expect? Or does it not provide a lot of value?

oddgrd (approved):
LGTM, thanks! I left one non-blocking question.

cargo-shuttle/src/lib.rs
jonaro00: expect makes sense since the error would be due to flawed logic in the code before this

PR #1656, opened 2024-03-04
feat(logger): clean old logs every hour

jonaro00:


oddgrd:
We can probably close this in favor of #1690, right?  We could clean at an interval in that one as well, if you think it's necessary, but I think it's fine without it.

PR #1424, opened 2023-11-25
fix(gateway): uppercase old ulids

jonaro00:
## How has this been tested? (if applicable)

Worked locally on two dummy rows.
Needs to test on unstable.

PR #1663, opened 2024-03-05
feat: migration for user_id

chesedo:
## Description of change
First phase to have proper user ids on our backend. This step is needed to be able to correctly link up the user ids in gw's project table.

Running the migrations locally

### DIFF
--- a/auth/src/lib.rs
+++ b/auth/src/lib.rs
@@ -10,6 +10,7 @@ use args::StartArgs;
 use shuttle_common::{claims::AccountTier, ApiKey};
 use sqlx::{migrate::Migrator, query, PgPool};
 use tracing::info;
+use user::User;
 
 use crate::api::serve;
 pub use api::ApiBuilder;
@@ -66,6 +67,22 @@ pub async fn pgpool_init(db_uri: &str) -> io::Result<PgPool> {
         .run(&pool)
         .await
         .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+
+    // Post-migration logic for 0003.
+    // This is done here to skip the need for postgres extensions.
+    let names: Vec<(String,)> =
+        sqlx::query_as("SELECT account_name FROM users WHERE user_id IS NULL")
+            .fetch_all(&pool)
+            .await
+            .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+    for (name,) in names {
+        sqlx::query("UPDATE users SET user_id = $1 WHERE account_name = $2")
+            .bind(User::new_user_id())
+            .bind(name)
+            .execute(&pool)
+            .await
+            .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+    }
 
     Ok(pool)
 }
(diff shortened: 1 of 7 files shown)


gateway/migrations/0007_user_ids.sql
```diff
@@ -0,0 +1,3 @@
+-- Add new user_id column for projects
+ALTER TABLE users
```
oddgrd, line 2: I think you meant this:
```suggestion
ALTER TABLE projects
```

oddgrd (approved):

jonaro00:
The users table after a test migration
```
 account_name |       key        | account_tier |             user_id             
--------------+------------------+--------------+---------------------------------
 admin        | dh9z58jttoes3qvt | admin        | user_01HR9T434GQ707YCP900FWJXVN
 test         | asdfasdfasdfasdf | basic        | user_01HR9T434KZH1JRQW8EHGYVDHE
 test2        | asdfasdfasdfasdg | basic        | user_01HR9T434NGMHXZETG1T46DAKY
```

jonaro00:
worked on staging

auth/src/lib.rs
```diff
@@ -67,5 +68,21 @@ pub async fn pgpool_init(db_uri: &str) -> io::Result<PgPool> {
...
+            .await
+            .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+    for (name,) in names {
+        sqlx::query("UPDATE users SET user_id = $1 WHERE account_name = $2")
+            .bind(User::new_user_id())
+            .bind(name)
+            .execute(&pool)
+            .await
+            .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+    }
```
oddgrd, line 85: What's the motivation for avoiding a Postgres extension? It seems we are adding a bit of complexity here, as well as running this query on every startup, rather than just once, which it would if it was done in a migration.

oddgrd: Oh, and doing it like this means we can't make the column NOT NULL, even though it technically is. It feels a bit awkward to me.

jonaro00: Since the IDs will fundamentally be strings, and will always be generated in our application, having a Ulid postgres extension is extra work with low benefit IMO.
Using the sqlite ulid extension led to the project id bug due to being able to generate an id of the correct format in two places https://github.com/shuttle-hq/shuttle/pull/1387. Also, that extension made them lowercase, which also added work #1424

Kazy: > What's the motivation for avoiding a Postgres extension? It seems we are adding a bit of complexity here, as well as running this query on every startup, rather than just once, which it would if it was done in a migration.

RDS doesn't support ULID extension, so that's kind of a no-go. 

A better solution in my opinion though would be to use something like that: https://github.com/geckoboard/pgulid/blob/master/pgulid.sql (Apache 2.0, I think we can use it ?), and generate IDs in Postgres.

chesedo: I like the suggestion by Jocelyn. Maybe we can use it instead?

chesedo: Call notes: we are going to remove this soon - ie this is only temp. So this will be fine for now.

Issue #1571, opened 2024-01-25
[Plugins]: Revamp the Shuttle Secrets plugin


jonaro00:
The current approach to Secrets is Shuttle-native.
Evolving the plugin could include support for `.env` files and allowing the secrets to be set as env variables in more stages of the Shuttle deployment.
Ideas are appreciated.

---
<sub>This issue is a part of the [Product Feature Poll](https://github.com/shuttle-hq/shuttle/issues/1546) ([all issues](https://github.com/shuttle-hq/shuttle/issues?q=label%3A%22Feature+Poll%22+sort%3Acreated-asc)). React with :+1: if you want this feature. Comment if you have suggestions related to this feature.</sub>

bavalpey:
I'm thoroughly happy with the simplicity of secrets.  It follows similar frameworks such as Cloudflare workers, except is even easier to work with.

Please don't change :(

DevKAVI:
Yeah I like the current implementation of using Secrets.toml as the main secret store. So there's no need to introduce yet another new file type to the workspace (.env).

bavalpey:
> The idea is mainly about making it easier to customize the approach, such as using a `.env` file if you want to

I think that having multiple ways to do this makes the feature more complicated, and makes reading other projects slightly more involved. There should really only be 2 files capable of setting secrets, one for local runs and one for deployments.
It would also be, well, difficult to differentiate between the deployment secrets and the dev secrets.  Having a .env file violates this.

> and the ability to automatically load all secrets into the environment, like `dotenv` crates do.

Now this is great.  Definitely would want it to be a deployment flag, though.

> Another improvement that is probably coming up is to have it built into the shuttle-runtime crate by default so that you don't even need to add a separate crate to use it.

This makes no difference to me.  The only reason I can see to not do this is to avoid pay the compilation cost for an unused module.  But I can't imagine the secrets module would have any real impact on compile times.


> There is also room to reduce the boilerplate required when getting secrets from the SecretStore. How does this sound?

Now this is something I would be very excited for.  It's a bit annoying having to pass around secrets, or throw them into a global static variable. Would love to see this.

jonaro00:
> It would also be, well, difficult to differentiate between the deployment secrets and the dev secrets. Having a .env file violates this.

Next.js uses `.env` + `.env.local` + a few other options. `Secrets[.dev].toml` is an echo of that pattern.

> a deployment flag

[...]

PR #1666, opened 2024-03-05
feat: Add support to parse a SecretStore as a struct using the config crate

JasterV:
## Description of change
The config crate is pretty standard nowadays on the Rust community to parse configuration files to a Rust type.
I didn't want to fetch one by one the secrets from the SecretStore manually since the config crate does that automatically and I figured out a way to pass the contents of the SecretStore as an `Environment` source to the `ConfigBuilder` as explained [here](https://docs.rs/config/0.14.0/config/struct.Environment.html#method.source).

I use this implementation locally on my production software.


JasterV:
I wanted to implement the `From` crate for any `T` that implement `DeserializeOwned` but Rust doesn't allow to implement a foreign trait to a foreign type so I guess I should create some sort of wrapper around T

JasterV:
Probably a proper thing to do could be to implement `Deserialize` on the `SecretsStore` to a NewType `SecretsConfig` that contains `T`

jonaro00:
Thanks for the suggestion! We will consider adding support for more opinionated ways of secrets/env loading like this one in a future update. In that case it can probably be built in without a feature flag. Will leave this PR open to keep the idea alive. More discussions can be found in #1571.

PR #1667, opened 2024-03-05
feat(cargo-shuttle): new template system for init

jonaro00:
## Description of change

New flow in the init command that allows browsing the full list of templates.

WIP preview:
image

updated tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,8 +47,9 @@ bollard = "0.15.0"
 bytes = "1.3.0"
 cargo_metadata = "0.18.1"
 chrono = { version = "0.4.23", default-features = false }
-colored = "2.0.0"
 clap = { version = "4.2.7", features = ["derive"] }
+colored = "2.0.0"
+comfy-table = "6.2.0"
 crossterm = "0.27.0"
 ctor = "0.2.5"
 dirs = "5.0.0"
@@ -59,6 +60,7 @@ futures = "0.3.27"
 headers = "0.3.8"
 home = "0.5.4"
 http = "0.2.8"
+http-body = "0.4.5"
 hyper = "0.14.23"
 # not great, but waiting for WebSocket changes to be merged
 hyper-reverse-proxy = { git = "https://github.com/chesedo/hyper-reverse-proxy", branch = "bug/host_header" }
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -108,5 +110,5 @@ ttl_cache = "0.5.1"
 ulid = "1.0.0"
 url = "2.4.0"
 uuid = "1.2.2"
-wiremock = "0.6.0-rc.3"
+wiremock = "0.6.0"
```
oddgrd, line 113: :partying_face:

cargo-shuttle/src/lib.rs
```diff
@@ -427,21 +437,150 @@ impl Shuttle {
...
-                    .default(0)
-                    .interact()?;
-                println!();
-                frameworks[index].template()
+                // /// Can be used during testing
+                // async fn get_schema() -> Result<TemplatesSchema> {
+                //     Ok(toml::from_str(include_str!(
+                //         "../../examples/templates.toml"
+                //     ))?)
+                // }
```
oddgrd, line 449: Looks like you forgot to remove this.

cargo-shuttle/src/lib.rs
```diff
@@ -427,21 +437,150 @@ impl Shuttle {
...
+                    let client = reqwest::Client::new();
+                    Ok(toml::from_str(
+                        &client
+                            .get(shuttle_common::constants::EXAMPLES_TEMPLATES_TOML)
+                            .send()
+                            .await?
+                            .text()
+                            .await?,
+                    )?)
+                }
```
oddgrd, line 460: It looks like this is just used once, can't we call it directly rather than having an inner function?

oddgrd (commented):
LGTM, just left a couple of minor comments.

cargo-shuttle/src/lib.rs
jonaro00: I intend to leave this here for anyone to use at will

cargo-shuttle/src/lib.rs
jonaro00: Having it like this makes it easier to hot-swap to another impl (see other comment)

jonaro00: It also groups the `?` error returns to be easily handled as one in `map_err`

oddgrd: I don't want to block the release, but it feels quite messy to me. Can we rather have a comment inside/over the function that you can use a template from the examples directly for testing? This function could also be external, even though it's only used here, since this parent function is very large already.

PR #1668, opened 2024-03-05
fix(gateway): proxy wait for service port to open

jonaro00:
## Description of change

Fixes (?) the `{"message":"Project returned invalid response","status_code":502}` error on the first hit on an idled project by checking the TCP port of the service before proxying the request.

An axum project that had a 2s sleep in the main function was deployed. (2s delay between "ready" and `bind`; simulates slow setup logic or other lag in that phase)
The wakeup proxy hits correctly waited and got 200.

### DIFF
--- a/gateway/src/api/latest.rs
+++ b/gateway/src/api/latest.rs
@@ -436,7 +436,10 @@ async fn route_project(
             .await?;
     }
 
-    let project = service.find_or_start_project(&project_name, sender).await?;
+    let project = service
+        .find_or_start_project(&project_name, sender)
+        .await?
+        .0;
     service
         .route(&project.state, &project_name, &scoped_user.user.name, req)
         .await
--- a/gateway/src/api/project_caller.rs
+++ b/gateway/src/api/project_caller.rs
@@ -34,7 +34,10 @@ impl ProjectCaller {
             service, sender, ..
         } = state;
         let project_name = scoped_user.scope;
-        let project = service.find_or_start_project(&project_name, sender).await?;
+        let project = service
+            .find_or_start_project(&project_name, sender)
+            .await?
+            .0;
 
         Ok(Self {
             project: project.state,
--- a/gateway/src/proxy.rs
+++ b/gateway/src/proxy.rs
@@ -30,7 +30,7 @@ use shuttle_common::models::project::ProjectName;
 use tokio::net::TcpSocket;
 use tokio::sync::mpsc::Sender;
 use tower_sanitize_path::SanitizePath;
-use tracing::{debug, debug_span, error, field, trace};
+use tracing::{debug, debug_span, error, field, trace, Instrument};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 
 use crate::acme::AcmeClient;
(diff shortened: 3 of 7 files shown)


gateway/src/api/project_caller.rs
```diff
@@ -34,7 +34,10 @@ impl ProjectCaller {
             service, sender, ..
         } = state;
         let project_name = scoped_user.scope;
-        let project = service.find_or_start_project(&project_name, sender).await?;
+        let project = service
+            .find_or_start_project(&project_name, sender)
+            .await?
+            .0;
```
iulianbarbu, line 40: We then need to use this directly when creating `Self {` below, right?

gateway/src/proxy.rs
```diff
@@ -97,10 +100,11 @@ async fn proxy(
...
-        let ip = state
+        let (proj, was_stopped) = state
             .gateway
             .find_or_start_project(&project_name, state.task_sender.clone())
-            .await?
+            .await?;
+        let ip = proj
             .state
             .target_ip()?
             .ok_or_else(|| Error::from_kind(ErrorKind::ProjectNotReady))?;
```
iulianbarbu, line 110: nit: I guess this can be another source of 502? It is probably less problematic than the deployer getting back up and starting the runtime.

gateway/src/proxy.rs
```diff
@@ -109,9 +113,33 @@ async fn proxy(
...
+                    if let Ok(socket) = TcpSocket::new_v4() {
+                        if socket.connect(addr).await.is_ok() {
+                            break;
+                        }
+                    }
+                    trace!("waiting for service port to open");
+                    // exponential backoff
+                    tokio::time::sleep(Duration::from_millis(ms)).await;
+                    ms *= 2;
+                }
```
iulianbarbu, line 135: Can we emit an event with how much we've waited for the port to open?

gateway/src/api/latest.rs
```diff
@@ -436,7 +436,10 @@ async fn route_project(
...
     }
 
-    let project = service.find_or_start_project(&project_name, sender).await?;
+    let project = service
+        .find_or_start_project(&project_name, sender)
+        .await?
+        .0;
     service
         .route(&project.state, &project_name, &scoped_user.user.name, req)
         .await
```
iulianbarbu, line 445: Again, aren't we supposed to use directly `project` for this function call?

gateway/src/api/project_caller.rs
jonaro00: no... ? wdym?

gateway/src/proxy.rs
jonaro00: Perhaps, but unrelated to this change

gateway/src/api/project_caller.rs
iulianbarbu: Oh, sorry. Missed the change you did in `find_or_start_project`.

gateway/src/proxy.rs
jonaro00: Added an event when we break the loop. The timing can be derived from the timestamps

iulianbarbu (approved):

gateway/src/proxy.rs
```diff
@@ -109,9 +113,37 @@ async fn proxy(
...
+            })
+            .await;
+        }
```
oddgrd, line 142: We can simplify the tracing a little bit, I quickly tested it locally and it's equivalent.
```suggestion
        if was_stopped {
            // wait until service has started and opens its port, give up after 10s
            span.in_scope(|| debug!("project waking up, checking service port"));
            let addr = SocketAddr::new(ip, DEPLOYER_SERVICE_HTTP_PORT);
            
            let _ = tokio::time::timeout(Duration::from_secs(10), async move {
                let mut ms = 5;
                loop {
                    if let Ok(socket) = TcpSocket::new_v4() {
                        if socket.connect(addr).await.is_ok() {
                                debug!("service port detected open");
                            break;
                        }
                    }
                    trace!("waiting for service port to open");
                    // exponential backoff
                    tokio::time::sleep(Duration::from_millis(ms)).await;
                    ms *= 2;
                }
            })
            .instrument(span.clone())
            .await;
        }
```

oddgrd (commented):
LGTM! Just one suggestion for the instrumentation.

PR #1674, opened 2024-03-08
refactor(auth, gateway): use user_id over account_name

jonaro00:
## Description of change

- Rename AccountName -> UserId but use a type alias instead of a newtype with no parsing logic (IMO user_id should not be stricly parsed to allow more flexibility)
- Rename header XShuttleAccountName -> XShuttleUserId
- Posting a user still uses account name (the only endpoint that takes it in), returns struct with a user id
- Admin call to gw still includes account_name
- Admin command logic not updated
- Auth's create user CLI command inserts user with blank name and the provided string as user_id
- New projects in gw get a blank account_name

### DIFF
--- a/auth/src/api/builder.rs
+++ b/auth/src/api/builder.rs
@@ -66,12 +66,12 @@ impl ApiBuilder {
             .route("/auth/key", get(convert_key))
             .route("/auth/refresh", post(refresh_token))
             .route("/public-key", get(get_public_key))
-            .route("/users/:account_name", get(get_user))
+            .route("/users/:user_id", get(get_user))
             .route("/users/:account_name/:account_tier", post(post_user))
             .route("/users/reset-api-key", put(put_user_reset_key))
-            .route("/users/:account_name/subscribe", post(post_subscription))
+            .route("/users/:user_id/subscribe", post(post_subscription))
             .route(
-                "/users/:account_name/subscribe/:subscription_id",
+                "/users/:user_id/subscribe/:subscription_id",
                 delete(delete_subscription),
             )
             .route_layer(from_extractor::<Metrics>())
@@ -79,7 +79,7 @@ impl ApiBuilder {
                 TraceLayer::new(|request| {
                     request_span!(
                         request,
-                        request.params.account_name = field::Empty,
+                        request.params.user_id = field::Empty,
                         request.params.account_tier = field::Empty,
                     )
                 })
--- a/auth/src/api/handlers.rs
+++ b/auth/src/api/handlers.rs
@@ -1,6 +1,6 @@
 use crate::{
     error::Error,
-    user::{AccountName, Admin, Key},
+    user::{Admin, Key},
 };
 use axum::{
     extract::{Path, State},
@@ -9,7 +9,7 @@ use axum::{
 use http::StatusCode;
 use shuttle_common::{
     claims::{AccountTier, Claim},
-    models::user::{self, SubscriptionRequest},
+    models::user::{self, SubscriptionRequest, UserId},
 };
 use tracing::instrument;
 
@@ -18,24 +18,25 @@ use super::{
     RouterState,
 };
 
-#[instrument(skip_all, fields(account.name = %account_name))]
+#[instrument(skip_all, fields(account.user_id = %user_id))]
 pub(crate) async fn get_user(
     _: Admin,
     State(user_manager): State<UserManagerState>,
-    Path(account_name): Path<AccountName>,
+    Path(user_id): Path<UserId>,
 ) -> Result<Json<user::Response>, Error> {
-    let user = user_manager.get_user(account_name).await?;
+    let user = user_manager.get_user(user_id).await?;
 
     Ok(Json(user.into()))
 }
 
-#[instrument(skip_all, fields(account.name = %account_name, account.tier = %account_tier))]
+#[instrument(skip_all, fields(account.account_name = %account_name, account.tier = %account_tier))]
 pub(crate) async fn post_user(
     _: Admin,
     State(user_manager): State<UserManagerState>,
-    Path((account_name, account_tier)): Path<(AccountName, AccountTier)>,
+    Path((account_name, account_tier)): Path<(String, AccountTier)>,
 ) -> Result<Json<user::Response>, Error> {
     let user = user_manager.create_user(account_name, account_tier).await?;
+    // TODO?: Add `user.id` to span's `account.user_id`?
 
     Ok(Json(user.into()))
 }
@@ -44,24 +45,19 @@ pub(crate) async fn put_user_reset_key(
     State(user_manager): State<UserManagerState>,
     key: Key,
 ) -> Result<(), Error> {
-    let account_name = user_manager.get_user_by_key(key.into()).await?.name;
+    let user_id = user_manager.get_user_by_key(key.into()).await?.id;
 
-    user_manager.reset_key(account_name).await
+    user_manager.reset_key(user_id).await
 }
 
 pub(crate) async fn post_subscription(
     _: Admin,
     State(user_manager): State<UserManagerState>,
-    Path(account_name): Path<AccountName>,
+    Path(user_id): Path<UserId>,
     payload: Json<SubscriptionRequest>,
 ) -> Result<(), Error> {
     user_manager
-        .insert_subscription(
-            &account_name,
-            &payload.id,
-            &payload.r#type,
-            payload.quantity,
-        )
+        .insert_subscription(&user_id, &payload.id, &payload.r#type, payload.quantity)
         .await?;
 
     Ok(())
@@ -70,10 +66,10 @@ pub(crate) async fn post_subscription(
 pub(crate) async fn delete_subscription(
     _: Admin,
     State(user_manager): State<UserManagerState>,
-    Path((account_name, subscription_id)): Path<(AccountName, String)>,
+    Path((user_id, subscription_id)): Path<(UserId, String)>,
 ) -> Result<(), Error> {
     user_manager
-        .delete_subscription(&account_name, &subscription_id)
+        .delete_subscription(&user_id, &subscription_id)
         .await?;
 
     Ok(())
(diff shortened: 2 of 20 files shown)


auth/migrations/0004_user_ids_part2.sql
chesedo, line 1: We should add the script to undo this in case this goes bad :sweat_smile:

auth/src/api/builder.rs
```diff
@@ -62,24 +62,24 @@ impl Default for ApiBuilder {
 impl ApiBuilder {
     pub fn new() -> Self {
         let router = Router::new()
-            .route("/", get(health_check))
+            .route("/", get(|| async move {})) // Health check: 200 OK
             .route("/auth/key", get(convert_key))
-            .route("/auth/refresh", post(refresh_token))
+            .route("/auth/refresh", post(|| async move {}))
```
chesedo, line 67: Does anything even call this?

auth/src/api/builder.rs
```diff
@@ -62,24 +62,24 @@ impl Default for ApiBuilder {
...
+                "/users/:user_id/subscribe/:subscription_id",
                 delete(delete_subscription),
             )
             .route_layer(from_extractor::<Metrics>())
             .layer(
                 TraceLayer::new(|request| {
                     request_span!(
                         request,
-                        request.params.account_name = field::Empty,
+                        request.params.user_id = field::Empty,
```
chesedo, line 82: Note: this will affect metrics

auth/src/api/handlers.rs
```diff
@@ -9,33 +9,34 @@ use axum::{
...
+    Path(user_id): Path<UserId>,
 ) -> Result<Json<user::Response>, Error> {
-    let user = user_manager.get_user(account_name).await?;
+    let user = user_manager.get_user(user_id).await?;
 
     Ok(Json(user.into()))
 }
 
-#[instrument(skip_all, fields(account.name = %account_name, account.tier = %account_tier))]
+#[instrument(skip_all, fields(account.account_name = %account_name, account.tier = %account_tier))]
```
chesedo, line 32: Double check this with Oddbjorn

auth/src/api/handlers.rs
```diff
@@ -9,33 +9,34 @@ use axum::{
...
 pub(crate) async fn post_user(
     _: Admin,
     State(user_manager): State<UserManagerState>,
-    Path((account_name, account_tier)): Path<(AccountName, AccountTier)>,
+    Path((account_name, account_tier)): Path<(String, AccountTier)>,
 ) -> Result<Json<user::Response>, Error> {
     let user = user_manager.create_user(account_name, account_tier).await?;
+    Span::current().record("account.user_id", &user.id);
```
chesedo, line 39: The span should have an empty field for this to work.

auth/src/args.rs
```diff
@@ -37,9 +38,9 @@ pub struct StartArgs {
 
 #[derive(clap::Args, Debug, Clone)]
 pub struct InitArgs {
-    /// Name of initial account to create
+    /// User id of account to create
     #[arg(long)]
-    pub name: String,
+    pub user_id: UserId,
```
chesedo, line 43: Wait, the init should still take in the account name right?

common/src/backends/headers.rs
```diff
@@ -33,15 +33,15 @@ impl Header for XShuttleAdminSecret {
     }
 }
 
-pub static X_SHUTTLE_ACCOUNT_NAME: HeaderName = HeaderName::from_static("x-shuttle-account-name");
+pub static X_SHUTTLE_USER_ID: HeaderName = HeaderName::from_static("x-shuttle-user-id");
```
chesedo, line 36: Changing the header name will break old deployers iirc. So we can't do that

gateway/src/api/latest.rs
```diff
@@ -368,10 +368,10 @@ async fn override_create_service(
     scoped_user: ScopedUser,
     req: Request<Body>,
 ) -> Result<Response<Body>, Error> {
-    let account_name = scoped_user.user.claim.sub.clone();
+    let user_id = scoped_user.user.claim.sub.clone();
     let posthog_client = state.posthog_client.clone();
     tokio::spawn(async move {
-        let event = async_posthog::Event::new("shuttle_api_start_deployment", &account_name);
+        let event = async_posthog::Event::new("shuttle_api_start_deployment", &user_id);
```
chesedo, line 374: Note: will affect metrics

gateway/src/api/latest.rs
```diff
@@ -869,9 +869,9 @@ impl ApiBuilder {
             TraceLayer::new(|request| {
                 request_span!(
                     request,
-                    account.name = field::Empty,
+                    account.user_id = field::Empty,
```
chesedo, line 872: Note: will affect metrics

gateway/src/project.rs
```diff
@@ -1497,7 +1497,7 @@ impl Service {
                 .method(Method::PUT)
                 .uri(uri)
                 .header(AUTHORIZATION, format!("Bearer {}", jwt))
-                .header(X_SHUTTLE_ACCOUNT_NAME.clone(), "gateway")
+                .header(X_SHUTTLE_USER_ID.clone(), "gateway")
```
chesedo, line 1500: The user id for gateway won't be gateway right?

gateway/src/project.rs
```diff
@@ -1518,7 +1518,7 @@ impl Service {
         let req = Request::builder()
             .uri(uri)
             .header(AUTHORIZATION, format!("Bearer {}", jwt))
-            .header(X_SHUTTLE_ACCOUNT_NAME.clone(), "gateway")
+            .header(X_SHUTTLE_USER_ID.clone(), "gateway")
```
chesedo, line 1521: Same here

chesedo (commented):
Seems good. The metrics notes can be ignored - I just want to make sure we update them all

auth/src/api/handlers.rs
jonaro00: Messed this one up. It should remain unchanged (not sure if that field will be of any use tho)

auth/src/args.rs
jonaro00: The account name should be seen as the auth0 name, so the custom names that some accounts use should ideally be in user_id as `user_mycustomname` (prefix being just a convention and not a restriction)

gateway/src/project.rs
jonaro00: For those special accounts, we should also update their user id to be non-standard.

auth/src/api/builder.rs
oddgrd: I believe this is a remnant of the skeleton we created when we started working on the auth service, we never ended up using it. I cannot see any calls to this endpoint in our Datadog history either.

common/src/backends/headers.rs
jonaro00: This one is only used between auth and gateway, so we can

auth/migrations/0004_user_ids_part2.sql
jonaro00: A rolled back auth service can't use a down script (it did not exist in that version).
If the up migration fails, it will be aborted and fall back to previous version, in which case we need to roll back gw to match that.
A test on staging should give high confidence.

auth/migrations/0004_user_ids_part2.sql
```diff
@@ -0,0 +1,27 @@
+-- All rows should have user_ids at this point
```
oddgrd, line 1: Nit: should we have a comment about how this was done in the application? So someone who reads the migrations down the line won't be confused.

deployer/src/handlers/mod.rs
```diff
@@ -430,11 +427,6 @@ pub async fn delete_deployment(
     }
 }
 
-/// Deprecated.
-/// Now always starts the last running deployment on start up / wake up.
-/// Kept around for compatibility.
-pub async fn start_deployment() {}
-
```
oddgrd, line 437: I'm out of the loop here, why can we remove this now?

common/src/models/user.rs
```diff
@@ -2,9 +2,12 @@ use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use strum::{Display, EnumString};
 
+pub type UserId = String;
```
oddgrd, line 5: This type alias is just for readability, right? Should we add a doc comment explaining the format of the ID (`user_<ULID>`).

oddgrd (commented):
The code changes LGTM, but we should figure out exactly how our metrics will be affected by these changes, and what we will need to do after the release.

auth/src/args.rs
chesedo: Ok, sounds good :+1:

auth/src/lib.rs
```diff
@@ -38,17 +38,18 @@ pub async fn init(pool: PgPool, args: InitArgs, tier: AccountTier) -> io::Result
         None => ApiKey::generate(),
     };
 
-    query("INSERT INTO users (account_name, key, account_tier) VALUES ($1, $2, $3)")
-        .bind(&args.name)
+    query("INSERT INTO users (account_name, key, account_tier, user_id) VALUES ($1, $2, $3, $4)")
+        .bind("")
```
chesedo, line 42: Will a unique constraint fail if we end up inserting two machine accounts with no account_name? Maybe this should be `NULL` instead?

chesedo (commented):
Quick extra question

common/src/backends/headers.rs
jonaro00: Fixed in 1edf4073

auth/src/lib.rs
jonaro00: There is no UNIQUE constraint on account_name remaining in my local db after the migrations, but it is still NOT NULL, hence this.

.circleci/config.yml
```diff
@@ -731,14 +731,14 @@ workflows:
...
+          #   branches:
+          #     only: main
       - approve-build-and-push-images-unstable:
           type: approval
-          filters:
-            branches:
-              only: main
+          # filters:
+          #   branches:
+          #     only: main
```
jonaro00, line 741: reminder

PR #1677, opened 2024-03-12
gateway: delete project restarts them first if oudated

iulianbarbu:
## Description of change

Check the deployer version based on its container inspect response and restart the project before trying to delete it.
We assume it doesn't matter whether the project will be running afterward, or if the deployment will crash, given the
user wants to delete it. What's important is for the deployer to be able to handle the delete request.

N/A

### DIFF
--- a/gateway/src/api/latest.rs
+++ b/gateway/src/api/latest.rs
@@ -267,12 +267,26 @@ async fn delete_project(
 
     let project_name = scoped_user.scope.clone();
     let project = state.service.find_project(&project_name).await?;
+
     let project_id =
         Ulid::from_string(&project.project_id).expect("stored project id to be a valid ULID");
 
-    // Try to startup destroyed or errored projects
+    // Try to startup destroyed, errored or outdated projects
     let project_deletable = project.state.is_ready() || project.state.is_stopped();
-    if !(project_deletable) {
+    let lowest_restartable_version = semver::Version::new(0, 39, 0);
+    let version = project
+        .state
+        .container()
+        .and_then(|container_inspect_response| {
+            container_inspect_response.image.and_then(|inner| {
+                inner
+                    .strip_prefix("public.ecr.aws/shuttle/deployer:v")
+                    .and_then(|x| x.parse::<semver::Version>().ok())
+            })
+        })
+        .unwrap_or(lowest_restartable_version.clone());
+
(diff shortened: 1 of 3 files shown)


gateway/src/api/latest.rs
```diff
@@ -267,12 +267,26 @@ async fn delete_project(
...
     let project = state.service.find_project(&project_name).await?;
+
     let project_id =
         Ulid::from_string(&project.project_id).expect("stored project id to be a valid ULID");
 
-    // Try to startup destroyed or errored projects
+    // Try to startup destroyed, errored or outdated projects
     let project_deletable = project.state.is_ready() || project.state.is_stopped();
-    if !(project_deletable) {
+    let lowest_restartable_version = semver::Version::new(0, 39, 0);
```
jonaro00, line 276: We should just compare it to the current version of gateway IMO

iulianbarbu: So restart everything less than 0.41 you mean?

jonaro00: Yeah everything less than `env!("CARGO_PKG_VERSION")`

iulianbarbu: Yeh, good point. Given projects will be deleted, it doesn't matter if we restart it and it turns out the deployment will crash for any reason (given the app is supposed to be stopped and deleted with all its resources).

iulianbarbu: Addressed here: https://github.com/shuttle-hq/shuttle/pull/1677/commits/f767676db566e550bef54d7e09b0d3f7528c1df0. What do you think?

gateway/src/api/latest.rs
```diff
@@ -284,9 +287,16 @@ async fn delete_project(
                     .and_then(|x| x.parse::<semver::Version>().ok())
             })
         })
-        .unwrap_or(lowest_restartable_version.clone());
-
-    if !project_deletable || version.le(&lowest_restartable_version) {
+        // Defaulting to a version that introduced a breaking change.
+        // This was the last one that introduced it at the present
+        // moment.
+        .unwrap_or(semver::Version::new(0, 39, 0));
```
jonaro00, line 293: non-blocking: The image tag failing to be read should always lead to a restart, so this might as well be (0, 0, 0)

jonaro00 (approved):

PR #1682, opened 2024-03-15
feat: fall back to finding Secrets.toml in workspace root

jonaro00:
## Description of change

We look for Secret.toml in the *crate* directory, making placing it in the workspace root a common papercut.
This allows you to do both, with the crate dir taking prio.

Tested with local run and local deployment

### DIFF
--- a/deployer/src/deployment/queue.rs
+++ b/deployer/src/deployment/queue.rs
@@ -200,7 +200,7 @@ impl Queued {
         let built_service = build_deployment(&project_path, tx.clone()).await?;
 
         // Get the Secrets.toml from the shuttle service in the workspace.
-        let secrets = get_secrets(built_service.crate_directory()).await?;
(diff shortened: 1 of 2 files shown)


deployer/src/deployment/queue.rs
```diff
@@ -539,21 +548,4 @@ ff0e55bda1ff01000000000000000000e0079c01ff12a55500280000",
             "barfoo"
         );
     }
-
-    #[tokio::test]
-    async fn get_secrets() {
```
oddgrd, line 544: Hmm, mind clarifying why we can remove this?

oddgrd (approved):
Good thinking! I have one question about the test removal, otherwise LGTM!

deployer/src/deployment/queue.rs
jonaro00: Thought it would be more of an integration test at this point, but no. Added it back.

PR #1691, opened 2024-03-19
feat(cargo-shuttle): remove retry client, add version header

jonaro00:
## Description of change

Our APIs should be stable enough where the retry logic is no longer relevant.
Adds a version header that can be used for better version checking later on.

against netcat


oddgrd (approved):
LGTM!

PR #1694, opened 2024-03-20
chore: update libsql dep

Jaco-Minnaar:
## Description of change
I bumped the dependency on `libsql` in the `shuttle-turso` create to `v0.3.1`. I also refactored the code to replace the deprecated `Database::open_remote` and `Database::open` with the new `Builder` API.

I ran the unit tests on the crate. If I need to do anything else, please let me know!


jonaro00 (approved):

Jaco-Minnaar:
> Awesome, thank you @Jaco-Minnaar! It would be great to know what we need to change in the example as well, even if it's just bumping the version. Would you mind taking a look at that (https://github.com/shuttle-hq/shuttle-examples/tree/main/axum/turso)? A PR would be most welcome too, if you have time!

I'll take a quick look in the morning 👍

PR #1695, opened 2024-03-20
feat: update services api client

jonaro00:
## Description of change

- use reqwest for a simpler api + more convenience


common/Cargo.toml
```diff
@@ -57,11 +57,12 @@ backend = [
     "axum/matched-path",
     "axum/json",
     "claims",
-    "hyper/client",
+    "hyper",
```
chesedo, line 60: Should this remove `hyper` instead?

common/src/backends/client/mod.rs
```diff
@@ -17,94 +18,145 @@ pub use resource_recorder::ResourceDal;
...
 #[derive(Clone)]
 pub struct ServicesApiClient {
-    client: Client<HttpConnector>,
+    client: Client,
     base: Uri,
 }
 
 impl ServicesApiClient {
-    fn new(base: Uri) -> Self {
+    fn _builder() -> ClientBuilder {
```
chesedo, line 39: Why the underscore, this function is not `pub`?

common/src/backends/client/mod.rs
```diff
@@ -17,94 +18,145 @@ pub use resource_recorder::ResourceDal;
...
-        let bytes = body::to_bytes(resp.into_body()).await?;
-
-        Ok(bytes)
+        Ok(resp)
     }
 }
 
+pub fn header_map_with_bearer(token: &str) -> HeaderMap {
+    let mut h = HeaderMap::new();
+    h.append(
```
chesedo, line 151: Was the `typed_insert` possible?

deployer/src/deployment/gateway_client.rs
```diff
@@ -17,17 +16,11 @@ pub trait BuildQueueClient: Clone + Send + Sync + 'static {
...
         let body = models::stats::LoadRequest { id: deployment_id };
         let load: models::stats::LoadResponse = self
-            .public_client()
-            .request(
-                Method::POST,
-                "stats/load",
-                Some(body),
-                None::<Authorization<Bearer>>,
-            )
+            .request(Method::POST, "stats/load", Some(body), None)
```
chesedo, line 23: These can use the helpers right?

chesedo (commented):
Looks good! Just a few small  nits

common/Cargo.toml
jonaro00: it's used in other places in backend feature

PR #1696, opened 2024-03-20
feat: permit client skeleton

jonaro00:


common/src/backends/client/permit.rs
```diff
@@ -0,0 +1,383 @@
...
+    pdp: ServicesApiClient,
+    /// The project id used in API paths
+    proj_id: String,
+    /// The environment id used in API paths
+    env_id: String,
+}
+
+impl Permit {
+    // "https://api.eu-central-1.permit.io".parse().unwrap(),
+    // "http://localhost:7000".parse().unwrap(),
```
chesedo, line 23: Comments? :confused:

common/src/backends/client/permit.rs
```diff
@@ -0,0 +1,383 @@
...
+        Self {
+            api: ServicesApiClient::new_with_bearer(api_uri, api_key),
+            pdp: ServicesApiClient::new(pdp_uri),
+            proj_id,
+            env_id,
+        }
+    }
+
+    fn endpoint_facts(&self) -> String {
+        format!("/v2/facts/{}/{}", self.proj_id, self.env_id)
```
chesedo, line 34: Maybe put this in the constructor so that we don't remake this the whole time

chesedo (approved):
Seems good. Two nit items

PR #1697, opened 2024-03-20
feat: permit pdp in docker stack

jonaro00:
todo:
- add args that pass api url and key and pdp url to auth+gateway
- test on staging?


Makefile
```diff
@@ -42,6 +42,10 @@ MONGO_INITDB_ROOT_PASSWORD?=password
 STRIPE_SECRET_KEY?=""
 AUTH_JWTSIGNING_PRIVATE_KEY?=""
 
+PERMIT_API?=https://api.eu-central-1.permit.io
+PERMIT_API_KEY?=""
+PERMIT_DEBUG?="False"
```
chesedo, line 47: I think this one can just be a dev compose override

chesedo (approved):
Just one nit

PR #1698, opened 2024-03-21
refactor: shuttle-common/backend -> shuttle-backends

jonaro00:


common/src/limits.rs
```diff
@@ -8,7 +8,7 @@ use crate::{
 #[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq)]
 pub struct Limits {
```
chesedo, line 9: Only the backends need to know about this right?

deployer/src/handlers/mod.rs
```diff
@@ -56,15 +54,16 @@ pub struct PaginationDetails {
 #[derive(Clone)]
 pub struct RouterBuilder {
     router: Router,
-    _project_name: ProjectName,
+    // might be used for tracing instruments?
+    _project_name: String,
```
chesedo, line 58: No idea. Should be better to delete

proto/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "shuttle-proto"
-version.workspace = true
+version = "0.42.0"
```
chesedo, line 3: Is this expected?

chesedo (commented):
A few nits

proto/Cargo.toml
jonaro00: I made it follow the same pattern as the rest of the crates. No real world difference

deployer/src/handlers/mod.rs
jonaro00: Don't want to risk breaking anything with this PR. No risk in keeping it.

iulianbarbu (approved):
LGTM!

deployer/src/handlers/mod.rs
chesedo: Fair enough

proto/Cargo.toml
chesedo: Oky

common/src/limits.rs
jonaro00: proto uses claims, not limits, but the limits file is part of the claims feature

common/src/models/error.rs
```diff
@@ -6,6 +6,13 @@ use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tracing::{error, warn};
 
+#[cfg(feature = "axum")]
+impl axum::response::IntoResponse for ApiError {
+    fn into_response(self) -> axum::response::Response {
+        (self.status(), axum::Json(self)).into_response()
+    }
+}
```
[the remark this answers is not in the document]
jonaro00: Yes, but the trait impl must exist in the same crate, and the model is shared with the "frontends". This is a simple compromise, but can be solved in other ways.

PR #1699, opened 2024-03-21
feat: generated Permit client, project permission logic

jonaro00:
Adds Permit logic to the project create+delete actions. This will be released without taking control of permission checking to allow a background migration between releases.

Things for upcoming PRs:
- Implement organizations (commented parts) in the permit client + tests
- Command in gateway to trigger a sync of project data
- Let Permit take over the Permission checks in ScopedUser
- Add CORS + update Console

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -649,6 +649,10 @@ workflows:
                 - shuttle-common
       - test-workspace-member-with-integration:
           name: << matrix.crate >>
+          filters:
+            branches:
+              ignore:
+                - /pull\/.*/
           matrix:
             alias: test-workspace-member-with-integration-medium
             parameters:
@@ -680,6 +684,10 @@ workflows:
                 - shuttle-deployer
       - test-workspace-member-with-integration-docker:
           name: << matrix.crate >> with docker
+          filters:
+            branches:
+              ignore:
+                - /pull\/.*/
           matrix:
             alias: test-workspace-member-with-integration-docker-medium
             parameters:
@@ -687,6 +695,7 @@ workflows:
                 - medium
               crate:
                 - shuttle-auth
+                - shuttle-backends
                 - shuttle-provisioner
                 - shuttle-logger
       - test-workspace-member-with-integration-docker:
--- a/auth/src/lib.rs
+++ b/auth/src/lib.rs
@@ -4,9 +4,8 @@ mod error;
 mod secrets;
 mod user;
 
-use std::io;
-
-use args::{StartArgs, SyncArgs};
+use anyhow::Result;
+use args::{CopyPermitEnvArgs, StartArgs, SyncArgs};
 use shuttle_backends::client::{permit, PermissionsDal};
 use shuttle_common::{claims::AccountTier, ApiKey};
 use sqlx::{migrate::Migrator, query, PgPool};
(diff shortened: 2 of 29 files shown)


.circleci/config.yml
```diff
@@ -649,6 +649,10 @@ workflows:
                 - shuttle-common
       - test-workspace-member-with-integration:
           name: << matrix.crate >>
+          filters:
+            branches:
+              ignore:
+                - /pull\/.*/
```
iulianbarbu, line 655: What is this for?

auth/src/lib.rs
```diff
@@ -81,7 +83,9 @@ pub async fn sync(pool: PgPool, args: SyncArgs) -> io::Result<()> {
                     }
                 }
             }
-            Err(shuttle_backends::client::Error::RequestError(StatusCode::NOT_FOUND)) => {
+            Err(_) => {
+                // FIXME: Make the error type better so that this is only done on 404s
```
jonaro00, line 77: This is fixed in #1705

gateway/src/api/latest.rs
```diff
@@ -188,6 +188,12 @@ async fn create_project(
         .await?;
     let idle_minutes = project.state.idle_minutes();
 
+    service
+        .permit_client
+        .create_project(&id, &project.project_id)
+        .await
+        .map_err(|_| Error::from(ErrorKind::Internal))?;
```
iulianbarbu, line 195: Should we delete it from the internal state in case this fails?

docker-compose.yml
```diff
@@ -280,7 +280,7 @@ services:
     image: docker.io/permitio/pdp-v2:0.2.37
     restart: always
     environment:
-      - PDP_CONTROL_PLANE=${PERMIT_API}
+      - PDP_CONTROL_PLANE=https://api.eu-central-1.permit.io
```
iulianbarbu, line 283: nit: I would still keep this behind an env variable. It is not necessarily a secret, but it doesn't need to be hardcoded either.

docker-compose.dev.yml
```diff
@@ -68,7 +68,9 @@ services:
       - "--stripe-secret-key=${STRIPE_SECRET_KEY}"
       # used only for local development
       - "--jwt-signing-private-key=LS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCk1DNENBUUF3QlFZREsyVndCQ0lFSUR5V0ZFYzhKYm05NnA0ZGNLTEwvQWNvVUVsbUF0MVVKSTU4WTc4d1FpWk4KLS0tLS1FTkQgUFJJVkFURSBLRVktLS0tLQo="
-      - "--permit-api=https://api.eu-central-1.permit.io"
+      - "--permit-api-uri=https://api.eu-central-1.permit.io"
```
iulianbarbu, line 71: I imagine this can differ based on how a Permit account is configured. Makes me think that it is better to hide it behind an env var.

backends/src/client/permit.rs
```diff
@@ -1,640 +1,545 @@
...
-        self.api
-            .post(
-                &format!("{}/users", self.facts),
-                json!({"key": user_id}),
-                None,
-            )
-            .await
+// Helpers for trait methods
+impl Client {
+    // /// Assigns a user to an org directly without creating the org first
```
iulianbarbu, line 224: Should the commented code be removed?

iulianbarbu (commented):
The permit client changes and some of the tests are heavy to go through in this review. It is probably best to be reviewed by @chesedo who has more context, but apart from this, left a few qs for now.

backends/src/client/permit.rs
jonaro00: No, will be added later

PR #1702, opened 2024-03-25
fix(codegen): use full path for std types

jonaro00:
## Description of change

- Fixes compilation error if `Result` or `Vec` is being shadowed in the user namespace.
- Remove old stuff + update docs, names, comments, tests.

CI


oddgrd (approved):
LGTM!

PR #1705, opened 2024-03-27
feat: gateway command to sync permit

jonaro00:
## Description of change

Improves the error model from permit and adds a project sync command

Tested against permit's dev env from local stack

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -356,6 +356,9 @@ jobs:
       gateway-admin-key:
         description: "Admin API key that authorizes gateway requests to auth service, for key to jwt conversion."
         type: string
+      permit-api-key:
+        description: "Permit.io API key for the Permit environment that matches the current ${SHUTTLE_ENV}."
+        type: string
     steps:
       - checkout
       - set-git-tag
@@ -383,6 +386,7 @@ jobs:
             AUTH_JWTSIGNING_PRIVATE_KEY=${<< parameters.jwt-signing-private-key >>} \
             CONTROL_DB_POSTGRES_URI=${<< parameters.control-db-postgres-uri >>} \
             GATEWAY_ADMIN_KEY=${<< parameters.gateway-admin-key >>} \
+            PERMIT_API_KEY=${<< parameters.permit-api-key >>} \
             make deploy
       - when:
           condition:
@@ -748,6 +752,7 @@ workflows:
           jwt-signing-private-key: DEV_AUTH_JWTSIGNING_PRIVATE_KEY
           control-db-postgres-uri: DEV_CONTROL_DB_POSTGRES_URI
           gateway-admin-key: DEV_GATEWAY_ADMIN_KEY
+          permit-api-key: PERMIT_API_KEY # TODO: switch to staging key
           requires:
             - build-and-push-unstable
             - approve-deploy-images-unstable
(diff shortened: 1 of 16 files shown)


.circleci/config.yml
```diff
@@ -748,6 +752,7 @@ workflows:
           jwt-signing-private-key: DEV_AUTH_JWTSIGNING_PRIVATE_KEY
           control-db-postgres-uri: DEV_CONTROL_DB_POSTGRES_URI
           gateway-admin-key: DEV_GATEWAY_ADMIN_KEY
+          permit-api-key: PERMIT_API_KEY # TODO: switch to staging key
```
iulianbarbu, line 755: what do you mean by staging key? Is that a thing you need to do in CircleCI?

auth/src/lib.rs
```diff
@@ -54,37 +58,38 @@ pub async fn sync(pool: PgPool, args: SyncArgs) -> Result<()> {
...
+                ..
+            })) => {
+                // Add users that are not in permit
                 println!("creating user: {}", user.id);
 
-                // Add users that are not in permit
                 permit_client.new_user(&user.id).await?;
-
-                if user.account_tier == AccountTier::Pro {
+                if user.account_tier.as_permit_account_tier() == AccountTier::Pro {
```
iulianbarbu, line 86: why is this only for pro?

gateway/src/lib.rs
```diff
@@ -571,13 +571,15 @@ pub mod tests {
                     cch_container_limit: 1,
                     soft_container_limit: 2,
                     hard_container_limit: 3,
+
+                    // Allow access to the auth on the host
+                    extra_hosts: vec!["host.docker.internal:host-gateway".to_string()],
```
iulianbarbu, line 576: I am just seeing this, why do we need it?

jonaro00: Don't know, I just moved it

auth/src/lib.rs
jonaro00: Only basic and pro tiers exist in Permit. (other ones don't affect permissions)

backends/src/client/permit.rs
```diff
@@ -143,6 +147,7 @@ impl PermissionsDal for Client {
     }
 
     async fn create_project(&self, user_id: &str, project_id: &str) -> Result<(), Error> {
+        // TODO?: Ignore error if 409?
```
chesedo, line 150: todo

backends/src/client/permit.rs
```diff
@@ -1,10 +1,13 @@
-use anyhow::Error;
+use std::fmt::{Debug, Display};
+
+// use anyhow::Error;
```
chesedo, line 3: Comment?

gateway/src/lib.rs
chesedo: Fyi, this is for tests. Since the gateway is not started in a container, auth and the other service fail to reach it in the tests. So this makes docker resolve the `host-gateway` lookups to the host machine correctly.

gateway/src/main.rs
```diff
@@ -63,34 +55,70 @@ async fn main() -> io::Result<()> {
...
+            .fetch_all(&db)
+            .await
+            .unwrap();
+
+    for (uid, pid) in projects {
+        match client.get_user_projects(&uid).await {
+            Ok(projs) => {
+                if !projs
+                    .iter()
+                    .any(|p| p.resource.as_ref().unwrap().key == pid)
```
chesedo, line 88: Wait, we get all the projects from permit to check if just one is correct in each loop? Maybe we should as permit for just the one project. Or group the DB results first.

chesedo (commented):
Left some notes :smile:

PR #1706, opened 2024-03-27
feat(gateway): compare permission checks to permit

jonaro00:
## Description of change

Compare project access to Permit project access.
Use Permit for project list.
Remove project list pagination.

Staging

### DIFF
--- a/gateway/src/auth.rs
+++ b/gateway/src/auth.rs
@@ -87,7 +87,11 @@ where
             || user.claim.is_deployer()
             || service
                 .permit_client
-                .allowed(&user.id, &scope.to_string(), "develop") // TODO?: make this configurable per endpoint?
+                .allowed(
+                    &user.id,
+                    &service.find_project_by_name(&scope).await?.id,
+                    "develop",
+                ) // TODO?: make this configurable per endpoint?
                 .await
                 .map_err(|_| {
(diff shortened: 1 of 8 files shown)


gateway/src/auth.rs
```diff
@@ -32,18 +32,15 @@ where
 {
     type Rejection = Error;
 
-    async fn from_request_parts(parts: &mut Parts, state: &S) -> Result<Self, Self::Rejection> {
+    async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
         let claim = parts.extensions.get::<Claim>().ok_or(ErrorKind::Internal)?;
         let user_id = claim.sub.clone();
```
chesedo, line 37: Does it make sense to still have a `User` struct? We can just implement `FromRequestParts` for the claim right?

chesedo (commented):
Small nit. If all tests are still passing then I don't see the need for updating any more tests.

gateway/src/auth.rs
jonaro00: *only traits defined in the current crate can be implemented for a type parameter*
Claim can technically be moved to backends (it seems), but I'd skip it for now.

chesedo (approved):
LGTM

PR #1709, opened 2024-04-02
feat: merge runtime updates in main ecs branch

oddgrd:
## Description of change

This adds a healthcheck to the runtime, as well as some logic we can use to ensure the runtime shuts down if it doesn't start the service within a certain timeframe. The logic to make the runtime handle failure well will need some further iteration in the future, since there are a few ways having it as a server that is called by the runner sidecar can lead to failures. 

This also changes the runtime to start on the unspecified IP, which is necessary in the new architecture, but the current platform should still use localhost.

These changes have been tested on the headroom stack on their own, alongside changes in ecs-shuttle and shuttle-aws.


iulianbarbu (approved):
lgtm!

PR #1721, opened 2024-04-04
refactor: only use permit to check project permissions

chesedo:
## Description of change
We saw no errors in the last 24 hours nor any results different from our old internal checks. So it is safe to switch fully to permit.

The compiler :grin:

### DIFF
--- a/gateway/src/auth.rs
+++ b/gateway/src/auth.rs
@@ -82,39 +82,21 @@ where
 
         let RouterState { service, .. } = RouterState::from_ref(state);
 
-        let has_bypass = user.claim.is_admin() || user.claim.is_deployer();
-
-        let allowed = has_bypass
-            || {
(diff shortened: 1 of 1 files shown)


gateway/src/auth.rs
```diff
@@ -82,39 +82,21 @@ where
...
+                    &user.id,
+                    &service.find_project_by_name(&scope).await?.id,
+                    "develop", // TODO: make this configurable per endpoint?
+                )
+                .await
+                .map_err(|_| {
+                    error!("failed to check Permit permission");
+                    // Error::from_kind(ErrorKind::Internal)
+                })
+                .unwrap_or_default();
```
jonaro00, line 99: This error should be put back in with a `?` instead of unwrap.

jonaro00 (approved):

PR #1725, opened 2024-04-05
feat(admin): change project owner

jonaro00:
## Description of change

Admin command to change project owner.
Some improvements to admin internals and permit client errors.

- On local stack with permit dev env

### DIFF
--- a/admin/src/client.rs
+++ b/admin/src/client.rs
@@ -1,4 +1,4 @@
-use anyhow::{Context, Result};
+use anyhow::{bail, Context, Result};
 use bytes::Bytes;
 use serde::{de::DeserializeOwned, Serialize};
 use shuttle_common::models::{admin::ProjectResponse, stats, ToJson};
@@ -141,12 +141,16 @@ impl Client {
     }
 
     async fn get_raw(&self, path: &str) -> Result<Bytes> {
-        reqwest::Client::new()
+        let res = reqwest::Client::new()
             .get(format!("{}{}", self.api_url, path))
             .bearer_auth(&self.api_key)
             .send()
             .await
-            .context("making request")?
+            .context("making request")?;
+        if !res.status().is_success() {
+            bail!("API call returned non-2xx: {:?}", res);
+        }
+        res
             .bytes()
             .await
             .context("getting response body")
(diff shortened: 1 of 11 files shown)


admin/src/client.rs
```diff
@@ -130,15 +140,19 @@ impl Client {
             .context("failed to extract json body from delete response")
     }
 
-    async fn get<R: DeserializeOwned>(&self, path: &str) -> Result<R> {
+    async fn get_raw(&self, path: &str) -> Result<Bytes> {
```
chesedo, line 143: Hmm, this can just switch to that service client we made recently right?

backends/src/client/permit.rs
```diff
@@ -572,6 +572,7 @@ pub enum Error {
     #[error("response error: {0}")]
     ResponseError(ResponseContent),
 }
+pub type Result<T> = std::result::Result<T, Error>;
```
chesedo, line 575: This should be closer to the top of the file. And maybe even in the service client file.

gateway/src/service.rs
```diff
@@ -492,6 +492,30 @@ impl GatewayService {
...
+        let (project_id,) = query_as::<_, (String,)>(
+            "UPDATE projects SET user_id = ?1 WHERE project_name = ?2 RETURNING project_id",
+        )
+        .bind(new_user_id)
+        .bind(project_name)
+        .fetch_one(&mut *tr)
+        .await?;
+
+        self.permit_client.delete_project(&project_id).await?;
+        self.permit_client
```
chesedo, line 510: There should be an unassign & assign role / tuple thing

chesedo (commented):
A few nits

admin/src/client.rs
jonaro00: Yeah I will rewrite admin to re-use the rest of the codebase instead of being its own thing. But not now 🙈

backends/src/test_utils/gateway.rs
```diff
@@ -13,7 +13,7 @@ use wiremock::{
 };
 
 use crate::client::{
-    permit::{Error, Organization},
+    permit::{Organization, Result},
```
chesedo, line 16: Having `Result` come from permit is weird

chesedo (commented):
One small nit :sweat_smile:

backends/src/test_utils/gateway.rs
jonaro00: skill issue :D

PR #1731, opened 2024-04-09
cargo-shuttle: add platform version switches

iulianbarbu:
## Description of change

Added a `platform_version` cargo-shuttle argument which can be set as environment as well. Up to this point we can use it against the new platform for:
- project start
- project status
- project list
- project delete

Tested locally.

## Notes

`cargo shuttle deploy` is not supported yet, but that will happen in a follow up PR.

### DIFF
--- a/cargo-shuttle/src/args.rs
+++ b/cargo-shuttle/src/args.rs
@@ -42,6 +42,9 @@ pub struct ShuttleArgs {
 
     #[command(subcommand)]
     pub cmd: Command,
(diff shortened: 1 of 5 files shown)


Kazy:
Wouldn't it be better to put all of that behind a trait, and just switch the implementation ? With anything that can be shared being in standalone functions ? It's a larger rework, but this would make things clearer rather than if/else and v1/v2.

PR #1732, opened 2024-04-10
feat(cargo-shuttle): beta flag, remove project list pagination logic

jonaro00:
Supersedes #1731

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -75,7 +75,7 @@ use crate::args::{
     DeployArgs, DeploymentCommand, InitArgs, LoginArgs, LogoutArgs, LogsArgs, ProjectCommand,
     ProjectStartArgs, ResourceCommand, TemplateLocation,
 };
-use crate::client::ShuttleAPIClient;
+use crate::client::ShuttleApiClient;
 use crate::provisioner_server::LocalProvisioner;
 
 const VERSION: &str = env!("CARGO_PKG_VERSION");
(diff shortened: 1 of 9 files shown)


Kazy (commented):
Curious how this will scale if we start to have more modification from the current CLI. I was thinking of having two different implementation of the API instead (`ShuttleApiClient` and `ShuttleBetaApiClient` basically). This would also make it clearer what isn't implemented vs the other one.

jonaro00:
Decided to go with a boolean property like this since the amount of differences thus far is very small.
The pattern becomes:
- If only the API call is different, change behaviour in Client (can also refactor to `<T>` if the return model is different),
- If more of the command is different, change behaviour in the `Shuttle` command method,
- If lots changed, add new methods where needed.

Kazy:
> Decided to go with a boolean property like this since the amount of differences thus far is very small. The pattern becomes:

Okay, let's keep it that way for now then.

cargo-shuttle/src/lib.rs
```diff
@@ -244,9 +240,7 @@ impl Shuttle {
             Command::Project(ProjectCommand::Status { follow }) => {
                 self.project_status(follow).await
             }
-            Command::Project(ProjectCommand::List { page, limit, raw }) => {
-                self.projects_list(page, limit, raw).await
-            }
+            Command::Project(ProjectCommand::List { raw, .. }) => self.projects_list(raw).await,
             Command::Project(ProjectCommand::Stop) => self.project_stop().await,
             Command::Project(ProjectCommand::Delete(ConfirmationArgs { yes })) => {
```
iulianbarbu, line 245: Delete should call `DELETE /projects/:project_name` instead of DELETE `/projects/:project_name/delete`.

iulianbarbu (commented):
Project start, list, status works as expected.

iulianbarbu (approved):
Thanks!

PR #1733, opened 2024-04-11
feat: get project owners and only personal projects

chesedo:
## Description of change
Add more details to projects to know who their owners are
Allow get the details of a single org
And allow only getting projects the user is the direct owner of

Added more tests and started the local stack

### DIFF
--- a/backends/src/client/permit.rs
+++ b/backends/src/client/permit.rs
@@ -29,10 +29,14 @@ use permit_pdp_client_rs::{
         policy_updater_api::trigger_policy_update_policy_updater_trigger_post,
         Error as PermitPDPClientError,
     },
-    models::{AuthorizationQuery, Resource, User, UserPermissionsQuery, UserPermissionsResult},
+    models::{AuthorizationQuery, Resource, User, UserPermissionsQuery},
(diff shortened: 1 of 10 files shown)


backends/src/client/permit.rs
```diff
@@ -276,24 +302,31 @@ impl PermissionsDal for Client {
...
-                tenants: Some(vec!["default".to_owned()]),
-                ..Default::default()
-            },
+    async fn get_personal_projects(&self, user_id: &str) -> Result<Vec<String>> {
+        let projects = list_role_assignments(
+            &self.api,
+            &self.proj_id,
+            &self.env_id,
+            Some(user_id),
+            Some("admin"),
```
jonaro00, line 311: What if they are not admin? Is this now a "get projects where I'm admin" method?

chesedo: Not quite. This is a "get projects where I have the admin role assigned method". So it won't return projects where the user is an "admin" because they are the admin of the org. I need this to be able to show users the projects they own which is not in some org.

jonaro00 (approved):

PR #1734, opened 2024-04-11
cargo-shuttle: update deploy cmd for beta platform 

iulianbarbu:
## Description of change

Output for first time deploy:
```
Current deployment: 'depl_01HV9V3C1CVPMN88ENE67FJ121' - in progress (http://pdo1-ms6ln917sadw.app.iulian.staging.shuttle.rs)
```

For subsequent deploys:
```
Current deployment: 'depl_01HV9V5GMT8WCF5NQHJD03VS3R' - in progress
Running deployment: 'depl_01HV9QJS9JEH5RKMAE7J65MAC5' - running (http://pdo-lum30rp5jxe3.app.iulian.staging.shuttle.rs)
```

Locally.

### DIFF
--- a/cargo-shuttle/src/client.rs
+++ b/cargo-shuttle/src/client.rs
@@ -83,7 +83,7 @@ impl ShuttleApiClient {
             .context("parsing name check response")
     }
 
-    pub async fn deploy_alpha(
+    pub async fn deploy(
         &self,
         project: &str,
         deployment_req: DeploymentRequest,
@@ -93,11 +93,7 @@ impl ShuttleApiClient {
             .context("serialize DeploymentRequest as a MessagePack byte vector")?;
 
         let url = format!("{}{}", self.api_url, path);
-        let mut builder = if self.beta {
-            self.client.put(url)
-        } else {
-            self.client.post(url)
-        };
+        let mut builder = self.client.post(url);
         builder = self.set_auth_bearer(builder);
 
         builder
@@ -120,11 +116,7 @@ impl ShuttleApiClient {
             .context("serialize DeploymentRequest as a MessagePack byte vector")?;
 
         let url = format!("{}{}", self.api_url, path);
-        let mut builder = if self.beta {
-            self.client.put(url)
-        } else {
-            self.client.post(url)
-        };
+        let mut builder = self.client.put(url);
         builder = self.set_auth_bearer(builder);
 
         builder
--- a/cargo-shuttle/src/client.rs
+++ b/cargo-shuttle/src/client.rs
@@ -7,6 +7,7 @@ use percent_encoding::utf8_percent_encode;
 use reqwest::header::HeaderMap;
 use reqwest::RequestBuilder;
 use reqwest::Response;
+use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use shuttle_common::constants::headers::X_CARGO_SHUTTLE_VERSION;
 use shuttle_common::models::deployment::DeploymentRequest;
@@ -24,7 +25,7 @@ pub struct ShuttleApiClient {
     api_url: String,
     api_key: Option<ApiKey>,
     /// alter behaviour to interact with the new platform
-    beta: bool,
+    pub beta: bool,
 }
 
 impl ShuttleApiClient {
@@ -84,11 +85,11 @@ impl ShuttleApiClient {
             .context("parsing name check response")
     }
 
-    pub async fn deploy(
+    pub async fn deploy<T: DeserializeOwned>(
         &self,
         project: &str,
         deployment_req: DeploymentRequest,
-    ) -> Result<deployment::Response> {
+    ) -> Result<T> {
         let path = if self.beta {
             format!("/projects/{project}")
         } else {
--- a/cargo-shuttle/src/client.rs
+++ b/cargo-shuttle/src/client.rs
@@ -23,7 +23,7 @@ pub struct ShuttleApiClient {
     api_url: String,
     api_key: Option<ApiKey>,
     /// alter behaviour to interact with the new platform
-    pub beta: bool,
+    beta: bool,
 }
 
 impl ShuttleApiClient {
--- a/cargo-shuttle/src/client.rs
+++ b/cargo-shuttle/src/client.rs
@@ -5,9 +5,7 @@ use anyhow::{Context, Result};
 use headers::{Authorization, HeaderMapExt};
 use percent_encoding::utf8_percent_encode;
 use reqwest::header::HeaderMap;
-use reqwest::RequestBuilder;
-use reqwest::Response;
-use serde::de::DeserializeOwned;
+use reqwest::{RequestBuilder, Response};
 use serde::{Deserialize, Serialize};
 use shuttle_common::constants::headers::X_CARGO_SHUTTLE_VERSION;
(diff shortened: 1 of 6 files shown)


cargo-shuttle/src/client.rs
```diff
@@ -24,7 +25,7 @@ pub struct ShuttleApiClient {
     api_url: String,
     api_key: Option<ApiKey>,
     /// alter behaviour to interact with the new platform
-    beta: bool,
+    pub beta: bool,
```
Kazy, line 28: nit: it shouldn't need to be `pub` ?

cargo-shuttle/src/client.rs
```diff
@@ -84,11 +85,11 @@ impl ShuttleApiClient {
             .context("parsing name check response")
     }
 
-    pub async fn deploy(
+    pub async fn deploy<T: DeserializeOwned>(
         &self,
         project: &str,
         deployment_req: DeploymentRequest,
-    ) -> Result<deployment::Response> {
+    ) -> Result<T> {
```
Kazy, line 92: suggestion: I would really just have two functions at this point. Ideally we would have two client, one for each version, even though that's a big refactor. The issue here is that it makes the function more difficult to understand (what's the expected output ? It varies between call site), and a change at the call site could make the function fail in non-obvious way.

common/src/models/service.rs
```diff
@@ -26,7 +26,6 @@ impl Display for Summary {
 Service Name:  {}
 Deployment ID: {}
 Status:        {}
-Last Updated:  {}
```
Kazy, line 29: Why remove it from the old platform ?

common/src/models/deployment.rs
```diff
@@ -47,6 +56,48 @@ impl Display for Response {
     }
 }
 
+impl Display for EcsResponse {
```
Kazy, line 59: suggestion: IMHO we shouldn't use `Display` for that, as it includes color, which is something that only concern the CLI display. Here it means that if we log this value, we will get an output with escape code.

cargo-shuttle/src/lib.rs
```diff
@@ -1701,8 +1702,19 @@ impl Shuttle {
             );
         }
 
+        // End this early for beta platform.
+        if client.beta {
```
jonaro00, line 1706: The flag is available in both places ;)
```suggestion
        if self.beta {
```

common/src/models/deployment.rs
```diff
@@ -29,15 +29,24 @@ pub struct Response {
...
+    pub id: String,
+    pub latest_deployment_state: EcsState,
+    pub running_id: Option<String>,
+    pub updated_at: DateTime<Utc>,
+    pub uri: String,
+    pub git_commit_id: Option<String>,
+    pub git_commit_msg: Option<String>,
+    pub git_branch: Option<String>,
+    pub git_dirty: Option<bool>,
+}
```
jonaro00, line 43: Don't we want to return and print raw JSON from beta commadns before we make a frontend?

common/src/models/deployment.rs
```diff
@@ -163,13 +226,11 @@ pub fn get_deployments_table(
...
                 table.add_row(vec![
-                    Cell::new(deploy.id),
+                    Cell::new(deploy.id.clone()),
                     Cell::new(&deploy.state)
                         // Unwrap is safe because Color::from_str returns the color white if str is not a Color.
                         .fg(Color::from_str(deploy.state.get_color()).unwrap())
                         .set_alignment(CellAlignment::Center),
-                    Cell::new(deploy.last_update.format("%Y-%m-%dT%H:%M:%SZ"))
-                        .set_alignment(CellAlignment::Center),
                     Cell::new(truncated_commit_id),
```
jonaro00, line 234: we should avoid modifying old frontend's behaviour

cargo-shuttle/src/client.rs
iulianbarbu: I needed it to be pub for a check outside the client. This comment is related to this one too: https://github.com/shuttle-hq/shuttle/pull/1734#discussion_r1565290876, where I guess what we'll do is have two different functions. I'll update it based on this remark.

common/src/models/deployment.rs
iulianbarbu: Good thought. I'll add this logic outside of `Display` then, which will be the same except for the coloring.

common/src/models/deployment.rs
iulianbarbu: We can, but I prefer to get some initial feedback of the frontend too. We can modify the frontend as we wish afterward. WDYT?

common/src/models/service.rs
iulianbarbu: Left over :see_no_evil:. Will fix.

cargo-shuttle/src/lib.rs
iulianbarbu: Done.

common/src/models/deployment.rs
iulianbarbu: Done.

cargo-shuttle/src/client.rs
iulianbarbu: Split this in two functions (beta & alpha) in the same client for now. I suggest to create a client for each platform later on, after we clarify which is the differentiating logic for all commands.

common/src/models/service.rs
iulianbarbu: Done.

cargo-shuttle/src/client.rs
```diff
@@ -84,16 +83,39 @@ impl ShuttleApiClient {
...
+            .context("serialize DeploymentRequest as a MessagePack byte vector")?;
+
+        let url = format!("{}{}", self.api_url, path);
+        let mut builder = if self.beta {
+            self.client.put(url)
         } else {
-            format!("/projects/{project}/services/{project}")
+            self.client.post(url)
         };
+        builder = self.set_auth_bearer(builder);
```
Kazy, line 101: question: why do you still `if self.beta` when this is supposed to only be for alpha ? And there seems to be a mix between `alpha`/`beta`, because personally I would have used `alpha` for the current one, and `beta` for the new one (but better is my other suggestion to just keep the current `deploy` as-is).

common/src/models/deployment.rs
```diff
@@ -47,6 +60,47 @@ impl Display for Response {
     }
 }
 
+impl EcsResponse {
+    pub fn formatted_println(&self) {
```
Kazy, line 64:
```suggestion
    pub fn colored_println(&self) {
```

cargo-shuttle/src/client.rs
```diff
@@ -84,16 +83,39 @@ impl ShuttleApiClient {
             .context("parsing name check response")
     }
 
-    pub async fn deploy(
```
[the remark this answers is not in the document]
iulianbarbu: Done.

cargo-shuttle/src/client.rs
iulianbarbu: Good catch! My mistake, removed the if..else and kept only the corresponding method call (post for `deploy` and put for `deploy_beta`).

common/src/models/deployment.rs
iulianbarbu: Done.

Kazy (approved):

PR #1742, opened 2024-04-17
feat(cargo-shuttle): beta deploy zip file, package name

jonaro00:

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -57,6 +57,7 @@ use shuttle_proto::{
     provisioner::{provisioner_server::Provisioner, DatabaseRequest},
     runtime::{self, LoadRequest, StartRequest, StopRequest},
 };
+use shuttle_service::builder::{async_cargo_metadata, find_shuttle_packages};
 use shuttle_service::{
     builder::{build_workspace, BuiltService},
     runner, Environment,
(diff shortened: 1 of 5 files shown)


common/src/models/deployment.rs
```diff
@@ -269,7 +269,11 @@ pub fn get_deployments_table(
 
 #[derive(Default, Deserialize, Serialize)]
 pub struct DeploymentRequest {
+    /// Alpha: tar archive. Beta: zip archive.
     pub data: Vec<u8>,
+    /// The cargo package name to compile and run. Required on beta.
+    pub package_name: Option<String>,
+    /// Ignored on beta.
```
oddgrd, line 276: Thanks for adding these!

cargo-shuttle/src/lib.rs
```diff
@@ -1656,11 +1658,26 @@ impl Shuttle {
...
+            }
+            let metadata = async_cargo_metadata(manifest_path.as_path()).await?;
+            let packages = find_shuttle_packages(&metadata)?;
+            let package_name = packages
+                .first()
+                .expect("at least one shuttle crate in the workspace")
+                .name
+                .to_owned();
+            deployment_req.package_name = Some(package_name);
+        }
```
oddgrd, line 1679: Could we have a problem if there are more and the chosen name is not the one with the shuttle main function? Perhaps I'm not seeing the bigger picture here, and it doesn't matter. :thinking:

oddgrd (approved):
LGTM! Just left one question for my own understanding.

PR #1747, opened 2024-04-19
fix: change TursoOutput to provide Database

### LINKED ISSUE #1746: [Bug]: Turso connection closes too early
### What happened?

Ref: https://github.com/tursodatabase/libsql/issues/925
Ref Discord help thread: https://discord.com/channels/803236282088161321/1228486587864584293

The Turso connection closes too early because it expects a ping/query every 10-15s or it will close to ensure low resource usage. We should therefore not use `Connection` and instead use `Database` so that users can connect to the database as needed.

### Version

v0.43.0

macOS, Windows, Linux, In deployment

Other

### Relevant log output

_No response_

### Duplicate declaration

joshua-mo-143:
## Description of change

Closes #1746

### DIFF
--- a/resources/turso/src/lib.rs
+++ b/resources/turso/src/lib.rs
@@ -1,5 +1,5 @@
 use async_trait::async_trait;
-use libsql::{Builder, Connection};
+use libsql::{Builder, Database};
 use serde::{Deserialize, Serialize};
 use shuttle_service::{
     error::{CustomError, Error as ShuttleError},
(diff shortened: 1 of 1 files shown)


oddgrd (commented):
Thanks @joshua-mo-143! It looks like the example will need to be updated as well, since you've got some context on this, would you mind doing that as well?

oddgrd (approved):

PR #1753, opened 2024-04-23
chore: merge shuttle-ecs-common to main

jonaro00:


backends/src/trace.rs
```diff
@@ -57,7 +57,7 @@ where
         .with_log_config(Config::default().with_resource(Resource::new(resources.clone())))
         .with_exporter(
             opentelemetry_otlp::new_exporter()
-                .tonic()
+                .http()
```
jonaro00, line 60: TODO: check that the otel config change for this still works on staging

runtime/src/alpha.rs
```diff
@@ -29,20 +29,65 @@ use tokio_stream::wrappers::ReceiverStream;
...
+                }
+                "--address" => {
+                    let address = args_iter
+                        .next()
+                        .context("missing address value")?
+                        .parse()
+                        .context("invalid address value")?;
+                    args.address = Some(address);
+                }
+                _ => {}
```
oddgrd, line 73: Maybe we add some message about invalid args here?

oddgrd (commented):
The runtime compatibility changes LGTM!

runtime/src/alpha.rs
jonaro00: We don't want old runtimes to freak out if they recieve new args that they don't support yet.

oddgrd: Ah, that's a good point.

runtime/Cargo.toml
```diff
@@ -22,6 +22,7 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 strfmt = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
+tokio-util = "0.7.10"
```
Kazy, line 25: Should it use the workspace version ?

runtime/src/alpha.rs
```diff
@@ -223,6 +300,27 @@ where
...
+        let state = self.state.clone();
+        let cancellation_token = self.cancellation_token.clone();
+
+        // State and cancellation is not used in alpha
+        if self.beta {
+            // Ensure that the runtime is set to unhealthy if it doesn't reach the running state after
+            // it has sent a load response, so that the ECS task will fail.
+            tokio::spawn(async move {
+                // Note: The timeout is quite low as we are not actually provisioning resources after
+                // sending the load response.
```
Kazy, line 314: @oddgrd this isn't the case anymore now, is it ? Should we increase the timeout or do something else ?

cargo-shuttle/src/client.rs
```diff
@@ -41,14 +39,25 @@ impl Client {
                     )]))
                     .unwrap(),
                 )
-                .timeout(Duration::from_secs(60))
+                .timeout(Duration::from_secs(300)) // TODO: adjust
```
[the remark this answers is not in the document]
jonaro00: What is the LB timeout?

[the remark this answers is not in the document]
Kazy: 60 seconds.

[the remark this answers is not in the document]
Kazy: The change can be removed, not the timeout (in case that wasn't clear).

runtime/src/alpha.rs
oddgrd: We have a 120 second timeout in the provisioner rds sdk, so I think we can keep it like this for now, and we'll see if we need to change it as I test the RDS resource implementation.

oddgrd: We can update the comment, though, to give the reader an understanding of why we need it, which is the fact that some resources can take a while to provision, and at the time of writing the main offender is RDS.

oddgrd (approved):

backends/src/trace.rs
```diff
@@ -43,7 +43,7 @@ where
         .tracing()
         .with_exporter(
             opentelemetry_otlp::new_exporter()
-                .tonic()
+                .http()
```
chesedo, line 46: Why are we making this change?

cargo-shuttle/src/client.rs
```diff
@@ -5,34 +5,32 @@ use anyhow::{Context, Result};
...
 #[derive(Clone)]
-pub struct Client {
-    api_url: ApiUrl,
-    api_key: Option<Secret<ApiKey>>,
+pub struct ShuttleApiClient {
     client: reqwest::Client,
+    api_url: String,
+    api_key: Option<ApiKey>,
+    /// alter behaviour to interact with the new platform
+    beta: bool,
```
chesedo, line 28: I would recommend rather creating a new client for beta rather than passing this flag around everywhere

chesedo (commented):
Doing `if beta` checks everywhere seems to be a recipe for disaster. We should try to reduce this by making new structs for the beta code and diverting the code flow to them as early as possible.

backends/src/trace.rs
jonaro00: Making it compatible with an ALB on beta platform IIRC

cargo-shuttle/src/client.rs
jonaro00: The number of beta changes thus far is small enough for me to justify this. We might extract it later, but there will still be many places outside of this client that will do the same thing (modify behaviour based on passed flag).

backends/src/trace.rs
Kazy: Yep, I had some trouble making gRPC work in front of the load balancer. I can try again though, I'll let you know how it goes.

Kazy: Yep can confirm, it's a PITA. We have an ALB in front of OTEL, which requires an HTTPS listener when targeting a gRPC target group. The issue is that to have an HTTPS listener, you need to have a valid ACM certificate attached to it, and to get a valid ACM certificate, you need to be able to verify it, which isn't possible for internal facing address. 

I think we might be able to use ACM PCA (Private Certificate Authority), but that's not my priority right now.

cargo-shuttle/src/client.rs
chesedo: Hmm, that more points to why it would have been a good idea to do it now instead of later when it will be harder.

PR #1757, opened 2024-04-26
service: ask question if package not a binary

iulianbarbu:
## Description of change

Make the error also suggest what might be the issue.

## How has this been tested? (if applicable)

N/A

PR #1758, opened 2024-04-26
fix(service): improve workspace member discovery and error

jonaro00:
## Description of change

Supersedes https://github.com/shuttle-hq/shuttle/pull/1757

Improves logic for which workspace member(s) that are chosen for building by ignoring crates with no binary target.

Tested on a local example.
Having libs with runtime now works. Having no members with binaries now prints the new error.


iulianbarbu (approved):
LGTM! Thanks.

PR #1775, opened 2024-05-17
feat: update beta logs logic

jonaro00:


cargo-shuttle/src/args.rs
```diff
@@ -430,6 +429,9 @@ pub struct LogsArgs {
     /// View all log lines
     #[arg(long, group = "output_mode")]
     pub all: bool,
+    /// Get logs from all deployments instead of one deployment
+    #[arg(long)]
+    pub dump_it_all: bool,
```
Kazy, line 434: Let's maybe use a more explicit name ? :p
```suggestion
    pub all_deployments: bool,
```

cargo-shuttle/src/client.rs
```diff
@@ -254,11 +254,25 @@ impl ShuttleApiClient {
             .await
             .context("Failed parsing logs. Is your cargo-shuttle outdated?")
     }
+    pub async fn get_deployment_logs_beta(
+        &self,
+        project: &str,
+        deployment_id: &str,
+    ) -> Result<Vec<LogItemBeta>> {
```
Kazy, line 261: Can we return a JSON here instead of lines directly ? Same for the project logs actually. Something like this for now, but that we could easily change to add pagination for instance:
```rust
struct {
  logs: Vec<LogItemBeta>,
}
```

cargo-shuttle/src/lib.rs
```diff
@@ -850,31 +850,23 @@ impl Shuttle {
...
+        let proj_name = self.ctx.project_name();
+        let logs = if args.all_deployments {
+            client.get_project_logs_beta(proj_name).await?.logs
+        } else {
+            // TODO:
+            // let depl = client.get_current_deployment_beta(proj_name).await?;
+            let depls = client.get_deployments_beta(proj_name).await?;
+            let depl = depls
+                .first()
+                .expect("at least one deployment in this project");
```
Kazy, line 862: suggestion(non-blocking): we might want to return a proper error to the user instead of panicking. Can be done in an other PR.

Kazy (approved):

PR #1782, opened 2024-05-22
feat: beta deployment response, current deployment & logs

jonaro00:
uses fixed /current deployment endpoint for deployment status and logs command.

updates deployment response model with more relevant fields:
- created_at
- URIs is a Vec
- leaving out git metadata to leave room to rethink: it probably belongs in build metadata

### DIFF
--- a/common/src/deployment.rs
+++ b/common/src/deployment.rs
@@ -32,6 +32,8 @@ pub enum EcsState {
     Stopped,
     Stopping,
     Failed,
+    /// Fallback
+    Unknown,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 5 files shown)


common/src/deployment.rs
```diff
@@ -32,6 +32,8 @@ pub enum EcsState {
     Stopped,
     Stopping,
     Failed,
+    /// Fallback
+    Unknown,
```
Kazy, line 36: Do we really want to introduce that ? When can it happen ?

Kazy (approved):

common/src/deployment.rs
jonaro00: Made it a fallback when we are parsing an ECS string response into the enum. We don't cover states that we don't expect to see.

PR #1783, opened 2024-05-22
refactor: better errors

chesedo:
## Description of change
Improve `ApiError` to make error handling on headroom easier

Na

### DIFF
--- a/common/src/models/error.rs
+++ b/common/src/models/error.rs
@@ -30,13 +30,22 @@ impl ApiError {
     }
 
     /// Creates an internal error without exposing sensitive information to the user.
-    pub fn internal_safe(error: impl std::error::Error) -> Self {
-        error!(error = error.to_string(), "");
+    #[inline(always)]
+    pub fn internal_safe<E>(message: &str, error: E) -> Self
+    where
+        E: std::error::Error + 'static,
+    {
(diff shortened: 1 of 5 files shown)


jonaro00 (approved):

common/src/models/error.rs
```diff
@@ -80,6 +82,131 @@ impl ApiError {
     }
 }
 
+pub trait ErrorContext<T> {
```
Kazy, line 85: What do you think of having a generic `context_error(self, message: &str, status_code: StatusCode)` (and the corresponding `with` variant) instead of dedicated method each time ?

chesedo: Nah, then each caller would have type out the status code each time. Not a fan of passing `booleans` as args to functions and this seems close to the same thing.

chesedo: That's an idea, but I can abstract it more though. Ie I won't be able to make them use the generic one since they log different `warn!` and `error!`.

I mean they are the same in that they use a dynamic control (the args) when we know which concrete type we want when we write the code. So their dynamic nature at runtime feels the same.

Kazy: Ah, make sense. Ok let's keep it that way for now then, we'll see when the need arise.

PR #1784, opened 2024-05-23
refactor: implement permission dal when behind an arc

chesedo:
## Description of change
Headroom uses the `PermissionDal` behind an `Arc` and needs this

The compiler


jonaro00 (approved):
The crazy things we do to use traits :smiling_face_with_tear:

chesedo:
> The crazy things we do to use traits 🥲

Haha, this is not really about using traits, but what using traits will allow you to do easier. It is impossible to test business logic without the use of traits.

PR #1791, opened 2024-06-03
fix: improve delete command reliability

jonaro00:


gateway/src/task.rs
```diff
@@ -161,7 +161,8 @@ pub fn restart(project_id: Ulid) -> impl Task<ProjectContext, Output = Project>
             .container()
             .and_then(|container| ProjectCreating::from_container(container, 0).ok())
             .unwrap_or_else(|| {
-                ProjectCreating::new_with_random_initial_key(ctx.project_name, project_id, 1)
+                // idle minutes 1 is generally buggy, therefore use 3.
+                ProjectCreating::new_with_random_initial_key(ctx.project_name, project_id, 3)
```
oddgrd, line 165: Hmm, can you remind me why we set idle minutes like this here?

cargo-shuttle/src/client.rs
```diff
@@ -37,7 +37,8 @@ impl ShuttleApiClient {
                     )]))
                     .unwrap(),
                 )
-                .timeout(Duration::from_secs(60))
+                // some operations, like project delete, can take a while.
+                .timeout(Duration::from_secs(60 * 5))
```
oddgrd, line 41: I'm not sure if we should increase the timeout across the board. If we have some request that isn't making progress, the user would need to wait a lot longer before they get a failure. I suppose the alternative is refactoring the project delete call to polling, is that going to be too time-consuming to implement?

gateway/src/task.rs
jonaro00: I think P set it like that since we know the project is about to get deleted. I'm thinking there's a small chance the 1 minute setting (which acts like 0 minutes) could start conflicting when state machine is slow (delete command trying to start/stop + ambulance trying to idle).

cargo-shuttle/src/client.rs
jonaro00: Or I set the high timeout for the delete request only.

gateway/src/task.rs
oddgrd: Ah, I see, thanks!

cargo-shuttle/src/client.rs
oddgrd: Nice, thanks!

oddgrd (approved):

PR #1795, opened 2024-06-10
feat: provisioner idempotence changes

oddgrd:
## Description of change

This moves resourcestate to common, since the runner will also need it.

Tested on my stack.

### DIFF
--- a/common/src/resource.rs
+++ b/common/src/resource.rs
@@ -64,7 +64,7 @@ pub struct ShuttleResourceOutput<T> {
     pub custom: Value,
 
     /// The state of the resource.
-    pub state: ResourceState,
+    pub state: Option<ResourceState>,
 }
 
 /// Common type to hold all the information we need for a generic resource
--- a/common/src/resource.rs
+++ b/common/src/resource.rs
@@ -51,7 +51,6 @@ pub enum ResourceState {
     Failed,
     Ready,
     Deleting,
-    Deleted,
 }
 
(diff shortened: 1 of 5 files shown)


common/src/resource.rs
```diff
@@ -166,6 +183,30 @@ mod _sqlx {
...
+    }
+
+    impl<'r> sqlx::Decode<'r, Postgres> for ResourceState {
+        fn decode(value: PgValueRef<'r>) -> Result<Self, sqlx::error::BoxDynError> {
+            let value = <&str as sqlx::Decode<Postgres>>::decode(value)?;
+
+            let state = ResourceState::from_str(value)?;
+            Ok(state)
+        }
+    }
```
jonaro00, line 209: nit: Feels like there should be a better way to do this. :thinking: Not a must.

deployer/src/deployment/run.rs
```diff
@@ -549,6 +550,7 @@ async fn provision(
                 *bytes = serde_json::to_vec(&ShuttleResourceOutput {
                     output: new_secrets.clone(),
                     custom: shuttle_resource.custom,
+                    state: ResourceState::Ready
```
jonaro00, line 553: Might not be worth changing this model on alpha. Consider using a `Beta` copy of it, but we also have time to re-think how we record resources.

Kazy (approved):

jonaro00 (approved):

PR #1796, opened 2024-06-11
feat(cargo-shuttle): beta: support image deployment

jonaro00:


cargo-shuttle/src/args.rs
```diff
@@ -259,6 +259,10 @@ pub struct LogoutArgs {
 }
 #[derive(Parser, Default)]
 pub struct DeployArgs {
+    /// BETA: Deploy this Docker image instead of building one
+    #[arg(long, short = 'i', hide = true)]
+    pub image: Option<String>, // TODO?: Make this a subcommand instead? `cargo shuttle deploy image ...`
```
Kazy, line 264: I'm not sure we want to enable that just yet. I know it's hidden, but it might give the wrong impression. I wonder if it could lead to someone using images that are way too big, or that are clearly made to assault the platform in different ways. 

I've commented on the endpoint itself, if we limit that to admins, that's fine.

Kazy (approved):

PR #1809, opened 2024-06-19
fix: local development

jonaro00:
fixes small things discovered when helping a new user run the stack

### DIFF
--- a/auth/src/main.rs
+++ b/auth/src/main.rs
@@ -17,7 +17,8 @@ async fn main() {
         return;
     }
 
-    let pool = pgpool_init(args.db_connection_uri.as_str())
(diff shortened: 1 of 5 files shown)


auth/src/main.rs
```diff
@@ -17,7 +17,8 @@ async fn main() {
         return;
     }
 
-    let pool = pgpool_init(args.db_connection_uri.as_str())
+    // do the migrations in local development
+    let pool = pgpool_init(args.db_connection_uri.as_str(), cfg!(debug_assertions))
```
oddgrd, line 21: Would it be safer to use an arg?

oddgrd (approved):

PR #1814, opened 2024-06-28
feat(cargo-shuttle): beta archive deployment separation

jonaro00:


cargo-shuttle/src/lib.rs
```diff
@@ -1893,7 +1893,8 @@ impl Shuttle {
         }
 
         deployment_req.data = self.make_archive(args.secret_args.secrets.clone(), self.beta)?;
-        if deployment_req.data.len() > CREATE_SERVICE_BODY_LIMIT {
+        // TODO: Make size warning server-side on beta.
```
oddgrd, line 1896: Should we just remove the check here now, then, and let the API send a 413 if the body is too large?

oddgrd (approved):

cargo-shuttle/src/lib.rs
jonaro00: Will leave alpha behaviour as is. I think the server error was quite ugly.

PR #1839, opened 2024-07-24
fix(resource-recorder): check project permission by project id

jonaro00:
## Description of change

- allow checking project endpoint with the project id. this simplifies and corrects the check in r-r for project permission.

tested on staging

### DIFF
--- a/gateway/src/auth.rs
+++ b/gateway/src/auth.rs
@@ -82,17 +82,20 @@ where
         let RouterState { service, .. } = RouterState::from_ref(state);
 
         // Enables checking HEAD at /projects/{ulid}, used by res-rec to check permission for a proj id
-        let scope = if let Ok(Path(ulid)) = Path::<Ulid>::from_request_parts(parts, state).await {
-            let p = service.find_project_by_id(&ulid.to_string()).await?;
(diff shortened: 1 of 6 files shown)


gateway/src/auth.rs
```diff
@@ -78,17 +79,27 @@ where
...
-        };
-
         let RouterState { service, .. } = RouterState::from_ref(state);
 
-        let allowed = claim.is_admin()
+        // Enables checking HEAD at /projects/{ulid}, used by res-rec to check permission for a proj id
+        let scope = match Path::<Ulid>::from_request_parts(parts, state).await {
+            Ok(Path(ulid)) => {
+                let p = service.find_project_by_id(&ulid.to_string()).await?;
+                ProjectName::new(&p.name).expect("valid project name")
```
oddgrd, line 88: Should we mapp to invalidprojectname err instead?

oddgrd (approved):

gateway/src/auth.rs
jonaro00: Would rather have an internal error than a user facing one here. It should be safe regardless since the name is coming out of the db.

PR #1844, opened 2024-07-25
feat(runtime): beta runtime as client

jonaro00:


runtime/src/beta.rs
```diff
@@ -0,0 +1,145 @@
...
+            .await
+        {
+            Ok(o) => o,
+            Err(e) => {
+                eprintln!("Runtime Provisioning phase failed: {e}");
+                exit(131);
+            }
+        };
+        *bytes = serde_json::to_vec(&o).expect("to serialize struct");
+    }
```
oddgrd, line 125: This needs to be polled until all the resources are ready.

service/src/runner.rs
```diff
@@ -12,17 +11,13 @@ use tracing::info;
 pub async fn start(
     beta: bool,
     port: u16,
-    // only used on beta. must match port.
-    address: SocketAddr,
     runtime_executable: PathBuf,
     project_path: &Path,
 ) -> anyhow::Result<(process::Child, runtime::Client)> {
```
oddgrd, line 16: Will this break anything? For example if someone upgrades deployer, but not runtime? :thinking:

oddgrd: Only the args passed to beta runtime changes, though, so it should be fine..

jonaro00: Yeah that line was added as part of adding the beta runtime flag

runtime/src/beta.rs
```diff
@@ -0,0 +1,171 @@
...
+        .expect("invalid shuttle environment");
+    let ip = std::env::var("SHUTTLE_RUNTIME_IP")
+        .expect("runtime ip env var")
+        .parse()
+        .expect("invalid ip");
+    let port = std::env::var("SHUTTLE_RUNTIME_PORT")
+        .expect("runtime port env var")
+        .parse()
+        .expect("invalid port");
+    let api_url = std::env::var("SHUTTLE_API").expect("api env var");
```
oddgrd, line 42: nit: Maybe we could still use a Config struct in a separate file, that is created by getting all these env vars? To clean up the function a bit.

oddgrd (approved):

PR #1850, opened 2024-08-01
feat(cargo-shuttle): beta updated local run

jonaro00:
extends https://github.com/shuttle-hq/shuttle/pull/1844

tested local run with:
rocket/hello-world
rocket/postgres
rocket/secrets


cargo-shuttle/src/lib.rs
```diff
@@ -1416,9 +1435,7 @@ impl Shuttle {
                 service_name.as_str(),
                 false,
                 false,
-                // Set beta to false to avoid breaking local run with beta changes.
-                // TODO: make local run compatible with --beta.
-                false
```
oddgrd, line 1421: Is the comment no longer accurate?

cargo-shuttle/src/lib.rs
```diff
@@ -1787,6 +1820,182 @@ impl Shuttle {
...
+        );
+        let mut runtime = tokio::process::Command::new(
+            dunce::canonicalize(runtime_executable).context("canonicalize path of executable")?,
+        )
+        .current_dir(&service.workspace_path)
+        .args(["--run"])
+        .envs([
+            ("SHUTTLE_BETA", "true"),
+            ("SHUTTLE_PROJECT_ID", "proj_LOCAL"),
+            ("SHUTTLE_PROJECT_NAME", "TODO"),
```
oddgrd, line 1888: What is the plan for this, and does the user see it?

cargo-shuttle/src/lib.rs
```diff
@@ -1787,6 +1820,182 @@ impl Shuttle {
...
+                }))
+            }
+        });
+        let server = Server::bind(&api_addr).serve(make_svc);
+        tokio::spawn(async move {
+            if let Err(e) = server.await {
+                eprintln!("Provisioner server error: {}", e);
+                exit(1);
+            }
+        });
```
oddgrd, line 1866: Should we abstract this into a BetaProvisionerServer struct?

oddgrd (approved):

cargo-shuttle/src/lib.rs
jonaro00: Updating resource tables (if we want tables at all) for beta is still TODO, so whatever happens here will change.

cargo-shuttle/src/lib.rs
jonaro00: TY for reminder. It's passed into ResourceFactory and Metadata.

PR #1858, opened 2024-08-15
feat: instrument permit calls

oddgrd:
## Description of change

Instrument permit client calls, and add a common field to group them.

### DIFF
--- a/backends/src/client/permit.rs
+++ b/backends/src/client/permit.rs
@@ -37,7 +37,7 @@ use shuttle_common::{
     claims::AccountTier,
     models::{error::ApiError, project, team},
 };
-use tracing::error;
+use tracing::{error, instrument};
 
 #[async_trait]
 pub trait PermissionsDal {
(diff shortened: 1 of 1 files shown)


backends/src/client/permit.rs
```diff
@@ -488,6 +505,7 @@ impl PermissionsDal for Client {
         Ok(res.allow.unwrap_or_default())
     }
 
+    #[instrument(skip_all, fields(shuttle.user.id = user_id, shuttle.permit.client = true))]
     async fn create_team(&self, user_id: &str, team: &Team) -> Result<()> {
```
jonaro00, line 509: add team id

backends/src/client/permit.rs
```diff
@@ -462,6 +474,11 @@ impl PermissionsDal for Client {
         Ok(projects)
     }
 
+    #[instrument(skip_all, fields(
+        shuttle.user.id = user_id,
+        shuttle.project.id = project_id,
+        shuttle.permit.client = true
+    ))]
     async fn allowed(&self, user_id: &str, project_id: &str, action: &str) -> Result<bool> {
```
jonaro00, line 482: add the action

jonaro00 (commented):
found these. also a nit to have traling comma on the ones with multiple lines

backends/src/client/permit.rs
```diff
@@ -477,7 +477,8 @@ impl PermissionsDal for Client {
     #[instrument(skip_all, fields(
         shuttle.user.id = user_id,
         shuttle.project.id = project_id,
-        shuttle.permit.client = true
+        shuttle.team.action = action,
```
jonaro00, line 480: would have called it `shuttle.permit.action` or just `action` but no worries :)

PR #1860, opened 2024-08-20
feat(cargo-shuttle): beta certificate command

jonaro00:


oddgrd (approved):
Nice!

common/src/certificate.rs
```diff
@@ -6,4 +6,13 @@ pub struct AddCertificateRequest {
...
+pub struct DeleteCertificateRequest {
+    pub domain: String,
+}
+
+#[derive(Deserialize, Serialize, Debug)]
+pub struct CertificateResponse {
+    pub subject: String,
+    pub serial_hex: String,
+    pub not_after: String,
+}
```
oddgrd, line 18: Are these duplicated in ecs-shuttle, or are they used for different things?

oddgrd (approved):

common/src/certificate.rs
jonaro00: The req models are similar but different due to these not needing project id

Issue #1810, opened 2024-06-19
[Bug]: Intall fails on Arch Linux


mael-bomane:
### What happened?

I tried to run the one-liner to install from the docs `curl -sSfL https://www.shuttle.rs/install | bash` but the install script doesn't goes through :<

### Version

v0.46.0-1

### Which operating system(s) are you seeing the problem on?

Linux

### Which CPU architectures are you seeing the problem on?

x86_64

### Relevant log output

```sh
[dev@terminal ~]$ curl -sSfL https://www.shuttle.rs/install | bash
     _           _   _   _
 ___| |__  _   _| |_| |_| | ___
/ __| '_ \| | | | __| __| |/ _ \
\__ \ | | | |_| | |_| |_| |  __/
|___/_| |_|\__,_|\__|\__|_|\___|

https://www.shuttle.rs
https://github.com/shuttle-hq/shuttle

Please file an issue if you encounter any problems!
===================================================
Detected Linux!
Checking distro...
lsb_release could not be found. Falling back to /etc/os-release
Arch Linux detected!
[sudo] password for dev:
Installing with pacman
resolving dependencies...
:: There are 2 providers available for cargo:
:: Repository extra
   1) rust  2) rustup

Enter a number (default=1): looking for conflicting packages...

Packages (2) rust-1:1.79.0-1  cargo-shuttle-0.46.0-1

Total Download Size:    68.66 MiB
Total Installed Size:  281.11 MiB

:: Proceed with installation? [Y/n] cat <<EOF
```

### Duplicate declaration

- [X] I have searched the issues and there are none like this.

oddgrd:
@ez-krk thanks for the report! What do you mean it doesn't go through? Does it complete, but then cargo-shuttle wasn't installed? Did you get an error message?

PR #1861, opened 2024-08-23
fix: failing install on Arch Linux (#1810)

lielfr:
## Description of change
Added flag for install.sh, such that for Arch Linux, it does not try to ask for confirmation.

I modified the Dockerfile I used for replicating the bug, such that it runs install.sh instead of downloading the script from the Shuttle website. It seems to work fine


oddgrd (approved):
Thanks for the fix, and thanks for adding steps to reproduce the issue!

PR #1871, opened 2024-09-06
feat(common): beta models update, add typeshare

jonaro00:
- renames models
- simplifies beta resource models
- adds typeshare to typescript

TODO:
- verify all models work
- update beta backend to use new models
- mark ts file generated
- add ci workflow + check for ts file


resources/shared-db/src/lib.rs
oddgrd, line 1: I guess this could be a breaking change? If someone is using just the postgres-rustls feature. It would surprise me, though.

oddgrd (approved):
LGTM, we may just want a display impl for the resource type, see comment in other PR

resources/shared-db/src/lib.rs
jonaro00: The feature does not exist. New cargo warning caught it

PR #1872, opened 2024-09-10
fix: remove teams call for beta project list

oddgrd:
## Description of change

The teams feature is disabled on the beta platform, so we get a 404 if we try to call that endpoint.

Tested in beta dev environment.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -2703,19 +2703,22 @@ impl Shuttle {
         println!("{}", "Personal Projects".bold());
         println!("{projects_table}\n");
 
-        let teams = client.get_teams_list().await?;
+        if !self.beta {
+            let teams = client.get_teams_list().await?;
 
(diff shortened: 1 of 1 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -2703,19 +2703,22 @@ impl Shuttle {
...
+
+            for team in teams {
+                let team_projects_table = if self.beta {
+                    let team_projects =
+                        client.get_team_projects_list_beta(&team.id).await?.projects;
+                    project::get_projects_table_beta(&team_projects, table_args.raw)
+                } else {
+                    let team_projects = client.get_team_projects_list(&team.id).await?;
+                    project::get_projects_table(&team_projects, table_args.raw)
+                };
```
jonaro00, line 2717: inner branch can be flattened, or not :man_shrugging:

jonaro00 (approved):

PR #1875, opened 2024-09-11
feat: beta models update, fix tables and prints

jonaro00:


common/src/lib.rs
```diff
@@ -176,10 +176,11 @@ impl DatabaseInfo {
 
 /// Holds the data for building a database connection string on the Beta platform.
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[typeshare::typeshare]
 pub struct DatabaseInfoBeta {
     engine: String,
     role_name: String,
-    role_password: Secret<String>,
```
oddgrd, line 182: There is a risk that we expose it with the debug impl if we remove Secret. maybe create a custom impl? I assume the removal is to make it usable with typeshare/console.

oddgrd (approved):

PR #1885, opened 2024-09-27
feat(cargo-shuttle): beta project linking

jonaro00:


cargo-shuttle/src/config.rs
```diff
@@ -249,10 +262,75 @@ impl RequestContext {
...
+                config.id = Some(id_from_args.clone());
+            }
+            // If key exists in config then keep it as it is
+            (None, Some(_)) => {
+                trace!("using .shuttle/config.toml project id");
+            }
+            (None, None) => {
+                trace!("no project id in args or config found");
+            }
+        };
```
oddgrd, line 308: It's a bit confusing that this is name or id, but we expect it to be id.

cargo-shuttle/src/init.rs
```diff
@@ -274,11 +274,11 @@ fn edit_shuttle_toml(path: &Path, set_name: Option<&str>) -> Result<()> {
 }
 
 /// Adds any missing recommended ignore rules to .gitignore or .ignore depending on if git is used.
-fn create_ignore_file(path: &Path, name: &str) -> Result<()> {
+pub fn create_ignore_file(path: &Path, name: &str) -> Result<()> {
```
oddgrd, line 277: Rename to create_or_update_ignore_file?

oddgrd (approved):

cargo-shuttle/src/config.rs
jonaro00: If a name is given, it is translated to its id on the go in `load_project`

PR #1907, opened 2024-10-21
feat: dump db endpoint, fixes

jonaro00:
## Description of change

- Add dump endpoint in provisioner that uses pg_dump
- Add dump endpoint in deployer that calls it
- Add cli command that calls it
- cli will uppercase proj ids
- AccountTier::Employee 
- minor fixes

Dumped on local stack


provisioner/src/lib.rs
```diff
@@ -565,6 +573,31 @@ impl Provisioner for ShuttleProvisioner {
...
+        let claim = request.get_claim()?;
+        let request = request.into_inner();
+        if !ProjectName::is_valid(&request.project_name) {
+            return Err(Status::invalid_argument("invalid project name"));
+        }
+        self.verify_ownership(&claim, &request.project_name).await?;
+
+        let database_name = format!("db-{}", request.project_name);
+        let output = tokio::process::Command::new("pg_dump")
+            .arg(format!("{}{}", self.shared_pg_uri_stripped, database_name))
```
oddgrd, line 591: Why do we first strip the database name, then add it back here?

oddgrd (approved):
Cool, thanks!

provisioner/src/lib.rs
jonaro00: The incoming name from url is 'postgres', but here we need to change it to the name of the db to dump.

PR #1922, opened 2024-11-25
fix: admin and cli fixes

jonaro00:


admin/src/args.rs
```diff
@@ -60,6 +57,13 @@ pub enum Command {
     },
 
     RenewCerts,
+
+    Gc {
+        days: u32,
+    },
+    GcShuttlings {
+        minutes: u32,
+    },
```
oddgrd, line 66: We should have some comments for these

cargo-shuttle/src/lib.rs
```diff
@@ -868,23 +868,20 @@ impl Shuttle {
...
                         }
+                        Err(e) => {
+                            trace!(error = %e, "did not find project by name");
+                            if create_missing_beta_project {
+                                trace!("creating project since it was not found");
+                                let proj = client.create_project_beta(name).await?;
+                                eprintln!("Created project '{}' with id {}", proj.name, proj.id);
+                                self.ctx.set_project_id(proj.id);
+                            }
+                        }
```
oddgrd, line 884: We're reverting this, right?

PR #1923, opened 2024-11-26
chore: move ApiKey type to backends, remove AccountTier sqlx::Type

jonaro00:
- debloats common
- prevents mistake of binding an AccountTier and db not finding the type


oddgrd (commented):
See auth CI failure

oddgrd (approved):

PR #1924, opened 2024-11-27
feat(common): Add utoipa attribute macro types

fussybeaver:
## Description of change
Adds utoipa types for the control layer in the new platform. 

Used so that we can generate openapi types from code.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ executors:
   docker-rust:
     docker:
       # Note: Let CI use our MSRV, rather than latest
-      - image: cimg/rust:1.78.0
+      - image: cimg/rust:1.81.0
     resource_class: small
   machine-ubuntu:
     machine:
@@ -103,7 +103,7 @@ commands:
       - run:
           name: Install Rust (MSRV)
           # Note: Let CI use our MSRV, rather than latest
-          command: which cargo || curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.78.0
+          command: which cargo || curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.81.0
   install-protoc:
     steps:
       - run:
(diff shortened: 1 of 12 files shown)


.circleci/config.yml
```diff
@@ -10,7 +10,7 @@ executors:
   docker-rust:
     docker:
       # Note: Let CI use our MSRV, rather than latest
-      - image: cimg/rust:1.78.0
+      - image: cimg/rust:1.81.0
```
jonaro00, line 13: Does this change bump MSRV? If so, we also need to bump it in `common/Cargo.toml`.

common/src/certificate.rs
```diff
@@ -1,6 +1,7 @@
 use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
 
-#[derive(Deserialize, Serialize, Debug)]
+#[derive(Deserialize, Serialize, Debug, ToSchema)]
```
jonaro00, line 4: To keep `common` deps light, we want to feature gate "utoipa" and use `#[cfg_attr("utoipa", derive(utoipa::ToSchema))]` (inlined to not need `use` with cfg flag ;) )

jonaro00 (commented):
The permit test is failing on other branches, so no need to fix it here and now

.circleci/config.yml
fussybeaver: Thanks for pointing out, addressed in cdfb19760f91

common/src/certificate.rs
fussybeaver: Put that into 28e195d46791

common/Cargo.toml
```diff
@@ -39,6 +39,7 @@ tracing-subscriber = { workspace = true, optional = true }
 typeshare = { workspace = true }
 url = { workspace = true, features = ["serde"] }
 uuid = { workspace = true, features = ["v4", "serde"], optional = true }
+utoipa = { version = "5", features = ["chrono"], optional = true }
```
jonaro00, line 42: For clarity/visibility, I would also add `utoipa = ["dep:utoipa"]` under features, like tonic has.

fussybeaver: :+1:  Addressed in d81ddcb9b5a8

PR #1930, opened 2024-12-19
fix(cargo-shuttle): don't override user provided RUST_LOG on local run

jonaro00:
## Description of change

Fixes RUST_LOG override in `shuttle run` reported in https://discord.com/channels/803236282088161321/1318703796007735306

Local run with `RUST_LOG=trace` now shows `trace!()` logs from the user code.

### DIFF
--- a/runtime/src/start.rs
+++ b/runtime/src/start.rs
@@ -79,7 +79,13 @@ pub async fn start(loader: impl Loader + Send + 'static, runner: impl Runner + S
                 // let user override RUST_LOG in local run if they want to
                 EnvFilter::try_from_default_env()
                     // otherwise use our default
-                    .or_else(|_| EnvFilter::try_new("info,shuttle=trace"))
+                    .or_else(|_| {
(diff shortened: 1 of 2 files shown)


runtime/src/start.rs
```diff
@@ -79,7 +79,13 @@ pub async fn start(loader: impl Loader + Send + 'static, runner: impl Runner + S
                 // let user override RUST_LOG in local run if they want to
                 EnvFilter::try_from_default_env()
                     // otherwise use our default
-                    .or_else(|_| EnvFilter::try_new("info,shuttle=trace"))
+                    .or_else(|_| {
+                        EnvFilter::try_new(if args.beta {
+                            "info"
+                        } else {
+                            "info,shuttle=trace"
+                        })
```
oddgrd, line 87: Why is this different for beta?

jonaro00: The default level was reduced to have less verbose output in logs when starting the runtime

oddgrd (approved):

PR #1931, opened 2024-12-19
feat: deprecate new project creation

oddgrd:
## Description of change

We announced that new project creation will be disabled on the 19th of December. This change will enforce that restriction.

Will test on staging.

### DIFF
--- a/gateway/src/service.rs
+++ b/gateway/src/service.rs
@@ -26,7 +26,7 @@ use shuttle_backends::headers::XShuttleAdminSecret;
 use shuttle_backends::project_name::ProjectName;
 use shuttle_common::constants::SHUTTLE_IDLE_DOCS_URL;
 use shuttle_common::models::error::{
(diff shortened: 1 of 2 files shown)


gateway/src/service.rs
```diff
@@ -718,11 +722,12 @@ impl GatewayService {
...
-            // outright if the project already exists (this happens if
-            // it belongs to another account).
-            self.insert_project(project_name, Ulid::new(), user_id, idle_minutes)
-                .await
+            // Project creation is deprecated on the legacy platform as of the 19th of December
+            // 2024.
+            Err(Deprecated(
+                "New project creation on the shuttle.rs platform is deprecated.".to_string(),
+            )
+            .into())
```
jonaro00, line 730: nit: maybe add admin bypass, but not important

jonaro00 (approved):

Issue #1959, opened 2025-01-15
[Improvement]: allow using docker locally configured for DOCKER_HOST tcp/http values


AndreiOrmanji:
### Describe the improvement

My local setup requires using `DOCKER_HOST=tcp://localhost:2375`. Such configuration can not be used with `shuttle run` command at least for versions `0.51.0`, because it relies on `connect_with_local_defaults` method, that uses unix sockets for `unix` and named pipe for `windows`. `bollard` crate (starting from version 0.16.0) has `connect_with_defaults` method, that tries all possible variants of `DOCKER_HOST` value types (sockets, named pipes, http, https).

I suggest updating to the latest version of `bollard` crate 
```diff
- bollard = "0.15.0"
+ bollard = { version = "0.18.1", features = ["ssl_providerless"] }
```
replace calls of `connect_with_local_defaults` to `connect_with_defaults` to fix cases when DOCKER_HOST is set up as tcp/http/https
replace calls of `connect_with_unix` to `connect_with_socket`, because it was renamed

### Duplicate declaration

- [X] I have searched the issues and this improvement has not been requested before.

PR #1960, opened 2025-01-15
DOCKER_HOST handling fix

AndreiOrmanji:
## Description of change

fixes https://github.com/shuttle-hq/shuttle/issues/1959
1. update bollard to latest version
- modify `Cargo.toml`
```diff
- bollard = "0.15.0"
+ bollard = { version = "0.18.1", features = ["ssl_providerless"] }
```
- replace calls of `connect_with_unix` to `connect_with_socket`, because it was renamed in new versions
2. Fix DOCKER_HOST handling
replace calls of `connect_with_local_defaults` to `connect_with_defaults` to fix cases when DOCKER_HOST is set up as tcp/http/https

By running tests locally


cargo-shuttle/src/provisioner_server.rs
```diff
@@ -41,7 +41,7 @@ impl LocalProvisioner {
         // This only constructs the client and does not try to connect.
         // If the socket is not found, a "no such file" error will happen on the first request to Docker.
         Ok(Self {
-            docker: Docker::connect_with_local_defaults()?,
+            docker: Docker::connect_with_defaults()?,
```
jonaro00, line 44: @fussybeaver Is the comment above this line still true for the new version? CI is failing for tests that shouldn't try to connect to Docker.

fussybeaver: I believe `connect_with_defaults` is a new method that tries to infer from environment variables whether you are connecting to a socket or https or a named pipe - IIRC it was added for better compatibility with testcontainers and co. I would stick to the old method if CI is failing, because we may not be setting those env vars appropriately.

jonaro00 (approved):
LGTM! Seems to work as expected now.

PR #1961, opened 2025-01-15
feat: user project telemetry

the-wondersmith:
## Description of change
PR expands on the existing `tracing` setup/configuration feature in `shuttle-runtime`, adding an enhanced subscriber that includes an [OpenTelemetry](https://crates.io/crates/tracing-opentelemetry) layer and enables exporting all telemetry data (traces, metrics, and logs) generated by a Shuttle project to any compatible OTLP collector.

The new behavior is gated behind a new `setup-telemetry` feature flag.

> [!IMPORTANT]
> PR changes the default `RUST_LOG` configuration from `info` to `info,{crate_name}=debug`

### DIFF
--- a/common/src/resource.rs
+++ b/common/src/resource.rs
@@ -352,7 +352,7 @@ mod test {
         ];
 
         for input in inputs {
-            let actual = ResourceTypeBeta::from_str(&input.to_string()).unwrap();
+            let actual = ResourceTypeBeta::from_str(input.as_ref()).unwrap();
             assert_eq!(input, actual, ":{} should map back to itself", input);
         }
     }
--- a/common/src/resource.rs
+++ b/common/src/resource.rs
@@ -182,10 +182,6 @@ pub enum ResourceTypeBeta {
     #[strum(to_string = "database::aws_rds::mariadb")]
     #[serde(rename = "database::aws_rds::mariadb")]
     DatabaseAwsRdsMariaDB,
-    /// Project-specific telemetry configuration data
-    #[strum(to_string = "project::telemetry::config")]
-    #[serde(rename = "project::telemetry::config")]
-    ProjectTelemetryConfig,
     /// (Will probably be removed)
     #[strum(to_string = "secrets")]
     #[serde(rename = "secrets")]
@@ -254,7 +250,7 @@ impl Display for Type {
 // this can be removed when deployers AND r-r no longer hold resources in sqlite state
 #[cfg(feature = "sqlx")]
 mod _sqlx {
-    use std::{borrow::Cow, boxed::Box, str::FromStr};
+    use std::{borrow::Cow, str::FromStr};
 
     use sqlx::{
         postgres::{PgArgumentBuffer, PgValueRef},
@@ -289,16 +285,16 @@ mod _sqlx {
         }
     }
 
-    impl<DB: Database> sqlx::Type<DB> for ResourceState
+    impl<DB: sqlx::Database> sqlx::Type<DB> for ResourceState
     where
         str: sqlx::Type<DB>,
     {
-        fn type_info() -> <DB as Database>::TypeInfo {
+        fn type_info() -> <DB as sqlx::Database>::TypeInfo {
             <str as sqlx::Type<DB>>::type_info()
         }
     }
 
-    impl sqlx::Encode<'_, Postgres> for ResourceState {
+    impl sqlx::Encode<'_, sqlx::Postgres> for ResourceState {
         fn encode_by_ref(&self, buf: &mut PgArgumentBuffer) -> sqlx::encode::IsNull {
             #[allow(clippy::needless_borrows_for_generic_args)]
             <&str as sqlx::Encode<Postgres>>::encode(&self.to_string(), buf)
@@ -311,28 +307,6 @@ mod _sqlx {
 
             let state = ResourceState::from_str(value)?;
             Ok(state)
-        }
-    }
-
-    impl<DB: Database> sqlx::Type<DB> for super::ResourceTypeBeta
-    where
-        str: sqlx::Type<DB>,
-    {
-        fn type_info() -> <DB as Database>::TypeInfo {
-            <str as sqlx::Type<DB>>::type_info()
-        }
-    }
-
-    impl sqlx::Encode<'_, Postgres> for super::ResourceTypeBeta {
-        fn encode_by_ref(&self, args: &mut PgArgumentBuffer) -> sqlx::encode::IsNull {
-            <&str as sqlx::Encode<Postgres>>::encode(self.to_string().as_str(), args)
-        }
-    }
-
-    impl<'r> sqlx::Decode<'r, Postgres> for super::ResourceTypeBeta {
-        fn decode(value: PgValueRef<'r>) -> Result<Self, sqlx::error::BoxDynError> {
-            <&str as sqlx::Decode<Postgres>>::decode(value)
-                .and_then(|value| super::ResourceTypeBeta::from_str(value).map_err(Into::into))
         }
     }
 }
--- a/common/src/resource.rs
+++ b/common/src/resource.rs
@@ -182,6 +182,10 @@ pub enum ResourceTypeBeta {
     #[strum(to_string = "database::aws_rds::mariadb")]
     #[serde(rename = "database::aws_rds::mariadb")]
     DatabaseAwsRdsMariaDB,
+    /// Project-specific telemetry configuration data
+    #[strum(to_string = "project::telemetry::config")]
+    #[serde(rename = "project::telemetry::config")]
+    ProjectTelemetryConfig,
     /// (Will probably be removed)
     #[strum(to_string = "secrets")]
     #[serde(rename = "secrets")]
@@ -250,7 +254,7 @@ impl Display for Type {
 // this can be removed when deployers AND r-r no longer hold resources in sqlite state
 #[cfg(feature = "sqlx")]
 mod _sqlx {
-    use std::{borrow::Cow, str::FromStr};
+    use std::{borrow::Cow, boxed::Box, str::FromStr};
 
     use sqlx::{
         postgres::{PgArgumentBuffer, PgValueRef},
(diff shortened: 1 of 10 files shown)


runtime/Cargo.toml
```diff
@@ -40,6 +62,26 @@ default = ["setup-tracing"]
...
 setup-tracing = [
-    "tracing-subscriber/default",
+    "tracing-subscriber/ansi",
     "tracing-subscriber/env-filter",
+    "tracing-subscriber/fmt",
+    "tracing-subscriber/smallvec",
+    "tracing-subscriber/std",
+    "tracing-subscriber/tracing-log",
+]
+setup-telemetry = [
```
jonaro00, line 72: I had an idea: This can be called `telemetry`, and we can rename `setup-tracing` to `tracing-subscriber`.

common/src/resource.rs
```diff
@@ -250,7 +254,7 @@ impl Display for Type {
 // this can be removed when deployers AND r-r no longer hold resources in sqlite state
 #[cfg(feature = "sqlx")]
 mod _sqlx {
```
jonaro00, line 256: FYI This module is getting :bomb: in #1962

runtime/src/start.rs
```diff
@@ -69,33 +76,41 @@ pub async fn start(loader: impl Loader + Send + 'static, runner: impl Runner + S
...
             .with(fmt::layer().without_time())
             .with(
                 // let user override RUST_LOG in local run if they want to
-                EnvFilter::try_from_default_env()
+                EnvFilter::try_from_default_env().unwrap_or_else(|_| {
                     // otherwise use our default
-                    .or_else(|_| {
-                        EnvFilter::try_new(if args.beta {
+                    Into::into(format!(
+                        "{},{}=debug",
```
jonaro00, line 89: Changing the log level for user crates is good to note down in the PR so we remember it in release notes.

runtime/src/start.rs
```diff
@@ -48,7 +48,14 @@ impl Args {
...
 }
 
-pub async fn start(loader: impl Loader + Send + 'static, runner: impl Runner + Send + 'static) {
+pub async fn start(
+    loader: impl Loader + Send + 'static,
+    runner: impl Runner + Send + 'static,
+    #[cfg_attr(not(feature = "setup-telemetry"), allow(unused_variables))]
+    project_name: &'static str,
+    #[cfg_attr(not(feature = "setup-telemetry"), allow(unused_variables))]
+    project_version: &'static str,
```
jonaro00, line 57: nit: Name them `crate_name` and `package_version` to match cargo's names

jonaro00: We can also make this more verbose: (a few lines below)
```rs
println!("{} starting: {} {}", crate::VERSION_STRING, crate_name, package_version);
```

jonaro00 (commented):
Thanks for the typo fixes :)
Remember to keep changes to legacy code few.

common/src/resource.rs
```diff
@@ -182,6 +182,10 @@ pub enum ResourceTypeBeta {
     #[strum(to_string = "database::aws_rds::mariadb")]
     #[serde(rename = "database::aws_rds::mariadb")]
     DatabaseAwsRdsMariaDB,
+    /// Project-specific telemetry configuration data
+    #[strum(to_string = "project::telemetry::config")]
+    #[serde(rename = "project::telemetry::config")]
+    ProjectTelemetryConfig,
```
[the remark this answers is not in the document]
the-wondersmith: Reverted in favor of a server-side fix

common/src/resource.rs
the-wondersmith: 🎉

runtime/src/start.rs
the-wondersmith: Both good suggestions, applied

runtime/src/start.rs
the-wondersmith: Added to PR description 🙂

runtime/Cargo.toml
jonaro00: Hmm yeah, and `telemetry` sounds like you'll be opting in to the unwanted kind of telemetry... `setup-` will be good.

jonaro00 (commented):
Looks good so far!

runtime/src/trace.rs
```diff
@@ -0,0 +1,498 @@
...
+}
+
+// Create a Resource that captures information about the entity for which telemetry is recorded.
+pub fn resource(crate_name: &'static str, package_version: &'static str) -> Resource {
+    Resource::from_schema_url(
+        [
+            Some(KeyValue::new(SERVICE_NAME, crate_name)),
+            Some(KeyValue::new(SERVICE_VERSION, package_version)),
+            Some(KeyValue::new(TELEMETRY_SDK_NAME, "opentelemetry")),
+            Some(KeyValue::new(TELEMETRY_SDK_VERSION, "0.27.1")),
```
oddgrd, line 373: Should we set this from a constant, with some notes to remind us to update it when we update the sdk?

runtime/src/trace.rs
```diff
@@ -0,0 +1,498 @@
...
+    builder.build()
+}
+
+// Construct MeterProvider for MetricsLayer
+pub fn init_meter_provider(endpoint: &Option<String>, resource: Resource) -> SdkMeterProvider {
+    let mut builder = MeterProviderBuilder::default().with_resource(resource);
+
+    if let Some(endpoint) = endpoint {
+        let exporter = opentelemetry_otlp::MetricExporter::builder()
+            .with_temporality(Temporality::default())
```
oddgrd, line 407: Nit: This is already set by default in ::builder().

runtime/src/trace.rs
```diff
@@ -0,0 +1,498 @@
...
+
+// Construct TracerProvider for OpenTelemetryLayer
+pub fn init_tracer_provider(endpoint: &Option<String>, resource: Resource) -> TracerProvider {
+    // TODO(the-wondersmith): make trace sample rate & strategy configurable
+    let sampler = Sampler::ParentBased(Box::new(Sampler::TraceIdRatioBased(1.0)));
+
+    let mut builder = TracerProvider::builder()
+        .with_sampler(sampler)
+        .with_resource(resource)
+        .with_id_generator(RandomIdGenerator::default());
```
oddgrd, line 435: Nit: this is also set by default

runtime/src/trace.rs
```diff
@@ -0,0 +1,498 @@
...
+}
+
+impl EventFieldValues {
+    const REMAPPED_FIELDS: [(&'static str, &'static str); 4] = [
+        ("log.line", CODE_LINENO),
+        ("log.file", CODE_FILEPATH),
+        ("log.target", "code.target"),
+        ("log.module_path", "code.module_path"),
+    ];
+}
```
oddgrd, line 183: I see we have a lot of code to remap these fields to match otel conventions. Do you mind expanding on why we need to do that, and what would happen if we didn't? Is it a requirement, or a nice to have?

oddgrd (approved):
Nice!

runtime/src/trace.rs
the-wondersmith: [Working](https://github.com/open-telemetry/opentelemetry-rust/pull/2504) on it 🥲

jonaro00: A comment in Cargo.toml should do fine.

runtime/src/trace.rs
the-wondersmith: It's... sort of both? It's less "remapped to match convention" and more "standardized so it all comes out the same". If we *don't* do it, the queries on the user side get really annoying because the attributes show up differently based on whether an event was emitted by `log` or `tracing`. The convention is just a convenient target to standardize on 😅

runtime/src/trace.rs
the-wondersmith: It's included here because 1) I ripped it off wholesale from the example code in `opentelemetry_otlp`, and 2) so that we know what the buttons and levers we have available / might want to fiddle with.

Would you prefer it be omitted?

oddgrd: Not important, feel free to leave in.

runtime/src/trace.rs
oddgrd: Okay, sgtm, thanks!

runtime/src/trace.rs
jonaro00: Otel guy mentioned it is populated automatically. Did you observe otherwise?
Anyways, don't let this block. We can have it hardcoded until your otel PR is resolved.

jonaro00 (approved):

runtime/src/trace.rs
the-wondersmith: Yeah, it's not populated *at all* unless you do it here manually. Unless he means it's populated... somewhere else 🤔? Anything we could *not* do here would be nice. Do you have a link/reference to where he mentioned it?

the-wondersmith: > A comment in Cargo.toml should do fine.

[Added](https://github.com/shuttle-hq/shuttle/blob/feat/user-project-telemetry/runtime/Cargo.toml#L31)

oddgrd (approved):

runtime/src/trace.rs
```diff
@@ -0,0 +1,498 @@
...
+
+    global::set_meter_provider(provider.clone());
+
+    provider
+}
+
+// Construct TracerProvider for OpenTelemetryLayer
+pub fn init_tracer_provider(endpoint: &Option<String>, resource: Resource) -> TracerProvider {
+    // TODO(the-wondersmith): make trace sample rate & strategy configurable
+    let sampler = Sampler::ParentBased(Box::new(Sampler::TraceIdRatioBased(1.0)));
```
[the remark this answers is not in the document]
the-wondersmith: I'm honestly not sure one way or the other, so I commented out the overrides and we can do some more research as we evolve the feature? FTR, this is the default sample config when no additional customizations are applied:

Screenshot 2025-01-20 at 9 02 01 AM

PR #1962, opened 2025-01-15
chore: remove legacy code

jonaro00:
## Description of change

The :bomb: is here.
Removes all shuttle.rs code, keeping only what is used on shuttle.dev. :saluting_face: 
*Review at your own risk (not recommended).*
(Shoutout to https://github.com/twitter/the-algorithm/pull/122)

- Lines of code (`onefetch -e examples`): **40315 -> 9979**
- Axum hello-world dependency tree: **279 -> 217**
- cargo-shuttle release build dependency tree: **485 -> 432**

Breaking changes:
- Removed legacy CLI commands and options
  - `cargo shuttle` now targets shuttle.dev
- Removed ~~nearly~~ all legacy crate APIs
- Removed unsupported crates & resources
- Removed some crate features
- Changed module location of some shuttle-common items, added "tables" feature
- Bumped postgres version in local provisioner to 16, to match current RDS and shared databases
- Fixed `shuttle run` to use same secrets loading logic as `shuttle deploy`, which is more strict, instead of using legacy logic

- Updated CLI tests to use new platform. Added back `rocket_postgres` CLI local run test using DB.
- Other than that tests were just removed ;)


jonaro00:
@chesedo @oddgrd @brokad @iulianbarbu @christos-h @Kazy and other contributors,
It has been a pleasure :saluting_face:

cargo-shuttle/src/args.rs
```diff
@@ -108,10 +106,6 @@ pub enum Command {
     /// Manage deployments
     #[command(subcommand, visible_alias = "depl")]
     Deployment(DeploymentCommand),
-    /// View the status of a Shuttle service
-    Status,
-    /// Stop a Shuttle service
-    Stop,
     /// View build and deployment logs
     Logs(LogsArgs),
```
oddgrd, line 110: Should we also remove the logs args that are not supported on .dev?

oddgrd (approved):
Thanks! :boom:

PR #1973, opened 2025-01-30
feat: allow --no-follow and --raw on redeploy and stop

jonaro00:


fussybeaver (approved):

cargo-shuttle/src/args.rs
```diff
@@ -308,23 +311,29 @@ pub struct DeployArgs {
     /// WIP: Deploy this Docker image instead of building one
     #[arg(long, short = 'i', hide = true)]
     pub image: Option<String>,
-    /// Don't follow the deployment status, exit after the deployment begins
-    #[arg(long, visible_alias = "nf")]
-    pub no_follow: bool,
```
oddgrd, line 313: Don't we want to keep this for the regular deploy command? Or am I reading this removal wrong?

cargo-shuttle/src/args.rs
```diff
@@ -185,6 +185,9 @@ pub enum DeploymentCommand {
     Redeploy {
         /// ID of deployment to redeploy
         id: Option<String>,
+
+        #[command(flatten)]
+        tracking_args: DeploymentTrackingArgs,
```
oddgrd, line 190: Would you mind adding this to deployment stop as well?

cargo-shuttle/src/args.rs
jonaro00: It got moved into tracking_args

oddgrd: I see it now :see_no_evil:

PR #1989, opened 2025-02-25
build(shared-db): upgrade diesel-async to v0.5.2

valentinegb:
## Description of change

Upgrades `shuttle-shared-db`'s dependency on `diesel-async` from v0.4.1 to v0.5.2.

Clippy gives no warnings, compilation is successful.


jonaro00 (approved):
Thanks!

PR #1995, opened 2025-03-06
feat(common): team invite models

jonaro00:


admin/src/lib.rs
```diff
@@ -24,7 +24,8 @@ pub async fn run(args: Args) {
             let certs = client.get_old_certificates().await.unwrap();
             eprintln!("Starting renewals of {} certs in 5 seconds...", certs.len());
             tokio::time::sleep(tokio::time::Duration::from_millis(5000)).await;
-            for (cert_id, _) in certs {
+            for (cert_id, subject) in certs {
+                println!("--> {cert_id} {subject}");
```
the-wondersmith, line 28: What's with the three different methods of logging here ([`tracing::trace`](https://github.com/shuttle-hq/shuttle/blob/6212aa64851fa5d3ba906724108b473164728a53/admin/src/lib.rs#L14), [`println!()`](https://github.com/shuttle-hq/shuttle/blob/6212aa64851fa5d3ba906724108b473164728a53/admin/src/lib.rs#L28), [`eprintln!()`](https://github.com/shuttle-hq/shuttle/blob/6212aa64851fa5d3ba906724108b473164728a53/admin/src/lib.rs#L86))? Do we not just use `tracing::{LEVEL}` for a reason? Or is it just one of those "over time" things? 🤔

the-wondersmith (approved):

admin/src/lib.rs
jonaro00: (For cargo-shuttle) I try to use stdout for usable command output, stderr for other informative prints, and tracing for internals (since the log level makes printing them optional).
It matters even less in the admin cli if this is being followed, as we can update it any time.

PR #2012, opened 2025-03-19
feat: Add client capability to set replicas

fussybeaver:
## Description of change
Remove the update compute_tier functionality, replace with a freeform runtime configuration admin functionality.

### DIFF
--- a/common/src/models/project.rs
+++ b/common/src/models/project.rs
@@ -73,7 +73,7 @@ pub struct ProjectListResponse {
 }
 
 /// Set wanted field(s) to Some to update those parts of the project
-#[derive(Debug, Default, Deserialize, Serialize, Clone, PartialEq)]
+#[derive(Debug, Default, Deserialize, Serialize, Clone)]
 #[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
 #[typeshare::typeshare]
 pub struct ProjectUpdateRequest {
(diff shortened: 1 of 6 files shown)


common/types.ts
```diff
@@ -205,8 +205,8 @@ export interface ProjectUpdateRequest {
 	team_id?: string;
 	/** Transfer away from current team */
 	remove_from_team?: boolean;
-	/** Change compute tier */
-	compute_tier?: ComputeTier;
+	/** Project runtime configuration */
+	config?: RawValue;
```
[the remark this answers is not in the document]
fussybeaver: This is generated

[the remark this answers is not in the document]
jonaro00: It is right. Either add `"RawValue" = "any"` in typeshare.toml, or switch to `Value`.

common/src/models/project.rs
```diff
@@ -85,8 +85,9 @@ pub struct ProjectUpdateRequest {
     pub team_id: Option<String>,
     /// Transfer away from current team
     pub remove_from_team: Option<bool>,
-    /// Change compute tier
-    pub compute_tier: Option<ComputeTier>,
+    /// Project runtime configuration
+    #[cfg_attr(feature = "utoipa", schema(value_type = String, format = "json"))]
+    pub config: Option<Box<serde_json::value::RawValue>>,
```
jonaro00, line 90: I would use `Option<serde_json::Value>` and `serde_json::from_value()` on the other side. Saves the need of an extra feature.

fussybeaver: Unfortunately this serializes it as a string, so you need to deserialize it twice on the other side.

jonaro00: You sure? https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=fe5bd881adf39aefa656d176b5803988

fussybeaver: Seems to work now... I guess you can't have it as `serde_json::Value` in Clap's arguments.

PR #2028, opened 2025-05-05
fix(cargo-shuttle): exit with error when tracked deployment fails

### LINKED ISSUE #2027: Incorrect exit code when project build fails
Problem:
The project build failed because of a syntax error, but the workflow run was reported as success.
From the logs, we can see that the builder step in the Dockerfile returned "Reason: exit status 1".
Then, we can see "Phase complete: POST_BUILD State: SUCCEEDED": maybe the post-build step hides the erroneous exit of the builder.

Expected behaviour: The workflow run should fail.

Logs:
[logs_38047730048.zip](https://github.com/user-attachments/files/20025149/logs_38047730048.zip)

jonaro00:
Fixes #2027

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -1632,7 +1632,8 @@ impl Shuttle {
         raw: bool,
     ) -> Result<()> {
         let client = self.client.as_ref().unwrap();
-        if self.track_deployment_status(proj_id, depl_id).await? {
+        let failed = self.track_deployment_status(proj_id, depl_id).await?;
+        if failed {
             for log in client.get_deployment_logs(proj_id, depl_id).await?.logs {
                 if raw {
                     println!("{}", log.line);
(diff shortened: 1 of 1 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -1632,14 +1628,16 @@ impl Shuttle {
...
+        let failed = self.track_deployment_status(proj_id, depl_id).await?;
+        if failed {
             for log in client.get_deployment_logs(proj_id, depl_id).await?.logs {
                 if raw {
                     println!("{}", log.line);
                 } else {
                     println!("{log}");
                 }
             }
+            return Err(anyhow!("Deployment failed"));
```
oddgrd, line 1640: Should we add suggested actions for debugging the failure, or will that be apparent from the printed logs?

oddgrd (approved):

PR #2045, opened 2025-05-15
refa(common): better ApiError constructors & getters, feature gate tracing

jonaro00:
- add `new()` for brevity
- allow constructing 400, 401, 403, 404, and 503 with Error or string as the message
- feature gate tracing events and dependency
  - emit a `tracing::error!` on `ApiError::internal()` (prev only on `ApiError::internal_safe()`)
- remove complicated context constructors that are unused

### DIFF
--- a/common/src/models/error.rs
+++ b/common/src/models/error.rs
@@ -1,3 +1,5 @@
+use std::fmt::Display;
+
 use http::{status::InvalidStatusCode, StatusCode};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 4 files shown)


common/src/models/error.rs
```diff
@@ -1,226 +1,171 @@
...
+    }
+
+    /// Create a one-off internal error with a string message exposed to the user.
+    #[inline(always)]
+    pub fn internal(message: impl AsRef<str>) -> Self {
+        #[cfg(feature = "tracing-in-errors")]
+        {
+            /// Dummy wrapper to allow logging a string `as &dyn std::error::Error`
+            #[derive(Debug)]
+            struct InternalError(String);
```
jonaro00, line 50: @oddgrd I've assumed this helper is needed in order to make a string error correctly show up in DD through the `&dyn std::error::Error` path. Thoughts?

oddgrd: Correct, we use a visitor in the tracing layer, and the type is how we know which field to record: https://docs.rs/tracing-core/latest/tracing_core/field/trait.Visit.html#method.record_error

oddgrd (approved):

PR #2057, opened 2025-06-06
feat: added updated user usage structs

nathanjclark:
## Description of change
- updated user usage response to include new daily data and user overview


fussybeaver:
@nathanjclark you'll need to run `cargo make types` to generate the typescript compatible type signatures

PR #2058, opened 2025-06-10
feat(api-client, cargo-shuttle): json output mode

jonaro00:
Adds `--output json` so relevant commands print the json response instead. Not 100% coverage but the important ones have it.

breaking: generate command's `--output` arg renamed to `--output-file` (somewhat matches `deploy --output-archive`)


jonaro00:
Keeping this branch around until builder is rebased on main

PR #2063, opened 2025-06-16
feat(admin): fail error in looped jobs, notify gc'ed community projects

jonaro00:


admin/src/lib.rs
```diff
@@ -74,15 +74,24 @@ pub async fn run(args: Args) {
...
+                    client.renew_certificate(&cert_id).await.inspect_err(|_| {
+                        any_error = true;
+                    })
+                );
                 // prevent api rate limiting
                 tokio::time::sleep(tokio::time::Duration::from_millis(1000)).await;
             }
+            if any_error {
+                panic!("Error in loop occured");
+            }
```
oddgrd, line 94: Would this be cleaner in a utility function? Also, while we know what the error(s) was?

fussybeaver, line 94: Is there any capacity to instead of blanket `sleep` 5s initial and `sleep` 1s per loop:
 - retries with backoff (i.e max retries with a longer sleep each retry)
 - check the status code for the correct one
 - only panic on max retries

Issue #2055, opened 2025-06-01
Auto update project id in `.shuttle/config.toml` on project creation


dcodesdev:
### Describe the improvement

It would be nice to automatically update the `.shuttle/config.toml` file to the new project id once a new project is created.

### Current Behavior

If I delete a project with:

```bash
shuttle project delete
```

And then I want to create another one:

```bash
shuttle project create
```

The `.shuttle/config.toml` project id will still be the old project id.

### Desired behavior

Automatically update the project id in `.shuttle/config.toml` to the new project id.

### Duplicate declaration

- [x] I have searched the issues and this improvement has not been requested before.

PR #2098, opened 2025-08-09
feat(cargo-shuttle): maintain project id when deleting/creating project

iksuddle:
Addresses Issue #2055 by updating `.shuttle/config.toml` as follows:

- Removes `id` when a project is deleted using `shuttle project delete`
- Inserts/updates `id` when a project is created using `shuttle project create --name ...`

This should prevent potential mismatches and provide a smoother user experience.

### DIFF
--- a/cargo-shuttle/src/lib.rs
+++ b/cargo-shuttle/src/lib.rs
@@ -289,7 +289,7 @@ impl Shuttle {
                 } => self.delete_certificate(domain, yes).await,
             },
             Command::Project(cmd) => match cmd {
-                ProjectCommand::Create => self.project_create(args.project_args.name).await,
+                ProjectCommand::Create => self.project_create(&args.project_args).await,
                 ProjectCommand::Update(cmd) => match cmd {
                     ProjectUpdateCommand::Name { new_name } => self.project_rename(new_name).await,
                 },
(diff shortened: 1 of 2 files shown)


cargo-shuttle/src/lib.rs
```diff
@@ -1798,24 +1798,31 @@ impl Shuttle {
...
             bail!("Provide a project name with '--name <name>'");
         };
 
+        self.ctx.load_local_internal_config(project_args)?;
+
         let client = self.client.as_ref().unwrap();
         let r = client.create_project(name).await?;
 
+        let raw_json = r.raw_json.clone();
+        let project = r.into_inner();
```
jonaro00, line 1812: These 3 lines can be combined with into_parts (see other places)

cargo-shuttle/src/lib.rs
```diff
@@ -1798,24 +1798,31 @@ impl Shuttle {
...
                 println!("Created project '{}' with id {}", project.name, project.id);
             }
             OutputMode::Json => {
-                println!("{}", r.raw_json);
+                println!("{}", raw_json);
             }
         }
 
+        self.ctx.set_project_id(project.id);
+        self.ctx.save_local_internal()?;
```
jonaro00, line 1824: If we are calling `project create` (which can be called from any dir) from a directory without a rust project, this would write a `.shuttle/config.toml` file in a location where it is not wanted. We might need a way to save the config only if it existed when trying to load it.

cargo-shuttle/src/lib.rs
```diff
@@ -1933,6 +1940,9 @@ impl Shuttle {
 
         let res = client.delete_project(pid).await?.into_inner();
 
+        self.ctx.remove_project_id();
+        self.ctx.save_local_internal()?;
```
jonaro00, line 1944: If a `--id` different from the id in the config file is provided, we don't want to clear the id in the config file.

jonaro00 (commented):
Thanks for the PR! Unfortunately, there are many edge-cases that need to be considered. Leaving my thoughts:

iksuddle:
Thanks for all the feedback! I'll aim to fix everything you mentioned soon.

